PR #339, opened 2014-10-07
Spikegenerator

yger:


mstimberg:
Given that `SpikeGeneratorGroup` can currently give actually wrong results due to rounding errors (see https://groups.google.com/forum/#!topic/brian-development/Fyk4HtNaxMw), we should try to merge these changes soon.

How about simply raising a `NotImplementedError` for a period that does not correspond to a multiple of dt? We can always add support for it later if there is a clear use case for it (which I doubt there is). All that remains to do then is some cleaning up and merging it with latest master (I can have a go at it).

mstimberg:
I had another look at it and at least the naive implementation adds a whole new layer of complication to the already quite complicated templates, adding a new if condition and repeating the whole existing logic from re-starting at the beginning of the array. That also makes me think:  we currently do not complain if the period is *smaller* than `dt` either, though we do not handle this case.

In the interest of preserving everybody's sanity, I'll just go ahead adding a bunch of checks and raising a `NotImplementError` for periods that are not an integer multiple of `dt`.

mstimberg:
I merged master into this branch and added checks that raise errors for incompatible `dt` and `period` arguments. I didn't look any deeper into the code, but I think it should be certainly better than the previous one (which had some rounding issues) and also more performant (i.e. not slowing down for larger arrays).

So from my side this is ready to merge. Note that when you run the long test suite, you'll get errors for `SpatialNeuron` with Cython. I already fixed that in the spatialneuron branch and merged it into master, it will therefore work after the merge.

mstimberg:
@thesamovar: could you have a look at it when you find some time? It'd be nice if we could merge this. Actually, a number of bug fixes accumulated already, maybe we should do a second beta release soon?

thesamovar:
I just ran the test suite and get a large number of failures. I'm a bit busy right now to look into this in detail but I'll get to it as soon as possible.

Is it possible that merging master into this branch might fix some of the problems?

```
Running tests in "D:\programming\brian2\brian2" for targets numpy, weave, cython (excluding long tests)
Testing codegen-independent code 

Running tests that do not use code generation
......S.................S....................................................................................................................................................................
Ran 187 tests in 23.617s

OK (SKIP=2)
Running tests for target numpy:
.......S...........................................FFF..............................FF.....F........
FAIL: Test that `SpikeGeneratorGroup` connects properly.
    self.test(*self.arg)
  File "D:\programming\brian2\brian2\tests\test_spikegenerator.py", line 39, in test_spikegenerator_connected
    assert all(mon[idx].v[mon.t>=time] == 1)
AssertionError

FAIL: Basic test for `SpikeGeneratorGroup`.
    self.test(*self.arg)
  File "D:\programming\brian2\brian2\tests\test_spikegenerator.py", line 59, in test_spikegenerator_basic
    assert generator_spikes == recorded_spikes
AssertionError

FAIL: Basic test for `SpikeGeneratorGroup` with already sorted spike events.
    self.test(*self.arg)
  File "D:\programming\brian2\brian2\tests\test_spikegenerator.py", line 75, in test_spikegenerator_basic_sorted
    assert generator_spikes == recorded_spikes
AssertionError

FAIL: brian2.tests.test_synapses.test_transmission_scalar_delay
    self.test(*self.arg)
  File "D:\programming\brian2\brian2\tests\test_synapses.py", line 521, in test_transmission_scalar_delay
    assert_equal(mon[0].v[mon.t>=0.5*ms], 1)
  File "C:\Anaconda\lib\site-packages\numpy\testing\utils.py", line 275, in assert_equal
    return assert_array_equal(actual, desired, err_msg, verbose)
  File "C:\Anaconda\lib\site-packages\numpy\testing\utils.py", line 739, in assert_array_equal
    verbose=verbose, header='Arrays are not equal')
  File "C:\Anaconda\lib\site-packages\numpy\testing\utils.py", line 665, in assert_array_compare
    raise AssertionError(msg)
AssertionError: 
Arrays are not equal

(mismatch 100.0%)
 x: array([ 0.,  0.,  0.,  0.,  0.,  0.,  0.,  0.,  0.,  0.,  0.,  0.,  0.,
        0.,  0.])
 y: array(1)

FAIL: brian2.tests.test_synapses.test_transmission_scalar_delay_different_clocks
    self.test(*self.arg)
  File "D:\programming\brian2\brian2\tests\test_synapses.py", line 544, in test_transmission_scalar_delay_different_clocks
    assert_equal(mon[0].v[mon.t>=0.5*ms], 1)
  File "C:\Anaconda\lib\site-packages\numpy\testing\utils.py", line 275, in assert_equal
    return assert_array_equal(actual, desired, err_msg, verbose)
  File "C:\Anaconda\lib\site-packages\numpy\testing\utils.py", line 739, in assert_array_equal
    verbose=verbose, header='Arrays are not equal')
  File "C:\Anaconda\lib\site-packages\numpy\testing\utils.py", line 665, in assert_array_compare
    raise AssertionError(msg)
AssertionError: 
Arrays are not equal

(mismatch 100.0%)
 x: array([ 0.,  0.,  0.,  0.,  0.,  0.,  0.,  0.,  0.,  0.,  0.,  0.,  0.,
        0.,  0.])
 y: array(1)

FAIL: brian2.tests.test_synapses.test_external_variables
    self.test(*self.arg)
  File "D:\programming\brian2\brian2\tests\test_synapses.py", line 717, in test_external_variables
    assert target.v[0] == 2
AssertionError

Ran 100 tests in 22.253s

FAILED (SKIP=1, failures=6)
Running tests for target weave:
....S.S....................................S.......FFF..............................FF.....F........
FAIL: Test that `SpikeGeneratorGroup` connects properly.
    self.test(*self.arg)
  File "D:\programming\brian2\brian2\tests\test_spikegenerator.py", line 39, in test_spikegenerator_connected
    assert all(mon[idx].v[mon.t>=time] == 1)
AssertionError

FAIL: Basic test for `SpikeGeneratorGroup`.
    self.test(*self.arg)
  File "D:\programming\brian2\brian2\tests\test_spikegenerator.py", line 59, in test_spikegenerator_basic
    assert generator_spikes == recorded_spikes
AssertionError

FAIL: Basic test for `SpikeGeneratorGroup` with already sorted spike events.
    self.test(*self.arg)
  File "D:\programming\brian2\brian2\tests\test_spikegenerator.py", line 75, in test_spikegenerator_basic_sorted
    assert generator_spikes == recorded_spikes
AssertionError

FAIL: brian2.tests.test_synapses.test_transmission_scalar_delay
    self.test(*self.arg)
  File "D:\programming\brian2\brian2\tests\test_synapses.py", line 521, in test_transmission_scalar_delay
    assert_equal(mon[0].v[mon.t>=0.5*ms], 1)
  File "C:\Anaconda\lib\site-packages\numpy\testing\utils.py", line 275, in assert_equal
    return assert_array_equal(actual, desired, err_msg, verbose)
  File "C:\Anaconda\lib\site-packages\numpy\testing\utils.py", line 739, in assert_array_equal
    verbose=verbose, header='Arrays are not equal')
  File "C:\Anaconda\lib\site-packages\numpy\testing\utils.py", line 665, in assert_array_compare
    raise AssertionError(msg)
AssertionError: 
Arrays are not equal

(mismatch 100.0%)
 x: array([ 0.,  0.,  0.,  0.,  0.,  0.,  0.,  0.,  0.,  0.,  0.,  0.,  0.,
        0.,  0.])
 y: array(1)
-------------------- >> begin captured stdout << ---------------------
No module named msvccompiler in numpy.distutils; trying from distutils
Missing compiler_cxx fix for MSVCCompiler
No module named msvccompiler in numpy.distutils; trying from distutils
Missing compiler_cxx fix for MSVCCompiler

--------------------- >> end captured stdout << ----------------------

FAIL: brian2.tests.test_synapses.test_transmission_scalar_delay_different_clocks
    self.test(*self.arg)
  File "D:\programming\brian2\brian2\tests\test_synapses.py", line 544, in test_transmission_scalar_delay_different_clocks
    assert_equal(mon[0].v[mon.t>=0.5*ms], 1)
  File "C:\Anaconda\lib\site-packages\numpy\testing\utils.py", line 275, in assert_equal
    return assert_array_equal(actual, desired, err_msg, verbose)
  File "C:\Anaconda\lib\site-packages\numpy\testing\utils.py", line 739, in assert_array_equal
    verbose=verbose, header='Arrays are not equal')
  File "C:\Anaconda\lib\site-packages\numpy\testing\utils.py", line 665, in assert_array_compare
    raise AssertionError(msg)
AssertionError: 
Arrays are not equal

(mismatch 100.0%)
 x: array([ 0.,  0.,  0.,  0.,  0.,  0.,  0.,  0.,  0.,  0.,  0.,  0.,  0.,
        0.,  0.])
 y: array(1)
-------------------- >> begin captured stdout << ---------------------
No module named msvccompiler in numpy.distutils; trying from distutils
Missing compiler_cxx fix for MSVCCompiler
No module named msvccompiler in numpy.distutils; trying from distutils
Missing compiler_cxx fix for MSVCCompiler

--------------------- >> end captured stdout << ----------------------

FAIL: brian2.tests.test_synapses.test_external_variables
    self.test(*self.arg)
  File "D:\programming\brian2\brian2\tests\test_synapses.py", line 717, in test_external_variables
    assert target.v[0] == 2
AssertionError

Ran 100 tests in 29.388s

FAILED (SKIP=3, failures=6)
Running tests for target cython:
....S.SS...................................S........................................................
Ran 100 tests in 31.687s

OK (SKIP=4)
```

mstimberg:
master was merged into this branch quite recently, and the test suite runs fine on the branch alone on travis (and on my computer as well, of course), so I don't think that's the problem. Maybe double check that you have a clean working copy?

thesamovar:
I cleaned all my temp files out and rebuilt the spikequeue and I still get the errors. Is there anything else I need to clean? Master runs fine.

mstimberg:
> I cleaned all my temp files out and rebuilt the spikequeue and I still get the errors. Is there anything else I need to clean? Master runs fine.

I can't really think of anything else to do. It is odd that you get the errors for numpy and weave (it seems the `SpikeGeneratorGroup` is not generating any spikes?) but all works fine for Cython... Does the "openmp_consistency" test for standalone work (it uses `SpikeGeneratorGroup` as well)?

thesamovar:
OK I think I have isolated the problem: if the period is infinity, then ``t % period`` gives a nan rather than ``t``, and this messes everything up. Perhaps the behaviour of this has changed with different versions of numpy? I think a fix would be to set ``period=1e100`` or something like that instead of ``inf``. This appears to solve the problem for me, but I didn't commit a change. What do you think of this solution? I think 1e100 is OK given the lifetime of the universe is around 4e17 seconds.

thesamovar:
Actually I just committed the change: didn't see any reason why not (also added documentation on the period).

brian2/codegen/runtime/cython_rt/templates/spikegenerator.pyx
```diff
@@ -1,39 +1,45 @@
...
-    global _start_idx, _current_time
-    
-    # Check if we have been reset, if so reset the cache
-    if t<_current_time:
-        _start_idx = 0
-    
-    _current_time = t
+    {# USES_VARIABLES {_spikespace, N, t, dt, neuron_index, spike_time, period, _lastindex } #}
 
     # TODO: We don't deal with more than one spike per neuron yet
```
thesamovar: What happens if there is more than one spike in a timestep actually?

brian2/codegen/runtime/cython_rt/templates/spikegenerator.pyx
```diff
@@ -1,39 +1,45 @@
...
-    cdef int _cpp_numspikes = 0
+    cdef int _cpp_numspikes   = 0
+    cdef float padding_before = t % period
+    cdef float padding_after  = (t + dt) % period
+    cdef float epsilon        = 1e-3*dt
```
thesamovar: Were we thinking of making epsilon a user-defined constant?

brian2/codegen/runtime/numpy_rt/templates/spikegenerator.py_
```diff
@@ -1,12 +1,41 @@
...
+    else:
+        # If we are in the last timestep before the end of the period, we remove the first part of the
+        # test, because padding will be 0
+        test = abs(_spike_time - padding_after) < epsilon
+    if test:
+        break
+    _n_spikes += 1
+
+_indices = {{neuron_index}}[{{_lastindex}}[0]:{{_lastindex}}[0]+_n_spikes]
+
```
thesamovar: Have we checked the efficiency of the Python implementation yet? Is it better than the previous version?

mstimberg:
I don't think the `1e100` should cause any problems, it seems slightly less elegant but the fact that it works is more important ;)

brian2/codegen/runtime/cython_rt/templates/spikegenerator.pyx
mstimberg: It's basically undefined and all kind of things can happen, e.g. spikes can be missed. What happens most of the time is that the spike will be generated in a later time step I think.

thesamovar:
OK well I just finished running all the tests and they are fine except for the expected failure of the Cython versions of SpatialNeuron. Any thoughts on the inline comments above? Apart from that, I'm happy to merge.

brian2/codegen/runtime/cython_rt/templates/spikegenerator.pyx
thesamovar: I think perhaps we should add a check for this then?

mstimberg: I guess that would be better than a TODO in the code :) But probably a check before hand, not during the run, right? But this would involve applying the rounding/epsilon logic before. Hmm, maybe we should actually do a conversion from times to time bins before hand?

thesamovar: Maybe that's a good idea. It would simplify the logic within the template and I don't think it would make much difference computationally, right?

brian2/codegen/runtime/numpy_rt/templates/spikegenerator.py_
mstimberg: Actually, I ran some crude benchmarks and for big populations/many spikes it is considerably worse now... In fact, in the previous implementation I don't see any scaling with the number of spikes (it should be logarithmic because of the binary search). I guess the above code should be significantly faster if expressed in a vectorized way instead of using a loop. Obviously it would be nice if we don't decrease performance with the new, "faster" approach...

brian2/codegen/runtime/cython_rt/templates/spikegenerator.pyx
mstimberg: It's a bit of an opaque constant, I'm not sure this is something the user'd like to set. Is there any reason for not setting it to the smallest epsilon that still gives correct results for multiples of dt?

yger: Actually, this implementation without the logarithmic search has never been
benchmarked, but I always assumed that this would be faster. It "felt" as
if it was simpler.... If this is not the case, then this is problematic...
I see no reason why the used should change epsilon, and this can (should)
be set to the smallest possible value giving correct results for multiples
of dt.

We can talk about that tomorrow

2015-01-21 15:20 GMT+01:00 Marcel Stimberg <notifications@github.com>:

brian2/codegen/runtime/numpy_rt/templates/spikegenerator.py_
```diff
@@ -1,12 +1,41 @@
...
+# If there is a periodicity in the SpikeGenerator, we need to reset the lastindex 
+# when all spikes have been played and at the end of the period
+if not_first_spike and ({{spike_time}}[{{_lastindex}}[0] - 1] > padding_before):
+    {{_lastindex}}[0] = 0
+
+for _spike_time in {{spike_time}}[{{_lastindex}}[0]:]:
+    test = (_spike_time > padding_before) or (abs(_spike_time - padding_before) < epsilon)
+    if test:
+        break
+    {{_lastindex}}[0] += 1
```
mstimberg: What are lines 20-24 for, actually (same for the C code)? Forwarding the `_lastindex` if the spike times after it are in fact still before the current time -- can this situation ever occur? It does not seem to be triggered in any of the tests and they all pass if the code is deleted...

yger: As far as I remember, those tests were because I had problem when I had
spikes, with a non zero period, that were falling in the first or the last
time-step. I wanted to be sure that all were played and played only once.
But if you say that they can be removed...

2015-01-21 16:32 GMT+01:00 Marcel Stimberg <notifications@github.com>:

mstimberg: Well, either they can be removed or we are missing a test that tests that specific situation they are necessary for:)

brian2/codegen/runtime/cython_rt/templates/spikegenerator.pyx
thesamovar: I agree it's an opaque constant, but it somehow feels odd to have it written in each template rather than specified in a single place. It's not super important.

brian2/codegen/runtime/numpy_rt/templates/spikegenerator.py_
thesamovar: OK for the initial pull it doesn't matter too much about the performance, but we should keep it in mind that we want to optimise the Python version maybe.

brian2/codegen/runtime/cython_rt/templates/spikegenerator.pyx
mstimberg: About doing the conversion to integer time steps at the preparation stage: I'm wondering if we could do this in an elegant way, also incorporating the offset calculation we had in Brian 1.
The simple integer conversion would convert (for dt=0.1ms)
```
indices = [0, 1, 0, 1, 0, 1]
times = [0.1, 0.1, 0.2, 0.3, 1.0, 1.0] * ms
```
into
```
timebins = [1, 1, 2, 3, 10, 10]
```
In Brian 1, we'd calculate offsets:
```
offsets = [0, 0, 2, 3, 4, 4, 4, 4, 4, 4, 4, 6]
```
The Brian 1 solution isn't great for long runs with few spikes, because it stores an array of the size of the number of time steps. But we could a similar thing by just storing the number of spikes falling in the same time bin, but only for spikes that actually happen:
```
same_bin_spikes = [2, 1, 1, 2]
```
I think this could be calculated quite efficiently before hand and would simplify the templates significantly, in numpy we'd have to do neither looping nor bisection during runtime.

thesamovar: That sounds good: it just depends on how efficient the computation is beforehand. It has to be efficient enough that there is no advantage to using weave I think (which is probably true just for computing the timebins but maybe not for computing same_bin_spikes?).

mstimberg: I'm almost sure one can come up with a fancy numpy expression that calculates the `same_bin_spikes` and is pretty efficient. But if not, then of course it doesn't help. Also we should not spend too much time optimising this, I don't think any of this is the bottleneck in simulations.

thesamovar: Good points! I'm convinced. :)

mstimberg: It turns out numpy's `bincount` is pretty straightforward and efficient for this task:
```Python
bincounts = np.bincount(timebins);bincounts = bincounts[bincounts!=0]
```
I did some tests with "timebin" arrays of size 1000000, and it took never more than ~7ms, and it seems to scale linearly for bigger arrays. I'll go ahead and implement something based on this approach.

thesamovar: 7ms/1M spikes certainly seems reasonable. :)

thesamovar: Wait though doesn't that create an intermediate array with a rather large size? For example, if ``timebins=[1000000]`` then ``bincount(timebins)`` will have length 1M even though the final array is small. I think ``unique(timebins, return_counts=True)[1]`` will do the trick without the large intermediate array.

mstimberg: Yeah, I had a look at this as well, the problem is that it has only been introduced by numpy 1.9 and I would prefer not to force everyone to upgrade just for this. It also was about ten times slower in my example, but you are of course right that my proposal isn't great for big maximum spike times.Here's another variant that combines `unique` (without `return_count`) with `bincount`, which doesn't have the problem of generating an intermediate array of the size of `max_t / dt` (it only generates intermediate on the order of the length of `timebins`), this is significantly slower (about 20 times) for the densely filled arrays I used previously, but maybe that's still fine for a single preparatory step:
```Python
_, indices = np.unique(timebins, return_index=True)
bincounts = np.diff(np.hstack([indices, len(timebins)]))
```

thesamovar: Well if we wanted to be fancy we could compare ``amax(timebins)`` against ``len(timebins)`` somehow and select one of the two algorithms. Maybe this is getting out of hand though? For your solution, for a 1h long set of spike times with dt=0.1ms (i.e. ``amax(timebins)==10000*60*60``), it would use 275MB of intermediate RAM. I doubt that this will ever be much of a problem, especially given that it gets deallocated immediately after. So I'd say let's go with your original bincount solution.

mstimberg:
I'm adding this as a general comment now... I looked a bit into the pre-calculation bits and came up against a few issues:
* If we do the preparation in `before_run` (which we should, because `dt` could change between runs), then the code written in Python will not be executed for standalone, we'd therefore either need a different approach in standalone, or do the complete template-based approach like we do for the preparation of the spike queue. 
* We currently do not expose the integer time step of the clock. Therefore the pre-calculated integer time steps will not be that useful, we'll re-introduce rounding errors when determining the integer time step for the current `t`.
* The pre-calculated `same_bin_spikes` are not easy to fit into our current system, since we assume that arrays have a fixed length (and the length of this array would depend on `dt` and the actual spike times). We could simply use an array of the maximum length (number of spikes) and have unnecessary zeros in the end, though.

These are not huge problems, but I think it'd be nicer if we had a general approach for the first two issues (I'll open issues for them), they are not `SpikeGeneratorGroup`-specific. Maybe we just merge it as it is now, and deal with these questions later?

thesamovar:
I'm happy to put off those issues to later.

Did you talk to @yger about the lines that we weren't sure what they were doing? I think maybe we should resolve that and the issue of multiple spikes in a single timestep before merging.

mstimberg:
> Did you talk to @yger about the lines that we weren't sure what they were doing?

I did and he said it was about making sure that spikes were not played more than once. I think the particular situation where this could arise were situations with a period that was not a multiple of `dt`. Since we no longer allow this, I don't think it is necessary any more. Either way, the tests are passing without it, so if the code is necessary, we are missing a critical test here.

> issue of multiple spikes in a single timestep

You are right, I'll put a check for this in. I just also realized that the code in `CodeObject.before_run` *will* be executed for standalone, and since we are only dealing with static arrays and the clock's `dt`, which will be set in Python, we can actually do the timestep pre-binning in Python and it will be used in standalone afterwards (via the static array mechanism). But of course in the long run, this is something that should be done in the C++ code and not in Python.

mstimberg:
I added the check for multiple spikes in a time step and more tests. I did not quite find the correct slack to use based on machine epsilon (1000 times this epsilon was not big enough in some cases) so it uses 1e-6*dt for now. I guess we can clean this up when we do the pre-calculated bins (in which case we could also add the opportunity for the user to specify time bins instead of spike times directly, maybe?).

mstimberg:
Um, this epsilon issue is a nasty one, my choise of 1e-6 did actually lead to incorrect results in the Cython implementation. I reverted everything back to 1e-3, it should pass the tests now.

mstimberg:
There was a test failure because the `openmp_consistency` test used completely random spike times and it was therefore possible that a single neuron in the `SpikeGeneratorGroup` was supposed to spike more than once during a time step. I now changed the test so that all spikes fall into separate time bins.

thesamovar:
On the other hand, randomly selecting spike times in this way is the sort of thing users are very likely to do, so perhaps we need to fix the problem of multiple spikes in one bin? At the moment it raises an exception is that right? If so, that's OK for now and we could probably merge this (I should run the tests again first), but we'll want to fix that sooner rather than later.

mstimberg:
I can definitely imagine things like this coming up and I agree that we should take care of this. As you said, it currently raises an exception. This is mostly to be on the safe side, since there might be places in the code where we assume at most one spike per neuron per time step, given that a `NeuronGroup` cannot spike in any other way. We'd just have to check all the places around the spike transmission to be sure.

thesamovar:
OK I ran the tests OK here except for the expected Cython spatial neuron fails (``KeyError: u'The identifier "ab_plus" could not be resolved.'``). I think this is OK to merge. What shall we do for the remaining issues? Keep this branch and pull request open or create new issues and a new branch? I'm happy for you to go ahead and do the merging/etc.

mstimberg:
I think it's actually nicer to have new issues and eventually new branches for the all the other stuff, some of it is quite independent of `SpikeGeneratorGroup` (e.g. #389 and #390). I'll go ahead with the merge then.

Issue #145, opened 2013-10-17
New testing framework


thesamovar:
Some thoughts on what our testing framework should have:

* [ ] Simple unit tests of objects and functions.
* [ ] Tests of accuracy of simulation, including stochastic examples.
* [ ] Tests for devices that return a report on how much of the Brian API is supported and how accurate the simulation is compared to the reference.

thesamovar:
I think everything here is covered in other, newer issues.

PR #353, opened 2014-10-29
Device correctness

thesamovar:
This branch isn't ready to merge yet, but I'm creating this pull request for discussion.

See issue #145.


thesamovar:
Should we merge this branch at some point? It's not finished: more feature tests are needed, but the framework is OK I think? Not high priority, but I keep seeing this branch in the pull requests section...

mstimberg:
I'd be happy to merge it now, adding feature tests is something that would also be ok to do by directly pushing to master, it can't break anything.

thesamovar:
Unless we incorporate the feature tests into the test suite. ;)

mstimberg:
Sure, but for that we need the more complex test setup, I don't think we are ever going to run them as part of the after-every-commit test suite on travis.

thesamovar:
Indeed. OK shall I leave you to merge this then?

mstimberg:
Doing the merge now. I'll leave the branch open, I think it's actually nicer if new tests and other related things are first added to this branch and then added in chunks to the master branch instead of directly adding them to master one by one. If the commits are only new tests and the like, feel free to directly merge into master however, no need for a pull request.

I'll also merge master into this branch so that it is up-to-date with the rest.

PR #387, opened 2015-01-16
Spatialneuron

mstimberg:
Not quite ready for merging yet, I sometimes get segmentation faults with Cython... @thesamovar: could you have nevertheless a look at it whether it runs under Windows? I'm quite sure I sneaked in something Linux and/or gcc-specific...
There are no more Python-only bits in the simulation of `SpatialNeuron` now, so that means we can run it under standalone (yay :-) ). There is a small bug that make some examples not work where standalone does not like assignments like `neuron.I[0] = 1*nA`, but this is not spatialneuron-specific (and can be worked around by `neuron.I = 'int(i==0)*nA'` for now).

### DIFF
--- a/brian2/codegen/runtime/weave_rt/templates/spatialstateupdate.cpp
+++ b/brian2/codegen/runtime/weave_rt/templates/spatialstateupdate.cpp
@@ -13,16 +13,20 @@
     {{ common.insert_group_preamble() }}
 
 	double *_gtot_all=(double *)malloc(N*sizeof(double));
-	double *c=(double *)malloc(N*sizeof(double));
+	double *c=(double *)calloc(N, sizeof(double));
 	double ai,bi,_m;
 
+    int _vectorisation_idx = 1;
+
 	//// MAIN CODE ////////////
+	{{scalar_code|autoindent}}
+
 	// Tridiagonal solving
 	// Pass 1
 	for(int i=0;i<N;i++)
 	{
 		const int _idx = i;
-	    const int _vectorisation_idx = _idx;
+	    _vectorisation_idx = _idx;
 
 		{{vector_code|autoindent}}
 		_gtot_all[_idx]=_gtot;
@@ -44,8 +48,9 @@
 		}
 	}
 	for(int i=N-2;i>=0;i--)
+	{
 		{{v_star}}[i]={{v_star}}[i] - c[i]*{{v_star}}[i+1];
-	
+    }
 	// Pass 2
 	for(int i=0;i<N;i++)
 	{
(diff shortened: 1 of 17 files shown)


mstimberg:
I fixed the small standalone issue and found the source of the crash (it should also have affected weave and standalone, but for some reason it only led to segmentation faults in Cython for me). So this is ready to merge from my side now. @thesamovar, could run the  (long) test suite on Windows and if you are happy with it merge it? Let's continue to keep this branch open, though.

thesamovar:
I'm getting a lot of test failures for weave like this:

```
c:\users\dgoodman\appdata\local\temp\dgoodman\python27_compiled\sc_f7f9c58b7763abfd41df7ca0dd902b2c0.cpp(1060) : error C2065: 'M_PI' : undeclared identifier
```

The relevant section is:

```
        // Inverse axial resistance
        for (int _i=1; _i<N; _i++)
            _ptr_array_spatialneuron_spatialstateupdater__invr[_i] = (M_PI / (2 * _Ri) * (_ptr_array_spatialneuron_diameter[_i-1] * _ptr_array_spatialneuron_diameter[_i]) /
                           (_ptr_array_spatialneuron_length[_i-1] + _ptr_array_spatialneuron_length[_i]));
```

I guess a ``#define`` ought to do it. All other tests are passing, including Cython and standalone so hopefully that should fix it.

mstimberg:
Are you getting the `M_PI` error with msvc or also with gcc? If it is only with msvc, we could put a `#define` with the ones we already have for `NAN` and `INFINITY`. Apparently, we could also define `#define _USE_MATH_DEFINES` before importing `math.h`, but I'm not sure how this works with weave which automatically imports `math.h` (I think?).

thesamovar:
I only tested it with msvc so far because I'm on my 64 bit machine at the moment. I assume gcc will not have this problem.

mstimberg:
> I assume gcc will not have this problem.

I think so too. I merged latest master into this branch (the msvc-specific bits were not yet in the branch) and added a definition of `M_PI`, hopefully it should work now.

brian2/codegen/runtime/weave_rt/templates/spatialstateupdate.cpp
```diff
@@ -5,22 +5,28 @@
...
+                  _P, _B, _morph_i, _morph_parent_i, _starts, _ends,
+                  _invr0, _invrn} #}
 
 {% macro main() %}
     {{ common.insert_group_preamble() }}
 
 	double *_gtot_all=(double *)malloc(N*sizeof(double));
-	double *c=(double *)malloc(N*sizeof(double));
+	double *c=(double *)calloc(N, sizeof(double));
 	double ai,bi,_m;
```
thesamovar: Could these allocations be done just once rather than in the state update step? Also, why calloc in one case and malloc in the other case?

brian2/devices/cpp_standalone/templates/spatialstateupdate.cpp
```diff
@@ -0,0 +1,197 @@
...
+{# USES_VARIABLES { Cm, dt, v, N,
+                  ab_star0, ab_star1, ab_star2, b_plus,
+                  ab_plus0, ab_plus1, ab_plus2, b_minus,
+                  ab_minus0, ab_minus1, ab_minus2, v_star, u_plus, u_minus,
+                  _P, _B, _morph_i, _morph_parent_i, _starts, _ends,
+                  _invr0, _invrn} #}
+{% extends 'common_group.cpp' %}
+{% block maincode %}
+	double *_gtot_all=(double *)malloc(N*sizeof(double));
+	double *c=(double *)calloc(N, sizeof(double));
```
thesamovar: Same question about allocs here.

thesamovar:
Reviewed the code now and all looks good except that we do a memory allocation each time step which seems wasteful. Can this be avoided by allocating once at the beginning? This could be done by creating new arrays via the codegen system for example.

Still waiting for the full test run to complete but it got through weave OK so it should be fine.

thesamovar:
I ran the tests and they complete OK without error, but the Cython tests are taking an incredibly long time. Around 5m for the weave tests and 2h for the Cython ones, so I figure something is very badly optimised. I haven't narrowed it down yet. Did you not have the same issue? (This was for the long tests.)

brian2/codegen/runtime/weave_rt/templates/spatialstateupdate.cpp
mstimberg: I'd rather avoid having such internal variables as state variables, they are an implementation detail and only used within the template (they are not used in numpy, for example). But we could do the allocation only once by declaring the pointers as static and doing the allocation only if they haven't been allocated before. The only problem I'm seeing with that is that we can't de-allocate the memory. For standard runs this is certainly not an issue, but if a user creates multi-compartmental neurons in a loop, this would be a memory leak. In the interest of keeping things simple, I'd therefore tend to keep them as they are. I don't quite remember why I changed the `malloc` into the `calloc`, though, looking at the code it seems to be never used before it gets a value assigned.

mstimberg: Hmm, actually since this `N` is a constant, we could simply declare arrays instead of pointers and wouldn't have to care about the memory management. Maybe that's the simplest solution.

mstimberg:
> Around 5m for the weave tests and 2h for the Cython ones, so I figure something is very badly optimised. I haven't narrowed it down yet. Did you not have the same issue? (This was for the long tests.)

Um, no, I'm definitely not seeing this. I just ran the long test suite again, after deleting all Cython code files and it took less than 9m for the Cython tests including the compilcation time.
Curiously `test_rall` is now failing for me under Cython, though, but only as part of the test suite and not when I run the `test_spatialneuron` file individually (setting `prefs.codegen.target` to `'cython'` manually).

mstimberg:
Ug, I can reproduce the test error *sometimes*, apparently it fails every 5th run or so. When it fails, it fails quite dramatically. I guess it might use some values that are supposed to be zero but that are not properly initialized.

mstimberg:
Ok, the error I was seeing was a genuine "one element too far" access in a loop, it is fixed now. Can you run the long test suite again and see whether it still takes so long? Or maybe just run `test_spatialneuron` individually with `prefs.codegen.target = 'cython'`?

thesamovar:
OK, all tests pass in a reasonable time now so I guess the slow down was in some way mysteriously related to that bug.

Did you want to try static allocation or just stick with the current system? If you're happy with it as it is, I'm happy for it to be merged (go ahead).

mstimberg:
> Did you want to try static allocation or just stick with the current system? If you're happy with it as it is, I'm happy for it to be merged (go ahead).

I think I am going to merge it now, the static allocation complicates things a bit since for maximum C compiler compatiblity, I think we want to do it with a constant and not with the variable `N` that is passed in via the namespace. We can do this with something like `{{owner.N}}` in the jinja template, but I think that makes it less clear for others reading the code. I don't expect this allocation to make any meaningful performance difference anyway.

I'll do the merge then.

PR #392, opened 2015-01-29
Loop invariant optimisations

mstimberg:
I rebased this very old branch on current master and integrated the optimisation into the code generation mechanism. This was mostly straightforward, but getting all the details (e.g. dealing with existing scalar statements, etc.) needed some refactoring. It is now applied in `make_statements`. All the tests still pass and in some cases it leads to a non-negligible speed-up (using numpy or weave without `-ffast-math`). If you don't see anything wrong with it, I think this is ready to merge.


thesamovar:
Great! What examples show a non-negligible speed up and what sort of factor are we talking?

I'm running the tests now and will merge when they complete if they're fine.

thesamovar:
Found an error:

```
FAIL: Doctest: brian2.stateupdaters.exponential_euler.get_conditionally_linear_system
  File "C:\Anaconda\lib\doctest.py", line 2226, in runTest
    raise self.failureException(self.format_failure(new.getvalue()))
AssertionError: Failed doctest test for brian2.stateupdaters.exponential_euler.get_conditionally_linear_system
  File "D:\programming\brian2\brian2\stateupdaters\exponential_euler.py", line 10, in get_conditionally_linear_system

File "D:\programming\brian2\brian2\stateupdaters\exponential_euler.py", line 38, in brian2.stateupdaters.exponential_euler.get_conditionally_linear_system
Failed example:
    print(system['v'])
Expected:
    (-1/tau, w**2.0/tau)
Got:
    (-1/tau, 0)
```

mstimberg:
> Great! What examples show a non-negligible speed up and what sort of factor are we talking?

I only systematically tested one example, `IF_curve_Hodgkin_Huxley.py` (100 neurons, 2 seconds), and looked at the time the state update step takes, using our handy profiling information. Here's what I get (best of 3):

target | Before | after
------ | -------- | ------
numpy | 11.36s | 10.03s
weave (no `-ffast-math`) | 5.45 s | 3.61 s
weave (`-ffast-math`) | 0.84s | 0.85 s
cython | 5.18s | 3.58s

So for everything except weave with `-ffast-math` it makes some difference -- but, well, weave with `-ffast-math` is what you *should* use, anyway.

thesamovar:
Well the Cython speed-up is significant for people on Python 3 I guess.

mstimberg:
> Well the Cython speed-up is significant for people on Python 3 I guess.

Good point!

thesamovar:
OK I just ran the CUBA example and it is giving very wrong results: all neurons firing at the same time regularly.

mstimberg:
> OK I just ran the CUBA example and it is giving very wrong results: all neurons firing at the same time regularly.

Right, sorry, I was about to write that you shouldn't merge it yet. I realized that I forgot dealing with the non-stateless functions, i.e. I guess in the CUBA example it pulled out `rand()` as a scalar expression and only generates one random value for all neurons.

thesamovar:
Another thing noted from the CUBA example:

```
_lio_const_1 = - dt
...
_ge = ge * exp((_lio_const_1 * int(not_refractory)) / taue)
...
```

This isn't an error but shows up a limitation of this because you're still having to compute the ``exp`` as each neuron has potentially different values of the boolean ``not_refractory``. I wonder if there is a way we can handle boolean values explicitly in the LIO code?

thesamovar:
Ah, that would explain it!

thesamovar:
We should have a test for that. ;)

mstimberg:
> We should have a test for that. ;)

We should, but unfortunately testing with random numbers is a tricky subject. But I think we can stretch the definition of a deterministic test a bit and just check for something basic, e.g. that values aren't identical.

mstimberg:
About the doctest error: I think I introduced that one with a recent fix (not in this branch) :-/ It works fine with sympy 0.7.6 but I get the same error you are seeing with sympy 0.7.4. We should probably include tests with the minimal versions of requirements in our test suite...

I'll look into the problem soon, you should get the error on the master branch as well, no?

thesamovar:
Actually for the Hodgkin-Huxley IF-curve there is a problem too: it is flat. Is this related?

Also, I realise that we should maybe have a better way of generating constant names so that the generated code is more readable. For example:

```
_n = n * exp(dt * ((((_lio_const_14 / ((_lio_const_15 * exp(((-0.2) * v) / mV)) - _lio_const_9)) - (_lio_const_16 / ((_lio_const_15 * exp(((-0.2) * v) / mV)) - _lio_const_9))) + ((0.032 * v) / ((_lio_const_15 * exp(((-0.2) * v) / mV)) - _lio_const_9))) - ((_lio_const_17 * exp(((-0.025) * v) / mV)) / ms)))
```

At the moment we have:

```
_lio_const_6 = gl / Cm
```

This would be much nicer as:

```
_gl_over_Cm = gl / Cm
```

We could have fallbacks in case those generated names are not unique.

Maybe we should also have commented out the original expression? That requires some deeper changes to codegen but perhaps it's worth it actually? Readability of the generated code is important. (This could be a separate issue/pull request though.)

thesamovar:
Updating sympy fixed the issue with the IF-curve and the doctest. I guess we should bump the requirements?

thesamovar:
OK on reflection I think the two problems I mentioned (readability and boolean variables) are separate: I'll create new issues for them.

mstimberg:
> Updating sympy fixed the issue with the IF-curve and the doctest. I guess we should bump the requirements?

Maybe that's an easier solution than trying to fix it for sympy 0.7.5... Currently, we are only checking the version numbers at installation time, though, we should have a check for the requirements at runtime. I guess we can use some setuptools stuff for this?

mstimberg:
> OK on reflection I think the two problems I mentioned (readability and boolean variables) are separate: I'll create new issues for them.

Agreed, thanks.

thesamovar:
I think at this point it's ok for us to just keep bumping the requirements: we're still at beta.

mstimberg:
> I think at this point it's ok for us to just keep bumping the requirements: we're still at beta.

Ok, I agree. Actually, we could also consider dropping 2.6 support, we carry around a few workarounds for it.

thesamovar:
I would be OK with dropping support for 2.6, but maybe let's check with Romain about that? How bad are the workarounds?

thesamovar:
OK tests all pass now, so I think we just need a fix for the stateless functions and maybe a test for that, and we're ready to go.

mstimberg:
> I would be OK with dropping support for 2.6, but maybe let's check with Romain about that? How bad are the workarounds?

The workarounds aren't very bad, the only thing that comes to my mind right now is our own version of `OrderedDict` for Python 2.6. But I think by now (or by the release of 2.0), I think we'd be fairly safe to drop 2.6 support, Mac OS X comes with 2.7 since 2011 and also the only "long-term-stable" Ubuntu version that still came with 2.6 is no longer officially supported (except the server version, but even that will stop May this year).

thesamovar:
How about we ask on the Brian list?

mstimberg:
> How about we ask on the Brian list?

Good idea! But we should directly ask: is anyone still using Python 2.6 (and if yes, why, i.e. on what platform?), if not we will drop support to make sure that the affected users do respond. Do you want to do it? Only on the brian-development list would be fine, I guess.

Vaguely related: I saw some FOOS projects started using the https://gitter.im service as a chat room that is nicely integrated with github (and allows for markdown formatting, etc.), e.g. https://gitter.im/sympy/csympy Maybe this would be useful for some discussions that require a bit of back-and-forth? The mailing list is not always ideal for that. On the other hand, you can only keep a long history of the conversations if you pay for it...

thesamovar:
I sent emails to both lists: better to get a wider range of opinions I think.

I have used gitter.im on a different project. It's quite nice, but I hadn't realised you didn't get archives of the conversation. I'd say that's a problem for us: I often refer back to earlier discussions.

thesamovar:
We could pay for it, but even at $2/user/month I'm not sure it's worth it...

mstimberg:
> We could pay for it, but even at $2/user/month I'm not sure it's worth it...

I'm not sure either, not even sure what "/user" actually means in this case.

Anyway, maybe this issue is not the best place to discuss this ;)

mstimberg:
I fixed the issue with the `rand` function and added a simple test for the issue (that failed before the fix). If all works for you, I guess we can merge it?

thesamovar:
Everything passes, code looks good, merging now.

Issue #400, opened 2015-02-09
loop-invariant optimisation can turn integer expressions into floats


mstimberg:
When the loop-invariant optimisation pulls out constant expressions, it uses the standard dtype for those constants (i.e. a float). This will lead to bugs (in C) if the code uses an operation that is only defined for integers such as the modulo operator:
```Python
from brian2 import *
prefs.codegen.target = 'weave'
G = NeuronGroup(4, '''a : integer''',
                threshold='True',
                reset='a += i%(2*N)'  # 2*N will be pulled out
                )
run(defaultclock.dt)
```

Will lead to
```
[...]CompileError: error: Command "c++ -pthread -fno-strict-aliasing -DNDEBUG -g -fwrapv -O2 -fPIC -I/usr/include -I/home/marcel/.local/lib/python2.7/site-packages/scipy/weave -I/home/marcel/.local/lib/python2.7/site-packages/scipy/weave/scxx -I/home/marcel/.local/lib/python2.7/site-packages/numpy/core/include -I/usr/include/python2.7 -c /home/marcel/.cache/scipy/python27_compiled/sc_8e5301c5898c4a29b6a1650eadee6f641.cpp -o /tmp/scipy-marcel-sHtBfu/python27_intermediate/compiler_ede35212078c0beff706e86697d62e7e/home/marcel/.cache/scipy/python27_compiled/sc_8e5301c5898c4a29b6a1650eadee6f641.o -w -O3" failed with exit status 1
/home/marcel/.cache/scipy/python27_compiled/sc_8e5301c5898c4a29b6a1650eadee6f641.cpp: In function ‘PyObject* compiled_func(PyObject*, PyObject*)’:
/home/marcel/.cache/scipy/python27_compiled/sc_8e5301c5898c4a29b6a1650eadee6f641.cpp:754:26: error: invalid operands of types ‘const int’ and ‘const double’ to binary ‘operator%’
                 a += i % _lio_const_1;
                          ^
/home/marcel/.cache/scipy/python27_compiled/sc_8e5301c5898c4a29b6a1650eadee6f641.cpp: In function ‘PyObject* compiled_func(PyObject*, PyObject*)’:
/home/marcel/.cache/scipy/python27_compiled/sc_8e5301c5898c4a29b6a1650eadee6f641.cpp:754:26: error: invalid operands of types ‘const int’ and ‘const double’ to binary ‘operator%’
                 a += i % _lio_const_1;
```

thesamovar:
[...]

PR #401, opened 2015-02-10
Fix lio

mstimberg:
Applies the short-term fix for #400, i.e. will not pull out expressions that contain an integer or a boolean.


thesamovar:
OK I like the code and the tests pass on Windows so please go ahead and merge once the Travis build OK's it.

mstimberg:
Thanks, will do.

PR #403, opened 2015-02-11
Improve `Variable` objects

mstimberg:
This makes sure that all `Variable` objects, including `Constant` and `AttributeVariable` have an `owner` attribute, making it easier to check for the owner of variables in a general way. Also adds a `variables_by_owner` utility function that returns all variables that "belong" to a group:
```Python
from brian2 import *
from brian2.core.variables import variables_by_owner

S = Synapses(...)

# get all the synaptic variables
syn_vars = variables_by_owner(S.variables, S)
# get all the pre-synaptic variables
pre_vars = variables_by_owner(S.variables, S.source)
```

From my side: ready-to-merge

### DIFF
--- a/brian2/core/variables.py
+++ b/brian2/core/variables.py
@@ -85,6 +85,12 @@ def get_dtype_str(val):
     return 'unknown[%s, %s]' % (str(val), val.__class__.__name__)
 
 
+def variables_by_owner(variables, owner):
+    owner_name = getattr(owner, 'name', None)
+    return dict([(varname, var) for varname, var in variables.iteritems()
+                 if getattr(var.owner, 'name', None) is owner_name])
+
+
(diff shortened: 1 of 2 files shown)


thesamovar:
All good and tests pass. Shame about the drop of 0.01% in coverage though. ;)

Merging now.

thesamovar:
Can we also merge this into device_correctness so that @tnowotny can use it in brian2genn?

mstimberg:
We already merged device_correctness into master, so I guess that's what he should use now.

thesamovar:
I was going to say yes but the loop-invariant stuff is a problem but I'm just about to merge the fix for that into master too, so not necessary. :)

mstimberg:
Great :)

PR #404, opened 2015-02-12
Make the "loop-invariant optimisations" optional via a preference

mstimberg:
As the title says... I briefly looked into making this an attribute of the `Device`, but we apply the optimisations in `make_statements` which is an early function that does not yet know about code generation targets, devices, etc. so it seemed wrong to introduce such a dependence there. The preference is quite straightforward and it can also be a useful future workaround if we encounter further issues with the mechanism, i.e. we can just tell users to switch it off until we fix the problem.


thesamovar:
Jolly good. :)

Issue #379, opened 2014-12-11
Automatically chose a codegen target (and tell the user about it)


mstimberg:
Currently, we default to the `numpy` codegen target. Some users might therefore not benefit from the acceleration of `weave`/`cython` even though they could.

Instead, we could try to chose (at import time) a code generation target for the user if they did not declare one explicitly. We'd emit an "info" (which should soon be the default log level) about it, telling the user which codegen target was chosen.

thesamovar:
This seems like a good idea to me. One issue is whether or not there is a fast way to check if a codegen target is available. Seems like the only way to be sure (for Windows users) would be to try compiling something and see if it worked: but this might add a considerable amount of time to the initial import time?

mstimberg:
Maybe we could do it like this:
* a possible value for `prefs.codegen.target` is `auto`
* Whenever code is generated for the first time (i.e. not at import, but at the first `before_run`) and the code generation target is set to `auto`, we determine the best target (`weave` > `cython` ?). This is stored somewhere in the codegen module, so it is only done once per process
* We log on the info level that we are automatically determining a codegen target and tell the user to set it explicitly if they don't want that (this kind of requires that we increase the default log level to info)

This way, if it adds a significant delay to the process the user will know why and can do something about it.

thesamovar:
Actually how about we just write it to the user preference file the first time we determine it? We could have two prefs: ``prefs.codegen.target = 'auto'`` and ``prefs.codegen.auto_target = ...`` which gets filled in the first time it is run with target=auto? We still warn each time you run with auto but it doesn't take any longer to do so.

mstimberg:
It depends a bit on how long this determination takes. If it is just 2 seconds or so, I'd say if it is fine if it is repeated with every new script you run. It has the advantage that you don't have to dig around and change something in a configuration file just because you installed cython after you first ran Brian or something like that. Actually, putting it in the user preferences at the first run would be equivalent to a determination at installation time (if the user installs Brian in the first place...), wouldn't it?

thesamovar:
OK true: let's see how long it takes.

PR #408, opened 2015-02-17
Auto codegen target

mstimberg:
This implements the mechanism we discussed in #379. I implemented the determination whether weave/cython works with a little code snippet that will be compiled. This can indeed take a considerable time (on the order of 3-5s if done from scratch without any caches) but I don't think this is a problem because:

1. This is only done if `'auto'` is actually used, i.e. it doesn't take any time if you explicitly set a codegen target
2. In most cases where a target is not available (e.g. Python 3 for weave, or no Cython installation), we don't even get that far
3. Most importantly: weave and Cython cache their codes on disk automatically already, i.e. when this has been executed once on a machine it is basically instantaneous the next time

I didn't realize this before but point 3 does really take the burden of doing some across-scripts caching ourselves.

I'm happy with this solution, let me know what you think.

### DIFF
--- a/brian2/codegen/_prefs.py
+++ b/brian2/codegen/_prefs.py
@@ -10,19 +10,21 @@ prefs.register_preferences(
     'codegen',
     'Code generation preferences',
     target=BrianPreference(
-        default='numpy',
+        default='auto',
         docs='''
         Default target for code generation.
         
         Can be a string, in which case it should be one of:
         
-        * ``'numpy'`` by default because this works on all platforms, but may not
-          be maximally efficient.
+        * ``'auto'`` the default, automatically chose the best code generation
+          target available.
         * `'weave'` uses ``scipy.weave`` to generate and compile C++ code,
           should work anywhere where ``gcc`` is installed and available at the
           command line.
         * ``'cython'``, uses the Cython package to generate C++ code. Needs a
           working installation of Cython and a C++ compiler.
+        * ``'numpy'`` works on all platforms and doesn't need a C compiler but
+          is often less efficient.
         
         Or it can be a ``CodeObject`` class.
         ''',
@@ -36,7 +38,7 @@ prefs.register_preferences(
         default numpy target, because the overhead of compiling code is not
         worth the speed gain for simple expressions.
 
-        Accepts the same arguments as `codegen.target`.
+        Accepts the same arguments as `codegen.target`, except for ``'auto'``
         ''',
         validator=lambda target: isinstance(target, basestring) or issubclass(target, CodeObject),
         ),
--- a/brian2/codegen/codeobject.py
+++ b/brian2/codegen/codeobject.py
@@ -51,6 +51,15 @@ class CodeObject(Nameable):
         self.template_name = template_name
         self.template_source = template_source
 
+    @classmethod
+    def is_available(cls):
+        '''
+        Whether this target for code generation is available. Should use a
+        minimal example to check whether code generation works in general.
+        '''
+        raise NotImplementedError("CodeObject class %s is missing an "
+                                  "'is_available' method." % (cls.__name__))
+
     def update_namespace(self):
         '''
         Update the namespace for this timestep. Should only deal with variables
--- a/brian2/codegen/runtime/cython_rt/cython_rt.py
+++ b/brian2/codegen/runtime/cython_rt/cython_rt.py
@@ -7,6 +7,7 @@ from brian2.core.variables import (DynamicArrayVariable, ArrayVariable,
                                    AttributeVariable, AuxiliaryVariable,
                                    Subexpression)
 from brian2.core.preferences import prefs, BrianPreference
+from brian2.utils.logger import get_logger
 
 from ..numpy_rt import NumpyCodeObject
 from ...templates import Templater
@@ -17,6 +18,8 @@ from .extension_manager import cython_extension_manager
 
 __all__ = ['CythonCodeObject']
 
+
+logger = get_logger(__name__)
 
 # Preferences
 prefs.register_preferences(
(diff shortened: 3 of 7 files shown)


thesamovar:
The situation where it's slow is if you have weave installed (i.e. everyone on Python 2.x) but an incorrectly configured compiler (so that it tries and fails each time). Unfortunately I don't think this is an uncommon situation. 

Could we save the results of the check for just 24h? Like that you only lose a few seconds per day rather than every run? (Or a shorter period if you prefer.)

mstimberg:
> The situation where it's slow is if you have weave installed (i.e. everyone on Python 2.x) but an incorrectly configured compiler (so that it tries and fails each time). Unfortunately I don't think this is an uncommon situation.

What kind of configuration error do you have in mind? Shouldn't a missing compiler and other problems fail very quickly?

thesamovar:
It can be surprisingly slow if it manages to call the compiler but the paths are set up incorrectly so that compilation fails. This has happened to me sufficiently often that I'd consider it something worth taking seriously. 

Another suggestion: if checking for the compiler takes more than say 0.5s then we raise a warning and encourage the user to either sort out the problem or set the preference explicitly.

mstimberg:
> It can be surprisingly slow if it manages to call the compiler but the paths are set up incorrectly so that compilation fails. This has happened to me sufficiently often that I'd consider it something worth taking seriously.

I see. I'm still hesitant to introduce something based on some arbitrary time for caching or warning... How about always raising a warning when compilation fails? I.e., not in the cases where there's no weave (i.e., Python 3) or cython, those will already fail to import earlier and will not get to the compilation check. I think we do want to push users towards using weave/cython in general, having an explicit warning if it doesn't work for some non-trivial reason sounds reasonable to me.

thesamovar:
Sounds good. Or actually we could just always warn when auto is used?

mstimberg:
I'd say definitely no for weave/cython (similar to what you said about exact integration in the INFO log discussion: the user couldn't have made a better choice) but possibly for numpy (i.e. explicitly calling numpy the "fallback" that is not optimal performance-wise).

One strong reason for not always displaying a warning is that you'd get a warning for every Brian example -- we should try to keep them free of warnings to not scare of users :)

thesamovar:
You're right let's only have an error if there is a compilation failure. I think this will be the case for most Windows users though, as weave will be installed but no compiler (typically).

mstimberg:
Yeah, I wonder whether it's worth having a special case for that, i.e. only log a message at a lower level in that case. On the other hand, I think it would be good to push users a little bit towards installing a compiler, especially so that they can benefit from the standalone mode as well. We should give some recommendations/instructions in the docs, though.

thesamovar:
I wouldn't special case that: I would tell those users to either specify numpy or install a compiler properly. If they install from the recommendations in the docs they will have one.

mstimberg:
> If they install from the recommendations in the docs they will have one.

ah, sorry, I looked at the documentation for 2.0beta, not the latest from master. A bit besides the point of the current issue, but is the installation of the VisualStudio stuff really the simplest way to procede, given that you need to patch it for 64 Bit Python? Wouldn't cygwin be easier possibly? Did we ever test using mingw32, actually?

thesamovar:
Ah yeah I added that documentation recently.

Mingw32 produces 32 bit code so it wouldn't work. There is a mingw64 but as far as I know it's still beta/buggy. Unfortunately, 64 bit windows is just a pain and I don't think there's any way round it.

mstimberg:
> Unfortunately, 64 bit windows is just a pain and I don't think there's any way round it.

Ok, maybe we can rewrite the docs a bit and more clearly point out that we recommend using a 32 Bit Python for Windows (even on 64 Bit Windows) and that this isn't really a disadvantage apart from not being able to use more than 4GB for a simulation.

thesamovar:
It does also stop them from using >4GB (or 3GB on windows) for other tasks though, e.g. for data analysis. That's why I switched from 32 to 64 bit. I'd rather say just that using 32 bit Python on Win64 is easier and possible (since people might not know that), and leave it to them to decide which they prefer. We should point out the 3GB limit though.

mstimberg:
Ok, let's keep this in mind when we do the next release.

thesamovar:
OK so we're agreed for the changes to this branch then? I think it's just a matter of raising the warning if compilation fails.

mstimberg:
Just pushed such a change, including a warning when we are falling back to numpy. In the extreme case (weave and cython available but not working) this would lead to three warnings but I think it is good to have the detailed information about why compilation failed as well.

The only situation that I think is not optimally handled is the combination of a non-working weave installation with a working Cython installation which will now raise a warning even though it is fine.

Hmm, maybe there's another situation: if you are on Python 3 and you don't have Cython installed, you get a warning about the numpy fallback but without any indication why you can't have compiled code. But we should probably change our `setup.py` so that we require weave for Python 2 and Cython for Python 3, anyway -- now that we have weave as an independent package, that's easier than before. Or always require Cython? But then again, installing cython on Windows 64 Bit might not always work...?

mstimberg:
Oh, and: from my side it's ok to merge it as it is now.

thesamovar:
Weave and Cython shouldn't be requirements, right? They are optional extras. Is there any way to specify that in the setup script? i.e. not required but very beneficial to have? I guess not.

I don't think the non-working weave and working Cython should ever come up in practice, so that's OK.

mstimberg:
> Weave and Cython shouldn't be requirements, right? They are optional extras. Is there any way to specify that in the setup script? i.e. not required but very beneficial to have? I guess not.

We kind of require weave by requiring scipy, but I agree that they shouldn't be requirements in general -- it would be different if they were trivially installable (i.e., pure Python), but obviously they are not.

Unfortunately there are no suggested or recommended requirements in `setup.py`, you can define "extras" but they require manual selection by the user, e.g. we already have "docs" and "tests", if you install brian with:
```
pip install brian2[docs,tests]
```
you'll also install sphinx and nosetests alongside. I don't think that anyone ever used this feature, though :) We could add a "compiled" extra and recommend that people use `pip install brian2[compiled]`, maybe?

thesamovar:
If someone can't install weave or cython for some reason, will installation of Brian fail or will it still try to install everything else? I'd much rather avoid having to do ``pip install brian2[compiled]`` if possible.

thesamovar:
Tests pass and code looks good so I'm going to merge this and close/delete. If you think we need to discuss the installation stuff further open a new issue?

Issue #406, opened 2015-02-12
Raise a warning when using postsynaptic variables without a suffix (if there's a pre-synaptic variable of the same name)


mstimberg:
See the discussion in #348. Brief summary: if synaptic code uses a post-synaptic variable without the `_post` suffix and there is a pre-synaptic variable of the same name (e.g. when using `pre='v += w'`), emit an info-level log message telling the user that the variable is interpreted as a postsynaptic variable. This warning should come that the point of use, e.g. during the creation of the `Synapses` object for a `pre` code. See issue #405 for some refactoring that should be done before.

thesamovar:
This should be closed now, right?

mstimberg:
> This should be closed now, right?

This is still open, the following code does not raise a warning, but to be less ambiguous the user should have used `pre='v_post += 1'`:
```Python
source = NeuronGroup(10, 'dv/dt = -v/(10*ms) : 1', threshold='v > 1', reset='v = 0')
target = NeuronGroup(10, 'v: 1')
syn = Synapses(sources, targets, pre='v += 1', connect=True)
```

But I am a) not 100% convinced that we really need a warning here and b) there is some non-trivial work necessary to solve it correctly, since we cannot currently raise the warning at the time of the Synapses creation (see #405). I'll therefore remove this issue from the 2.0 milestone.

Issue #348, opened 2014-10-28
Raise errors for ambiguous synapse variables


thesamovar:
Working on @brian-team/brian2genn with @tnowotny we were thinking about the way pre/post/synaptic variables are referenced in synapse code. I think the current rule is that an unsuffixed name must refer unambiguously to either a postsynaptic neuron variable or a synaptic variable, with an error raised if there is a clash, but no error raised if there is a clash with a presynaptic variable. Perhaps we ought to be more rigorous here and raise an error if there is any clash and force the user to specify _pre, _post or _syn in this case?

mstimberg:
This is something about which we should get feedback from @romainbrette, I'll ask him later. I was initially a bit worried that we used the "postsynaptic variable without suffix" syntax in the Frontiers paper but it turns out we only used the `_post` syntax there, so this is not a problem. If we decide to go with the unambiguous syntax (no suffix --> synaptic variable): should we have it raise an error immediately or should we do a soft transition with a `DeprecationWarning`? We used statements such as `v+=w` in quite a few examples... On the other hand, with the error message telling you what to do to make it work again, it shouldn't really be an issue.

thesamovar:
I think we discussed this when I was in Paris and we decided to go ahead? I would say just make it an error and don't worry about ``DeprecationWarning`` but I'm not against having it if you think it's worthwhile.

mstimberg:
We discussed this today and @romainbrette wasn't too fond of getting rid of the "`v` instead of `v_post`" shorthand. The compromise solution we could all live with was: raise an error if there is a clash and force the user to rename the synaptic variable but continue to allow to not use the `_post` suffix if there's no synaptic variable with the same name.
One remaining question: should we only do this for a clash between synaptic and postsynaptic variable or also for a clash between synaptic and presynaptic variable? In the latter case there's not really room for ambiguity, you can never just write the name of a presynaptic variable to refer to it. I think I'd therefore prefer to only raise an error for the name conflict synaptic-postsynaptic.

thesamovar:
Strange: I thought we'd already talked about it and decided differently. I would say raise the error if there is any ambiguity at all (i.e. between pre, synaptic or post; including a clash between pre and post). We know that the user cannot write presynaptic variables without putting ``_pre`` but they don't necessarily know that.

mstimberg:
I don't quite remember what/whether we decided back then but I'm fine with raising an error in all cases. Should we go ahead with this then?

[...]

PR #409, opened 2015-02-18
Raise an error if the name of a synaptic variable clashes with a pre- or postsynaptic variable name

mstimberg:
Does what the title says. As discussed in #348, this only deals with the error case. Warnings about clashes between pre- and postsynaptic variables are still up for discussion (see #406)


thesamovar:
Tests pass, code looks good. Merging.

PR #410, opened 2015-02-23
Spatialneuron

mstimberg:
Two small bug fixes for `SpatialNeuron`, 1a768d6 fixes a quite severe problem, basically all simulations for complex morphologies did not work since a while ago (a regression introduced after the beta release). It should be safe to merge and is spatialneuron-specific, anyway. Let's keep the branch alive for future changes.


thesamovar:
I get several test failures along these lines:

```
ERROR: brian2.tests.test_spatialneuron.test_construction
  File "C:\Python27\lib\site-packages\nose\case.py", line 197, in runTest
    self.test(*self.arg)
  File "C:\Users\Dan\programming\brian2\brian2\tests\test_spatialneuron.py", line 48, in test_construction
    assert len(np.unique(neuron.diffusion_state_updater._morph_i)) == len(neuron.diffusion_state_updater._morph_i)
  File "C:\Python27\lib\site-packages\numpy\lib\arraysetops.py", line 166, in unique
    items = sorted(set(ar))
TypeError: unhashable type: 'numpy.ndarray'
```

I'm surprised that doesn't affect you too unless this is a numpy bug that has been fixed in a more recent version?

mstimberg:
> I'm surprised that doesn't affect you too unless this is a numpy bug that has been fixed in a more recent version?

Hmm, possibly. I'm using 1.9.1., the same version that travis uses for testing. Does the error stay if you replace
```Python
len(np.unique(neuron.diffusion_state_updater._morph_i))
```
by
```Python
len(np.unique(neuron.diffusion_state_updater._morph_i[:]))
```
or
```Python
len(np.unique(neuron.diffusion_state_updater._morph_i[:].flat))
```
?

thesamovar:
Yep, adding ``[:]`` did the trick. I committed the change and the tests pass on my machine now. Do you want to go ahead and merge once the tests pass on Travis? (I assume they will.)

mstimberg:
Ok, I will do that. For reference, what numpy version are you using? Since it disappeared with the addition of the `[:]`, there might be something for us to fix in `VariableView` for that numpy version. Oh, and we should probably add a test run with the minimum version requirements on travis.

thesamovar:
Ah yes, that's a very good idea.

My numpy version is 1.8.

mstimberg:
I added a test run for older requirement versions and it turned out that we were already using some features that were not supported by numpy versions < 1.8 but we only required numpy >= 1.4.1... Instead of trying to work around those issues, I just bumped up the requirement, let's see whether the tests pass now.

mstimberg:
After some fiddling with the exact version numbers, everything passes now. I also removed the warning when you don't have matplotlib installed, instead you simply get an error when you try to plot.

I'll merge.

PR #414, opened 2015-02-25
Standalone tests

mstimberg:
This introduces a new category of tests, "standalone-compatible" tests. In addition to tests marked as `@attr('cpp_standalone')` (renamed from the previous `standalone`), tests can be marked as `@attr('standalone-compatible')`. They are then executed with the new `CPPStandaloneSimpleDevice` which directly calls `build` after a run. Unfortunately, nose does not offer a mechanism to simply call a teardown function after every single test, so "standalone-compatible" tests also have to use `with_setup(teardown=restore_device)`. From my side this is ready to merge.

BTW: The `cpp_standalone_simple` device could also be useful for users in general (you don't have to add a `device.build` call for simple simulations), I wonder whether we should advertise this use case.

### DIFF
--- a/brian2/devices/cpp_standalone/device.py
+++ b/brian2/devices/cpp_standalone/device.py
@@ -8,6 +8,7 @@ import inspect
 import platform
 from collections import defaultdict
 import numbers
+import tempfile
 
 import numpy as np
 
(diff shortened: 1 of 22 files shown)


thesamovar:
Can it be made a bit more general so that we can adapt it to other devices once they're available without having to make further changes to Brian? e.g. GeNN and others.

mstimberg:
Done! The `test_standalone` keyword of `brian2.test` now expects the name of a device, e.g. `'cpp_standalone'` and assumes that there is a corresponding device named `cpp_standalone_simple`. You could do the same for `genn_device` now. One missing thing: it will only look for tests in the brian2 tests directory for now, don't know what the best approach for that would be, but maybe deal with this issue later?

thesamovar:
Just to make sure I understand: if you mark a test as standalone-compatible it will now be tested if you specify ``test_standalone='genn'`` or something like that?

Yes, I'm happy to just look for the tests in the brian2 tests directory for now. People can handle additional tests themselves. It's more about making the existing tests available to other devices.

mstimberg:
> Just to make sure I understand: if you mark a test as standalone-compatible it will now be tested if you specify test_standalone='genn' or something like that?

Exactly: it will assume that there exists a device that works in the same way as `CPPStandaloneSimpleDevice` and that it is named `'genn_simple'`. It will also execute all tests written *specifically* for GeNN and having the `genn` test attribute, but since it looks only in the Brian2 folder, it will obviously not find any such test.

thesamovar:
Great. OK, and the tests pass here so I'll go ahead and merge.

PR #415, opened 2015-02-26
Brian examples

mstimberg:
This branch includes a number of examples converted from Brian1 and a few small bug fixes, mostly related to error messages and warnings. I think it should be good to merge, we can keep the branch around to add more examples in the future.


thesamovar:
Looks good. Tests pass here but I'll also run ``run_examples.py`` which will take a while so I'll merge later when that completes.

In general, I think we can add new examples direct to master. Even if they have a bug they won't break the tests or users code.

Slightly unrelated. We could modify ``run_examples.py`` so that it can run on devices. We just need to automatically insert a ``set_device`` after the ``from brian2 import *`` and a ``device.build`` after the last ``run`` statement. Could be useful for device debugging.

thesamovar:
Some comments as I go through them:

* Some figures look bunched up, we could put in a ``tight_layout()`` before the ``show()`` for examples that use ``subplot`` maybe?
* Touboul_Brette example uses a huge amount of memory and actually runs out on my 32 bit machine. I think we should make sure this runs with limited memory. Maybe reduce the number of neurons or decrease the time a bit?
* ``synapses/nonlinear.py`` has an error that we might want to think about.

Will maybe come back and add more as it continues to run. Also I committed a small change to one script. I put [ci_skip] in the commit message but it should have been [ci skip]. Oops! Anyway, Travis had given it the green light before that so it should be ok.

thesamovar:
More details.

For ``run_examples`` it raises an error for the examples in ``frompapers/Brette_2012`` because it tries to import from ``params`` and ``run_examples`` doesn't run from that directory. However, they all run fine if you do it yourself, so this just needs a change to ``run_examples``.

For ``synapses/nonlinear`` we get the following error: ``ValueError: The pre-synaptic variable 'g' has the same name as a synaptic variable, rename the synaptic variable (for example to 'g_syn') to avoid confusion``. I think maybe we have a conflict here between the demands of the ``(summed)`` mechanism and the ambiguity issue. Thoughts on this?

``Touboul_Brette`` has the ``MemoryError`` I mentioned before.

mstimberg:
> For synapses/nonlinear we get the following error: ValueError: The pre-synaptic variable 'g' has the same name as a synaptic variable, rename the synaptic variable (for example to 'g_syn') to avoid confusion. I think maybe we have a conflict here between the demands of the (summed) mechanism and the ambiguity issue.

This is actually an error in line with our intentions, the example defined `dg/dt : 1` in the `Synapses` object and `g: 1` in the `NeuronGroup`, I changed the definition in `Synapses` to `g_syn` (e.g. it now reads `g_post = g_syn : 1  (summed)` and it works fine.

> For run_examples it raises an error for the examples in frompapers/Brette_2012 because it tries to import from params and run_examples doesn't run from that directory. However, they all run fine if you do it yourself, so this just needs a change to run_examples.

When you run the example on its own, Python will automatically add its directory to the path but in `run_examples` we use `execfile` which doesn't do that. We could either mess with the `PYTHONPATH` environment variable before calling `execfile` or add something like the following to all `Fig...` files in `Brette_2012`:
```Python
import sys, os
sys.path.append(os.path.dirname(os.path.realpath(__file__)))
```
Not sure which variant is less ugly...

> Touboul_Brette has the MemoryError I mentioned before.

I reduced the runtime and the number of runs, it should be fine now. The problem with this example is that it used `StateSpikeMonitor` which only records `w` at the time of spikes and in brian2 it has to record the variable at every timestep.

thesamovar:
OK I can confirm that fixes the problems.

For ``run_examples`` we generate and run code already so we could add the path to it in the generated code. Is that what you mean? I think that's fine. We shouldn't add it directly to the original code itself though: don't want users having to look at that.

Shall we do that before merging or just go ahead now? All tests pass on my machine so if you want to merge go for it.

mstimberg:
> For run_examples we generate and run code already so we could add the path to it in the generated code. Is that what you mean?

I didn't, but you are of course right :) This is implemented now. Could you maybe try it a last time since the path bit might potentially be OS-specific? To save some time, add `and 'Brette_2012' in name` to the condition in line 112 of `run_examples.py` :)

thesamovar:
I had to make a change: can you check it works on yours? I'm surprised it did before actually because you set the path using ``__file__`` which isn't related to the script being run I think.

Also I removed the printing of stdout and stderr because it was making the display ugly and it still captures the tracebacks. What do you think?

Finally: I forgot to include [ci skip] again! ;)

I'm happy to merge if you are.

thesamovar:
Oh also, I would say let's delete the branch after the merge and add more examples directly to master if we want to, because I generally don't like having too many branches lying around unless they're being actively worked on.

mstimberg:
> I had to make a change: can you check it works on yours? I'm surprised it did before actually because you set the path using __file__ which isn't related to the script being run I think.

Well, the first line of our generated code says `__file__ == '{fname}'`, so I'm surprised that your change makes a difference...? Anyway, it works for me like this as well.

> Also I removed the printing of stdout and stderr because it was making the display ugly and it still captures the tracebacks. What do you think?

stdout was captured by nosetest's capture plugin so you'll only see it for examples that failed (same as for our tests) but the capture mechanism does not work for the log messages (i.e., warnings) due to our complicated setup. I now simply redirected stderr to stdout, so you'll see both iff the example failed.

> Oh also, I would say let's delete the branch after the merge and add more examples directly to master if we want to, because I generally don't like having too many branches lying around unless they're being actively worked on.

Sounds reasonable, let's only do a new branch when we add a real feature, e.g. running examples on other devices.

thesamovar:
Oh, weird that setting ``__file__`` wasn't working on mine. Oh well, as you say: it works on both now. :)

Your changes work on mine too, so I'm going to go ahead and merge and delete the branch now.

PR #418, opened 2015-03-02
Change the system that decides which state updater algorithm is used

### LINKED ISSUE #351: Improve the mechanism that choses a numerical integration algorithm
Currently, if no integration algorithm is chosen, the list of state updaters is tried one after the other until a suitable updater is found. There are several issues with the current system:
* The exact integrators might take a long time until they realize they can't integrate the equations
* One might want to use different defaults for "normal" neurons and multicompartment neurons
* conditionally linear equations will use exponential Euler integration therefore users might think Brian 2 is slower than Brian 1 where the simpler forward Euler algorithm is used by default (but exponential Euler is also particularly slow in numpy or weave without `--fast-math` currently because we don't optimize the generated code, a term like `exp(-dt/tau)` will appear quite often)

We still want to keep the system simple and extensible but maybe something where we don't just order algorithms in a list but instead store some information about their accuracy (and assuming that this is inversely related to their run time)? The user could then chose between a "performance-first" or an "accuracy-first" strategy.

mstimberg:
Now, each object defines its own list of algorithms that are tried in order. Closes #351 

Well, I think that's all we wanted to do. It's straightforward and easy to explain and it also avoids the use of the fragile `'independent'` state updater without a user explicitly asking for it.

### DIFF
--- a/brian2/groups/neurongroup.py
+++ b/brian2/groups/neurongroup.py
@@ -310,7 +310,8 @@ class NeuronGroup(Group, SpikeSource):
     '''
     add_to_magic_network = True
 
-    def __init__(self, N, model, method=None,
+    def __init__(self, N, model,
+                 method=('linear', 'euler', 'milstein'),
                  threshold=None,
                  reset=None,
                  refractory=False,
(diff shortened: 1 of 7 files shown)


thesamovar:
Tests pass here and code looks good. Happy for you to merge once travis returns the OK.

mstimberg:
Travis is happy, merging now.

PR #419, opened 2015-03-02
Check dependency versions at runtime

mstimberg:
Does what the title says. I did not find a simple straightforward way to share the dependency information with `setup.py`, but I think given that we change our dependencies quite rarely, it isn't really worth the effort of coming up with an elaborate system.

### DIFF
--- a/brian2/__init__.py
+++ b/brian2/__init__.py
@@ -3,6 +3,7 @@ Brian 2.0
 '''
 # Check basic dependencies
 import sys
+from distutils.version import StrictVersion
 missing = []
 try:
     import numpy
@@ -59,3 +60,27 @@ __version__ = '2.0beta'
 __release_date__ = '2014-11-03'
 
 from brian2.only import *
+
+# Check for outdated dependency versions
+def _check_dependency_version(name, version):
+    from core.preferences import prefs
+    from utils.logger import get_logger
+    logger = get_logger(__name__)
+
+    module = sys.modules[name]
+    if not isinstance(module.__version__, basestring):  # mocked module
+        return
+    if not StrictVersion(module.__version__) >= StrictVersion(version):
+        message = '%s is outdated (got version %s, need version %s)' % (name,
+                                                                        module.__version__,
+                                                                        version)
+        if prefs.core.outdated_dependency_error:
+            raise ImportError(message)
+        else:
+
+            logger.warn(message, 'outdated_dependency')
+
+for name, version in [('numpy', '1.8.0'),
+                      ('scipy', '0.13.3'),
+                      ('sympy', '0.7.6')]:
+    _check_dependency_version(name, version)
--- a/brian2/core/core_preferences.py
+++ b/brian2/core/core_preferences.py
@@ -21,8 +21,15 @@ prefs.register_preferences('core', 'Core Brian preferences',
     default_integer_dtype=BrianPreference(
         default=int32,
         docs='''
-        Default dtype for all arrays of integer scalars'
+        Default dtype for all arrays of integer scalars.
         ''',
         representor=dtype_repr,
+        ),
+    outdated_dependency_error=BrianPreference(
+        default=True,
+        docs='''
+        Whether to raise an error for outdated dependencies (``True``) or just
+        a warning (``False``).
+        '''
         )
     )
--- a/docs_sphinx/user/import.rst
+++ b/docs_sphinx/user/import.rst
@@ -37,3 +37,14 @@ Note that it is safe to use e.g. ``np.sin`` and ``numpy.sin`` after a
 ``from brian2 import *``.
 
 .. _matplotlib: http://matplotlib.org/
+
+Dependency checks
+-----------------
+Brian will check the dependency versions during import and raise an error for
+an outdated dependency. An outdated dependency does not necessarily mean that
+Brian cannot be run with it, it only means that Brian is untested on that
+version. If you want to force Brian to run despite the outdated dependency, set
+the `core.outdated_dependency_error` preference to ``False``. Note that this
+cannot be done in a script, since you do not have access to the preferences
+before importing `brian2`. See :doc:`../advanced/preferences` for instructions
+how to set preferences in a file.


thesamovar:
Scipy dependency could be lower. I have 0.12.0 and it works fine but the dependency checker is asking for 0.13.3.

mstimberg:
> Scipy dependency could be lower. I have 0.12.0 and it works fine but the dependency checker is asking for 0.13.3.

Yes, there's a reason for that: 0.13.3 is the lowest version (compatible with numpy >= 0.18) that is available via anaconda and that we can therefore test on travis. I thought that it would be good to not allow versions that we don't actually test -- what do you think? In the mid-term, we should get rid of scipy anyway (it is only used for `SpatialNeuron` currently) and only use the `weave` package.

thesamovar:
Can we have a warning rather than an error maybe? It seems wrong to arbitrarily restrict it from working on configurations where it does actually work.

mstimberg:
Hmm, but still have it in `setup.py` as a requirement with the higher version? It would also be odd to not allow the installation (or automatically upgrade scipy) when using `python setup.py install` (or `develop`) with an older scipy version but only raising a warning when actually using it? And using `0.12` is also quite arbitrary, I guess older versions will work as well?

thesamovar:
I think it's OK if setup.py has a requirement of the recommended version but we don't actually raise an error in any case, only a warning. I don't see any need to make it an error.

mstimberg:
> I think it's OK if setup.py has a requirement of the recommended version but we don't actually raise an error in any case, only a warning.

setup.py will always raise an error if the requirement is not met, we cannot change that.

thesamovar:
That's fine: but  we don't have to raise our own error.

thesamovar:
For example, people might install by downloading the source, or using pip install with ``--no-dependencies``, right? Not everyone will have the option to upgrade all their packages so I wouldn't want to exclude them.

mstimberg:
If you download from source, you are normally supposed to use `python setup.py develop` which would upgrade scipy if it is required. And if someone uses pip install with `--no-deps` then they could also have outdated dependencies that *should* raise an error (e.g. numpy or sympy) -- so should we make an exception just for scipy?
We should get rid of the scipy dependency rather sooner than later (it takes by far the most time to install of all our standard dependencies). If we want to support older scipy versions, then I think we should be consistent and also not make it a requirement in `setup.py`. We would not actually test that version in travis, but given that we don't use scipy much, I guess that's ok (and until not too long ago, we only tested the most recent version anyway). I do understand that it is annoying if we force users to upgrade without any really strong need.

thesamovar:
I just don't see why we need to create a runtime error rather than just a warning (for scipy or any other package, no exception for scipy needed). If someone wants to install it in a different way, let them deal with the consequences of that. Or are there situations where it will give wrong results with an older version (rather than just not work)? In that case we should raise an error.

For scipy, it would be great if we could remove the dependency, but don't we want to keep it in the ``from brian2 import *`` namespace anyway?

mstimberg:
> I just don't see why we need to create a runtime error rather than just a warning (for scipy or any other package, no exception for scipy needed). If someone wants to install it in a different way, let them deal with the consequences of that. Or are there situations where it will give wrong results with an older version (rather than just not work)? In that case we should raise an error.

I just don't want to make a difference between users that use the "proper way" (`python setup.py develop`) and users that add the source directory manually to their path. Some outdated versions will raise errors (e.g. numpy < 0.18), but others will give wrong results (sympy 0.7.5) and we can't really be sure until we test a specific version. We could have some general warning for any version that we don't test, but I think that gets a bit too complicated. Do you think it's really worth extending our dependency check to support warnings for some packages (scipy) and errors for others (sympy)? As I said, I'd rather simply downgrade the requirement both in `setup.py` and in the runtime check to `0.7.0` as it was until recently...

> For scipy, it would be great if we could remove the dependency, but don't we want to keep it in the from brian2 import * namespace anyway?

It isn't in the namespace currently. scipy does not offer any toplevel functions, it only has subpackages which you have to import explicitly anyway (e.g. `scipy.interpolate`, etc.)

thesamovar:
I see what you mean but I think it's a bad idea to unnecessarily annoy users by forcing them to upgrade when they don't actually need to. Perhaps you're right though that if we're not testing versions we don't know if they give wrong results or not.

OK, how about we stick with raising an error and keeping the same requirements for this and for setup.py, but we also have a global preference for ignoring the dependency check? If this is set, it downgrades those errors to warnings. What do you think?

Great news that we're not including anything from scipy. Would be very happy to have this dependency removed (the fewer the better).

mstimberg:
> OK, how about we stick with raising an error and keeping the same requirements for this and for setup.py, but we also have a global preference for ignoring the dependency check? If this is set, it downgrades those errors to warnings. What do you think?

Ok, I'm happy with that. And which version of scipy do we require? 0.7.0 or 0.13.3?

thesamovar:
I thought it had to be 0.13.3 so that we could include it in the tests on travis?

mstimberg:
Not quite -- we can only test 0.13.3 (or newer) on travis, but this is hardcoded in `travis.yml` -- it would be nice if we could say that we tested the minimally required dependency version but we are free to require something less strict in `setup.py` (and of course in the runtime dependency check).

mstimberg:
Um, so our dependency checks are currently before we import anything else, we can't actually use preferences at this point :-/

thesamovar:
I now think we should make the dependency check based on the minimum we've tested so it should be 0.13.3, but we should have a way to switch those checks off (or rather, downgrade them to warnings). Is there no other way we can do this? I don't mind if it is a bit annoying to do it, but I think it's important that there should be a way. How about making it check for an environment variable or the existence of a file like ``~/.brian/no_runtime_dependency_checks``?

thesamovar:
What is the in progress label?

mstimberg:
> What is the in progress label?

I just played around with this: https://waffle.io/brian-team/brian2 :)

mstimberg:
> I now think we should make the dependency check based on the minimum we've tested so it should be 0.13.3, but we should have a way to switch those checks off (or rather, downgrade them to warnings). Is there no other way we can do this? I don't mind if it is a bit annoying to do it, but I think it's important that there should be a way. How about making it check for an environment variable or the existence of a file like ~/.brian/no_runtime_dependency_checks?

This is maybe too annoying. How about we separate the import as such (a *missing* dependency will always raise an import error) and the check of the dependencies? The latter could happen at the end of the `brian2` import and could therefore 1) already assume that the package has been imported successfully and 2) use preferences.

thesamovar:
That works for me.

Also, the waffle things looks potentially useful. :)

mstimberg:
I replaced the previous commit, you can now have warnings instead of errors if you set the`core.outdated_dependency_error` preference to `False` (which you can only do in a file, because it has to be done before the import). I also added a paragraph about this in the "Importing Brian" user documentation.

thesamovar:
Works for me. Merging now.

Issue #340, opened 2014-10-08
Images in automatically generated examples documentation


thesamovar:
See discussion near https://github.com/brian-team/brian2/pull/328#issuecomment-58352608.

The plan is to use git submodules to store the images.

We should also run and generate the tutorials, see #413.

thesamovar:
I wrote some code to do this via scripts although it's not used by readthedocs.org and it uses a dropbox folder. We should probably use briansimulator.org or github to store the images instead, and move the downloading code into conf.py.

thesamovar:
For tutorials (which are ipython notebooks) note that when you run the tutorial if a cell raises an exception it stops running the rest of the notebook. We need a way round that because part of the tutorial is showing what the exceptions look like.

mstimberg:
If we use the `runipy` script I mentioned in #413, it has an option to continue running after an exception (`--skip-exceptions`)

thesamovar:
I think this is all done now, so closing.

PR #420, opened 2015-03-03
Tutorial

thesamovar:
OK, I wrote two tutorials now, one on neurons and one on synapses. We should probably have a few more but that's a pretty good start. I think the tutorials are good to merge (and future tutorials can be added directly to master I think), but this branch also has the new ``start_scope`` functionality. We should probably add some tests for that before merging.


thesamovar:
OK I added a test - I think it's sufficient?

We should probably have some user documentation on ``start_scope`` at some point, but maybe that should be in a new 'Using Brian with IPython' section? Probably not necessary for this PR.

I think it's good to merge.

mstimberg:
The test looks fine, and I think it is ok to not document `start_scope` for now. Also the structure of the user documentation should change a bit, the "Getting started" section is mostly redundant now.

I started some work to integrate the tutorials into the documentation. If you run `dev/tools/doc/build_tutorials.py` then they will be run and converted to rst (no need for the `runipy` package I mentioned before, ipython can do it itself in version 3.0). They way it is included in the HTML output is very ugly, though, there's no difference between code cells in the output :-/ Maybe readthedocs renders it better? Anyway, we have some time to figure it out, for now it will not appear in the online documentation.

If you are ok with my changes, feel free to merge.

The new synapses tutorial looks great as well, by the way. I wonder whether it makes sense to introduce `PoissonGroup` and `SpikeGeneratorGroup` at the end of the neuron tutorial, it would make the input to the synapses a bit easier instead of using the "threshold hack" with a long refractory period.

thesamovar:
Well I had a few problems installing all the prequisites to get that working, but I sort of managed it in the end. The images wouldn't work, I get ``D:\programming\brian2\docs_sphinx\tutorials\1-intro-to-brian-neurons.rst:316: WARNING: image file not readable: tutorials\1-intro-to-brian-neurons_images%5Coutput_28_1.png``. Also, the equations in the synapses notebook are for some reason all repeated three times horizontally (I guess this is a sphinx problem though because the RST file looks ok).

Can we execute the notebook and have it save a new ipynb file and just use nbviewer? i.e. like this except with the images http://nbviewer.ipython.org/github/brian-team/brian2/blob/tutorial/tutorials/1-intro-to-brian-neurons.ipynb

thesamovar:
Oh also, yeah I should probably introduce ``SpikeGeneratorGroup`` and ``PoissonGroup`` at the end of the neurons tutorial so that they can be used in the synapses tutorial. Probably shouldn't encourage hacks in the tutorial. ;)

thesamovar:
Are you motivated to keep looking into a better solution for this or should we just merge now and think about it later?

mstimberg:
> Well I had a few problems installing all the prequisites to get that working, but I sort of managed it in the end.

Ah sorry, I think it should only require a `pip install ipython[nbconvert]`. The image paths does not seem to deal correctly with the backslash for the directory on windows -- we don't necessarily have to have them in another directory, we could also give them long names (`1-intro-to-brian-neurons-image_28_1.png`).

We could also simply write a new `ipynb` file with all the output in it, as you suggest. I thought it would be nice to have it really as part of the documentation, though.

thesamovar:
The advantage of having it not look like the rest of the documentation would be that it maybe emphasises that it's potentially interactive. I put quite a few things in there where I said "Try changing this value in the cell and see what happens" which only works if they're running it as a notebook. The nbviewer page has a prominent download link. On the other hand, we could say the same thing at the top of the tutorial. Something like: "this tutorial is meant to be used as an interactive ipython notebook, click here to download it".

I think that now that these two tutorials at least are written it would be good to have it included in the documentation as standard sooner rather than later. I think it could potentially help lots of first time users.

mstimberg:
I agree that it should be clear that it is an interactive notebook. Whatever approach we use in the end, adding a standard sentence at the top is probably a good idea.

I'll have a look at the issues tomorrow (including #340) -- how about aiming for a second beta release as soon as this is done?

Something purely aesthetic: if you install the `sphinx_rtd_theme` package from pypi, you'll get your locally generated docs in the readthedocs look now.

thesamovar:
Yep, a new beta would be a good idea.

mstimberg:
Ok, I want to give the submodule solution a try. For that we need two new repositories (tutorials and example images) or one common repository. I think one repository for tutorials and images would be less complicated and therefore better. Suggestions for a name? Something like `brian2-doc-resources` ?

thesamovar:
Sounds good to me. Maybe use underscores to fit with our other repos? 

How will versioning work with this? We might change the tutorials with new syntax but we'd like the docs for older versions to be still correct.

mstimberg:
> Maybe use underscores to fit with our other repos?

The only other repo with a compound name is `brian-material` :)

> How will versioning work with this? We might change the tutorials with new syntax but we'd like the docs for older versions to be still correct.

It depends a bit on whether we want to use http://nbviewer.ipython.org/ or not. If we use it, then we'd have to link to a specific version on github which makes our live a bit more complicated. We have to first run `build_tutorials.py`, then commit the tutorials and then update the `index.rst` links. If we are fine for now with the somewhat ugly integrated version (fixing the images, of course), then it would be easy: everything will be stored on readthedocs as the rest of the docs. I tend towards the latter solution, in the long run, we should be able to get a nicer rendering directly included in the docs by converting it to HTML and post-processing the output (in a somewhat hackish way) as done here: https://bitbucket.org/ngoldbaum/yt/src/48c4f6086dd24a3fd30e00ec9771195cf78d85ac/doc/extensions/notebook_sphinxext.py?at=yt
The result: http://yt-project.org/doc/quickstart/introduction.html

mstimberg:
No submodule yet, but this (last commit) is what I had in mind for tutorials that are included directly in the documentation but with an additional list of explicit `.ipynb` links. What do you think of this solution?

Also, do the figures work for you now? I don't see any repeated equations here, so hopefully it should build fine on readthedocs as well.

mstimberg:
Ok, I introduced the submodule, let me know whether this works for you. After pulling the changes, I think you'll have to do `git submodule init` or `git submodule update` or both... Whenever you add/change something to the resources submodule, you have to commit and push it individually (by going into the directory, you treat it as a repository from there). You should do this before pushing the main repository. The reason is that a submodule is referenced in the main repository with a specific commit: If you have this submodule commit only locally, others cannot update their repository as the referenced commit is not available. All this could in principle get a bit complicated, but we will probably update tutorials and example images quite rarely, so I think it should be fine. More than you ever wanted to know can be found here :) http://git-scm.com/book/en/v2/Git-Tools-Submodules

If all this works for you, I think we can merge this branch now (we'll see only then whether it works correctly on readthedocs as well).

mstimberg:
Um, great, so much for the good advice :) It turns out I did exactly what I warned you about -- I really thought that I pushed the example images to the resources repository but apparently I didn't, so now a `git submodule update` fails because the referenced commit is not available. I'll fix this on Monday (and if all this works for you, we could do a release on Monday as well?).

thesamovar:
Doesn't seem to work for me. I get an auth fail.

I will have another look at it but probably not this weekend. I'm wondering if maybe git submodules are too complicated actually (they seem to be more complicated than I thought)?

Also, we shouldn't need to store the IPython notebooks in the submodule I realise. They take sufficiently little time to run that we can run them on readthedocs, and save the generated binaries directly into the generated docs directory. Like this they're automatically correctly versioned without any additional work by us (and we could have both the version in the docs and a downloadable notebook). Right? The only reason we can't do this for the examples is that it takes too long to run them.

mstimberg:
> Doesn't seem to work for me. I get an auth fail.

You mean you get a different error from what I mentioned before, i.e. not the `fatal: reference is not a tree` which is due to a missing push to `brian2-doc-repository`? 

> I will have another look at it but probably not this weekend. I'm wondering if maybe git submodules are too complicated actually (they seem to be more complicated than I thought)?

I think we should give them a try for now. We basically have to do one update per release (maybe we could include some code for this in the release scripts), so it shouldn't be too much hassle.

> Also, we shouldn't need to store the IPython notebooks in the submodule I realise. They take sufficiently little time to run that we can run them on readthedocs, and save the generated binaries directly into the generated docs directory.

I'm not sure this is a good idea. readthedocs isn't meant to execute code, for example they don't allow the installation of Python packages with C extensions. They are providing numpy, scipy and matplotlib as system-installed debian packages, but this is not well-documented (e.g. which versions are actually installed) and it might break at some point when we require a newer version than the one installed on the servers. I'm also not sure about build timeouts on readthedocs, builds already take >10min, I could well imagine that at some point running all the tutorials takes quite a bit of time.

mstimberg:
I now did the missing push to the submodule -- are you still having problems?

thesamovar:
This is the error I'm getting:

```
git@github.com:brian-team/brian2-doc-resources.git: Auth fail
```

I'm guessing this is some problem with Eclipse's git client because it shouldn't be using ``git@github.com``.

Sorry I didn't have more time to look into this today, will try to get to it tomorrow.

Agreed on readthedocs and not running examples on them. Are submodule versions linked to the main repository they live in and versioned with them? If not, maybe we should have it as a separate module rather than a submodule (i.e. is there any advantage to having it as a submodule)?

mstimberg:
I see, sorry, that's again my fault. I haven't really thought that my choice of the submodule url will affect everyone. I changed it to the https url, after an update `git submodule init` and `git submodule update` should now work. If not, apparently the submodule url stored in the repository is only used for the initial setup, you can also manually change it with `git config submodule.docs_sphinx/resources.url https://github.com/brian-team/brian2-doc-resources.git`.

> Are submodule versions linked to the main repository they live in and versioned with them? If not, maybe we should have it as a separate module rather than a submodule (i.e. is there any advantage to having it as a submodule)?

It's a completely separate repository, but each version of the main brian2 repository has a versioned link to it, i.e. if you check out the current version of the tutorial branch at any later time, you will get exactly the same content of the resources directory as now.

mstimberg:
Um, and maybe you also need to do a `git checkout master` followed by a `git pull origin master` in the resources subdirectory, on my computer it otherwise got confused by the remote change :-/

thesamovar:
OK it works! (After I updated Eclipse to a newer version.) I even made a change (correctly I think).

I think it can be merged now unless there's something else that you want to work on?

thesamovar:
Also, happy for you to go ahead and do the new release once this is merged. Should I run all the long tests etc. on Windows before we release? Or is Appveyor working now?

mstimberg:
Cool, so the docs look fine for you (images and all)? We can indeed merge this then, I'm curious to find out whether it all works correctly on readthedocs.

> Also, happy for you to go ahead and do the new release once this is merged. Should I run all the long tests etc. on Windows before we release? Or is Appveyor working now?

I'm actually in the process of getting it to run -- there's some issue with the Cython spike queue that is quite impossible to debug like this, but it should work for everything else (numpy, weave, standalone) which is already quite cool since it tests with Windows MSVC on 32 and 64 bit, Pythyon 2 and 3. I'm still fixing issues, inf fact the "standalone-compatible" tests were not quite working (only the first one was *actually* executed in standalone...). But the Windows testing also already found a Windows-specific bug: we didn't include the `win_makefile` in the source distribution, i.e. someone installing from pypi would not have gotten it.

thesamovar:
Yep, docs build and images for examples and tutorials are there. If it was straightforward to include a link to the binary ``.ipynb`` from the corresponding ``.rst`` that would be nice, but not essential. (Because not everyone will click on the 'Tutorials' link but maybe go straight to the individual tutorial.)

Weird about the standalone tests: I think I've run them on mine without any problems. Perhaps it's a Python 3 issue on Windows? I never tried that. If you let me know what precise platform is causing a problem maybe I can debug it on one of mine? I have 32 and 64 bit Windows but both Python 2.7.

mstimberg:
I'm merging this now. I'll open a pull request for the windows testing soon and I'll comment on the issues there.

mstimberg:
There was still a path issue, but now it works: http://brian2.readthedocs.org/en/latest/
I'll try to add the link (directly to the master branch). It should be quite straightforward if the first cell of the tutorial only contains the title (we should document this somewhere, I guess).

thesamovar:
Nice!

One thing: it looks like the changes I made to the tutorials ``index.rst`` file are not included. I'm wondering if you added a commit to this branch without pulling the latest changes to the submodule and that meant that it was pointing at the older version? We'll have to watch out for that in the future if that's what happened.

mstimberg:
> I'm wondering if you added a commit to this branch without pulling the latest changes to the submodule and that meant that it was pointing at the older version?

Indeed... :-/ It should be fine now.

thesamovar:
We'll have to be careful about that. I think it's a good solution though this submodules thing (now that I have it working).

mstimberg:
Ok, I updated everything, the result looks ok I think: http://brian2.readthedocs.org/en/latest/resources/tutorials/1-intro-to-brian-neurons.html

One more thing: I realized that you edited the `index.rst` file directly, this file is autogenerated and therefore changes will be overwritten with every run of `build_tutorials.py`. I included your changes there and also added a comment to the rst file to make this clearer (we could probably do all this more elegant with templates etc., but...)

thesamovar:
Ah, I hadn't realised that the ``index.rst`` was auto-generated (although in retrospect that was pretty stupid of me).

I think the results look great! I'm very satisfied with that.

PR #423, opened 2015-03-10
Windows testing

mstimberg:
This branch adds automatic testing on Windows VMs. It uses MSVC and specific build configurations for 64 Bit (see `dev/appveyor/run_with_env.cmd` for details). This all seems to work for weave and standalone, I did also manage to build the Cython spike queue but tests crashed when using it and I did not find any way to properly debug it (the complete test simply stopped, no output, no error message). We should get back to this at some point but for now automatically testing numpy, weave and standalone on Python 2 and 3 and 32Bit and 64Bit is already a big step forward.

Ready-to-merge from my side (and then we can do the release).


thesamovar:
I don't fully understand all of these changes. In particular, all the standalone tests were passing on my machine before these changes so why were they necessary?

* Scalar delays: did we not previously have a test for this?
* ``our_stdint.h`` maybe this should be renamed to something clearer like ``brian_stdint.h``?
* Lots of pairs of ``get_device`` and ``set_device`` at the beginning and ends of tests. This seems messy, why is it necessary? Can we fix it in a better way that doesn't require us to put this on every test?

mstimberg:
The "standalone-compatible" tests were not actually run with the standalone device before (except for the first one), so that explains why they were all passing on your machine.

Also, I think you have two MSVC versions on your computer? The builds on appveyor only use one (for each Python/architecture combination) and they work both for C extensions and standalone (well, except that it doesn't quite work with the compiled spike queue).

> Scalar delays: did we not previously have a test for this?

Until the recent addition of the standalone-compatible tests, we only had 7 standalone tests in total, none of them was using scalar delays. And as I said before, the standalone-compatible tests weren't actually using the standalone device.

> our_stdint.h maybe this should be renamed to something clearer like brian_stdint.h?

Agreed, I'll change this (or maybe a more generic `stdint_compat.h`, `stdint_wrapper.h` or something like that? It's not very brian-specific.).

> Lots of pairs of get_device and set_device at the beginning and ends of tests. This seems messy, why is it necessary? Can we fix it in a better way that doesn't require us to put this on every test?

I agree this is messy (and it can even screw things up, since the reset of the device will not be done if the test fails), but it is only necessary for the explicit `cpp_standalone` tests (7 in total). I don't see a much better way for now, we can't reset the device to a runtime device in the `restore_device` teardown (as we did previously) because this would break the standalone-compatible tests. We could have another teardown function just for the `cpp_standalone` tests (but it would still not be perfect as it resets the device to runtime which is not always the correct thing to do). The long term solution is to add a plugin to nose that allows to run a setup/teardown combination for every test (this should be fairly straightforward to write, I was surprised it didn't exist yet). We could then get rid of all the explicit `with_setup` decorators.

thesamovar:
``stdint_compat.h``: yes!

For the other things, OK I understand now!

Yeah, it would be nice to have that nose plugin and get rid of the ``with_setup`` decorators. As you say: let's do that separately though. For now it's fine.

OK so with that name change I think we will be good to merge. Will be great not to have to always run the tests on my machine before merging. :)

mstimberg:
> Will be great not to have to always run the tests on my machine before merging. :)

I agree, we have to be patient to wait for all these tests to finish, though...

Let's wait for all the tests to turn green and then we can merge (I guess I'll rather do the release tomorrow...)

mstimberg:
Travis and appveyor are both happy, I'll merge.

PR #424, opened 2015-03-10
Example finder now finds tutorials as well as examples

### LINKED ISSUE #422: Include links to tutorials in auto-generated reference documentation along with examples

thesamovar:
Fixes #422.

### DIFF
--- a/brian2/sphinxext/examplefinder.py
+++ b/brian2/sphinxext/examplefinder.py
@@ -60,7 +60,7 @@ def auto_find_examples(obj, headersymbol='='):
     txt = txt+'\n'+headersymbol*len(txt)+'\n\n'
     for tutname, tutloc in tutorials:
         tutname = tutname.replace('.ipynb', '')
-        txt += '* Tutorial :doc:`%s </tutorials/%s>`\n' % (tutname, tutloc)
+        txt += '* Tutorial :doc:`%s </resources/tutorials/%s>`\n' % (tutname, tutloc)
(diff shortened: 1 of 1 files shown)


thesamovar:
Let's include this in the release too assuming it works for you and for readthedocs. It's a fairly simple change.

mstimberg:
Looks good, the links were not quite correct though (they referred to files in `docs_sphinx/tutorials` which you probably still had on your machine instead of the new `docs_sphinx/resources/tutorials`). I corrected this, it looks good to me. I'll merge, then we'll see whether readthedocs is also happy with it.

thesamovar:
Ah yes, I did have that directory still lying around! :)

mstimberg:
Well, me too so it worked all fine :) Luckily I realized that the note box on top was missing. 

Oh, and the readthedocs build is failing... I'll fix this directly in master, I think it's the same problem as previously (a path relative to `__file__` which refers to the installation and not the source directory on readthedocs)

mstimberg:
Ok it now works, you actually did prepare everything correctly, you only forgot to set the environment variable for the tutorials. I did however realize that the "tutorial finder" is less than optimal since it searches all of the text instead of restricting the search to the code cells, this is why the first tutorial is listed for `Synapses`. It should be fine most of the time, though, I'm not sure it is worth the effort to improve this...

thesamovar:
Good stuff! :)

I agree it would be painful to make it more accurate and clearly not worth the effort.

Issue #373, opened 2014-12-01
Add a binomial function


mstimberg:
This is needed for #221 and depends on #366

thesamovar:
Where did we get with this?

mstimberg:
I don't quite remember what we agreed to as a solution for the indirect call of numpy's rand function was. I think we have two options:
* Get completely rid of the call to numpy's rand function in weave, use the C implementation as in standalone (after #222 has been implemented)
* Use the alternative namespace approach for functions that should replace the `#define` hacking, as outlined in #380 

I think the second option would be better for now, we have to revise this mechanism to get a `TimedArray` implementation in Cython, anyway.

thesamovar:
I think until we have a good solution to #222 we should be happy with "any decent RNG". Certainly numpy's rand function is good enough, and so would the simple implementation of MT that we discussed in #222, but not stdlib's ``rand()`` function. In terms of implementing it, I don't think we need to worry about having the same implementation in standalone as weave, and I agree with what we decided in #380 so that seems best.

Issue #366, opened 2014-11-12
Allow functions to depend on other functions


mstimberg:
Some mathematical functions depend other functions, so our function mechanism needs to deal with that. One example is a binomial random number function that will make use of a uniformly distributed random number (i.e. `rand()`). We'll need that for #221

mstimberg:
Closed via #428

Issue #221, opened 2014-02-21
PoissonInput in Brian2


yger:
Just a quick question, after I spoke with Marcel. I think this would be good to think about having a PoissonInput object as Brian 1 used to have, that would allow sending multiple spikes to NeuronGroup in an efficient manner. The implementation may not be as simple as it sounds, and since it has to fit into the 'Brian 2' spirit, I'll let you think about it :-)

PR #428, opened 2015-03-17
Function improvements + PoissonInput

### LINKED ISSUE #380: Revise function-specific namespace mechanism
Some functions need an additional namespace, for example `TimedArray` needs to refer to the values. This is currently done using some `#define` hackery for weave and we don't have any solution for Cython yet.

I think we need a more principled approach, e.g. instead of relying on a `#define` we could have the renderer write the function call with all the namespace values (e.g. alphabetically ordered) as additional arguments.

mstimberg:
This branch adds several improvements  to the function system and closes #380, #366, #373, #221. Functions can now depend on other functions and the namespace is passed to functions using global variables instead of via `#define` hackery (we are actually not using `hash_define` anywhere anymore, maybe we should remove it?). There are also two new classes: `BinomialFunction` (with fixed n and p) and `PoissonInput` (also fixed number of inputs and rates) which uses the former.

This is basically done from my side, I'm not too sure about the placements in the package hierarchy (the binomial is in `brian2.utils.binomial` and `PoissonInput` is in `brian2.groups` even though it isn't a group).


mstimberg:
Oh, I forgot to mention: this branch includes commits from the `poisson_input` branch which is now obsolete (I therefore deleted it on github).

thesamovar:
This looks great. I actually ran the tests on my computer and then realised that I didn't need to thanks to AppVeyor! :)

For the packages, how about we have a new ``brian2.random`` package and put binomial there? At the moment it will be the only thing but it's going to happen at some point. Alternatively, we could have a new package that has slightly more general scope and we could binomial, poisson input, timedarray, etc. there. I'm still not very happy with timedarray being in utils. I originally thought of utils as being independent stuff that wouldn't rely on the rest of the brian2 package.

Maybe let's just leave it for the moment and think about having a package reorganisation at some point?

I'm happy for this branch to be merged if you're done with it?

mstimberg:
> For the packages, how about we have a new brian2.random package and put binomial there? At the moment it will be the only thing but it's going to happen at some point. Alternatively, we could have a new package that has slightly more general scope and we could binomial, poisson input, timedarray, etc. there. I'm still not very happy with timedarray being in utils. I originally thought of utils as being independent stuff that wouldn't rely on the rest of the brian2 package.

I agree, I also would like to find a better place for `timedarray`. How about something like `brian2.input`, `brian2.stimulus` (or `stimuli`), ...? In Brian 1 there was a module called `directcontrol`, though I never found that very clear. It could be a place for `SpikeGeneratorGroup`, `PoissonGroup`, `PoissonInput`, `TimedArray` and `BinomialFunction` -- but I'd be happy to leave the first two in `brian2.groups` as well, after all they are proper groups.

Apart from the reorganisation, I'm finished with the branch. But it would probably be a good time to move things around.

thesamovar:
I like ``brian2.input``, shall we go with that?

mstimberg:
> I like brian2.input, shall we go with that?

I'm happy with that. Do you think it should include `SpikeGeneratorGroup` and `PoissonGroup` as well?

thesamovar:
I would say yes.

mstimberg:
Ok, fine with me, I'll move things around accordingly.

mstimberg:
Ok, all done. Let's maybe wait again for the tests (they are not on the last commit, the last commit only adds docstrings) to complete and then merge.

mstimberg:
Tests pass, feel free to merge if you are fine with everything.

PR #432, opened 2015-03-23
Small change to the algorithm for `SpikeGeneratorGroup`

mstimberg:
I changed the algorithm slightly, and now I'm getting correct rounding behaviour for spike times at multiples of dt for long simulations. I added two tests that check this, they take quite some time to run and are therefore only tested when running the long test suite.

@thesamovar Could you run the long test suite on your machine and merge if it passes -- our automatic testing on travis and appveyor will not run those tests.

### DIFF
--- a/brian2/codegen/runtime/cython_rt/templates/spikegenerator.pyx
+++ b/brian2/codegen/runtime/cython_rt/templates/spikegenerator.pyx
@@ -12,13 +12,7 @@
     cdef double _spike_time
 
     # We need some precomputed values that will be used during looping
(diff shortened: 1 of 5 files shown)


thesamovar:
Long tests pass on my machine, but Travis hasn't finished yet. Should probably wait for that.

mstimberg:
Sure, and there are also the appveyor tests which will take a couple of hours, I think -- it only runs a single job at a time...

thesamovar:
Looks like everything has passed: merging now.

PR #433, opened 2015-03-23
Use type inference support for Cython

thesamovar:
Just added the ``infer_types=True`` argument for Cython. This can sometimes improve the quality of the generated C++ code (avoiding Python calls).

### DIFF
--- a/brian2/codegen/runtime/cython_rt/templates/common.pyx
+++ b/brian2/codegen/runtime/cython_rt/templates/common.pyx
@@ -1,6 +1,7 @@
 #cython: boundscheck=False
 #cython: wraparound=False
 #cython: cdivision=True
+#cython: infer_types=True
 
 import numpy as _numpy
 cimport numpy as _numpy


thesamovar:
AppVeyor was cancelled for some reason and isn't restarting. Not sure how I restart the build?

mstimberg:
We use the "rolling build" option for appveyor that cancels a test build on a branch if a new commit for the same branch has been committed (given that appveyor builds can take hours, adding a couple of commits to a branch would otherwise buildup a considerable backlog). This was the case for the master branch here (where it tested the merged pull request) but the test on this branch completed. Apparently, it does not automatically start a new pull request test in this case, but given the minimal change in this commit, I think it is safe to merge things without this additional test. But in general, you should be able to log in using your github account and link the `brian_team` organisation to your account under which the tests are run. You then have a re-run button to manually restart a test.

I'm merging this now, looks good to me.

PR #434, opened 2015-03-25
Add a `sign` function

mstimberg:
Well, does what the title says. I needed a `sign` function (returning -1, 0 or 1) and did not want to fiddle around with complicated `int(...)` expressions.

From my side, it's ready to merge.


thesamovar:
Looks good to me. Merging.

Issue #161, opened 2013-10-31
Use ufunc.at for improved Synapses efficiency


thesamovar:
With the release of numpy 1.8 today, there is a new ufunc method ``at`` that allows us to do things like ``x[idx]+=1`` where ``idx`` contains repeated indices. This should allow for huge speed improvements in the pure Python case. The new syntax for the above would be ``add.at(x, idx, 1)``.

To make this work in the general case I guess we will need some sort of ufunc rewriter that takes an expression and rewrites it in terms of a sequence of ufunc ``at`` calls.

We could also backport this to Brian 1.x although I'm not sure if it's worth it.

mstimberg:
We implemented this in #438 (in a more general way and not quite optimal yet, but working alright). I'm therefore closing this issue.

Issue #436, opened 2015-03-26
(Optionally) replace dynamic with static arrays, e.g. in `StateMonitor`


mstimberg:
On some architectures, e.g. FPGAs, using dynamic data structures is non-trivial. It would therefore be nice if for example a `StateMonitor` could use a static array instead of a dynamic array (since the total run time is known, we can pre-allocate an array of the correct size). This would also give performance benefits, in particular for weave and cython where we currently go via Python to resize dynamic arrays.

mstimberg:
Putting Dan's comments here that he did when we were discussing this issue via mail:

> I think for state monitor we should still use a dynamic array because if you did two runs you wouldn't be able to statically allocate. On the other hand, we could certainly (in before_run) resize the dynamic array to be sufficiently large to cover the run. The only situation in which that wouldn't be a good idea is if someone is doing something like run(1e10*second) and then using stop() for a non-deterministic stop time. So we'd need some way to solve that problem.

> For both spike and state monitor, we could imagine having a size keyword argument that determines the initial allocated size. For devices that don't support dynamic arrays, that initial size would also be the maximum total size and it would either raise an error or just stop recording if that capacity was filled. If you left size=None (default) then for state monitor it would use the length of the run. For spike monitor we could indeed make the initial size 1000 (or maybe even larger, it's not likely that people are going to instantiate hundreds of spike monitors so we're not wasting much memory).

mstimberg:
> I think for state monitor we should still use a dynamic array because if you did two runs you wouldn't be able to statically allocate.

We could still do a re-allocation in the same way we do for dynamic arrays, i.e. for the second run, create a new array of the correct size and copy over the data from the first run.

> On the other hand, we could certainly (in before_run) resize the dynamic array to be sufficiently large to cover the run.

Actually, I realized that this is not really the issue that affects the weave/cython performance, the problem rather seems to be that at every time step we call `resize` in Python, which goes through a few steps before it decides that the array does not need resizing. We'd obviously not have this problem if the state monitor was simply using static arrays.

> The only situation in which that wouldn't be a good idea is if someone is doing something like run(1e10*second) and then using stop() for a non-deterministic stop time. So we'd need some way to solve that problem.

Good point, is this something that is actually used by anyone? Given the possible simplifications of using static arrays, maybe we should rather focus on the common use case and make this use case the exception, i.e. one would for example have to manually specify the monitor size in this case?

[...]

PR #438, opened 2015-03-27
Numpy synapses

### LINKED ISSUE #435: Modifying presynaptic variables doesn't work using numpy
Bug reported by Owen Mackwood on the Brian development list:

If you increment a presynaptic variable in the Synapse pre/post code, it is only incremented once per presynaptic neuron, rather than once per synapse (as it presumably should). The following code exhibits this behaviour:

```Python
import os
from brian2 import *

on_pre = True
standalone = True
if standalone: set_device('cpp_standalone')
else: prefs.codegen.target = 'numpy'

tau_A = 20*ms
gl = 10.0*nsiemens
el = -60*mV
vt = -50.*mV
memc = 200.0*pfarad
bgcurrent = 200*pA

eqs_neurons='''
dA/dt = -A / tau_A : 1
dv/dt=(-gl*(v-el)+bgcurrent)/memc : volt (unless refractory)'''
neurons = NeuronGroup(10, model=eqs_neurons, threshold='v > vt',
                      reset='v=el', refractory=5*ms)

kwds = {'connect':True, ('pre' if on_pre else 'post'): 'A_pre += 1'}
con = Synapses(neurons[:1], neurons[1:], **kwds)
sm = StateMonitor(neurons[:1], "A", record=True)

run(1*second, report='text')
if standalone:
    device.build(compile=True, run=True, debug=False)

plot(sm.t, sm.A.T)
show()
```

The plot should show increments of 9, whether on_pre is True or False. The numpy version exhibits increments of 1 in both cases.

thesamovar:
This branch fixes #435 and #161. It is not yet ready to merge.

What i've done so far is to create two new modules/functions. The first (``brian2.synapses.permutation_analysis``) checks if a block of synaptic code is permutation-independent, i.e. that it gives the same results regardless of the ordering of the spiking synapses. Actually, it doesn't quite do this because it uses the algorithm discussed in https://github.com/brian-team/brian2/issues/435#issuecomment-86676986 which gives false negatives but could be expanded to give fewer false negatives. I also wrote some new tests (in ``test_synapses.py``) to check that this code gives the correct results (but none of the known false negatives are tested). This testing code should probably be further expanded.

The second (``brian2.codegen.runtime.numpy_rt.synapse_vectorisation``) converts a block of abstract code into a numpy vectorised form assuming that it has passed the permutation-independence test above. It is not as efficient as it could be because it indexes arrays in every line, however that should make it correct. We can add some further optimisations on top of that later. If the vectorisation cannot be performed, it currently falls back on a Python loop. I was thinking that we should also allow it to fall back on the current implementation if no pre-synaptic variables are modified as an intermediate case, but I haven't implemented that yet. I don't yet have any tests for this. Good tests would be to compare against the reference loop implementation (or just against weave/cython actually).

In both cases I haven't integrated the new functions into the code generation workflow. This is mostly because my train journey was just enough time to write the code above, but also because I honestly don't know the best place to slot this code in. @mstimberg, maybe you could take a look at that? Also, the functions at the moment take an explicit set of synaptic, pre- and post-synaptic variables as argument, as well as a list of ``Statement`` objects: it might be the case that this could be incorporated in a more consistent way into these functions directly from the ``variables`` dict. In the case of the numpy vectorisation, for the moment the variable names and indices are also probably not the best ones to use, and the template will need some work too.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,20 +1,20 @@
 language: python
 matrix:
   include:    
-#    - python: "2.7"
-#      env: STANDALONE=no CYTHON=yes MINIMAL_VERSIONS=no REPORT_COVERAGE=yes
-#    - python: "2.7"
-#      env: STANDALONE=no CYTHON=yes MINIMAL_VERSIONS=yes REPORT_COVERAGE=no
-#    - python: "3.3"
-#      env: STANDALONE=no CYTHON=yes MINIMAL_VERSIONS=no REPORT_COVERAGE=no
-#    - python: "3.3"
-#      env: STANDALONE=no CYTHON=yes MINIMAL_VERSIONS=yes REPORT_COVERAGE=no
-#    - python: "3.4"
-#      env: STANDALONE=no CYTHON=yes MINIMAL_VERSIONS=no REPORT_COVERAGE=no
+    - python: "2.7"
+      env: STANDALONE=no CYTHON=yes MINIMAL_VERSIONS=no REPORT_COVERAGE=yes
+    - python: "2.7"
+      env: STANDALONE=no CYTHON=yes MINIMAL_VERSIONS=yes REPORT_COVERAGE=no
+    - python: "3.3"
+      env: STANDALONE=no CYTHON=yes MINIMAL_VERSIONS=no REPORT_COVERAGE=no
+    - python: "3.3"
+      env: STANDALONE=no CYTHON=yes MINIMAL_VERSIONS=yes REPORT_COVERAGE=no
+    - python: "3.4"
+      env: STANDALONE=no CYTHON=yes MINIMAL_VERSIONS=no REPORT_COVERAGE=no
     - python: "2.7"
       env: STANDALONE=yes CYTHON=no MINIMAL_VERSIONS=no REPORT_COVERAGE=no
-#    - python: "2.7"  # test without installed cython
-#      env: STANDALONE=no CYTHON=no MINIMAL_VERSIONS=no REPORT_COVERAGE=no
+    - python: "2.7"  # test without installed cython
+      env: STANDALONE=no CYTHON=no MINIMAL_VERSIONS=no REPORT_COVERAGE=no
 
 # Use miniconda to install binary versions of numpy etc. from continuum
 # analytic's repository. Follows an approach described by Dan Blanchard:
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,20 +1,20 @@
 language: python
 matrix:
   include:    
-    - python: "2.7"
-      env: STANDALONE=no CYTHON=yes MINIMAL_VERSIONS=no REPORT_COVERAGE=yes
-    - python: "2.7"
-      env: STANDALONE=no CYTHON=yes MINIMAL_VERSIONS=yes REPORT_COVERAGE=no
-    - python: "3.3"
-      env: STANDALONE=no CYTHON=yes MINIMAL_VERSIONS=no REPORT_COVERAGE=no
-    - python: "3.3"
-      env: STANDALONE=no CYTHON=yes MINIMAL_VERSIONS=yes REPORT_COVERAGE=no
-    - python: "3.4"
-      env: STANDALONE=no CYTHON=yes MINIMAL_VERSIONS=no REPORT_COVERAGE=no
+#    - python: "2.7"
+#      env: STANDALONE=no CYTHON=yes MINIMAL_VERSIONS=no REPORT_COVERAGE=yes
+#    - python: "2.7"
+#      env: STANDALONE=no CYTHON=yes MINIMAL_VERSIONS=yes REPORT_COVERAGE=no
+#    - python: "3.3"
+#      env: STANDALONE=no CYTHON=yes MINIMAL_VERSIONS=no REPORT_COVERAGE=no
+#    - python: "3.3"
+#      env: STANDALONE=no CYTHON=yes MINIMAL_VERSIONS=yes REPORT_COVERAGE=no
+#    - python: "3.4"
+#      env: STANDALONE=no CYTHON=yes MINIMAL_VERSIONS=no REPORT_COVERAGE=no
     - python: "2.7"
       env: STANDALONE=yes CYTHON=no MINIMAL_VERSIONS=no REPORT_COVERAGE=no
-    - python: "2.7"  # test without installed cython
-      env: STANDALONE=no CYTHON=no MINIMAL_VERSIONS=no REPORT_COVERAGE=no
+#    - python: "2.7"  # test without installed cython
+#      env: STANDALONE=no CYTHON=no MINIMAL_VERSIONS=no REPORT_COVERAGE=no
 
 # Use miniconda to install binary versions of numpy etc. from continuum
 # analytic's repository. Follows an approach described by Dan Blanchard:
--- a/README.rst
+++ b/README.rst
@@ -16,6 +16,10 @@ Brian2 is released under the terms of the CeCILL 2.1 license.
 .. image:: https://img.shields.io/pypi/v/Brian2.svg?style=flat-square
         :target: https://pypi.python.org/pypi/Brian2
 
+.. image:: https://badges.gitter.im/Join%20Chat.svg
+   :alt: Join the chat at https://gitter.im/brian-team/brian2
+   :target: https://gitter.im/brian-team/brian2?utm_source=badge&utm_medium=badge&utm_campaign=pr-badge&utm_content=badge
+
 Testing status for master branch
 --------------------------------
 
--- a/brian2/__init__.py
+++ b/brian2/__init__.py
@@ -11,11 +11,6 @@ except ImportError as ex:
     sys.stderr.write('Importing numpy failed: %s\n' % ex)
     missing.append('numpy')
 try:
-    import scipy
-except ImportError as ex:
-    sys.stderr.write('Importing scipy failed: %s\n' % ex)
-    missing.append('scipy')
-try:
     import sympy
 except ImportError as ex:
     sys.stderr.write('Importing sympy failed: %s\n' % ex)
@@ -81,6 +76,5 @@ def _check_dependency_version(name, version):
             logger.warn(message, 'outdated_dependency')
 
 for name, version in [('numpy', '1.8.0'),
-                      ('scipy', '0.13.3'),
                       ('sympy', '0.7.6')]:
     _check_dependency_version(name, version)
--- a/brian2/codegen/generators/base.py
+++ b/brian2/codegen/generators/base.py
@@ -4,10 +4,16 @@ methods which should be overridden to implement a new language.
 '''
 from brian2.core.variables import ArrayVariable
 from brian2.utils.stringtools import get_identifiers
+from brian2.utils.logger import get_logger
 from brian2.codegen.translation import make_statements
-from brian2.codegen.permutation_analysis import check_for_order_independence
+from brian2.codegen.permutation_analysis import (check_for_order_independence,
+                                                 OrderDependenceError)
 
 __all__ = ['CodeGenerator']
+
+
+logger = get_logger(__name__)
+
 
 class CodeGenerator(object):
     '''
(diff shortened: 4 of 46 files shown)


mstimberg:
Ok, I put quite a bit of work into this and integrated it with the main code generation system (it's part of `numpy_generator` since everything is about transforming abstract code into code snippets).

I generalized it to not be specifically about pre/post for synapses but about variables using indices that are 1) different from `_idx` and 2) not guaranteed to be unique. While synapses are certainly the most important use case, we can create exactly the same problems using linked variables, e.g. the following is not well-defined and gives different results with numpy and weave:
```Python
G = NeuronGroup(10, '''v:1
                       x : 1 (linked)''',
                threshold='v>1',
                reset='''x+=1
                         v=x''')
G2 = NeuronGroup(2, 'x:1')
G.x = linked_var(G2.x, index=np.repeat([0, 1], 5))
G.v = 1.1
run(defaultclock.dt)
``` 

`AttributeVariable` now has a new attribute `unique`, which is only relevant for arrays acting as indices, and says whether values in the array are *guaranteed* to be unique. This is necessary because there are unproblematic situations where indices are mixed, e.g. the use of subgroups.

I had to change your code around the code vectorisation quite significantly as the "replace variable names with indexed variable names" approach did not work with all the other stuff we do for indexing (conditional write, etc.). It now tries to avoid all problems by doing a complete read/code/write block for every statement line. There are still a few open/unclear issue:
* Introducing new unique indices for statements such as `u_post = v_post` is not easy to combine with the code generation process -- but maybe we can ignore this for now, since it is only an optimisation?
* If code is not order-independent, it will still go through the `ufunc_at_vectorisation` (it should directly go to the fallback loop, right?)
* I was a bit confused by the difference the code makes between the different in-place operators. Isn't `x -= y` strictly equivalent to `x += -y` (and `x /= y` to `x *= 1/y`)?


The current code seems to be working alright (e.g. it will give a warning for the above example), the example in #435 now works correctly and running the STDP example in numpy now takes about 8m40s instead of 9m40s minutes on my machine (so the benefits of the `ufunc.at` already seem to outweigh the unnecessary indexing/copying we are currently doing). We should add a substantial amounts of tests before merging this, though (and run all the examples, etc.)

The branch contains also a change that does not only affect numpy:
In the general code generation process (i.e. for all targets), we analyze statements using sympy and substitute `x = x + ...` by `x += ...` and the same for `*=` now.

thesamovar:
I'll try to look at this in more detail tomorrow, but just a few comments for the moment.

Great idea to have this working more generally for repeated indices.

Sorry that you had to rewrite my code substantially. Was the best I could do on the train. ;) I knew that the indices stuff would have to change but I had hoped it wouldn't be a substantial rewrite but rather just a few small additions.

I'm happy not to worry about the unique indices optimisation. I put it in while I was writing it because I realised it was possible, but I doubt it makes a huge difference because it doesn't change the order of the number of operations at all.

Yes, if it fails the order independence it should fall back to the loop. I maybe didn't write it in the code but I intended the ufunc at vectorisation function to be called only if it had already passed the order independence test.

I think you're right about ``-=`` and ``/=`` I was just having a mental block about them while I was writing the code so decided to leave them out for the moment. The thing is that they're not symmetric functions where f(a,b)=f(b,a) (because a-b!=b-a) which I thought was a necessary condition for order independence, but I don't think it is. What is the correct condition if not symmetry of the function?

I'm surprised the speed improvement is so minimal. I seem to remember when we had the discussion before (with hand written examples) the speed improvement was much greater than that. I guess we would need to profile it more carefully.

Agreed that we need very substantial testing before merging. Although on the other hand, the current code is actually giving wrong results so I guess it can't be much worse. ;)

Great about the ``x=x+...``->``x += ...`` optimisation. Does it work for all of ``x=x+y+z``, ``x=y+x+z``, ``x=y+z+z``? How about with parentheses? e.g. ``x=(x+y)+z``.

mstimberg:
> Sorry that you had to rewrite my code substantially. Was the best I could do on the train. ;)

I wasn't complaining, I simply realized it was still quite a long way from the proof-of-concept to the fully integrated code :)

Let's not worry about the unique indices thing for now, I also doubt it makes a significant difference. 

About the speed improvement: The current code is far from optimal, since it loads/writes arrays for every statement instead of once for a block of statements. Either way, it's already faster than before, so I don't think it is necessary to improve this a lot before merging.

> I think you're right about -= and /= I was just having a mental block about them while I was writing the code so decided to leave them out for the moment. The thing is that they're not symmetric functions where f(a,b)=f(b,a) (because a-b!=b-a) which I thought was a necessary condition for order independence, but I don't think it is. What is the correct condition if not symmetry of the function?

I'm not sure either... I have the feeling we are stepping into complicated code analysis territory here, where we have to do dependency graphs etc... Maybe we are over-analyzing this, given that most of the use cases are not very complex in this regard? Can we come up with something simple and straightforward that is possibly more restrictive than necessary but we are sure that it works, e.g. something like "the RHS of assignments has to use only constants and variables with the same index or variables with other indices that are never written to", everything else has to be augmented assignments (maybe that's already close to what you did for the order-independence check?). Is there anything reasonable useful that this would not cover?

> Agreed that we need very substantial testing before merging. Although on the other hand, the current code is actually giving wrong results so I guess it can't be much worse. ;)

Sure, but it only gave wrong results for uncommon use cases, we have to make sure we are not introducing errors for stuff that worked fine before.

> Great about the x=x+...->x += ... optimisation. Does it work for all of x=x+y+z, x=y+x+z, x=y+z+z? How about with parentheses? e.g. x=(x+y)+z.

It uses symbolic evaluation with sympy so it should work for all of these. I just added a test for these replacements, it works fine for everything I tried.

thesamovar:
I might have a go at improving the speed a bit by only reading/writing arrays when necessary. I think this shouldn't be too hard to code. You're not working on this branch right now are you?

I think the correct condition for being able to reorder elements is as follows. If you are reducing a function f(a,b) so that at each stage a->f(a,b) for varying b, then it is sufficient if f(a,b)==f(b,a) but not necessary. The necessary condition is f(f(a, b), c)==f(f(a,c), b) I think. Both f(a,b)=a-b and f(a,b)=a/b satisfy this. I don't know if this class of functions has a name (I couldn't find anything googling it), but as you say, perhaps it's sufficient to just say that ``+=``, ``-=``, ``*=`` and ``/=`` all satisfy it, whereas e.g. ``**=`` does not.

I think the rest of the current reasoning is correct: it was just that we were missing out this case. I'd leave it as it is rather than simplify and lose generality.

mstimberg:
> I might have a go at improving the speed a bit by only reading/writing arrays when necessary. I think this shouldn't be too hard to code. You're not working on this branch right now are you?

No, please go ahead. The new tests I added are failing on Python 3 currently, by the way, but it isn't a big issue.

thesamovar:
Actually the speed is not as bad as I thought. I just ran the STDP example with profiling, and it reveals a slightly better picture. Running it on my machine for 10s of simulated time takes 38s on master with numpy, and 34s with the new ufunc.at code, which looks poor (11% improvement). However, if you look at the profiling, you can see that synapses_pre drops from 10.35s to 6.51s, which is great (59% improvement).

The difference is even more extreme if you bump the size of the simulation up to 10k neurons instead of 1k (dividing the synaptic weights by 10 to normalise the firing rates). Then you get the total times before and after for a 10s run being 120s -> 49s (2.4x faster), but the synapses_pre time goes from 70s -> 7.7s (9.1x faster). synapses_post only goes from 7.8s->6.4s (1.2x faster), but that's not surprising because it only modifies synaptic variables and so the ufunc.at optimisation doesn't really help. Interestingly, with the code in master synapses_pre is by far the longest running codeobj (taking 4x longer than poissongroup_thresholder) but with the ufunc.at optimisations it's now only the second longest, taking about half the time of the thresholder.

Going up to 100k neurons gives us a 4x total speed improvement and a 31x speed improvement for synapses_pre.

So given all that I might not bother with trying to improve the reading/writing of arrays for the moment.

mstimberg:
Ah, indeed, that seems to be good enough for now at least.

thesamovar:
I just did some testing of weave versus numpy.

For 100k neurons, 1s simulation:
- Numpy: total=33s, synapses_pre=3.1s, thresholder=22.9s
- Weave: total=36s, synapses_pre=1.8s, thresholder=28.7s

So, synapses_pre is still slower than weave, but only 1.7x slower. Weirdly, poissongroup thresholder is considerably slower in weave. Any ideas why this might be?

For 10k neurons, 10s simulation:
- Numpy: total=71s, synapses_pre=10.8s, thresholder=21.6s
- Weave: total=61s, synapses_pre=5.6s, thresholder=26.2s

So here synapses_pre is only 1.9x slower than weave, and again thresholder is slower.

For 1k neurons, 10s simulation:
- Numpy: total=50s, synapses_pre=9.7s, thresholder=5.1s
- Weave: total=37s, synapses_pre=4.8s, thresholder=4.4s

For this example, synapses_pre is about 2x slower than weave. Note that for this last example, synapses_post and statemonitor took more time than synapses_pre and thresholder, and that for the first time thresholder is faster in weave than numpy. Curious.

Anyway, I think these results are great. Basically, with ufunc.at we are getting within 2x of C++ speeds with numpy, despite all that indexing. I think that's great and means we've done a good job here.

Overall, I'm happy with this branch now. I think it remains to do some more testing, right?

thesamovar:
Looking at the implementations of threshold for numpy and weave, I can't see any reason why numpy would be faster, so I guess it must be the random number generation buffering stuff we use in weave. I'm a bit surprised that it makes such a difference because over a long run where many random numbers are generated, this ought to be efficient - right?

Can you check these results on your machine?

thesamovar:
Oh, the results above were not using the best compiler settings (I'm on a new laptop), but it doesn't change things too much.

mstimberg:
I didn't yet check the runtimes in detail on my machine. Just a quick remark regarding the random number generation: we currently use a fixed buffer size of 1024, so for the bigger networks it will call Python several times during a time step. I think at the time we did not yet have the dynamic implementation framework for functions, but with it we could relatively easily tune the buffer size to something appropriate, e.g. the number of neurons. Or maybe just go for a fixed-size but bigger buffer right away? Thinking about it, there shouldn't be any advantage in having exactly one call per time step -- if our group is small than having to call numpy's `rand` function only every 10th timestep or so would be an advantage.

thesamovar:
That sounds reasonable to me if the small buffer size is the problem. In any case, making it larger is unlikely to be a problem.

thesamovar:
I tried manually changing the buffer size to the size of the ``NeuronGroup`` but it didn't have a large effect on the time, so I guess the problem is somewhere else. I also tried inlining the function to remove the function call overhead and that didn't change it. I guess maybe the fact that it has to do a couple of if statements each time it's called could be the problem? Not sure I see a way to fix that though.

mstimberg:
A good baseline comparison is to simply remove the full `DEFAULT_FUNCTIONS['rand'] = ...` statement in `weave_rt.py` -- it will then use the `rand()/RAND_MAX` approach for random numbers. I tried this in a single run of the STDP example and it reduced the runtime of the thresholder by 30% or something like that, so significant but not huge. I guess the slow monitors (see discussion in #436) might be lower hanging fruits for optimization.

thesamovar:
Not exactly comparable because C's ``rand()`` presumably does much less work than MT. Agreed though, it's probably not the biggest worry. Still, it bugs me that it is slower in weave than in numpy. ;)

mstimberg:
Ignoring the performance aspect for now, we should probably merge this soon... I added one more test, could you have a look and maybe add tests for other situations that I didn't think of? Also, please run the long test suite and the examples (they are by default run using the numpy target), regressions should show up in the generated images -- I did not notice any.

While writing tests I also added a feature for standalone (getting the value of subexpressions) that allowed to run more tests as "standalone-compatible". This is not really well-placed in this branch but I was too lazy to create a new branch just for that...

thesamovar:
Starting to run the tests / etc. now but should we be worried about the Travis build failing?

thesamovar:
These are the test failures I'm getting for standalone.

```
ERROR: Test simulation of an infinite cable vs. theory for current pulse (Green function)
    self.test(*self.arg)
  File "C:\programming\brian2\brian2\tests\test_spatialneuron.py", line 86, in test_infinitecable
    net.run(0.02*ms)
  File "C:\programming\brian2\brian2\core\base.py", line 249, in device_override_decorated_function
    return getattr(curdev, name)(*args, **kwds)
  File "C:\programming\brian2\brian2\devices\cpp_standalone\device.py", line 930, in network_run
    with_output=False)
  File "C:\programming\brian2\brian2\devices\cpp_standalone\device.py", line 788, in build
    raise RuntimeError("Project run failed")
RuntimeError: Project run failed

ERROR: Test simulation of short cylinder vs. theory for constant current.
    self.test(*self.arg)
  File "C:\programming\brian2\brian2\tests\test_spatialneuron.py", line 132, in test_finitecable
    net.run(100*ms)
  File "C:\programming\brian2\brian2\core\base.py", line 249, in device_override_decorated_function
    return getattr(curdev, name)(*args, **kwds)
  File "C:\programming\brian2\brian2\devices\cpp_standalone\device.py", line 930, in network_run
    with_output=False)
  File "C:\programming\brian2\brian2\devices\cpp_standalone\device.py", line 788, in build
    raise RuntimeError("Project run failed")
RuntimeError: Project run failed
```

The two that failed built correctly, but crashed on running.

mstimberg:
There was indeed an incorrect handling of a quite special situation in the code generation that made the spatialneuron tests fail on standalone (they weren't tested on standalone until now). This should be fixed now, on my machine the full long/standalone test suite runs without errors. There was also a problem in the refractoriness test which was a bit difficult to debug since it only occurred on travis/appveyor but not on our machines. It was actually a bug in the test, it used `<` where it should have used `<=`. All tests are now passing on travis, there was unfortunately a timeout on appveyor so I canceled the test there. If it all the tests/examples work for you as well (and if you don't have any test cases to add), then I think this is ready to merge.

thesamovar:
Everything runs fine on my machine, and I'm happy with the code and the level of testing. Well, we should probably have more testing but I'm happy to do that separately. Merging now.

Issue #356, opened 2014-10-31
Refactor CPPStandaloneDevice for better derivability


thesamovar:
In particular, we should make it easier for @brian-team/brian2genn.

The main thing to do is to cut `Device.build` into smaller submethods.

thesamovar:
CNS meeting update; we might still improve on what was done in #439 but it's not critical for 2.0.

PR #439, opened 2015-04-01
refactoring CPPStandaloneDevice

Kwartke:
fixes: #356 
Suggestions for other function names, further refactoring, ... are also welcome

### DIFF
--- a/brian2/devices/cpp_standalone/device.py
+++ b/brian2/devices/cpp_standalone/device.py
@@ -368,7 +368,7 @@ class CPPStandaloneDevice(Device):
         self.code_objects[codeobj.name] = codeobj
         return codeobj
     
-    def check_OPENMP_compatible(self, nb_threads):
+    def check_openmp_compatible(self, nb_threads):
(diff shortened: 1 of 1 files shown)


mstimberg:
Thanks, looks good to me! I'll wait for feedback from Dan, he might have had specific things in mind when opening the issue (he should be back from vacation next week).

thesamovar:
Thanks for the PR!

The main point of the #356 was to make it easier to derive classes from CPPStandaloneDevice. I think the best way to see if this PR helps with that would be to look at a particular example, perhaps GeNN with @tnowotny.

thesamovar:
OK so on looking at it in more detail, I think it looks good. Could you make just one change: the method ``check_OPENMP_compatible`` doesn't match our naming style: could you make it ``check_openmp_compatible``?

thesamovar:
Thanks for making that change! I think this looks good now and it's passing the checks so I'm to go ahead and merge it.

Thanks again for the contribution!

PR #440, opened 2015-04-03
Scipy removal

### LINKED ISSUE #386: Try to import weave before scipy.weave
Weave is now deprecated in the latest scipy 1.5 so we should do something like:

```
try:
   import weave
except ImportError:
  try:
    from scipy import weave
  except ImportError:
    weave = None
```

mstimberg:
In this branch, I added a pure-numpy implementation of the `spatialstateupdate` template, i.e. one that does not use `solve_banded` from scipy. This was the only place in the code where we used scipy and listing it as a dependency meant that users installing brian via pip or easy_install had to download the full 10MB source code and wait for its compilation to finish (which takes a couple of minutes).

The problem is that the algorithm is really straightforward to implement in C (i.e., with loops) but its forward and backward passes over arrays are not vectoriseable using numpy syntax (at least I don't know how). Therefore, even after a bit of optimisation, the numpy solution is very slow -- about 10x as slow as the version using scipy... I therefore thought that the best approach is to:
* No longer list scipy as a dependency, but...
* ... still include both algorithms (pure numpy and scipy) in the `spatialstateupdate` template, using the scipy version if it is available
* emit a warning if the numpy-only algorithm is used, recommending the user to use weave/cython or to install scipy

While this makes the `spatialstateupdate` template a bit more complex than I'd like it to be, I think this is the best solution, it would be hard to justify to drastically reduce the performance of some code for users that are happily using numpy+scipy right now. @thesamovar, @romainbrette what do you think?

This branch also fixes #386, i.e. it allows for the independent `weave` package as an alternative to `scipy.weave`. I decided against raising a warning if `scipy.weave` is used because for example the Anaconda distribution does not ship the `weave` package.

### DIFF
--- a/brian2/__init__.py
+++ b/brian2/__init__.py
@@ -11,11 +11,6 @@ except ImportError as ex:
     sys.stderr.write('Importing numpy failed: %s\n' % ex)
     missing.append('numpy')
 try:
-    import scipy
-except ImportError as ex:
-    sys.stderr.write('Importing scipy failed: %s\n' % ex)
-    missing.append('scipy')
-try:
     import sympy
 except ImportError as ex:
     sys.stderr.write('Importing sympy failed: %s\n' % ex)
@@ -81,6 +76,5 @@ def _check_dependency_version(name, version):
             logger.warn(message, 'outdated_dependency')
 
 for name, version in [('numpy', '1.8.0'),
-                      ('scipy', '0.13.3'),
                       ('sympy', '0.7.6')]:
     _check_dependency_version(name, version)
--- a/brian2/codegen/generators/numpy_generator.py
+++ b/brian2/codegen/generators/numpy_generator.py
@@ -121,8 +121,13 @@ class NumpyCodeGenerator(CodeGenerator):
         return lines
 
     def determine_keywords(self):
-        # For numpy, no addiional keywords are provided to the template
-        return {}
+        try:
+            import scipy
+            scipy_available = True
+        except ImportError:
+            scipy_available = False
+
+        return {'_scipy_available': scipy_available}
 
 ################################################################################
 # Implement functions
--- a/brian2/codegen/generators/numpy_generator.py
+++ b/brian2/codegen/generators/numpy_generator.py
@@ -138,7 +138,8 @@ for func_name, func in [('sin', np.sin), ('cos', np.cos), ('tan', np.tan),
                         ('exp', np.exp), ('log', np.log), ('log10', np.log10),
                         ('sqrt', np.sqrt), ('arcsin', np.arcsin),
                         ('arccos', np.arccos), ('arctan', np.arctan),
-                        ('abs', np.abs), ('mod', np.fmod)]:
+                        ('abs', np.abs), ('mod', np.fmod),
+                        ('sign', np.sign)]:
     DEFAULT_FUNCTIONS[func_name].implementations.add_implementation(NumpyCodeGenerator,
                                                                     code=func)
 
--- a/brian2/codegen/runtime/numpy_rt/templates/spatialstateupdate.py_
+++ b/brian2/codegen/runtime/numpy_rt/templates/spatialstateupdate.py_
@@ -10,7 +10,6 @@
 Solves the cable equation (spatial diffusion of currents).
 This is where most time-consuming time computations are done.
 '''
-from scipy.linalg import solve_banded
 from numpy.linalg import solve
 from numpy import zeros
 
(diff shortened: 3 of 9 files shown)


romainbrette:
I think I agree with Marcel's choice. 10x slower is not great...

Le 03/04/2015 17:15, Marcel Stimberg a écrit :
>
> In this branch, I added a pure-numpy implementation of the 
> |spatialstateupdate| template, i.e. one that does not use 
> |solve_banded| from scipy. This was the only place in the code where 
> we used scipy and listing it as a dependency meant that users 
> installing brian via pip or easy_install had to download the full 10MB 
> source code and wait for its compilation to finish (which takes a 
> couple of minutes).
>
> The problem is that the algorithm is really straightforward to 
> implement in C (i.e., with loops) but its forward and backward passes 
> over arrays are not vectoriseable using numpy syntax (at least I don't 
> know how). Therefore, even after a bit of optimisation, the numpy 
> solution is very slow -- about 10x as slow as the version using 
> scipy... I therefore thought that the best approach is to:
>
>   * No longer list scipy as a dependency, but...
>   * ... still include both algorithms (pure numpy and scipy) in the
>     |spatialstateupdate| template, using the scipy version if it is
>     available
>   * emit a warning if the numpy-only algorithm is used, recommending
>     the user to use weave/cython or to install scipy
>
> While this makes the |spatialstateupdate| template a bit more complex 
> than I'd like it to be, I think this is the best solution, it would be 
> hard to justify to drastically reduce the performance of some code for 
> users that are happily using numpy+scipy right now. @thesamovar 
> <https://github.com/thesamovar>, @romainbrette 
> <https://github.com/romainbrette> what do you think?
>
> This branch also fixes #386 
> <https://github.com/brian-team/brian2/issues/386>, i.e. it allows for 
> the independent |weave| package as an alternative to |scipy.weave|. I 
> decided against raising a warning if |scipy.weave| is used because for 
> example the Anaconda distribution does not ship the |weave| package.
>
> ------------------------------------------------------------------------
>
>
>         You can view, comment on, or merge this pull request online at:
>
> https://github.com/brian-team/brian2/pull/440
>
>
>         Commit Summary
>
>   * Remove the use of scipy in the `spatialstateupdate` template (not
>     optimized yet)
>   * Some optimisations for the numpy version of `spatialstateupdate`
>     (still quite slow)
>   * Only use the pure numpy solution if scipy is not available and
>     raise a warning if it is used.
>   * Remove the dependency on weave and support the independent `weave`
>     package as an alternative to `scipy.weave`
>   * Add the `sign` function to the function implementations to
>     directly call the `numpy` version in executed code (instead of the
>     wrapped version with unit-checking)
>
>
>         File Changes
>
>   * *M* brian2/__init__.py
>     <https://github.com/brian-team/brian2/pull/440/files#diff-0> (6)
>   * *M* brian2/codegen/generators/numpy_generator.py
>     <https://github.com/brian-team/brian2/pull/440/files#diff-1> (12)
>   * *M*
>     brian2/codegen/runtime/numpy_rt/templates/spatialstateupdate.py_
>     <https://github.com/brian-team/brian2/pull/440/files#diff-2> (72)
>   * *M* brian2/codegen/runtime/weave_rt/weave_rt.py
>     <https://github.com/brian-team/brian2/pull/440/files#diff-3> (8)
>   * *M* brian2/spatialneuron/spatialneuron.py
>     <https://github.com/brian-team/brian2/pull/440/files#diff-4> (14)
>   * *M* brian2/tests/__init__.py
>     <https://github.com/brian-team/brian2/pull/440/files#diff-5> (6)
>   * *M* brian2/tests/test_parsing.py
>     <https://github.com/brian-team/brian2/pull/440/files#diff-6> (5)
>   * *M* brian2/utils/logger.py
>     <https://github.com/brian-team/brian2/pull/440/files#diff-7> (15)
>   * *M* setup.py
>     <https://github.com/brian-team/brian2/pull/440/files#diff-8> (1)
>
>
>         Patch Links:
>
>   * https://github.com/brian-team/brian2/pull/440.patch
>   * https://github.com/brian-team/brian2/pull/440.diff
>
> —
> Reply to this email directly or view it on GitHub 
> <https://github.com/brian-team/brian2/pull/440>.
>

-- 
Twitter: https://twitter.com/RomainBrette
Blog: http://romainbrette.fr/category/blog/

New paper:
Laudanski J, Zheng Y, Brette R (2014). A structural theory of pitch. eNeuro. DOI: 10.1523/ENEURO.0033-14.2014

thesamovar:
I agree with this solution too. Should I review and merge?

mstimberg:
> Should I review and merge?

Yes please.

thesamovar:
Looks good to me. Merging now.

One question: on github now the merge checks are failing because of coverage. I guess we should either (1) not include coverage in the checks, or (2) actually insist on writing enough tests to pass the coverage check. In principle we should do (2) I guess but perhaps this is a bit onerous. What do you think?

mstimberg:
> One question: on github now the merge checks are failing because of coverage. I guess we should either (1) not include coverage in the checks, or (2) actually insist on writing enough tests to pass the coverage check. In principle we should do (2) I guess but perhaps this is a bit onerous.

Yea, I noticed that, it's a bit confusing if it is displayed as "checks failed" when the coverage decreased by 0.11%... For now, I switched the setting to the system we had previously, where coveralls will add a comment instead of integrating as another test. In general I'm much in favor of having a higher test coverage but unfortunately it is not easy to get really useful numbers out of coveralls. What we'd like to have is the complete test coverage for long tests over all code generation targets and including standalone but we can only report the coverage from a single travis run (so it will be excluding long tests and excluding standalone). In the case of this pull request, the major reason for the reduced test coverage is the alternative code for `scipy.weave` vs. `weave` import -- on travis, `scipy` is installed and therefore `weave` will be never imported, leading to lines that are missing from the test coverage. All in all, not much we can do about it right now, I think...

thesamovar:
OK that sounds reasonable.

PR #454, opened 2015-04-21
Fix SpikeMonitor for subgroups (weave, cython, cpp_standalone)

mstimberg:
For subgroups not starting at 0, the code for searching through the list of all spikes for spikes that should be recorded was not correct. If some neuron in the full group spiked, but none of them >= the start index of the recorded subgroup, then all (instead of none) spikes were recorded. The good news is that this should not have led to undetected incorrect recordings because 1) it would have recorded spikes with negative indices, which would be incorrect in an obvious way and 2) it most of the time should have led to a segmentation fault due to accessing the `_count` array with negative indices (this is the way I encountered the issue). Nevertheless, an important bug.

### DIFF
--- a/brian2/codegen/runtime/cython_rt/templates/spikemonitor.pyx
+++ b/brian2/codegen/runtime/cython_rt/templates/spikemonitor.pyx
@@ -11,8 +11,8 @@
         # For subgroups, we do not want to record all spikes
         # We assume that spikes are ordered
         # TODO: Will this assumption ever be violated?
-        _start_idx = 0
-        _end_idx = - 1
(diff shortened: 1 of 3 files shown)


thesamovar:
Code looks good to me so happy for you to merge it as soon as the CIs report back.

mstimberg:
I'm going to merge this now. One of the travis test suites stalled for some reason, but since this is a single commit on top of current master, the two tests suites ran on the identical code and are therefore redundant. AppVeyor failed because of the standalone filename issue that we discussed, it will be fixed in a separate pull request (it also timed out, though...).

PR #455, opened 2015-04-21
Use filenames that are guaranteed to be unique for standalone

mstimberg:
Arrays were previously stored in the results directory using the name of the array as the filename. On file systems that are not case sensitive (e.g. Windows), this is not guaranteed to work. The most common situation were this leads to problems is with a variable "I" in a neuron group, since it also contains the variable "i" by default. As an easy fix, the filename now uses the name of the variable plus its hash value. Note that users should never have to construct such file names manually, if for any reasons they cannot access the values via the normal state variable mechanism, they should use `device.get_array_filename` to get the filename.

### DIFF
--- a/brian2/devices/cpp_standalone/device.py
+++ b/brian2/devices/cpp_standalone/device.py
@@ -186,6 +186,32 @@ class CPPStandaloneDevice(Device):
             raise TypeError(('Do not have a name for variable of type '
                              '%s') % type(var))
 
+    def get_array_filename(self, var, basedir='results'):
+        '''
+        Return a file name for a variable.
+
+        Parameters
+        ----------
+        var : `ArrayVariable`
+            The variable to get a filename for.
+        basedir : str
+            The base directory for the filename, defaults to ``'results'``.
+        Returns
+        -------
+        filename : str
+            A filename of the form
+            ``'results/'+varname+'_'+str(hash(varname))``, where varname is the
+            name returned by `get_array_name`.
+
+        Notes
+        -----
+        The reason that the filename is not simply ``'results/' + varname`` is
+        that this could lead to file names that are not unique in file systems
+        that are not case sensitive (e.g. on Windows).
+        '''
+        varname = self.get_array_name(var, access_data=False)
+        return os.path.join(basedir, varname + '_' + str(hash(varname)))
+
     def add_array(self, var):
         # Note that a dynamic array variable is added to both the arrays and
         # the _dynamic_array dictionary
@@ -317,8 +343,8 @@ class CPPStandaloneDevice(Device):
             # disk
             if self.has_been_run:
                 dtype = var.dtype
-                fname = os.path.join(self.project_dir, 'results',
-                                     array_name)
+                fname = os.path.join(self.project_dir,
+                                     self.get_array_filename(var))
                 with open(fname, 'rb') as f:
                     data = np.fromfile(f, dtype=dtype)
                 # This is a bit of an heuristic, but our 2d dynamic arrays are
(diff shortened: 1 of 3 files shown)


mstimberg:
Note that only the test suite of the first commit will run the tests that previously failed (but it may reach a time out after that). The second commit restricts the test suite to tests that are not marked as `long`, also for the standalone-compatible tests (it was previously ignored for all the standalone tests)

mstimberg:
Ah, damnit, that wasn't smart, appveyor canceled the test on the previous commit...

mstimberg:
Maybe run the long tests manually on your machine? If it runs fine on travis with the "hashed" filenames, it should run fine on appveyor as well, so I'm not too worried.

thesamovar:
I'm getting most tests failing with errors like this.

```
ERROR: Test the correct handling of referred scalar variables in subexpressions
    self.test(*self.arg)
  File "C:\programming\brian2\brian2\tests\test_neurongroup.py", line 127, in test_referred_scalar_variable
    assert_allclose(G2.out[:], np.arange(10)+1)
  File "C:\programming\brian2\brian2\core\variables.py", line 856, in __getitem__
    return self.get_item(item, level=1)
  File "C:\programming\brian2\brian2\core\variables.py", line 846, in get_item
    run_namespace=namespace)
  File "C:\programming\brian2\brian2\core\base.py", line 249, in device_override_decorated_function
    return getattr(curdev, name)(*args, **kwds)
  File "C:\programming\brian2\brian2\devices\cpp_standalone\device.py", line 384, in variableview_get_subexpression_with_index_array
    run_namespace=run_namespace)
  File "C:\programming\brian2\brian2\core\base.py", line 251, in device_override_decorated_function
    return func(*args, **kwds)
  File "C:\programming\brian2\brian2\core\variables.py", line 1195, in get_subexpression_with_index_array
    codeobj_class=get_default_codeobject_class('codegen.string_expression_target')
  File "C:\programming\brian2\brian2\codegen\codeobject.py", line 339, in create_runner_codeobj
    override_conditional_write=override_conditional_write,
  File "C:\programming\brian2\brian2\devices\device.py", line 264, in code_object
    name=name)
  File "C:\programming\brian2\brian2\codegen\runtime\numpy_rt\numpy_rt.py", line 52, in __init__
    self.variables_to_namespace()
  File "C:\programming\brian2\brian2\codegen\runtime\numpy_rt\numpy_rt.py", line 75, in variables_to_namespace
    value = var.get_value()
  File "C:\programming\brian2\brian2\core\variables.py", line 510, in get_value
    return self.device.get_value(self)
  File "C:\programming\brian2\brian2\devices\cpp_standalone\device.py", line 348, in get_value
    with open(fname, 'rb') as f:
IOError: [Errno 2] No such file or directory: 'c:\\users\\dgoodman\\appdata\\local\\temp\\tmpiqd9ya\\results\\_array_neurongroup_1_freq_171961427'
```

thesamovar:
The issue was backslashes and escapes. Windows filenames have backwards slash rather than forward slash, and when pasted directly into C++ code these were being interpreted as escape sequences. I modified the templates to handle this issue (because the get_array_filename method ought to return a value without the slashes escaped).

thesamovar:
Oops, that last commit messed up that file because apparently my git configuration on this machine is not correctly set up to handle CRLF, so it shows every line as having changed.

mstimberg:
Hmm, indeed, but now the file is actually using the recommended file ending (LF). If you check it out on a machine with `core.autocrlf` you should get it again with CRLF line endings. Which reminds me, should we maybe do the big clean-up commit for the line endings soon?

thesamovar:
Tests passed on my computer and apparently on appveyor, so I guess this is probably going to be good. Happy for it to be merged once they all report back.

thesamovar:
Oh and for the line endings, yeah now would be a good time I guess. Is the idea to have everything with just LF and we make sure to have core.autocrlf on to handle that?

mstimberg:
All tests passed, merging now.

PR #456, opened 2015-04-22
Normalize line endings

mstimberg:
This normalizes all the line endings in the repository to LF, as recommended here:
https://help.github.com/articles/dealing-with-line-endings/

On Linux/MacOSX machines, users should set the git setting `core.autocrlf`to `input`, on Windows machines it should be `true`. This way, Windows users will get CRLF line feeds on checkout that will be converted into LF when committing. 

Since this branch only changes whitespaces, it shouldn't make any test fail, but let's nevertheless wait for the test suite to finish before merging.


mstimberg:
Since it is a whitespace-only change, I'll go ahead with the merge. Note that all merges/rebases of already existing branches will most likely show conflicts because almost every file is changed by the second commit.

PR #457, opened 2015-04-23
Schedule improvements

### LINKED ISSUE #426: Add `before_` and `after_` slots to the schedule
In Brian 1, there were slots like `after_thresholds` and I think they are more convenient than writing `when='thresholds', order=1` for example. A good solution would be to not include them in `Network.schedule` (i.e. this would still just be `['start', 'groups', ...]`) but instead automatically add the `before_` and `after_` slots when sorting the objects in a network. This way, one can also conveniently add new slots to the schedule (e.g. for custom operations) and then use `before_`, `after_` relative to them.

### LINKED ISSUE #448: Deal with devices that do not support arbitrary schedules
Devices should have a possibility to define a fixed schedule, if they don't allow for arbitrary ones. If that fixed schedule is not the same as the user-defined schedule, it should raise a warning.

To make working with such devices easier, the default schedule should be made a preference.

mstimberg:
This branch introduces:
* a preference `core.network.default_schedule` to specify the default schedule for networks that don't set one explicitly
* the option to use `when` slots such as `before_thresholds` or `after_synapses` -- these slots are automatically taken into account when sorting objects, but they are not explicitly included in the schedule (closes #426)
* devices have a `network_schedule` attribute (defaulting to `None`). If it is set, a warning is raised if this schedule does not match the network's schedule (closes #448) 

The last point does probably not cover all issues surrounding devices that only support fixed schedules -- we still have the freedom of arbitrary ordering within one slot using the `order` attribute and devices might not support that (which of course only makes a difference if the objects are not independent, i.e. they are writing to some shared memory, e.g. synaptic pathways or code writing to linked variables).

Hmm, maybe this branch would also be a good place to set the order for `pre` and `post` pathways, I'll add a commit doing that.


mstimberg:
The two build failures are due to stalled tests in travis/appveyor, I restarted them.

mstimberg:
All tests pass now, @thesamovar feel free to merge if you are ok with the changes.

thesamovar:
Looks great to me, just two comments:
1. Should we document the pre before post behaviour? What do we do in the case where there are multiple pre/post pathways? I guess the user should be able to have control over that?
2. Should we validate the schedule? i.e. make sure that it contains at least groups, threshold, resets, synapses and that start and end are at the start/end? Maybe not necessary because any object in a missing slot will raise an error when you try to run the network?

mstimberg:
> 1. Should we document the pre before post behaviour? What do we do in the case where there are multiple pre/post pathways? I guess the user should be able to have control over that?

Yes, it should definitely be documented, I just added a commit that does that. I also mentioned how the user can change the order of multiple pre/post pathways -- I guess it would be in the Brian2 spirit to force the user to explicitly specify an ordering in this case, but I don't see a way to neatly integrate this into the existing syntax. What do you think?

> 2. Should we validate the schedule? i.e. make sure that it contains at least groups, threshold, resets, synapses and that start and end are at the start/end? Maybe not necessary because any object in a missing slot will raise an error when you try to run the network?

I considered that, but I think it is probably best to keep it simple. I could imagine (e.g. in a tutorial) to play around with the schedule for a single `NeuronGroup` and a `StateMonitor` and just use `['groups', 'thresholds', 'resets', 'end']` (or maybe a new slot name for `StateMonitor`) and change the order of those, not bothering with `start` and `synapses`. Getting an error in that case seems unnecessarily restrictive.

thesamovar:
OK, with that update I think it's looks good to merge. I think what we have now is fine and we don't need to force the user to specify the ordering (after all, ordering can potentially make a difference in other cases too and we don't want to force the user to specify everything). However, if you think it's worth doing I'm not against it either. It will presumably only make a difference in a very small number of cases anyway. No ideas for syntax.

Also agreed on not needing to validate the schedule.

If you're happy with it as it is and you don't want to add something to force users to specify synaptic orders then I'm happy for you to go ahead and merge.

mstimberg:
> OK, with that update I think it's looks good to merge. I think what we have now is fine and we don't need to force the user to specify the ordering (after all, ordering can potentially make a difference in other cases too and we don't want to force the user to specify everything). However, if you think it's worth doing I'm not against it either. It will presumably only make a difference in a very small number of cases anyway. No ideas for syntax.

I think you are right, if we force the user to provide an ordering here, we'd have to do it in other corner cases as well (e.g. several neuron groups writing to the same variable via a linked variable mechanism). I think it is ok as it is, the ordering is arbitrary but well defined and the most important thing is that the user can change it.

I'll go ahead with the merge then.

PR #458, opened 2015-04-24
Build docs from release

### LINKED ISSUE #453: Include sphinx conf.py in pypi tarball
I was just trying to build Brian2 from pypi. Since conf.py isn't included, the docs can't be built. Can you please include conf.py in the tarball pushed to pypi? Alternatively, you could drop the docs from the tar if you don't want users building them. Having them there but not being able to build them isn't nice :)

The docs don't seem to build without the examples. Can these please be included too? Previously, I think there was a separate tarball for all the docs.

mstimberg:
This includes `conf.py` and the example files in the release tarball (increasing its size from 487KB to 517KB, which is not a big deal I think). This will allow to build the documentation (using something like `sphinx-build docs_sphinx target_dir`) from the release tarball. Previously, we did include the documentation from `docs_sphinx` but you could not actually build the docs. The generated docs will not be exactly like the docs on http://brian2.readthedocs.org, since the example images are not included and neither are the tutorials -- but clearly we don't want to include them in the release tarball and I don't think providing a separate `brian2-doc` package is worth the effort, given that readthedocs allows  you to download the complete documentation for offline use.

Thanks again to @sanjayankur31 for making us aware of this issue. 
Closes #453 -- ready to merge from my side.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -41,9 +41,10 @@ install:
     fi
   - source activate travis_conda
   - if [[ $CYTHON == 'yes' ]]; then conda install --yes cython; SETUP_ARGS=--with-cython; else SETUP_ARGS=""; fi
-  - pip install -q coveralls --use-mirrors
+  - pip install -q coveralls
   - python setup.py install $SETUP_ARGS --fail-on-error
   - if [[ $DOCS_ONLY == 'yes' ]]; then
+    pip install -q sphinxcontrib-issuetracker;
     python setup.py sdist;
     fi
 # command to run tests (make sure to not run it from the source directory)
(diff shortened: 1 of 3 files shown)


thesamovar:
Happy to merge as far as the code goes, but tests are failing.

mstimberg:
Oh, sorry, that was my fault. It's actually not a problem of this branch: I did a "safe" change that I directly committed to master (it just added a warning) but I screwed up the test (it worked if run for a single codegen target, but if the test suite was run for multiple targets than the warning was only triggered once making the test fail). I already fixed this in master, so I'm going to merge this branch now.

PR #465, opened 2015-05-15
Standalone customization

mstimberg:
This branch introduces more options to customize the compilation for standalone (useful to interface with external libraries) and also tries to use a consistent interface for cython, weave and standalone. The preferences in `codegen.cpp` now allow:
* to add include, library and runtime library directories
* to add libraries (e.g. adding `mylibrary` will link with `-lmylibrary` on Linux)
* to include additional headers in all files that contain user code
* to define macros
* to add additional linker arguments

I tried to use the general distutils mechanism to generate the compiler/linker options also for standalone (in the same way as it is done for weave and cython), this might buy us support for additional compilers etc. in the future.

There's a new example (`examples/advanced/opencv_movie.py`) that demonstrates the use of some of those preferences to read in a movie file with OpenCV. 

@owenmackwood: you were missing this kind of configurability, is there anything in addition that you'd need?
@thesamovar: this is basically ready to merge but I did not test any of this on Windows...

### DIFF
--- a/brian2/codegen/cpp_prefs.py
+++ b/brian2/codegen/cpp_prefs.py
@@ -120,8 +120,7 @@ def get_compiler_and_args():
         compiler = get_default_compiler()
     extra_compile_args = prefs['codegen.cpp.extra_compile_args']
     if extra_compile_args is None:
-        if compiler == 'gcc' or compiler == 'unix':
-            compiler = 'gcc'
+        if compiler in ('gcc', 'unix'):
             extra_compile_args = prefs['codegen.cpp.extra_compile_args_gcc']
         if compiler == 'msvc':
             extra_compile_args = prefs['codegen.cpp.extra_compile_args_msvc']
--- a/brian2/codegen/cpp_prefs.py
+++ b/brian2/codegen/cpp_prefs.py
@@ -25,7 +25,7 @@ prefs.register_preferences(
         Should be gcc or msvc.
         '''
         ),
-    extra_compile_args = BrianPreference(
+    extra_compile_args=BrianPreference(
         default=None,
         validator=lambda v: True,
         docs='''
@@ -33,19 +33,26 @@ prefs.register_preferences(
         ``extra_compile_args_gcc`` or ``extra_compile_args_msvs``).
         '''
         ),
-    extra_compile_args_gcc = BrianPreference(
+    extra_compile_args_gcc=BrianPreference(
         default=['-w', '-O3'],
         docs='''
         Extra compile arguments to pass to GCC compiler
         '''
         ),
-    extra_compile_args_msvc = BrianPreference(
+    extra_compile_args_msvc=BrianPreference(
         default=['/Ox', '/EHsc', '/w'],
         docs='''
         Extra compile arguments to pass to MSVC compiler
         '''
         ),
-    include_dirs = BrianPreference(
+    extra_link_args=BrianPreference(
+        default=[],
+        docs='''
+        Any extra platform- and compiler-specific information to use when
+        linking object files together.
+        '''
+    ),
+    include_dirs=BrianPreference(
         default=[],
         docs='''
         Include directories to use. Note that ``$prefix/include`` will be
@@ -53,12 +60,39 @@ prefs.register_preferences(
         site-specific directory prefix as returned by `sys.prefix`.
         '''
         ),
-    msvc_vars_location = BrianPreference(
+    library_dirs=BrianPreference(
+        default=[],
+        docs='''
+        List of directories to search for C/C++ libraries at link time.
+        '''
+    ),
+    runtime_library_dirs=BrianPreference(
+        default=[],
+        docs='''
+        List of directories to search for C/C++ libraries at run time.
+        '''
+    ),
+    libraries=BrianPreference(
+        default=[],
+        docs='''
+        List of library names (not filenames or paths) to link against.
+        '''
+    ),
+    define_macros=BrianPreference(
+        default=[],
+        docs='''
+        List of macros to define; each macro is defined using a 2-tuple,
+        where 'value' is either the string to define it to or None to
+        define it without a particular value (equivalent of "#define
+        FOO" in source or -DFOO on Unix C compiler command line).
+        '''
+    ),
+    msvc_vars_location=BrianPreference(
         default='',
         docs='''
         Location of the MSVC command line tool (or search for best by default).
         '''),
-    msvc_architecture = BrianPreference(
+    msvc_architecture=BrianPreference(
         default='',
         docs='''
         MSVC architecture name (or use system architectue by default).
--- a/brian2/codegen/cpp_prefs.py
+++ b/brian2/codegen/cpp_prefs.py
@@ -78,6 +78,15 @@ prefs.register_preferences(
         List of library names (not filenames or paths) to link against.
         '''
     ),
+    headers=BrianPreference(
+        default=[],
+        docs='''
+        A list of strings specifying header files to use when compiling the
+        code. The list might look like ["<vector>","'my_header'"]. Note that
+        the header strings need to be in a form than can be pasted at the end
+        of a #include statement in the C++ code.
+        '''
+    ),
     define_macros=BrianPreference(
         default=[],
         docs='''
--- a/brian2/codegen/runtime/cython_rt/cython_rt.py
+++ b/brian2/codegen/runtime/cython_rt/cython_rt.py
@@ -6,7 +6,7 @@ import numpy
 from brian2.core.variables import (DynamicArrayVariable, ArrayVariable,
                                    AttributeVariable, AuxiliaryVariable,
                                    Subexpression)
-from brian2.core.preferences import prefs, BrianPreference
+from brian2.core.preferences import prefs
 from brian2.utils.logger import get_logger
 
 from ..numpy_rt import NumpyCodeObject
@@ -14,32 +14,13 @@ from ...templates import Templater
 from ...generators.cython_generator import (CythonCodeGenerator, get_cpp_dtype,
                                             get_numpy_dtype)
 from ...targets import codegen_targets
+from ...cpp_prefs import get_compiler_and_args
 from .extension_manager import cython_extension_manager
 
 __all__ = ['CythonCodeObject']
 
 
 logger = get_logger(__name__)
-
-# Preferences
-prefs.register_preferences(
-    'codegen.runtime.cython',
-    'Cython runtime codegen preferences',
-    extra_compile_args = BrianPreference(
-        default=['-w', '-O3'],
-        docs='''
-        Extra compile arguments to pass to compiler
-        '''
-        ),
-    include_dirs = BrianPreference(
-        default=[],
-        docs='''
-        Include directories to use. Note that ``$prefix/include`` will be
-        appended to the end automatically, where ``$prefix`` is Python's
-        site-specific directory prefix as returned by `sys.prefix`.
-        '''
-        )
-    )
 
 
 class CythonCodeObject(NumpyCodeObject):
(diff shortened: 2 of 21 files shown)


thesamovar:
For the ``extra_link_args`` this will presumably have the same problem as the compiler args, i.e. that the syntax is different for MSVC and gcc. Should we play the same trick with ``extra_link_args_gcc`` and ``extra_link_args_msvc``? Maybe it's not necessary since we're not providing any default values here.

thesamovar:
OK, Cython isn't working on my machine. I'll attempt to diagnose this as soon as I can. It appears to be trying to use gcc but with the msvc arguments, which is a little odd.

thesamovar:
The new example looks very cool and great to have something like this to show people the sort of thing you can do. It runs in pure Python mode on Windows using Christopher Gohlke's build of OpenCV. Haven't tested it with weave/standalone yet because I need to install the C++ version first.

mstimberg:
> For the extra_link_args this will presumably have the same problem as the compiler args, i.e. that the syntax is different for MSVC and gcc. Should we play the same trick with extra_link_args_gcc and extra_link_args_msvc? Maybe it's not necessary since we're not providing any default values here.

I'd say we can leave this as it is for now. As you said, we don't have any default linker values, so this will always be something that the user provides explicitly in whatever syntax they want.

mstimberg:
> OK, Cython isn't working on my machine. I'll attempt to diagnose this as soon as I can. It appears to be trying to use gcc but with the msvc arguments, which is a little odd.

Oh and about this problem: it is certainly possible that I screwed something up there...

thesamovar:
I've looked into this a bit and I have a bit better of an idea as to what's wrong and a partial fix, but it's still not fixed satisfactorily. The issue is that Cython decided on its own what compiler to use (I think it just uses the distutils choice), and we weren't passing it anything. However, we were independently deciding what compiler to use, and coming to a different choice. So what was happening was that we were deciding msvc was the best choice whereas Cython was deciding gcc was the best choice, so we were passing arguments like ``/W`` which of course gcc had no idea how to handle. In the commit above, I found a way to force Cython to compile with a particular compiler (our choice), but now at least on the machine I'm on at the moment (with 32 bit Python) it raises an error about "stdint.h". This is a known problem with MSVC but I don't understand why it didn't occur on my Windows machine that only has 64 bit Python, because it couldn't have been using gcc there and the same versions of MSVC are installed on both (2008, 2010 and 2013).

So, in conclusion, I think I still need to do some investigation here.

mstimberg:
Ok, great. I have been working the last two days quite a bit on Windows as well,getting the conda packages to build on appveyor. I finally got around setting up a VM with Windows (64 Bit), so I could do proper debugging. It took me quite a while to figure out while the compiled Cython SpikeQueue was crashing, I expected it to be a 64 Bit issue but in the end it turned out that the problem was a `&(vec[0])` for a vector of size 0 -- I can well imagine that this is undefined. I went as far as having a conda package with compiled SpikeQueue and weave compilation working (with the C++ compiler for Python), for Cython I then ran into the `stdint.h` issue. This should be easily fixable, are you already working on this or should I do it?
Oh and there's still a bug in Python's distutil (that will be fixed in Python 2.7.10 apparently) that prevents the msvc compiler to be found without additional patching, but it is fixed by monkey patching from setuptools, i.e. an import of a recent version of setuptools before importing brian makes it work. So we should simply require/import setuptools in brian to fix this issue.

About the compiler: I think it were better if we have as little code about guessing which compiler to use etc. as possible, we should probably completely rely on distutils for that. This way, we won't have the situation that weave/cython runs fine on its own but does not work when run from Brian. Of course, you always have the option to set the compiler preference manually (and this is what we'd recommend on Windows if users encounter problems). I also wonder whether we should not replace the `extra_compile_args_gcc` and `..._msvc` mechanism by something more general -- maybe `extra_compile_args` could be a dictionary by default with `{'gcc': [...], 'msvc': [...]}` and the user can either change this dictionary or simply overwrite it with a list? At some point, we might support `mingw`, `icc`, etc. and we don't want to keep adding preferences for these cases.

mstimberg:
Oh, speaking about using distutils more: I think we can have it search for `vcvarsall.bat` by using
```Python
from distutils import msvc9compiler
vcvars_loc = msvc9compiler.find_vcvarsall(msvc9compiler.get_build_version())
```
Is there a reason why the docs say that standalone mode needs MSVC 2010 -- wouldn't other versions work as well?

thesamovar:
>  the problem was a &(vec[0]) for a vector of size 0 -- I can well imagine that this is undefined

Ugh, ``&(vec[0])`` should be optimised so that it doesn't actually have to extract ``vec[0]`` and then take the address. I thought that was safe (I saw it recommended on several discussions).

> for Cython I then ran into the stdint.h issue. This should be easily fixable, are you already working on this or should I do it?

Well I don't know how to easily fix it, so I'd prefer if you did it! :)

> So we should simply require/import setuptools in brian to fix this issue.

OK!

> About the compiler: I think it were better if we have as little code about guessing which compiler to use etc. as possible, we should probably completely rely on distutils for that. 

Yes, if it works reliably. On the other hand, we might not want to do that for standalone where we need a newer version of MSVC than the one used for Python (because we use some C++ features that weren't implemented in the 2008 version).

> At some point, we might support mingw, icc, etc. and we don't want to keep adding preferences for these cases.

Happy to have it more general but I doubt we'll ever add any more compilers. :)

> Is there a reason why the docs say that standalone mode needs MSVC 2010 -- wouldn't other versions work as well?

I don't remember what feature it is, but something needs 2010 or later. The point is mainly that 2008 doesn't work and this is the version you need for compatibility with Python (up to 3.3 or 3.4 I think where they switched to 2010).

thesamovar:
Also, my attempted fix is not working on either of my two machines, but if you can fix the stdint.h problem I think we'll be OK. I'm confused as to why it wasn't causing a problem before.

mstimberg:
> Ugh, &(vec[0]) should be optimised so that it doesn't actually have to extract vec[0] and then take the address. I thought that was safe (I saw it recommended on several discussions).

Yeah, I thought so, too, and it worked fine for us so far. I did definitely pinpoint it down to this line in the Cython script, however, and only executing it for a non-empty queue made the crash go away. I did not check the generated C++ code in detail, it's also possible that it's Cython's fault.

> Well I don't know how to easily fix it, so I'd prefer if you did it! :)

I commited the change, it seems to work fine for me with MSVC 2008 now.

>> About the compiler: I think it were better if we have as little code about guessing which compiler to use etc. as possible, we should probably completely rely on distutils for that.
>> 

> Yes, if it works reliably. On the other hand, we might not want to do that for standalone where we need a newer version of MSVC than the one used for Python (because we use some C++ features that weren't implemented in the 2008 version).

I agree. Actually, as this might be a common source of error for Windows users, how about we require the mscv version explicitly? E.g. we could set it by default to the version we recommend installing in the docs (MSVC 2010 Express) and if the user uses something else, they have to specify it. We then use `msvc9compiler.find_vcvarsall` with the given version. This seems cleaner too me than having an explicit list of potential directories.

> I don't remember what feature it is, but something needs 2010 or later. The point is mainly that 2008 doesn't work and this is the version you need for compatibility with Python (up to 3.3 or 3.4 I think where they switched to 2010).

Maybe I should go through the old issues for that... It would be nice if one single MSVC compiler worked. I did not see any fundamental issues, apart from a few remaining `stdint.h` imports and the use of `std::move` (which I did not quite get, anyway) I was able to compile the `test_cpp_standalone` file with MSVC 2008. The main.exe then crashed, so it is certainly not working correctly, but I was also not using the latest version of the spike queue and maybe it just needs something similar to my previous Cython fix.

mstimberg:
About the crash: it seems indeed to be the same issue, but it's not that trivial to fix, because we have to change the code generation part. Whenever we get the underlying data for a dynamic array, we do the `&(vec[0])` trick and this reliably crashes for me (with MSVC 2008) if the vector has size 0.

mstimberg:
So, in fact, after dealing with those `&vec[0]` calls (all of this in the `appveyor_conda` branch), I can run the complete test suite, including standalone, weave and Cython on Windows 7 64 Bit, Python 2.7, all only after installing anaconda and the C++ for Python compiler, without any manual patching.
The only things I *did* change manually for now (and that's not included as a commit in the branch) are the removal of `std::move` and the use of `msvc9compiler.find_vcvarsall(msvc9compiler.get_build_version())` for the `vcvarsall.bat` location. Oh and OpenMP does not work, but I did not yet look into the reason for this yet.

I'll clean things up and do further work on the conda packaging, the release scripts etc., but maybe let's discuss all that in a separate pull request (I'll rebase the `appveyor_conda` branch before that, so rather don't work with it now).

For the current pull request, is there anything around the configurability of standalone that we still want/need to change? I agree that generalizing the compiler options is not really important for now...

thesamovar:
OK I had to make one more change (Cython was ignoring the compiler option) but now it works on both my machines, using either gcc or msvc on the 32 bit Python, or msvc on the 64 bit Python.

> I agree. Actually, as this might be a common source of error for Windows users, how about we require the mscv version explicitly? E.g. we could set it by default to the version we recommend installing in the docs (MSVC 2010 Express) and if the user uses something else, they have to specify it. We then use msvc9compiler.find_vcvarsall with the given version. This seems cleaner too me than having an explicit list of potential directories.

What exactly does msvc9compiler find? By the name, it only finds 2008? I think that's a little restrictive. For example, it stops people using newer C++ features in code they link, and it stops them from being able to compile code for the CPU they're running on (e.g. AVX extensions).

About ``std::move`` I don't remember what it does either. I suspect that on newer versions of MSVC it fails to compile without that because I think I was testing on MSVC 2013 Express when I wrote that.

> For the current pull request, is there anything around the configurability of standalone that we still want/need to change? 

No, I think it's good to go once the tests pass.

mstimberg:
> What exactly does msvc9compiler find? By the name, it only finds 2008? I think that's a little restrictive. For example, it stops people using newer C++ features in code they link, and it stops them from being able to compile code for the CPU they're running on (e.g. AVX extensions).

Ah, no, the name is quite misleading (and the documentation is sparse...). `msvccompiler9` is for all compilers starting with 2008, the `find_vcvarsall` function takes a version argument (i.e. `9` for VS 2008) and then looks for the corresponding compiler, first in the registry and then in an environment variable.

thesamovar:
Ah OK so for standalone, if we wanted to, we could use the most recent VS on the users system but use distutils to find it. I'd be happy with that - would it work for you too?

mstimberg:
There was a problem with the compiler setting, it used `'gcc'` on Linux instead of `'unix'`. I removed the replacement of `'unix'` by `'gcc'` (and the other way round that I introduced in another place), so we should always check for `compiler in ('gcc', 'unix')` if we want to do something specific for gcc. For now, we always do `if compiler == 'msvc' ... else ... `, so this is not really an issue.

I also changed standalone to use distutils to find `vcvarsall.bat` for us. If no explicit location is given as a preference, then it starts with version 16 and goes down to 9 until it finds something, so it should pick the newest compiler available. That's ok, right? At least it works on my Windows VM with the C++ for Python (i.e. VS 2008) compiler. I guess it would be nicer if one could explicitly ask for a MSVC version, for now the only way is to give the full path to `vcvarsall.bat` as a preference. On the other hand, this is something that only very few users will ever want to do.

Unfortunately, the appveyor service has some problem currently, builds are not being processed. But maybe we can go ahead merging it if it works on your machine(s) as well? Travis is passing.

mstimberg:
@thesamovar see my previous comment -- I guess we are good to merge?

thesamovar:
I need to test on my machines, will do that later this afternoon / early evening and merge if it is OK.

mstimberg:
Ok great, thanks.

thesamovar:
All tests pass on 3 different windows machines, so I'm going to merge now.

PR #468, opened 2015-05-22
Prevent race condition when making cython_extensions/.

dabliss:
When running parallel simulations and using cython as the code generator, I've been encountering a race condition in the creation of the ~/.brian/cython_extensions directory.  This commit avoids that race condition.

### DIFF
--- a/brian2/codegen/runtime/cython_rt/extension_manager.py
+++ b/brian2/codegen/runtime/cython_rt/extension_manager.py
@@ -44,8 +44,11 @@ class CythonExtensionManager(object):
         code = deindent(code)
 
         lib_dir = os.path.expanduser('~/.brian/cython_extensions')
-        if not os.path.exists(lib_dir):
+        try:
             os.makedirs(lib_dir)
+        except OSError as e:
+            if not os.path.exists(lib_dir):
+                raise OSError(e)
 
         key = code, sys.version_info, sys.executable, Cython.__version__
             
--- a/brian2/codegen/runtime/cython_rt/extension_manager.py
+++ b/brian2/codegen/runtime/cython_rt/extension_manager.py
@@ -8,6 +8,12 @@ import imp
 import os
 import sys
 import time
+import errno
+try:
+    import msvcrt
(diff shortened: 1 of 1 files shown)


thesamovar:
Thanks again! This looks good to me once the tests pass.

dabliss:
hold on -- there's one more race condition that follows that one.  second commit is on the way.

dabliss:
third commit for windows support on the way.

dabliss:
OK, this is all set.  i've added a decent amount of code.  let me know if you have any questions or concerns.  i'm happy to adjust things as you see fit.

mstimberg:
Hi and many thanks for working on this! The approach looks well thought out, but we are getting into dangerous territory here and I think we should be very careful to not break things for users (especially as this is an area where automatic testing is difficult). For example, does `flock` work correctly if the directory is shared NFS drive? I think it should on modern systems, but there might be situations where it does not (I have no idea about Mac OS X for example) and in this case we would break Cython for users running a single process. Maybe we have to wrap the lock calls in another try/except to catch situation where it fails? And/or have a global preference `codegen.runtime.cython.multiprocess_safe` that can be switched off (but defaults to on) if not needed and will remove all the file locking/exclusive opening?

A few stylistic comments:
* Let's not use `NameError` to detect that there's no `msvcrt`. In other places in the code we use the following idiom:
```Python
try:
    import msvcrt
except ImportError:
    msvcrt = None
...
if msvcrt:
    ...
else:
    ...
```
* Why `import fcntl as F`? A simple `import fcntl` would be clearer, I think
* The combination of `msvcrt.locking`/`fcntl.lock` that is used twice should be wrapped in a little helper function (e.g. `def _lock_file(...):`)

dabliss:
thanks for taking the time to read this over and for your edits.  i'm not sure how this would perform in the various NFS situations you mentioned.  both of the workarounds you suggested seem reasonable to me, but i'm not sure which option would make the most sense.  if you decide on one, let me know; i'd be happy to write it up.

PR #471, opened 2015-05-27
Prevent cython race condition.

dabliss:
I made a couple of git faux pas and ended up with the code in a new branch (with the same name).  The implementation here is cleaner than was the original pull request (though it retains the same dangers).  (I didn't mean to have this end up as a new request.)


mstimberg:
Just for reference, previous pull request/discussion was here: #468 
Looks very good, so you simplified the approach to always use a lock file instead of trying to exclusively open the file, right?

With this approach, it should be straightforward to include a preference that decides whether this mechanism is used or not. It should be switched on (i.e. use the lock file) by default but that way, in case of problems it will be easy to switch it off.

The preference should go into the `codegen.runtime.cython` category, i.e. added to the other preferences in the `cython_rt.py` module. I'm not sure about a good name `multiprocess_safe` maybe? @thesamovar, what do you think?

mstimberg:
Just one more minor thing: you don't seem to need the `import errno` anymore.

thesamovar:
Yep, preference and name seem good to me.

dabliss:
update: i removed the `errno` import, moved the code within the `with` block for the opened lock file to a separate method (`_load_module`), and added `multiprocess_safe` as a preference.

happy to incorporate new edits, stylistic or otherwise.

dabliss:
i'm not sure what to make of the AppVeyor build failure.  could this have something to do with my amending one commit and then shortly afterwards pushing another (before the checks had finished)?

mstimberg:
> i'm not sure what to make of the AppVeyor build failure. could this have something to do with my amending one commit and then shortly afterwards pushing another (before the checks had finished)?

No, that should be fine. It seems the conda installer crashed while installing the dependencies, no idea why... I restarted the test, let's see whether it builds fine now.

Otherwise it looks good to me now, so from my side this is ready to merge (pending the successful test run).

mstimberg:
Ok, the problem disappeared after a restart, so I'm going to merge this now. Note that there are some issues with Cython on Windows currently (fixed in another branch), so the builds on appveyor are not actually testing Cython...

Thanks again for your contribution.

thesamovar:
I never tested this on Windows: did you @mstimberg? If Appveyor is not testing Cython does that mean this has been merged without testing?

mstimberg:
I tested it on Windows, though not on the current master branch which does not work with Cython for me (before and after the merge). It does work with the fixes in the `appveyor_conda` branch however (which I'll soon prepare for merging) and it continues to work after merging it with the changes from this pull request.

I did not test multiprocessing with Cython, but since this was apparently broken before, it shouldn't do any harm.

Issue #169, opened 2013-11-14
Address inefficiencies in the SpikeQueue data structure


mstimberg:
Currently, there are two inefficiencies in the `SpikeQueue` implementations (both C++ and Python):
- [ ]  The "neurons to synapses" data structure is stored once per `SpikeQueue` -- if several pre or post pathways exist, this information is redundant across pathways
- [ ] In the `prepare` stage before a run, the integer delays and "neurons to synapses" data structures are re-created from scratch every time, even if nothing changed.

thesamovar:
This is fairly important for a 2.0 release I think. I don't remember what the current status of work on this is. Have we solved some or all of the problems?

mstimberg:
Only the scalar inefficiency is fixed, the other two still remain.
I don't think they are show-stoppers, though, the first one only affects multiple pre (or multiple post) pathways, a feature not used very often. The second is an inefficiency for multiple runs, but the unnecessary computation that is done involves a single pass over all synapses, this should be less than the time it takes to simulate a single time step.

thesamovar:
OK!

PR #475, opened 2015-06-08
Scalar delays in standalone

mstimberg:
In standalone mode, you could previously not use `Synapses(..., delay=1*ms)`, which sets a scalar delay for all synapses. Also, setting a scalar delay in this way was very inefficient because at the initialization phase of the C version of the spike queue, this delay was always stored once per synapse (see #169).

This pull request fixes those two issues, scalar delays are now stored as single values and they work for standalone as well (this also makes the "Brunel_Hakim_1999.py" and "Sturzl_et_al_2000.py" examples work in standalone mode now).

### DIFF
--- a/brian2/core/variables.py
+++ b/brian2/core/variables.py
@@ -1473,7 +1473,7 @@ class Variables(collections.Mapping):
 
     def add_dynamic_array(self, name, unit, size, dtype=None, constant=False,
                           constant_size=True, read_only=False, unique=False,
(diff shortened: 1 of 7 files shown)


thesamovar:
Code looks good to me, happy for it to be merged as soon as tests pass.

mstimberg:
Ok thanks, all tests pass. Merging it now.

PR #476, opened 2015-06-10
[MRG] Allow to change the spikes generated by SpikeGeneratorGroup between runs

mstimberg:
A feature that was often requested is to change the spike times generated by a `SpikeGeneratorGroup` after its creation. This makes it easier to run several trials of an experiment, for example, or to simulate a multi-layer feedforward network by simulating a single layer and feeding its output into the layer.

The changes necessary to allow this are not that big, spike times and indices are now stored in dynamic arrays instead of static arrays. I'm not sure whether `SpikeGeneratorGroup.set_spikes` is the best name for the method that allows to change the spike times, though.

Oh, and I sneaked in a convenient access to get the spikes from a single neuron in a `SpikeMonitor` with a dictionary approach. In Brian 1 this was possible using `spike_mon.spiketimes[index]` but I went directly for `spike_mon[index]`, I did not see a clear reason for using the extra attribute.


thesamovar:
Maybe write a more efficient version of the getitem for SpikeMonitor? At the moment we have O(numspikes) Python operations and it returns a list. Could do this with O(1) and return an array by using numpy. Also, it might be better to cache the results. I have some very efficient code for doing this in KlustaKwik, here: https://github.com/kwikteam/klustakwik2/blob/master/klustakwik2/clustering.py#L615

Apart from that, code looks good to me and I'm happy with the name.

thesamovar:
One option, if we are happy to have spikes not in temporal order, is when you request getitem, you do the sorting operation as in the partition cluster method in KK2, like that you can access all the spikes from the same neuron without having to copy the data.

mstimberg:
> Maybe write a more efficient version of the getitem for SpikeMonitor? At the moment we have O(numspikes) Python operations and it returns a list.

It did not return a list (`[...]*second` results in a `Quantity`) but it was certainly slow, I improved this a bit. It remains slow if the array is very big, though.

> Could do this with O(1) and return an array by using numpy.

I don't quite see how it could be `O(1)`, even the creation of the result array will already be `O(numspikes)`, no? And the array is not sorted by spike index but instead by spike time, so we can't do a fast search. I'm not sure it is really worth improving on this, though, since this is only meant as a convenience syntax and it certainly is fast enough if you are interested in the values for a single or a few neurons.

> Also, it might be better to cache the results.

I thought about this, but this will complicate things significantly, we'd need a "dirty" flag etc. As above, not sure this is worth the effort.

thesamovar:
Oops I meant O(1) Python operations, i.e. what your new version does.

I seem to remember we went through this on Brian 1, initially not cacheing and then having people complain that it was very slow. The thing is, if you provide the facility for getting just the spikes of one neuron, people will just use it. So I think if we provide the facility to get just the spikes of neuron n, we should make it efficient.

I think the cacheing is not too hard. You just:
- Sort into index order rather than spike order
- Store the offsets into this array (i.e. a ragged array with a contiguous store), this means you get very fast retrieval and no memory costs
- Use the length of the array as the 'dirty' flag (but make sure to set it as dirty when you resize to 0): this works because you only resize to 0 or add spikes, and it only means making a fairly minimal number of changes

I'm happy to do this but probably won't get around to it for a few days.

An alternative option is that we remove getitem, but we provide a convenience function that takes an unsorted array of spikes and returns a partitioned array.

mstimberg:
> The thing is, if you provide the facility for getting just the spikes of one neuron, people will just use it.

I see what you mean, that's a valid point.

> Sort into index order rather than spike order
> Store the offsets into this array (i.e. a ragged array with a contiguous store), this means you get very fast retrieval and no memory costs

Right, but that means re-sorting it after the run -- when would we do this, only when acessing the spikes in this way? And would we store an additional copy or replace the original data? I don't really like the idea of changing stored data when accessing it...

> Use the length of the array as the 'dirty' flag (but make sure to set it as dirty when you resize to 0): this works because you only resize to 0 or add spikes, and it only means making a fairly minimal number of changes

I'm sure it is doable but I'm still worried that in between all the possibilities of runtime and standalone mode, restoring previous network states etc. we forget about something and end up with surprising errors.

> An alternative option is that we remove getitem, but we provide a convenience function that takes an unsorted array of spikes and returns a partitioned array.

Hmm, actually that might be a good idea. Or maybe not a function that takes spikes but rather a `SpikeMonitor.to_dict` that returns a dictionary of all the data (or maybe even with an optional `indices` argument that allows you to get a subset). This way you could write rather efficient and convient code like this:
```Python
spike_dict = mon.to_dict()
for idx in xrange(N):
   do_something(spike_dict[idx])
```

Of course users could still shoot themselves in the foot by doing
```Python
 for idx in xrange(N):
   do_something(spike_mon.to_dict()[idx])
```
but there the inefficiency would be quite obvious, much more obvious than in
```Python
 for idx in xrange(N):
   do_something(spike_mon[idx])
```

I think I might prefer this solution, actually. It's straight-forward and easy to understand and we are not messing with any internal data structures.

thesamovar:
That's a great solution, let's go with that. Probably the most efficient algorithm would be sorting on a copy of the data (that way we don't change the original data, which indeed is probably a good idea). I'd call the method something like ``.partition()`` maybe rather than ``.to_dict()``? Or ``.spike_trains()``? Something semantic rather than the data structure of the return type.

mstimberg:
> I'd call the method something like .partition() maybe rather than .to_dict()? Or .spike_trains()? 
Something semantic rather than the data structure of the return type.

I fully agree, `to_dict` is a crappy name. `partition` still sounds a bit too technical for me, I prefer `spike_trains` (or maybe `as_spike_trains`...?). I'll put something together following this approach.

I should not have added this change to this branch, now we are having a full-fledged discussion about this in an unrelated pull request :-/

mstimberg:
I removed the previous approach and added a `spike_trains` method. The implementation is certainly not optimal but it ran reasonably fast on big numbers of spikes. We can always do non-straightforward optimizations later. The code has a somewhat inelegant loop, mostly because I wanted to make sure that `spike_trains[idx]` will not raise a `KeyError` if neuron `idx` did not spike but instead return an empty array (and conversely, I did not want to use a `defaultdict` because that would not raise a `KeyError` for neurons that were not recorded/illegal indices).
From my side this is ok to merge.

thesamovar:
Looks great to me. I think you can avoid the inelegant part of the loop with the trick I used in the KK code I posted, but it's really not important because it won't be any more efficient and we have something that works now! :) I'm happy for this to be merged as soon as the tests all pass.

mstimberg:
Tests pass, I'm merging now. I'm happy to postpone making the code more elegant, especially given that it is more of a convenience function than an elementary part of Brian.

PR #483, opened 2015-06-12
[MRG] Build and upload binary conda packages (for Windows and Linux)

### LINKED ISSUE #466: Test and recommend MS VC for Python
Microsoft now has a compiler specifically for Python 2.6-3.2 that claims to work for 32/64 bit without patching. We should test this on a fresh install, and if it's true then we should recommend it in our docs. http://www.microsoft.com/en-us/download/details.aspx?id=44266

mstimberg:
This package fixes various issues with the compilation on Windows that prevented the use of the C++ for Python compiler and builds binary conda packages (for Windows/Linux + 32/64 Bit + Python 2.7/3.4). It also adds two features to the test suite:
* Tests can be run in parallel (this is enabled by default for all targets except for weave) -- in addition to dramatically speeding up local testing it also means we are finally testing that the Cython patch for multiprocessing actually works  (note that parallel testing is disabled on appveyor, on the small single core VM used there, parallel testing slowed testing down)
* The test suite can optionally not reset all the preferences to their default values. This is important for testing a 32Bit package on 64 Bit (on travis, we only have 64 Bit machines), since code generation needs to pass flags to the compiler for 32 bit compilation

This is not final yet, mostly documentation is lacking. Note that the conda packages currently end up in a "dev" channel on binstar to avoid confusion with released versions. The idea is that every commit/merge to the master branch (but not other branches or pull request) will update the conda package in the dev channel. While it is nice to have the option to point users that want to use the latest Brian development version to this channel, the main reason to build these packages is that we can spot problems early instead of building the binary package for the first time when we do a release (building the final release packages is not implemented here yet). With these packages, all that is needed for installing brian2 is:
```
conda install -c brian-team/channel/dev brian2
```

There's still the issue of `std::move` which does not work with the old C++ compiler -- @thesamovar can you check whether simply leaving it away works with a recent version of visual studio?

Pending an update of the installation instructions, this closes #466


thesamovar:
I'll check this, but I think it's necessary on the newer versions of VS, so we might need to do a #define to get around the problem.

thesamovar:
I ran the parallel nose tests on my multicore Windows machine and I get the same behaviour as on AppVeyor, namely it took 672s to run the codegen independent tests compared to 22s on a single core with the no-parallel option specified. It was also only using one core (based on the CPU usage reported by task manager). Additionally, there were 4 test failures on the 'codegen independent' tests, each:

```
No module named msvccompiler in numpy.distutils; trying from distutils
Missing compiler_cxx fix for MSVCCompiler
```

This didn't happen with the no-parallel option, and I can verify that it was calling MSVC because cl.exe appeared in the task manager. So something strange is going on. It may be due to the way multiprocessing works on Windows, it creates a new process that has to reimport all the packages rather than forking (because forking doesn't exist on Windows). This could mean that setuptools doesn't get to do its monkey patching because the order of imports is different, I'm not sure. It also explains why it's slow: I'm guessing that nose is creating a new process for each task, rather than doing a one off creation of a pool of N processes and then reusing them for the different tests. If I'm right about all of this, the parallel testing is unavoidably broken on Windows due to the design of nose, and should be disabled.

It's also worth noting parenthetically that doctests are not codegen-independent. Perhaps these should be run in a separate block for clarity?

On Visual Studio 2010, it works without std::move. I still need to check 2013.

thesamovar:
OK so ``std::move`` leads to an error on MSVC2013. I haven't tried 2012 but apparently 2013 is a largely cosmetic rerelease of 2012 so I guess the same problem will be there. Given this, I suggest we do a ``#if defined(_MSC_VER) && (_MSC_VER>=1700)``. I'll do that now.

thesamovar:
OK I fixed the two problem I had on my Windows machines, and I'm happy with the rest of the code, so if the tests pass then I'm happy for this to be merged.

thesamovar:
I'm actually getting some test failures so don't merge yet. Will update when I have more info.

thesamovar:
The problem was that the symbol INFINITY wasn't defined and this was causing main.cpp not to compile. I added the common_math.h header file to main.cpp and now it seems to be running. Waitin on a full test run on my computer, but I think it's fixed now.

mstimberg:
Ok great, I think I'll first update the install documentation in this branch before merging it. And maybe add code so that we can do uploads for releases to conda as well.

thesamovar:
OK complete tests pass on my machine.

mstimberg:
I rewrote the installation instructions, the recommended installation is now based on Anaconda. I also simplified the installation instructions, and tested the instructions for Windows (64 Bit), both for Python 2 and Python 3. Feel free to change anything that you find unclear, from my side this is now ready to merge.

PS: Ignore the "failed" builds due to the non-working circleCI testing on Mac OS X, I deactivated it now.

thesamovar:
This all looks great to me and happy for it to be merged as soon as the tests pass.

thesamovar:
Is it just me or is Travis getting slower and slower?

PR #484, opened 2015-06-12
[MRG] Fix (and test) the HTML representation for `NeuronGroup`

### LINKED ISSUE #460: NeuronGroup's "rich representation" for ipython does no longer work
It refers to a non-existing `method` attribute of `StateUpdater`

mstimberg:
Closes #460

This is a trivial fix that simply removes the lines that stated the used state update method from the HTML output of a `NeuronGroup`. The state update method is not necessarily fixed at this point yet and I don't think getting it to work here for the cases when it is (the user explicitly provided a `method` argument) is worth the effort.


thesamovar:
Good to merge when tests pass.

PR #485, opened 2015-06-12
[MRG] Allow instance methods as network operations

### LINKED ISSUE #474: Allow instance methods as network operations
Currently, only module-level functions can be used as network operations. For object methods, the `self` reference is counted as an argument and therefore the method will receive the clock's time (instead of the correct reference). In complex frameworks it might be helpful to have network operations as object methods of a big "Simulation" object and there's no real reason why we could not support this.

mstimberg:
Not a very important feature, but in more complex frameworks it might be useful to have a network operation that accesses some parent object state via `self`. This pull request fixes this issue but more importantly does more strict error checking (before, an instance method with the `self` argument would be interpreted as a function with a single argument, i.e. the current time was passed as `self`, and functions with more than one argument did not raise an error until executed).
 
Closes #474

### DIFF
--- a/brian2/core/operations.py
+++ b/brian2/core/operations.py
@@ -1,3 +1,5 @@
+import inspect
+
 from brian2.core.base import BrianObject
 
 __all__ = ['NetworkOperation', 'network_operation']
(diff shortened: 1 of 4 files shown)


thesamovar:
Looks good to me. Happy for you to merge it when the tests pass.

PR #486, opened 2015-06-16
[MRG] Faster resize for dynamic array

mstimberg:
This branch improves the performance of the resize operation for dynamic arrays quite significantly (in a long running complex example the time for `statemonitor` went down from 45s to 17s). It mostly does it by including a special function for resizing along the first dimension for multi-dimensional arrays.

I briefly played around with storing the size of the underlying data array in a separate variable (so that the weave code can directly check whether a resize is necessary), but this makes things very fragile because not calling `resize` also means that the numpy array view on the underlying data has the wrong size, so we have to update this in the end of the run, etc. I got the impression that it is easy to introduce hard-to-debug problems (say a user wants to access the data stored in a `StateMonitor` in a network operation) this way, and this should certainly not be rushed just for better performance. I therefore only did changes I'm reasonably confident to be safe.

This pull request also contains a small change that marks the `dt` variable as constant during a run, this prevents an unnecessary namespace update every time step.

Ready to merge from my side.

### DIFF
--- a/brian2/codegen/runtime/cython_rt/templates/statemonitor.pyx
+++ b/brian2/codegen/runtime/cython_rt/templates/statemonitor.pyx
@@ -8,10 +8,8 @@
     cdef int _curlen = {{_dynamic_t}}.shape[0]
     cdef int _new_len = _curlen + 1
 
-    # Resize the arrays
-    _owner.resize(_new_len)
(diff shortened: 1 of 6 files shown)


thesamovar:
Looks good to me, happy for this to be merged once tests pass.

mstimberg:
Tests pass, merging now.

Issue #106, opened 2013-08-24
Rename 'reset'?


thesamovar:
The name 'reset' is actually slightly misleading in the sense that we will often use it not just as a reset but as a general spike-triggered event code to do things like modify adaptation variables. Perhaps a better name that reflects that? I can't immediately think of a good idea.

romainbrette:
I think we should first see a few examples.

thesamovar:
My thoughts:
- I'm happier with threshold than reset
- I still think there's a good case to leave it as it is to not break backwards compatbility
- I think on_spike is potentially confusing because it suggests that you'll be defining what happens when a spike occurs, which sounds like it would include the synaptic behaviour to me
- How about ``spike_if='v>1', after_spike='v=0'``?

thesamovar:
Or maybe ``spike_when='v>1'``?

romainbrette:
spike_when and after_spike are not bad!

There's indeed the issue of backwards compatibility. Perhaps we could 
issue a "Deprecated" warning when using threshold/reset.

mstimberg:
Sorry, to be the annoying guy here, but I think that `spike_when` and `after_spike` are one step forward but also one step back at the same time: since they have `spike` in the name, they will look odd for models where it is not actually about a spike but some other general "regime change". For example, in our ongoing modeling work for astrocytes we use the threshold/reset mechanism to release gliotransmitters when the Calcium levels crosses a threshold. I feel that `threshold` is a more general term here (even though `threshold_condition` would be more appropriate, but it's a bit long). But I also agree that `reset` is not appropriate in many contexts (including the astrocyte modeling) though.

Something that we should keep in mind is that we want to generalize to other event types in the future (#105). Will we do this with a separate `events` keyword arguments or generalize `threshold`/`reset`, i.e. allow to pass dictionaries to them?

Whatever we decide, `threshold`/`reset` are very central in almost every Brian model, so we should definitely have a transition period where they still work but lead to a deprecation warning (and later have a helpful error message)

romainbrette:
event_when and after_event would be more general. Possibly more obscure 
too though.

thesamovar:
We did decide to do more general types of events, so my suggestion would be to use ``spike_when``, ``after_spike`` for the spike event, and have a more general syntax for different types of events (maybe ``events=...``). At the moment, we're using the spike event for non-spike events as a sort of hack, but that should disappear if we allow for other types of events.

mstimberg:
[...]

PR #487, opened 2015-06-18
[MRG] Support for general events

mstimberg:
This introduces support for general events using the `events = {eventname: (condition, statement)}` syntax and an `EventMonitor` for recording such events. It is lacking tests, error-checking (e.g. for valid event names) and documentation but it seems to work fine.
Many of the tests fail currently, but that's because of a little change that we won't create a `_spikespace` array unconditionally anymore, but only if the `NeuronGroup` has a threshold. This makes quite a number of `Synapses` tests fail, where for simplicity, `Synapses` is connected to a `NeuronGroup` without a threshold. I'm not sure whether that's a bug or a feature, though -- but the error message is certainly not helpful at the moment.
One thing that came to my mind is that in the same way that this pull request generalizes threshold/reset/monitors for multiple events, synaptic pathways could easily connect to a different synaptic event type than spikes, which could be potentially useful (in our discussion in #106 we decided to not do this for simplicity, but the necessary changes are really not that big). Maybe worth opening a new issue for that (and/or discuss it in our meeting tomorrow).

This branch does not (yet):
* Rename reset/threshold ( #106)
* Rename the scheduling slots `resets` and `thresholds`
* Rename the templates `reset`, `threshold`, `spikemonitor` (which is potentially annoying for other people already working on devices, on the other hand we should make this change rather now or never)

### DIFF
--- a/brian2/codegen/runtime/cython_rt/templates/reset.pyx
+++ b/brian2/codegen/runtime/cython_rt/templates/reset.pyx
@@ -1,16 +1,16 @@
 {% extends 'common.pyx' %}
 
 {% block maincode %}
-    {# USES_VARIABLES { _spikespace } #}
-    
     # scalar code
     _vectorisation_idx = 1
     {{scalar_code|autoindent}}
+    {#  Get the name of the array that stores these events (e.g. the spikespace array) #}
+    {% set _eventspace = get_array_name(eventspace_variable) %}
 
-    cdef int _num_spikes = {{_spikespace}}[_num{{_spikespace}}-1]
-    for _index_spikes in range(_num_spikes):
+    cdef int _num_events = {{_eventspace}}[_num{{_eventspace}}-1]
+    for _index_events in range(_num_events):
         # vector code
-        _idx = {{_spikespace}}[_index_spikes]
+        _idx = {{_eventspace}}[_index_events]
         _vectorisation_idx = _idx
         {{ vector_code | autoindent }}
 {% endblock %}
--- a/brian2/codegen/runtime/cython_rt/templates/spikemonitor.pyx
+++ b/brian2/codegen/runtime/cython_rt/templates/spikemonitor.pyx
@@ -1,33 +1,37 @@
 {% extends 'common.pyx' %}
 
 {% block maincode %}
-    {# USES_VARIABLES { t, i, _clock_t, _spikespace, _count,
+    {# USES_VARIABLES { t, i, _clock_t, _count,
                         _source_start, _source_stop} #}
-    cdef int _num_spikes = {{_spikespace}}[_num{{_spikespace}}-1]
+
+    {#  Get the name of the array that stores these events (e.g. the spikespace array) #}
+    {% set _eventspace = get_array_name(eventspace_variable) %}
+
+    cdef int _num_events = {{_eventspace}}[_num{{_eventspace}}-1]
     cdef int _start_idx, _end_idx, _curlen, _newlen, _j
     cdef double[:] _t_view
     cdef int32_t[:] _i_view
-    if _num_spikes > 0:
+    if _num_events > 0:
         # For subgroups, we do not want to record all spikes
         # We assume that spikes are ordered
         # TODO: Will this assumption ever be violated?
-        _start_idx = _num_spikes
-        _end_idx = _num_spikes
-        for _j in range(_num_spikes):
-            _idx = {{_spikespace}}[_j]
+        _start_idx = _num_events
+        _end_idx = _num_events
+        for _j in range(_num_events):
+            _idx = {{_eventspace}}[_j]
             if _idx >= _source_start:
                 _start_idx = _j
                 break
-        for _j in range(_start_idx, _num_spikes):
-            _idx = {{_spikespace}}[_j]
+        for _j in range(_start_idx, _num_events):
+            _idx = {{_eventspace}}[_j]
             if _idx >= _source_stop:
                 _end_idx = _j
                 break
-        _num_spikes = _end_idx - _start_idx
-        if _num_spikes > 0:
+        _num_events = _end_idx - _start_idx
+        if _num_events > 0:
             # Get the current length and new length of t and i arrays
             _curlen = {{_dynamic_t}}.shape[0]
-            _newlen = _curlen + _num_spikes
+            _newlen = _curlen + _num_events
             # Resize the arrays
             _owner.resize(_newlen)
             # Get the potentially newly created underlying data arrays
@@ -35,7 +39,7 @@
             _i_view = {{_dynamic_i}}.data
             # Copy the values across
             for _j in range(_start_idx, _end_idx):
-                _idx = {{_spikespace}}[_j]
+                _idx = {{_eventspace}}[_j]
                 _t_view[_curlen + _j - _start_idx] = _clock_t
                 _i_view[_curlen + _j - _start_idx] = _idx - _source_start
                 {{_count}}[_idx - _source_start] += 1
--- a/brian2/codegen/runtime/cython_rt/templates/spikemonitor.pyx
+++ b/brian2/codegen/runtime/cython_rt/templates/spikemonitor.pyx
@@ -1,7 +1,7 @@
 {% extends 'common.pyx' %}
 
 {% block maincode %}
-    {# USES_VARIABLES { t, i, _clock_t, _count,
+    {# USES_VARIABLES { t, i, _clock_t, count,
                         _source_start, _source_stop} #}
 
     {#  Get the name of the array that stores these events (e.g. the spikespace array) #}
@@ -42,6 +42,6 @@
                 _idx = {{_eventspace}}[_j]
                 _t_view[_curlen + _j - _start_idx] = _clock_t
                 _i_view[_curlen + _j - _start_idx] = _idx - _source_start
-                {{_count}}[_idx - _source_start] += 1
+                {{count}}[_idx - _source_start] += 1
                 
 {% endblock %}
--- a/brian2/codegen/runtime/cython_rt/templates/threshold.pyx
+++ b/brian2/codegen/runtime/cython_rt/templates/threshold.pyx
@@ -2,17 +2,19 @@
 {# USES_VARIABLES { N } #}
 
 {% block maincode %}
-    {# USES_VARIABLES {_spikespace } #}
-    # t, not_refractory and lastspike are added as needed_variables in the
-    # Thresholder class, we cannot use the USES_VARIABLE mechanism
-    # conditionally
+    {# t, not_refractory and lastspike are added as needed_variables in the
+       Thresholder class, we cannot use the USES_VARIABLE mechanism
+       conditionally #}
 
     # scalar code
     _vectorisation_idx = 1;
     {{ scalar_code | autoindent }}
 
-    cdef long _cpp_numspikes = 0
-    
+    cdef long _cpp_numevents = 0
+
+    {#  Get the name of the array that stores these events (e.g. the spikespace array) #}
+    {% set _eventspace = get_array_name(eventspace_variable) %}
+
     for _idx in range(N):
         
         # vector code
@@ -20,13 +22,13 @@
         {{ vector_code | autoindent }}
 
         if _cond:
-            {{_spikespace}}[_cpp_numspikes] = _idx
-            _cpp_numspikes += 1
+            {{_eventspace}}[_cpp_numevents] = _idx
+            _cpp_numevents += 1
             {% if _uses_refractory %}
             {{not_refractory}}[_idx] = False
             {{lastspike}}[_idx] = t
             {% endif %}
             
-    {{_spikespace}}[N] = _cpp_numspikes
+    {{_eventspace}}[N] = _cpp_numevents
     
 {% endblock %}
--- a/brian2/codegen/runtime/numpy_rt/templates/reset.py_
+++ b/brian2/codegen/runtime/numpy_rt/templates/reset.py_
@@ -11,6 +11,6 @@ _vectorisation_idx = 1
 {% set _eventspace = get_array_name(eventspace_variable) %}
 
 # vector code
-_idx = {{_eventspace}}[:{{_spikespace}}[-1]]
+_idx = {{_eventspace}}[:{{_eventspace}}[-1]]
 _vectorisation_idx = _idx
 {{vector_code|autoindent}}
--- a/brian2/codegen/runtime/numpy_rt/templates/reset.py_
+++ b/brian2/codegen/runtime/numpy_rt/templates/reset.py_
@@ -1,5 +1,3 @@
-{# USES_VARIABLES { _spikespace } #}
-
 {# Note that we don't write to scalar variables conditionally. The scalar code
    should therefore only include the calculation of scalar expressions
    that are used below for writing to a vector variable #}
@@ -9,7 +7,10 @@ import numpy as _numpy
 _vectorisation_idx = 1
 {{scalar_code|autoindent}}
 
+{#  Get the name of the array that stores these events (e.g. the spikespace array) #}
+{% set _eventspace = get_array_name(eventspace_variable) %}
+
 # vector code
-_idx = {{_spikespace}}[:{{_spikespace}}[-1]]
+_idx = {{_eventspace}}[:{{_spikespace}}[-1]]
 _vectorisation_idx = _idx
 {{vector_code|autoindent}}
--- a/brian2/codegen/runtime/numpy_rt/templates/spikemonitor.py_
+++ b/brian2/codegen/runtime/numpy_rt/templates/spikemonitor.py_
@@ -1,4 +1,4 @@
-{# USES_VARIABLES {i, t, _count, _clock_t, _source_start, _source_stop} #}
+{# USES_VARIABLES {i, t, count, _clock_t, _source_start, _source_stop} #}
 import numpy as _numpy
 
 {#  Get the name of the array that stores these events (e.g. the spikespace array) #}
@@ -17,4 +17,4 @@ if _n_events > 0:
     {{_dynamic_t}}[_curlen:_newlen] = _clock_t
     {{_dynamic_i}}[_curlen:_newlen] = _events
 
-    {{_count}}[_events] += 1
+    {{count}}[_events] += 1
--- a/brian2/codegen/runtime/numpy_rt/templates/spikemonitor.py_
+++ b/brian2/codegen/runtime/numpy_rt/templates/spikemonitor.py_
@@ -1,17 +1,20 @@
-{# USES_VARIABLES {i, t, _spikespace, _count, _clock_t, _source_start, _source_stop} #}
+{# USES_VARIABLES {i, t, _count, _clock_t, _source_start, _source_stop} #}
 import numpy as _numpy
 
-_spikes = {{_spikespace}}[:{{_spikespace}}[-1]]
+{#  Get the name of the array that stores these events (e.g. the spikespace array) #}
+{% set _eventspace = get_array_name(eventspace_variable) %}
+
+_events = {{_eventspace}}[:{{_eventspace}}[-1]]
 # Take subgroups into account
-_spikes = _spikes[(_spikes >= _source_start) & (_spikes < _source_stop)]
-_spikes -= _source_start
-_n_spikes = len(_spikes)
-if _n_spikes > 0:
+_events = _events[(_events >= _source_start) & (_events < _source_stop)]
+_events -= _source_start
+_n_events = len(_events)
+if _n_events > 0:
 
     _curlen = len({{_dynamic_t}})
-    _newlen = _curlen + _n_spikes
+    _newlen = _curlen + _n_events
     _owner.resize(_newlen)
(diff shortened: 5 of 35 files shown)


thesamovar:
I rather like the idea of getting rid of ``_spikespace`` and just having threshold/reset be shorthands for the default ``spikes`` event. I also like your idea of Synapses connecting to arbitrary events (with ``spikes`` being just the default). Like that, we have one mechanism for events, and spikes and Synapses are just a special case (with special syntax to make it simpler, but not special in terms of implementation).

mstimberg:
I opened #488 for the synapses. I do rather like the idea of general events as well :)

mstimberg:
Ok, this all seems to work now. It is undocumented and lacks tests and nice error message, but let's make first sure that we all (CC @romainbrette) agree on the syntax:

**NeuronGroup**
`events` keyword takes a dictionary mapping event names to `(condition, statements)` tuples:
```Python
group = NeuronGroup(N, '...', threshold='...', reset='...',
                    events={'my_event': ('x>1', 'x=0')})
```
As a convenience syntax, if there's no "reset statement" for a custom event, a single string instead of a tuple can be used:
```Python
group = NeuronGroup(N, '...', threshold='...', reset='...', events={'my_event': 'x>1'})
```
Note that all this means that `threshold` and `reset` are only there for convenience (but we all agree to keep them there), specifying `threshold=threshold` and `reset=reset` is completely equivalent to:
```Python
group = NeuronGroup(N, '...', , events={'spike': (threshold, reset)})
```

**EventMonitor**
Custom events can be recorded with an `EventMonitor`:
```Python
event_mon = EventMonitor(group, 'my_event')
```
Again, `SpikeMonitor(group)` is redundant (but we leave it there), it is simply implemented as `EventMonitor(group, 'spike')`

**Synapses**
The `Synapses` class now takes a `events` keyword as well, specifying either an event for all pathways (by default: `spike`):
```Python
S = Synapses(group, group, '...', pre='...', post='...', events='my_event')
```
or a dictionary mapping pathway names to event names, where only non-standard names have to be specified, i.e. everything unspecified defaults to `'spike'`:
```Python
S = Synapses(group, group, '...', pre={'transmission': '...', 'plasticity': '...'},
             post='...', events={'plasticity': 'my_event'})
# transmission and post pathways use "spike" event
# plasticity pathway uses "my_event"
```

The nice thing about all this is that it is 100% backwards-compatible from the user point of view, but as Dan said, on the implementation side there's nothing special about spike events at all.

thesamovar:
I like everything except the tuple (condition, statements). Two alternative options:

```
1. event_conditions={'name': 'condition'}, event_code={'name':'statements'}
2. events={'name': Event(condition='condition', code='statements')}
```

I think I prefer (1) of these.

mstimberg:
I agree it's not great syntax, but it has the advantage over your proposal (1) that the event name is not repeated (and we don't need checks that each statement corresponds to a condition). I think we don't want to introduce a dummy `Event` class that is just a simple container and doesn't have any functionality. If I had to chose between your two options, I'd go for (1).

mstimberg:
Another option, given that there's already a quite large number of arguments to the `NeuronGroup` initializer, we could also take the events definition out of it completely:
```Python
G = NeuronGroup(...)
G.add_event('my_event', condition='v>1', code='v=0')

**Comment by thesamovar** [comment]
Not a big fan of the method, feels inconsistent with the way most of Brian works (particularly threshold/reset). As an alternative that I think I don't like, but just throwing it out there for consideration:

```
events = '''
name: condition
    statement_1
    statement_2
    ...
name2: condition2
   ....
'''
G = NeuronGroup(..., events=events)
```

But I think I would actually prefer the tuple over this option.

Another couple of variants on my option (1). Think of the 'event' as being the thing itself, and the code as being action taken in response to the event. In which case, it makes sense to group the name and condition together, but keep the code separate. For example, threshold and reset and synapses are the definition of the event (threshold), the NeuronGroup action (reset) and the synaptic action (Synapses), so it would make less sense to keep definition and NeuronGroup action tied together but to keep the Synapses separate. So, variants:

```
1a. G  = NeuronGroup(..., events={'name':'condition'}, on_event={'name':'code'})
1b. G = NeuronGroup(..., events={'name':'condition'}, actions={'name':'code'})
```

And maybe actually this would fit with your method based idea:

```
3.
G = NeuronGroup(..., events={'name':'condition'})
G.on_event('name', 'code')
```

Now that I think about it, I like keeping the event handling code separate from the event definition. It feels more logical to me.

**Comment by mstimberg** [comment]
I see, good point. I quite like variant 3; given that `on_event` would be similar in spirit to `custom_operation`, maybe we could find a name that reflects that? One more thing: if we use the `events={'name': 'condition'}'` syntax in `NeuronGroup` then we should maybe not use the word `event` in `Synapses`, since this does not describe an event but binds the pre/post actions to events. Not sure about this though and I don't have a good alternative name.

**Comment by thesamovar** [comment]
We could also consider changing the name of custom operation, it's not particularly elegant. Why is network operation Python code, but custom operation is abstract code? I don't even remember why we chose it! :) So how about we leave network operation for backwards compatibility and because it's defined at the network level and not attached to a particular group, change custom operation to ``NeuronGroup.regular_operation`` and the new one would be ``NeuronGroup.event_operation``? The good thing about this is that it's a bit more consistent, the bad things are that it breaks syntax for custom operation (but that's allowed still) and that it makes key behaviour of the neuron group be defined potentially far from the definition of the group. But, this last point is true of custom operations already, and partly true of Synapses too.

For Synapses the name event is OK I think. Events just refers to events: a NeuronGroup is naturally a source of events whereas Synapses is naturally a handler of them so there's no ambiguity. Although, you were talking at some point about having synapses be able to trigger events? If so, then indeed we need a new name. How about ``event_source``, ``event_name``, ``source_event``, or ``from_event``?

**Comment by mstimberg** [comment]
> We could also consider changing the name of custom operation, it's not particularly elegant. Why is network operation Python code, but custom operation is abstract code? I don't even remember why we chose it! :)

All the evidence is out there, you can't deny that one year ago it "seemed good" to you :) (https://github.com/brian-team/brian2/issues/220#issuecomment-44647129). Anyway, I like `regular_operation` and `event_operation` -- maybe we could still allow `custom_operation` for now and have it raise a `DeprecationWarning`, we can then remove it later. 
I just realized there's one problematic thing though, the consistency is misleading:
```Python
G = NeuronGroup(N, '...')
G.regular_operation('...')
G.event_operation('event', '...')
```
spot the mistake? The `regular_operation` returns a `BrianObject` that needs to be added to the `Network`, so it would only be correct like this:
```Python
G = NeuronGroup(N, '...')
updater = G.regular_operation('...')
G.event_operation('event', '...')
```
On the other hand, this is something that users get wrong and maybe it wasn't a good idea in the first place. The rename would be a good opportunity to fix this, how about instead having the `BrianObject` added to the `contained_objects` of the group? To make this clearer, we could use `add_regular_operation`/`add_event_operation`? A `remove_...` would then make sense as well, I guess, but while for the event operation this can be easily called with the name of the event, for the regular operation we'd have to use the name of the `BrianObject`, which the user might not know. Then again, it would be a quite rarely used feature and the user could always explicitly specify a name in this case.

Regarding `Synapses`
> How about event_source, event_name, source_event, or from_event?

The term `source` is a bit problematic here, since it has a specific meaning in the context of synapses. But actually we could use that and have two arguments instead of one: `source_event` and `target_event`. I can imagine that in most use cases where you have both a `pre` and a `post` statement and you want to specify a non-standard event trigger, it will be different for `pre` and `post`. This slightly more verbose syntax would allow doing that without using dictionary syntax (instead of `event={'post': 'custom_event'}` you'd write `target_event='custom_event'`).

thesamovar:
I had a moment of weakness! ;)

Yeah we should be consistent about whether we return a ``BrianObject`` or we add to contained objects. Is there any obvious use case where you'd want to define a custom/regular operation and not add it to the Network? I guess there is actually, if you were using it to generate activity for a particular period, for example. Would the same be true for event operations? I don't think so: it would be weird to just ignore some events.

OK how about they are automatically added to contained objects but they also return the object. Like that you can remove it.

For Synapses I worry that we're overengineering it and that it would be simpler to just use ``event``. Otherwise, I'm happy with ``source_event`` and ``target_event``, but we'd still need ``event`` right? For those cases where you needed more control.

mstimberg:
You could still do the equivalent of creating a `custom_operation` and not adding it to the network with the approach I suggested: you'd do something like:
```Python
# ...
run(runtime1)
G.add_regular_operation('...', name='generate_activity')
run(runtime2)
G.remove_regular_operation('generate_activity')
run(runtime3)
```
> OK how about they are automatically added to contained objects but they also return the object. Like that you can remove it.

Sounds good, or have it return the name for an explicit remove method, I guess. But I'd also be fine with not having `remove` methods and making the user to call `G.contained_objects.remove(...)` directly.

> Otherwise, I'm happy with source_event and target_event, but we'd still need event right? For those cases where you needed more control.

No, I meant that `source_event` and `target_event` would still allow dictionaries, but you'd only need them if there's more than one `pre` or more than one `post` pathway (in contrast to the `event` solution, where you'd have to use them whenever not all pathways listen to the same event).

I think which solution is cleaner from the user point of view depends on the following question: if the user specifies a synaptic model with `pre` and `post` (e.g. STDP) *and* some pathway in this model is not using spike events, what is more likely:
* all pathways will use the same custom event (then `event=...` is better)
* pre and post will use different events (then `source_even`/`target_event` is better)

I think the second option is more likely since source and target population are often not identical and therefore probably do not have the same (non-standard) events. Actually, they might even have different events with the same name, and the distinction would be clearer with separate arguments. I therefore tend to prefer `source_event`/`target_event`, but I don't have a strong objection against `event` either. After all, we are talking about some syntax here that most users will never use.

thesamovar:
I don't really like the ``add_`` in the name, it feels unnecessarily long. So how about this?

```
ro = G.regular_operation(...)
eo = G.event_operation(name, ...)
G.remove(ro)
```

For Synapses, I think that as you say, almost all the time people are just using spikes, and the small fraction of the remainder, they're probably only doing events in one direction, so it doesn't really matter much. I'd say just go with ``event`` because having two keywords for something that will almost never be used seems weird. But, I'm not bothered if you prefer a different approach.

thesamovar:
Just had a look at the code which seems good to me. Is the plan to do the reset/threshold template renames in this branch? Also, do we add new slots for events or rename threshold/reset slots?

mstimberg:
> I don't really like the add_ in the name, it feels unnecessarily long.

it makes clear that something is added, though. Otherwise, especially if it also has a return value (for the remove later), I would expect that I have to do something with it so that it gets used.

I don't have a strong opinion on the keyword(s) for `Synapses`, either is fine for me.

About threshold/reset: if we decide to rename it, then we should do it here, I think. I'm happy to rename the templates, I'm not sure about the schedule slots, though. This is something user-facing and since in 99% of the cases there will be no custom events it would feel weird, if e.g. to record the membrane potential including the threshold crossing you'd have to use something general such as `when='after_conditions'` instead of `when='after_thresholds'`.
But that makes me think that in the specific use case of custom events, users might well need precise control over the slot where the condition is tested (for the `event_operation` we can simply have the `'when'` argument in the method so that's fine). I can't really think of a nice syntax for that...

thesamovar:
Yeah you're right, the method name should have a verb that indicates that something changes. How about some of these:

```
G.run_regularly(custom_op_code)
G.run_on_event(event, event_code)
G.regularly_run(custom_op_code)
G.on_event_run(event, event_code)
```

For slots, I'm happy for us to keep threshold/reset and add new event slots too. For deciding when the event condition is tested, I don't mind having an ugly syntax for that since it's going to be fairly rare.

mstimberg:
I'd be happy with `run_regularly` and `run_on_event` but let's wait for feedback from @romainbrette 

> For slots, I'm happy for us to keep threshold/reset and add new event slots too.

I don't think we need new slots here, if the user wants they can always create their own slot easily. I think it's fine if all events are checked in the `thresholds` slot and executed in the `resets` slots by default -- I don't see any logical choice to check/execute general events before or after spike events.

> For deciding when the event condition is tested, I don't mind having an ugly syntax for that since it's going to be fairly rare.

I basically see two options: 1) yet another keyword argument, e.g. `events_when` 2) the key in the dictionary provided for `events` can optionally be a tuple `(condition, when)` instead of just a condition, i.e. `NeuronGroup(..., events={'my_event': ('x>0', 'after_thresholds')})`

thesamovar:
So should we allow for a synonym in the slot names then? Like ``event_conditions==thresholds`` and ``event_code==resets``? It seems weird to go to the effort of having different names for general events compared to threshold/reset but then just insist that the slots be threshold/reset. Special casing threshold/reset slots like this also fits in with the general idea of special casing threshold/reset.

Ugh, neither of those two options for specifying when are very nice, but perhaps we shouldn't worry too much. I think I'd rather go for ``events_when`` because it's a bit more explicit and since it'll be used rarely you want something to remind you what it means.

mstimberg:
> So should we allow for a synonym in the slot names then? Like event_conditions==thresholds and event_code==resets? It seems weird to go to the effort of having different names for general events compared to threshold/reset but then just insist that the slots be threshold/reset. Special casing threshold/reset slots like this also fits in with the general idea of special casing threshold/reset.

So you mean that in the schedule it says `...'groups', 'thresholds', 'synapses'...]` but the `when` setting for general event conditions is `event_conditions` which will be interpreted as an alias to `thresholds` if it is not present in the schedule itself? I think this would be a bit opaque, the `Network` should not have to know anything about all of this. And I'm not sure what you mean by "insist that the slots be threshold/reset" -- we'd always give the option to change the `when` of the event to something else, be it something like `after_thresholds`, which does not require any changes to the schedule, or something like `my_event_check`, which makes it necessary to add a new slot to the threshold.

I think we have to chose between the following options:
* general events use  `when='thresholds'` and `when='resets'` by default, so no difference to standard thresholds/resets (which makes their relative ordering depend on their names)
* general events use `when='after_thresholds'` and `when='after_resets'` (or the same with "before") -- the advantage is that for a single custom event, there's a clear ordering
* general events use new slots `when='event_conditions'`, and `when='event_code'` that are also included in the schedule (before or after thresholds/resets)

I don't have a strong preference, but I think I prefer the second option for pragmatic reasons.

> I think I'd rather go for events_when because it's a bit more explicit and since it'll be used rarely you want something to remind you what it means.

I agree, the tuple syntax is a bit opaque, I'd be fine with `events_when` (now for full completeness, we'd also need `events_order`, but I think that would be overdoing it. Also, the check for a condition should probably not be that order-sensitive, it's different for the executed code but there we'll have the standard `when` and `order` arguments).

thesamovar:
OK, so how about we go with default for events is after thresholds and resets (I like that solution), and for when and order, we have a method for that in order to simplify the constructor, e.g.:

```
1. G.set_event_schedule(when={'name':'slot'}, order={'name':order})
2. G.set_event_schedule(name=('slot', order))
3. G.set_event_schedule(name=dict(when='slot', order=order))
```

I know I argued against the method before, but now I'm thinking that if it's very rare to use it that's OK, and it simplifies the constructor arguments quite a lot. Also, even though option (2) has a tuple in it, for some reason I don't find it so bad. Putting when and order together into a tuple feels OK to me. I don't have a strong preference between the options above.

mstimberg:
Well, strictly speaking you argued against having the basic event definition in a separate method, so this is fine. I'm happy with this approach, but I think we could have it simpler at the expense of calling it several times for multiple general events with non-standard scheduling slots (but I think this will be really, really rare):
```Python
G.set_event_schedule('my_event', when='before_thresholds', order=42)
```

thesamovar:
We could either say that I'm inconsistent, or that I'm willing to revise my opinion in the light of new information / good arguments. I will argue for the latter. ;)

Yes, your final version looks good to me. Especially the order.

mstimberg:
Ok, let's summarize what we agreed on (and ask @romainbrette for a third opinion). The general idea is that the only difference between a `spike` event and general user-defined events is that there is some convenience syntax for the former and several objects assume it as the default event (e.g. `Synapses`). Internally, all events are handled in the same way.

### NeuronGroup

General events can be declared in a new `events` keyword in `NeuronGroup`:
```Python
G = NeuronGroup(N, '...', threshold='...', reset='...', events={'my_event': 'x > 1'})
```
Internally, defining a standard threshold is completely equivalent to specifying the `'spike'` event.

Code that is executed triggered on an event (i.e. the reset equivalent) is added in a separate method call instead of in the initializer call to avoid having too many arguments there, especially when detailed control of the schedule is necessary:
```Python
G.run_on_event('x = 0', when='before_resets', order=2)
```
For more consistency, `custom_operation` will be renamed to `run_regularly` and will work in the same way, i.e. you no longer have to add the object that is returned to the network (which users often forgot):
```Python
G.run_regularly('x += rand()', dt=10*ms, when='groups', order=1)
```

The default schedule slots for custom events are `after_threshold` for the check of the condition and `after_resets` for the `run_on_event` code. The latter can be easily specified in the `run_on_event` but for the condition check (which very rarely will need precise scheduling anyway) you have to call another method:
```Python
G.set_event_schedule('my_event', when='before_thresholds', order=42)
```

Finally, note that for full flexibility `run_on_event` and `run_regularly` will both return an object, even though those objects do not have to be added to the network (they are part of `NeuronGroup` in the same way as the standard "thresholder/resetter"). The returned object can be used to call `G.remove(...)` and remove the operation from the `contained_objects` list in `NeuronGroup`. This makes it possible to dynamically add/remove custom operations between runs (another option is to set `active` to `False` on the returned object to disable it temporarily).

### Synapses

The `pre` and `post` pathway by default react to the `spike` event in the connected groups. If this is not the event that `Synapses` should react to, then the `events` keyword can be used to specify another name:
```Python
S = Synapses(G, G, '...', pre='...', post='...', events='my_event')
```
If there is more than one pathway (e.g. `pre` and `post` as above but also possibly several pathways of each kind) then `events` can be specified as a dictionary (where unspecified pathways default to `spike`):
```Python
# The pre-pathway is triggered by spikes, post by "my_event"
S = Synapses(G, G, '...', pre='...', post='...', events={'post': 'my_event'})
```

### Monitors

A new class `EventMonitor` can be used to record arbitrary events:
```Python
event_mon = EventMonitor(G, 'my_event')
```
`SpikeMonitor` is just a convenient short-hand for `EventMonitor(G, 'spike')`

thesamovar:
Exactly so.

romainbrette:
I think it's excellent! For "run_on_event" I suppose you forgot the event name as the first argument?

mstimberg:
> For "run_on_event" I suppose you forgot the event name as the first argument?

Oh, yes indeed.

mstimberg:
Everything discussed should now be included in the branch and I also added tests and documentation.

While implementing things I had one thought, though: is there any reason why we should restrict events to `NeuronGroup` and not make it a general properties of `Group`, i.e. also allow event checking for `Synapses`? In most of the cases, checking a condition for every synapse and every time step will be too costly, but I guess there are situations where this could be interesting (say to trigger something if `t-lastupdate>some_time`).

This is not something that we have to build in right now, but if we add it, `events` in `Synapses` should probably mean the same thing as it does for `NeuronGroup` and not have the meaning it now has in the branch (trigger event for pre/post). Maybe we should therefore rename the current keyword argument to be compatible with future changes?

thesamovar:
We might also want to change the ``eventspace`` mechanism if allowing it for Synapses otherwise we'd have a big array. Happy to change the name for events in Synapses if you can think of a good alternative. I can't at the moment. The source/target events was too complicated I think, and your argument that events_source is confusing because of source/target being concepts for Synapses is correct too. We could go with ``event_type`` but then it's a new term 'type' that we don't use in the definition of the events. Although, we could have the document say that events are specified as ``events={'event_type': 'event_condition'}`` and then it would make sense I guess.

mstimberg:
You are right about `eventspace` -- as said before, not something I'd want to implement now but it would be good if we don't have to change the syntax if we do. I don't really have a good idea for a name, I thought of `trigger_event` but that does not help since it could be the event that triggers or the event that is triggered...  Maybe `on_event` (somewhat similar to NeuroML) or `receive_events` (~ NineML)?Actually, `on_event` isn't too bad, it's similar to `run_on_event`.

thesamovar:
``on_event`` is good I think.

thesamovar:
So do we want to change ``eventspace`` or not? I think if we might want to change it, it might be better to do it sooner rather than later because it will change the templates a lot, so if anyone makes a new device it would be invalid if we changed it later.

mstimberg:
Well, do we have an idea of what data structure to use instead? We can't use dynamic arrays since we'd have to resize them in Python for runtime mode. In the long run, we should probably use standalone's dynamic array implementation for weave/Cython as well and use a Cython wrapper around it so it can be accessed from Python, but this is a bit of non-trivial work.
I think the changes to the templates will not that big, though, and this is certainly not the last change that will break them...

mstimberg:
I committed the rename change, so this is now ready-to-merge from my side (if we don't do something about `eventspace` right now).

thesamovar:
Dynamic array (e.g. using C++ vector) is the only sensible option I think. If it's possible to wrap that in weave/Cython then great. Let's make another issue for that maybe and think about it another time.

I'll review/merge this later.

thesamovar:
Great. Good to go. Merging now.

PR #495, opened 2015-07-06
[MRG] Don't change the underlying value for in-place operations on scalar quantities

### LINKED ISSUE #469: Unexpected behaviour of units with inline operators
```
a = b = 1
b += 1
print a, b
a = b = 1*second
b += 1*second
print a, b
```

gives

```
(1, 2)
(2*second, 2*second)
```

This is a bug that actually bit me and I think the difference between Python standard behaviour here is a problem. The reason it happens is that ``1*second`` is actually an array so inplace operations modify the underlying data rather than rebinding the name. I don't think there's anything we can do about this is there? Could we raise a NotImplementedError for inplace operations for scalar values only and force it to go back to the Python default implementation (i.e. ``a+=b`` is ``a=a+b``)? If not, I think we probably need a big warning about this in several places in the documentation. What do you think?

mstimberg:
Closes #469

### DIFF
--- a/brian2/tests/test_units.py
+++ b/brian2/tests/test_units.py
@@ -4,7 +4,7 @@ import pickle
 
 from nose.plugins.attrib import attr
 import numpy as np
-from numpy.testing import assert_raises, assert_equal
+from numpy.testing import assert_raises, assert_equal, assert_allclose
 
 from brian2.core.preferences import prefs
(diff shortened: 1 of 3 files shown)


thesamovar:
Looks good to me and since it's marked as MRG I'm going to merge it now.

mstimberg:
Thanks, it was indeed intended to be "ready to merge" :)

PR #498, opened 2015-07-07
[MRG] Monitor other variables event-based

### LINKED ISSUE #181: Discrete event monitoring
See #168 for details. Depends on #105 before we can make progress on this issue.

mstimberg:
This adds the feature to record arbitrary other variables with `SpikeMonitor`/`EventMonitor`, i.e. this replaces `StateSpikeMonitor` from Brian 1. The syntax is close to `StateMonitor` syntax (you specify a single variable name or a list of variable names for the `variables` argument; you access the values as attributes of the `SpikeMonitor` class). I also added an example that makes somewhat useful use of this feature. I think this closes #181 -- we can always add more features (reductions, intervals, ...) later.

The only thing useful that is missing is an equivalent to the `spike_train` method, i.e. something that partitions the data. To access all the recorded values for a specific neuron you currently have to do `spike_mon.var['i == idx']` or something like that. The potential syntax is not quite as straightforward as for `spike_trains`, though -- should it map indices to arrays of values of one variable, i.e. you'd call something like `spike_mon.values('var')`? Or something more complex like a mapping of indices to a dictionary of variable names to values, so that you'd do `all_values = spike_mon.values(); var_values_0 = values[0]['var']`?

From my side, I'd be happy to merge this now and to postpone this question.

Oh, one thing that this PR also changes: unitl now, `SpikeMonitor` recorded in the `'end'` slot by default, which of course did not matter since it only recorded times/indices. I changed this so that it records by default after the condition for the triggering event is checked, i.e. by default a `SpikeMonitor` will check in the `thresholds` slot with order=1 and an `EventMonitor` for a custom event will check in the `after_thresholds` slot with order=1.


thesamovar:
This looks good to me. Two possible changes:

1. I like the idea of ``spikemon.values('var')`` returning a dict like ``spike_trains()`` does (so that ``spike_trains()==values('t')``. I'd also be happy to have a method ``all_values()`` which returns ``dict((varname, values(varname)) for varname in ...)``. I can imagine myself using this (``all_values()[varname][idx]``) but the other way round less frequently (``all_values()[idx][varname]``).

2. I wonder if we ought to introduce an events slot after all? The thing is, if we make ``SpikeMonitor`` record in ``thresholds`` with ``order=0`` this silently breaks expectations if the user changes the ``order`` of the event. The more natural place would be for it to go in ``after_thresholds``, but this leaves us stuck for events (we'd need ``after_after_thresholds``). So if by default the ``events`` slot was just after the ``thresholds`` slot we'd be OK. On the other hand, this is all stuff that will happen extremely rarely and only in the case of the sort of advanced users that make use of the ``order`` keyword, so maybe it's not worth worrying.

mstimberg:
> I like the idea of spikemon.values('var') returning a dict like spike_trains() does (so that spike_trains()==values('t'). I'd also be happy to have a method all_values() which returns dict((varname, values(varname)) for varname in ...). I can imagine myself using this (all_values()[varname][idx]) but the other way round less frequently (all_values()[idx][varname]).

Ok, I'll add a `values` method. I don't have a strong opinion on the `all_values` semantics, I could imagine having a use for all variables of a specific neuron as well as for the values for a variable across all neurons. But I think your proposal is more consistent, since it simply means that `all_values` is a short-cut to calling `values` for each variable (probably only from a user's point of view, though, it would be inefficient to do the sorting again for each variable). Small detail for that: should `all_values` also return the values for `t`? I guess yes, why not.

About the event slot: I might not have explained it very well, but the current code is not hard-coding `thresholds`/`order=1` for `SpikeMonitor` and `after_thresholds`/`order=1` for `EventMonitor` -- instead it uses the same slot as the recorded event and adds 1 to its order. So this would only break if the user changes the `when`/`order` attribute of the event *after* creating the monitor, and I think it is reasonable to expect the user to change the monitor's scheduling manually in this (somewhat weird and certainly rare) use case.

thesamovar:
Yes, I think record ``t`` as well as the other variables.

For thresholds/order I misunderstood. OK in that case I think it's fine. Anyone who changes the when and order after creating it deserves whatever they get. Actually, should we make those read only attributes after they're created? Is there any possible use case for changing that?

mstimberg:
> Actually, should we make those read only attributes after they're created? Is there any possible use case for changing that?

I'm hesitant to do this, I think it is in general a nice feature that you can change almost anything until a run. I don't have really convincing use cases, but one thing that comes up every now and then (and will come up more often in the future) is projects where model components are created in functions. It might be useful to change `when`/`order` in that case, simply by setting the attributes of the returned object. Otherwise, you'll have to add `when`/`order` arguments to such functions, and this can be very inconvenient especially when the functions create more than one object.

One improvement to make all this even more convenient (even though we are talking about optimisations for rare use cases...) is to not "resolve" the `when` at the creation time of the monitor, but to wait until `before_run`, this way it could even pick up a changed `when`/`order` of the event. But in this case, we should probably do this for other objects that have some kind of dependency as well. Not sure this is really worth it.

thesamovar:
OK!

mstimberg:
"OK" in the sense of: let's not bother with the more complicated "when/order resolution in `before_run`" approach right now?
In that case it is now ready to merge, I just added the `values` and `all_values` methods.

thesamovar:
Exactly so. Happy for this to be merged as soon as tests pass.

PR #501, opened 2015-07-12
[MRG] Fix multi-step state updates for non-autonomous equations

### LINKED ISSUE #477: Non-autonomous equations are not integrated correctly with multi-step solvers
Multi-step solvers don't use the `t` of intermediate time steps for non-autonomous equations:
```Python
>>> eqs = Equations('dv/dt = sin(2*pi*freq*t) / tau : 1')
>>> print rk2(eqs)
__k_v = dt*sin(2.0*pi*freq*t)/tau
_v = dt*sin(2.0*pi*freq*t)/tau + v
v = _v
```
In the statement for `_v`, `t` should have been replaced by `t + 0.5*dt`.

This is a regression that has been introduced more than one year ago in commit 1a33ce581 ...

mstimberg:
Closes #477
The test is very crude, though...

### DIFF
--- a/brian2/stateupdaters/explicit.py
+++ b/brian2/stateupdaters/explicit.py
@@ -21,14 +21,15 @@ __all__ = ['milstein', 'euler', 'rk2', 'rk4', 'ExplicitStateUpdater']
 # Class for simple definition of explicit state updaters
 #===============================================================================
 
-def _symbol(name):
+def _symbol(name, positive=None):
     ''' Shorthand for ``sympy.Symbol(name, real=True)``. '''
-    return sympy.Symbol(name, real=True)
+    return sympy.Symbol(name, real=True, positive=positive)
 
 #: reserved standard symbols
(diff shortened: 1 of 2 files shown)


thesamovar:
Looks good to me although I didn't quite follow the changes to the code. What was the significance of the change in the way ``t`` was substituted? In any case, happy for this to be merged.

mstimberg:
In the state updater description, we can have expressions such as `f(x, t + dt*0.5)`, so we have to substitute `t` in equations with `t + dt*0.5`. A while ago, we had a problem where having a `t` in state updater descriptions (which refers to the actual time) and a `t` in equations (which possibly has to be substituted by an expression) led to confusion. This is why we now replace `x` and `t` in state updater descriptions by `__x` and `__t` in the very beginning. With the introduction of this change, the substitution of `t` in the equations was broken, though, it was always just treated as `t`. The code is quite convoluted and not easy to follow, hopefully we'll get around fixing that one day...

The tests pass, I'm merging it now.

PR #502, opened 2015-07-15
[MRG] Spatialneuron improvements

mstimberg:
This branch contains several performance improvements for spatialneuron and adds some basic OpenMP support (which only does the trivial things in parallel for now). It also adds a few tests (in particular the "Rallpack" benchmarks: http://www.ncbi.nlm.nih.gov/pubmed/1281353).
It also includes a few general improvements for standalone that I fixed in the process, in particular less need for static arrays. Previously, simple code such as `G.lastupdate = -np.inf*second` was translated into a static array that was read from disk. Now it will be directly included as code. Profiling now also works for standalone (if not using OpenMP).

This can be merged from my side, please run the full long test suite because otherwise the Rallpack tests will not be run.

After merging, I think we can also delete this branch now (instead of keeping it up for future improvements to multicompartmental modeling). We should open new feature branches for everything we'd like to add/fix as for other parts of the code.


thesamovar:
Code looks good. Trying to run tests now. At the moment, numpy and weave both pass but it's segfaulting on Cython. Will investigate.

thesamovar:
``run_nose_tests_standalone`` works, so it seems as though it's just cython.

thesamovar:
OK, strange. On a second run the tests pass. A random error? Go ahead and merge or investigate further?

thesamovar:
Just to be clear: happy for you to merge this if you think it doesn't need more investigation.

mstimberg:
I had a segmentation fault (or sometimes other types of memory error) with weave, but it was terribly hard to debug (it only triggered when it ran as part of the test suite, not when I ran tests individually, the memory error was triggered in Python code, not in the weave code, etc.). I *think* the issue was the `static double *c = malloc(...)` allocation -- at least the problem went away when I replaced it by the earlier approach (allocate/free every time). No idea whether this has anything to do with your Cython problem, though, but without a way to reproduce it there's not much that we can do now.

PR #506, opened 2015-07-16
[MRG] Let `StateMonitor` record in the start slot by default

### LINKED ISSUE #452: Record at beginning rather than end by default?
At the moment, we record variables with when='end' by default. However, the corresponding time is t, the start of the interval [t, t+dt) being computed. It might be more logical to record at the beginning of the interval, although this has some drawbacks. Firstly, it is backwards incompatible, although Brian 2 is a good time to make these sorts of changes. Secondly, it means that the last value recorded in a run of 100ms would be the value at 99.9ms instead of the value at 100ms. (On the other hand, the last time recorded is already 99.9ms, so there's already something counterintuitive going on there.)

@romainbrette

mstimberg:
Also adds a convenience method for recording at the end.

Updates tests and documentation, ready-to-merge from my side.
Closes #452

### DIFF
--- a/brian2/monitors/statemonitor.py
+++ b/brian2/monitors/statemonitor.py
@@ -81,13 +81,13 @@ class StateMonitor(Group, CodeRunner):
     '''
     Record values of state variables during a run
     
-    To extract recorded values after a run, use `t` attribute for the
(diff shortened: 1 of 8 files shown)


thesamovar:
Great! Merging.

PR #507, opened 2015-07-16
[MRG] `get_states` improvements

mstimberg:
...aaand this is ready to merge as well. It makes the `get_states` method more useful, in particular it fixes it for `StateMonitor`. This is a small detail but it makes it much easier to dump everything that a `StateMonitor` recorded to disk, which is something that quite a few users want to do, I think.


thesamovar:
Good good.

PR #514, opened 2015-07-22
[MRG] Treat `i` and `t` in the same way as other variables in `StateMonitor`

### LINKED ISSUE #510: record argument disappeared from SpikeMonitor

mstimberg:
Do not record them with `record=False`.

Closes #510

Unfortunately this will break templates for other devices (CC @tnowotny @moritzaugustin @Kwartke). For full support (including the new support for recording arbitrary variables introduced with #498), the templates have to be adapted analogue to the C++ standalone templates. Alternatively, a quick fix would be to raise a `NotImplementedError` if `SpikeMonitor.record` is False and/or `SpikeMonitor.record_variables` is not simply `{'i', 't'}`.

### DIFF
--- a/brian2/codegen/runtime/cython_rt/templates/spikemonitor.pyx
+++ b/brian2/codegen/runtime/cython_rt/templates/spikemonitor.pyx
@@ -1,7 +1,7 @@
 {% extends 'common.pyx' %}
 
 {% block maincode %}
-    {# USES_VARIABLES { t, i, _clock_t, count,
+    {# USES_VARIABLES { N, _clock_t, count,
                         _source_start, _source_stop} #}
 
     {#  Get the name of the array that stores these events (e.g. the spikespace array) #}
@@ -9,8 +9,6 @@
 
     cdef int _num_events = {{_eventspace}}[_num{{_eventspace}}-1]
     cdef int _start_idx, _end_idx, _curlen, _newlen, _j
-    cdef double[:] _t_view
-    cdef int32_t[:] _i_view
     {% for varname, var in record_variables.items() %}
     cdef {{cpp_dtype(var.dtype)}}[:] _{{varname}}_view
     {% endfor %}
@@ -35,14 +33,10 @@
             # scalar code
             _vectorisation_idx = 1
             {{ scalar_code|autoindent }}
-            # Get the current length and new length of t and i arrays
-            _curlen = {{_dynamic_t}}.shape[0]
+            _curlen = {{N}}[0]
             _newlen = _curlen + _num_events
             # Resize the arrays
             _owner.resize(_newlen)
-            # Get the potentially newly created underlying data arrays
-            _t_view = {{_dynamic_t}}.data
-            _i_view = {{_dynamic_i}}.data
             {% for varname, var in record_variables.items() %}
             _{{varname}}_view = {{get_array_name(var, access_data=False)}}.data
             {% endfor %}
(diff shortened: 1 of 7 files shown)


tnowotny:
Hi Marcel,
I am struggling with this one. Basically, I think I can accommodate monitoring other variables in GeNN (even though at potentially high cost in the performance) but I am not sure I can figure out this new notion of “events”. What other variables other than _spikespace are you expecting to be eventspace_variables? I fear that there would not be equivalents in GeNN (even though we have two versions of events - “true spikes” and otherwise defined “spike like events”, the latter however defined per synapse population …)
Can we have a brief chat – or do you think you can sort it out here or by email?

thesamovar:
Marcel's on holiday at the moment, but I'll do my best.

We generalised our events scheme so that users can create arbitrary events and code to respond to them, but you don't necessarily need to implement that. I think if you just use the event type 'spikes' and maybe raise an error if you ever see an event type that isn't spikes that will be OK for the moment. We expect this to be a feature that is not used by the majority of users.

tnowotny:
Hi Dan,
I see – and interesting as we are doing the exact same thing … only that synapse populations are defining the events they want to see pre-synaptically because we were thinking graded synapses etc.
It does feel a bit as if it wasn’t miles apart but might nevertheless be hard to map onto each other …
On a practical level, we are only allowing one event definition per synapse pop – that probably differs. And name conventions obviously differ a lot …
So, in your case, if it’s not “spikes”, do users define their own name for it? And would the event variable then be named “_<name>space”?
For now I probably should go with the warning –
Best,
Thomas.

--
Prof. Thomas Nowotny
CCNR, Sussex Neuroscience                   Phone: +44-1273-678593
Engineering and Informatics                  Fax:   +44-1273-877873
University of Sussex
Falmer, Brighton BN1 9QJ http://sussex.ac.uk/informatics/tnowotny

thesamovar:
The way it works for names is that each code object can only respond to one event type, so in the template you see ``{{_eventspace}}`` but it will get expanded to the name of the array (which I don't remember off the top of my head).

We allow for a single event type and multiple Synapses responding to that event, which I think wouldn't be possible in GeNN right? We also allow for an equivalent of a reset in response to an arbitrary event (i.e. NeuronGroup responding to its own event), which is also not possible with your scheme as I understand it? (It's Synapses only.) However, I think to map what we do onto what you do, we could explicitly check for the situation where each Synapse is responding to a unique event type, and only raise an error if this mapping can't be done. But yeah, for now a warning/error would probably do the trick.

tnowotny:
ok, when we have multiple synapse populations with the same pre-synaptic neuron population they will define events and we will take the logical "or" of all of them and record them. The synapses then have to check whether any of the recorded events fit their more narrow definition. In a sense this is what you describe: the "or"-d definition is a single event type and all synapse react (differently) to it.
But it does sound as if there may be a few snags in sorting this all out - so will stay away from it for now.
The neurongroups only can reset/act upon the proper spikes in GeNN. Or, obviously, they can define whatever they want within their update rule (as we are not restrictive in terms of what an update rule can entail).

tnowotny:
I have another problem: How can I extract the name of the relevant neuron population (or other owner) of a monitored variable in "record_variables" within the spikemonitor template? var.owner.name is the name of the spike monitor, not of the object that otherwise owns the variable ...?

thesamovar:
It's probably not the correct way but does var.owner.source.name work?

mstimberg:
Using `eventspace_variable.owner.name` should work for most of the cases, it will only break if the recorded variable is a linked variable from somewhere else (but I think that brian2genn does not support linked variables in the first place, currently). The correct information would be, e.g. for a variable `v`, in the `_source_v` variable, i.e. `_source_v.owner.name`. However, this information is currently not available to the template, as this is dealt with in the code generation process. We create some dummy code `_recorded_v = _source_v` and then let the code generation machinery take care of it, this allows for subgroups, linked variables with different indices and all that.

I don't think brian2genn has to deal with all this, a nice thing would be to have a clear check for it, but this is something that is quite cumbersome to do currently. We should represent this information more explicitly in the `SpikeMonitor` object so that it can be easily accessed by devices.

Issue #29, opened 2013-04-02
Save and load network state


thesamovar:
We should have methods to save and load network state that are more portable than pickle, for example for use in standalone C++ code. Two options occur to me:

* New simple binary format of our own, the advantage being that it would be easy to read for Brian standalone C++ code (no external dependencies), the disadvantage being that nothing else could read it without writing something explicitly.
* Existing format, probably HDF5, the advantage being that it could be read in multiple pieces of software, the disadvantage being it would make a large dependency for our standalone C++ code, and may be problematic on, for example, Android/Java.

Or we can do both, but then we have effectively three file format (pickle, internal, HDF5) to support, which seems like a bad idea.

thesamovar:
CNS meeting update: it might be enough to just give access to the results of store and restore?

Could do this, e.g., by adding a ``Network.get_states`` and ``Network.set_states`` methods.

mstimberg:
I looked a bit at this, see #515 for add support of `Network.get_states`. I realized that this is not quite a solution, though: `get_states` is to have a covenient syntax to get all the data out of an object (e.g. a monitor) or to set the initial state (e.g. for a `NeuronGroup`). It is not meant to completely set the state of an object in the way `store`/`restore` is. The way the latter mechanism is implemented is by calling a `_store`/`_restore` method on the respective objects recursively, and these objects take care of storing the data in some way themselves. There's internal stuff, the state of a `SpikeQueue`, clocks, etc. which I don't really see can easily work with `get_states`/`set_states`. It would also expose a lot of internals to the user, and I fear that some users then start using this internal representation explicitly, assuming it stays implemented in the same way. The call of `_store`/`_restore` nicely encapsulates that, each object can simply change its own implementation of it without changing anything on the user-level.

I am not really sure what a good solution would look like here, it should be something that is possible to do easily from the C++ SpikeQueue for example. Maybe it is only an issue of making clear that `get_states`/`set_states` is something high-level, where the format is fixed and there is another mechanism that returns a serialized low-level representation that should not be assumed to stable. But maybe we should just use `__getstate__` and `__setstate__` for that and wrap the current store/restore mechanism around it. This way, we get pickling for free and don't implement our own serialization mechanism in parallel.

thesamovar:
I think dumping the output of store/restore is fine. If we want to emphasise that it's not for user consumption, let's pickle the output of store and only give the user a pickled string (i.e. raw binary data). In the documentation we can add that it's subject to change.

[...]

PR #515, opened 2015-07-22
[MRG] Provide `get_states`/`set_states`  methods for networks

mstimberg:
This allows to conveniently get/set the states for all objects in a network. This is related to #29 but it is not a solution for it (see my comment there).


thesamovar:
Code looks good to me, happy for it to be merged when tests pass.

PR #516, opened 2015-07-22
[MRG] Better support for mod operation with mixed types

### LINKED ISSUE #286: Mod operator in CPP vs numpy
Hi all,
Thanks to @mstimberg, I figured out that the `mod` function in numpy and `fmod` in C have different semantics. As a result, the following code does not do the same thing with different codegen targets:

```python
brian_prefs.codegen.target = 'numpy'
G  = NeuronGroup(1, 'x:1')
G.x = '-1 % 10'
run(defaultclock.dt)
print G.x
```

Indeed, in numpy `mod(x, N)` always returns an integer between 0 and N-1, even if x is negative, whereas C's `fmod` returns negative values if `x` is negative.
A way out is to make the `mod` operation in brian semantically consistent with numpy by writing `fmod(N+x, N)` instead of `fmod(x, N)`.

Note that sticking to numpy's semantics is maybe more convenient in the case where mod is evaluated for array indices, which shouldn't be negative.

### LINKED ISSUE #509: mod doesn't work for integer values
With weave code generation, ``i%N`` in a ``NeuronGroup`` confuses the compiler (because it is trying to use ``fmod`` and there is no directly matching signature).

thesamovar:
Needed to change the handling of mod, previously sympy was converting it to Mod and thence to mod. I found a workaround to get a single function to work for both integers and floats in C++ but not in Cython. So I changed it so that mod is left as % and now CPPNodeRenderer converts this to _brian_mod, which is always put into the support code (I created a new attribute universal_support_code to put things that are always in the support code).

Ready to merge if the tests pass (they do on my win64 machine).

Fixes #509. Fixes #286.

### DIFF
--- a/brian2/codegen/generators/cpp_generator.py
+++ b/brian2/codegen/generators/cpp_generator.py
@@ -110,6 +110,12 @@ class CPPCodeGenerator(CodeGenerator):
 
     class_name = 'cpp'
 
+    universal_support_code = '''
+    template<class T1, class T2> long _brian_mod(T1 x, T2 y) { return x%y; };
+    template<class T1> double _brian_mod(T1 x, double y) { return fmod((double)x, (double)y); };
+    template<class T2> double _brian_mod(double x, T2 y) { return fmod((double)x, (double)y); };
+    '''
+
     def __init__(self, *args, **kwds):
         super(CPPCodeGenerator, self).__init__(*args, **kwds)
         self.c_data_type = c_data_type
@@ -330,6 +336,9 @@ class CPPCodeGenerator(CodeGenerator):
             if func_namespace is not None:
                 self.variables.update(func_namespace)
 
+        support_code.append(self.universal_support_code)
+
+
         keywords = {'pointers_lines': stripped_deindented_lines('\n'.join(pointers)),
                     'support_code_lines': stripped_deindented_lines('\n'.join(support_code)),
                     'hashdefine_lines': stripped_deindented_lines('\n'.join(hash_defines)),
@@ -350,7 +359,6 @@ for func in ['sin', 'cos', 'tan', 'sinh', 'cosh', 'tanh', 'exp', 'log',
 
 # Functions that need a name translation
 for func, func_cpp in [('arcsin', 'asin'), ('arccos', 'acos'), ('arctan', 'atan'),
-                       ('mod', 'fmod'),
                        ]:
     DEFAULT_FUNCTIONS[func].implementations.add_implementation(CPPCodeGenerator,
                                                                code=None,
--- a/brian2/codegen/generators/cpp_generator.py
+++ b/brian2/codegen/generators/cpp_generator.py
@@ -86,6 +86,27 @@ prefs.register_preferences(
     )
 
 
+mod_support_code = ''
+typestrs = ['unsigned char', 'char', 'unsigned short', 'short', 'unsigned int', 'int', 'unsigned long', 'long',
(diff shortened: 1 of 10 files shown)


mstimberg:
I have to look closer what this change means on the sympy level. For now the branch breaks the modulo operator for floating point values on my machine (we never test it on the servers apparently...). The problem is that for two doubles, all the three variants are a possible match so gcc complains about ambiguity.

mstimberg:
Actually, there's a TODO note about a floating point modulo test in `test_math_functions_short`, it should be activated there.

thesamovar:
I enabled the more extensive test of mod support. I actually get test failures now, but not errors. It still compiles fine on my machine, but there is a difference in the value of mod for python and C++/Cython, namely that ``-1%2==1`` for Python but ``-1`` for C++/Cython.

Question: do we want to make a guarantee about this or just say that mod returns a valid modulo?

Maybe the tests will now show this compile error, but it's strange because it isn't causing any problems on my computer. Going to try out a few things now anyway.

thesamovar:
OK so the builds are failing because of the negative number issue, but appears to be compiling on both linux and windows.

mstimberg:
Just quickly (I'm on vacation now :) ). I think that we should have the same semantics here for Python and C++ if at all possible. I think Victor was using something like that for brian2hears. That reminds me, this is obviously very related to #286.

thesamovar:
Oh I'd forgotten that issue. I added it to the 2.0 milestone. I'll see what I can do, but it's surprisingly tricky. Perhaps Victor's solution could work: for C++/Cython interpret ``x%y==((x%y)+y)%y``. It's computationally a bit inefficient but it's unlikely to be a critical part I would think.

thesamovar:
OK tests all passing now. @mstimberg is your issue persisting?

thesamovar:
Actually should we have an option to allow for language-specific semantics (for a marginal speed improvement)? Could make a new issue for this.

thesamovar:
OK made a new issue for this (#519) and it's not critical for 2.0, so this one can be merged.

thesamovar:
Aha! OK, I added a new test and now I can reproduce the problem you were talking about. Still no fix though...

thesamovar:
Trying another approach: explicitly listing all pairs of argument types (64 function definitions in total). If this passes tests, I'm happy for it to be merged.

mstimberg:
I merged the current master into this branch to trigger a new test run. Generally I'm fine with the changes, I'm not sure about using `x%y==((x%y)+y)%y`, though. It's certainly deals with the most common use case, but it is not fully equivalent, e.g. `-13 % 10` would still get differing values. Not sure it is worth the additional computation (though it is probably minimal), but according to http://docs.scipy.org/doc/numpy/reference/generated/numpy.remainder.html (numpy's remainder is equivalent to Python's `%`), `x - floor(x / y) * y` would be the correct thing to do.

thesamovar:
I think ``x%y==((x%y)+y)%y`` is equivalent (it works for ``-13%10``). C's ``x%y`` is guaranteed to return something from -y to +y, so if you add y it is guaranteed to be between 0 and 2y and then taking mod again you're guaranteed to be between 0 and y. It's a shame about the extra computation time necessary but I can't imagine there are many cases where it would be critical.

mstimberg:
Ah, sorry, my mistake. I think it is fine then. I also don't think that the extra computation time will matter much. I am waiting for the tests to finish and then I'll merge.

Issue #417, opened 2015-02-27
Support for float32 type


thesamovar:
At the moment, setting `prefs.core.default_float_dtype = float32` works if you're in numpy mode but not any of the other modes (which make assumptions about doubles). Also, if you use the compiled `SpikeQueue` it doesn't work because it assumes `float64`. When we get round to addressing this, we should possibly include a new test configuration for Travis that checks everything using float32 as well as float64. (Maybe just one Python 2.x and one Python 3.x).

mstimberg:
For the `SpikeQueue`, I think we have two reasonable options:
1. Just don't allow float32 synaptic delays -- currently, the user cannot change the data type of the delays specificially, they are always assumed to be the "default" type so we could as well always have it be of float64 type. It doesn't change anything for simulation speed since delays will be converted before the simulation into integers. Also, `Clock.dt` is currently not user changeable (and I don't really see any advantage of doing it)
2. If we want to allow both float32 delays and float64 delays, then I think the best would be to have a `prepare_float32` and a `prepare_float64` method in `SpikeQueue`, we could then code generate the correct call.

thesamovar:
I like option 2 best. I think it's good to allow delays to be float32 because they are per-synapse which makes them one of the candidates for largest users of memory. Incidentally, for the same reason we might want to make the int variables be potentially of different dtypes. Generally speaking, a lot of our memory use is going to be in Synapses and at the moment we're using quite a lot of memory (I think it's around 40 bytes per synapse without any synaptic variables).

Can we not use templates with Cython? So that it automatically selects the right version based on the dtype?

mstimberg:
> Can we not use templates with Cython? So that it automatically selects the right version based on the dtype?

Ah right, we should probably be able to simply use Cython's fused types.

thesamovar:
Like that we could have support for different int types easily too.

thesamovar:
CNS meeting update: as a short term solution for 2.0, we should raise an error if you try to change the default dtype to float32, until it gets sorted.

mstimberg:
I'll leave this issue open (since we don't have float32 support yet) but remove the 2.0 milestone since the short-term fix is now in place (setting the default dtype to float32 will raise an error).

denisalevi:
Any plans of merging the `float32_support` branch @mstimberg ? I will update the brian2 version for brian2CUDA in the next days and would  be keen on those single precision floats.

mstimberg:
[...]

PR #517, opened 2015-07-22
[MRG] Change some tests from long to short and raise error for float32

### LINKED ISSUE #504: Remove 'long' from some tests
Now that travis/appveyor are completing well within timeouts, we could afford to move some more tests into the standard test suite.

thesamovar:
Somehow my changes for these two separate issues ended up in the same branch, but I think it's fine because they're both small changes. For float32 it's very simple, no more comment needed I think. When this is merged we can remove the 2.0 milestone for issue #417 but not close it. Detailed comments for long to short tests below.

I changed some tests from long to short. Here are the test timings on my computer. All times include compilation from scratch for numpy/weave/cython. The ones I changed are marked with CHANGED TO SHORT. Some tests probably don't need to be run for all of numpy/weave/cython, and I've marked those with ONLY NEEDS ONE. Perhaps we could have a new attr like 'long-for-cython', etc.

Total time added on my machine: ~9m. That's quite annoying I guess but it's valuable and I think that probably keeps us under the limit still for the CIs?

```
test_functions.py
	test_math_functions: CHANGE TO SHORT (because it's an important test for regressions)
		numpy: 1.5s
		weave: 29s
		cython: 77s
		
test_spatialneuron.py:
	6 tests marked long, didn't check yet because it's not clear if this will be
	officially marked as 'experimental' in 2.0
	
test_spikegenerator.py:
	test_spikegenerator_rounding_long: total is too long
		numpy: ~5m
		weave: ~3m
		cython: ~3m
	test_spikegenerator_rounding_period: total is too long
		numpy: ~4m
		weave: ~4m
		cython: ~3m
		
test_stateupdaters.py:
	test_multiple_noise_variables_extended: CHANGE TO SHORT, ONLY NEEDS ONE
		numpy: 1.8s
		weave: 16s
		cython: 33s
	test_multiple_noise_variables_deterministic_noise: CHANGE TO SHORT, ONLY NEEDS ONE
		numpy: 1.8s
		weave: 14s
		cython: 16s
	test_subexpressions: CHANGE TO SHORT, ONLY NEEDS ONE
		numpy: 3.5s
		weave: 23s
		cython: 32s
		
test_synapses.py:
	test_connection_string_deterministic: CHANGE TO SHORT
		numpy: 1.2s
		weave: 23s
		cython: 47s
	test_connection_random: CHANGE TO SHORT
		numpy: 1.3s
		weave: 22s
		cython: 52s
	test_transmission: CHANGE TO SHORT
		numpy: 2.3s
		weave: 45s
		cython: 50s
	test_event_driven: CHANGE TO SHORT, ONLY NEEDS ONE
		numpy: 4.3s
		weave: 16s
		cython: 24s
```

Fixes #504.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -72,10 +72,9 @@ install:
     conda create -n travis_conda --yes pip python=$PYTHON numpy scipy nose sphinx ipython sympy pyparsing jinja2 setuptools coverage;
     fi
   - source activate travis_conda
-  - conda update --yes pip setuptools
   - if [[ $CYTHON == 'yes' ]]; then conda install --yes cython; SETUP_ARGS=--with-cython; else SETUP_ARGS=''; fi
   - if [[ $REPORT_COVERAGE == 'yes' ]]; then pip install -q coveralls; fi
-  - python setup.py install $SETUP_ARGS --fail-on-error
+  - python setup.py install $SETUP_ARGS --fail-on-error --single-version-externally-managed --record=record.txt
   - if [[ $DOCS_ONLY == 'yes' ]]; then
       pip install -q sphinxcontrib-issuetracker;
       python setup.py sdist;
--- a/brian2/core/core_preferences.py
+++ b/brian2/core/core_preferences.py
@@ -10,13 +10,19 @@ from brian2.core.preferences import BrianPreference, prefs
 def dtype_repr(dtype):
     return dtype.__name__
 
+def default_float_dtype_validator(dtype):
+    return dtype is float64
+
 prefs.register_preferences('core', 'Core Brian preferences',
     default_float_dtype=BrianPreference(
         default=float64,
         docs='''
         Default dtype for all arrays of scalars (state variables, weights, etc.).
+
+        Currently, this is not supported (only float64 can be used).
         ''',
         representor=dtype_repr,
+        validator=default_float_dtype_validator,
         ),
     default_integer_dtype=BrianPreference(
         default=int32,
--- a/brian2/tests/test_functions.py
+++ b/brian2/tests/test_functions.py
@@ -30,17 +30,21 @@ def test_constants_values():
     assert G.v == np.inf
 
 
-def test_math_functions_short():
+def test_math_functions():
     '''
     Test that math functions give the same result, regardless of whether used
-    directly or in generated Python or C++ code. Test only a few functions
+    directly or in generated Python or C++ code.
     '''
     default_dt = defaultclock.dt
     test_array = np.array([-1, -0.5, 0, 0.5, 1])
 
     with catch_logs() as _:  # Let's suppress warnings about illegal values
         # Functions with a single argument
-        for func in [exp, np.sqrt]:
+        for func in [cos, tan, sinh, cosh, tanh,
+                     arcsin, arccos, arctan,
+                     log, log10,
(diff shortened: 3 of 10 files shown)


mstimberg:
The runtime is fine, but one of the tests seems to be failing in a few configurations on travis. A small thing: now that `test_math_functions` is no longer a long test, it should be merged again with `test_math_functions_short`. About the tests that should only be run once: they could be marked as `codegen-independent`, this way they'd be only run with numpy. I guess it is not strictly speaking the correct label since they need code generation to run, on the other hand whether they fail or not should not depend on *which* codegen is used, so in this way they are independent.

thesamovar:
The error is:

```
FAIL: brian2.tests.test_synapses.test_event_driven

  File "/home/travis/miniconda/envs/travis_conda/lib/python2.7/site-packages/nose/case.py", line 197, in runTest

    self.test(*self.arg)

  File "/home/travis/miniconda/envs/travis_conda/lib/python2.7/site-packages/Brian2-2.0b4+git-py2.7-linux-x86_64.egg/brian2/tests/test_synapses.py", line 829, in test_event_driven

    assert_equal(S1.w[:], S2.w[:])

  File "/home/travis/miniconda/envs/travis_conda/lib/python2.7/site-packages/numpy/testing/utils.py", line 275, in assert_equal

    return assert_array_equal(actual, desired, err_msg, verbose)

  File "/home/travis/miniconda/envs/travis_conda/lib/python2.7/site-packages/numpy/testing/utils.py", line 739, in assert_array_equal

    verbose=verbose, header='Arrays are not equal')

  File "/home/travis/miniconda/envs/travis_conda/lib/python2.7/site-packages/numpy/testing/utils.py", line 665, in assert_array_compare

    raise AssertionError(msg)

AssertionError: 

Arrays are not equal

(mismatch 50.0%)

 x: array([ 0.005009,  0.004718])

 y: array([ 0.005009,  0.004718])
```

Since the arrays look identical, I'm assuming this is a sort of order of operations / fast math type of an error that manifests randomly. Thoughts on what the best way to resolve it are? It should probably be a separate issue/pr from this one. Are you against merging it while there are still test failures (even though they'll be present in master too)?

thesamovar:
OK so I think I know why it happens, and just changing the test from ``assert_equal`` to ``assert_allclose`` is justified in this case I think.

In the test, providing an explicit solution for STDP is compared against using (event driven) to find it. The solutions in the two cases are ``Apre=Apre*exp((lastupdate-t)/taupre)+dApre`` and ``Apre = Apre*exp(-(t - lastupdate)/taupre); ...; Apre += dApre``. The difference is in what is inside the exp: ``(lastupdate-t)/taupre`` versus ``-(t-lastupdate)/taupre``. I think in principle these don't have to be the same although I couldn't find an explicit example where they weren't. For integers it's easy ``(-2)/3!=-(2/3)``.

Just about to push this change in the test to the branch, and if it passes then I think this is good to merge.

mstimberg:
I merged `test_math_functions_short` into `test_math_functions` and fixed some build issues on travis (related to some conda update, I think). Tests pass so I'm merging this now. About your last change: in general we should always use `assert_allclose` instead of `assert_equal` when comparing floating point values, so this change is perfectly justified.

PR #520, opened 2015-07-25
[WIP] Derivable templates

### LINKED ISSUE #357: Allow templates to be derived
This would be useful for devices that build on top of existing devices, for example, or when only a small new feature is being added. You wouldn't need to make a copy of the whole template directory, just add the new bits. This could be used by @brian-team/brian2genn.

See also #356 and #115.

thesamovar:
This PR fixes #357.

It adds a new method ``derive(...)`` to ``Templater`` that creates a new ``Templater`` derived from the original one with an additional template package name and (optionally) new global environment variables. The old templates will be used as a fallback if the new package doesn't define them. This is defined at the Jinja level using ``ChoiceLoader`` so everything should work smoothly (e.g. template inheritance should still work, etc.). To use it, you simply do this:

```
templater = CPPCodeGenerator.templater.derive('new.package.name')
```

Still needs documentation and tests, but would be happy to take comments on the approach.

### DIFF
--- a/brian2/codegen/generators/cpp_generator.py
+++ b/brian2/codegen/generators/cpp_generator.py
@@ -110,6 +110,12 @@ class CPPCodeGenerator(CodeGenerator):
 
     class_name = 'cpp'
 
+    universal_support_code = '''
+    template<class T1, class T2> long _brian_mod(T1 x, T2 y) { return x%y; };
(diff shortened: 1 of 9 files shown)


thesamovar:
Oops, I accidentally based this branch on the mod branch. Hopefully that should be included before this one is done, so not a problem.

thesamovar:
Actually it bugged me so I've made a new branch and PR based on master.

PR #521, opened 2015-07-25
[MRG] Templates derivable

### LINKED ISSUE #357: Allow templates to be derived

thesamovar:
This PR fixes #357.

It adds a new method ``derive(...)`` to Templater that creates a new Templater derived from the original one with an additional template package name and (optionally) new global environment variables. The old templates will be used as a fallback if the new package doesn't define them. This is defined at the Jinja level using ChoiceLoader so everything should work smoothly (e.g. template inheritance should still work, etc.). To use it, you simply do this:

```
templater = CPPCodeGenerator.templater.derive('new.package.name')
```

Still needs documentation and tests, but would be happy to take comments on the approach.


thesamovar:
OK I added documentation and tests so this is ready to merge as far as I'm concerned.

In #357 we did discuss simplifying the templates, but this seems to be a separate issue so I'll open a new one (and I don' tthink this is critical for 2.0).

mstimberg:
I merged in current master as there have been some changes since the last tests, but this approach looks good to me and I'm happy to merge it as soon as the tests pass. And I agree that simplifying the templates is a separate issue.

PR #529, opened 2015-08-04
[MRG] Allow multiple objects with the same name at creation time

### LINKED ISSUE #216: Reuse old names if object no longer exists
Creating a new ``NeuronGroup`` (or other object) without specifying the name gives a series of names like ``neurongroup_1``, ``neurongroup_2``, etc. This happens even if the first object is deleted before the second one is created. This is problematic because it leads to unnecessary recompiles of weave code (the name of the group is embedded in the array name).

One thing we might also consider is making weave code not have the name of the group embedded in it, since it's not clear if this is necessary.

mstimberg:
For wildcard names, the same mechanism as before is used but explicitly given names are not checked for uniqueness at creation time. Instead, `Network.before_run` checks that all names within the network are unique. In addition, `CPPStandaloneDevice.build` checks that names are globally unique. Closes #216 

This should be particularly useful in interactive consoles and ipython notebooks whenever explicit names for objects are used. Should also make @owenmackwood happy.

### DIFF
--- a/brian2/core/names.py
+++ b/brian2/core/names.py
@@ -11,11 +11,13 @@ logger = get_logger(__name__)
 
 
 def find_name(name):
-    if name.endswith('*'):
-        name = name[:-1]
-        wildcard = True
-    else:
-        wildcard = False
+    if not name.endswith('*'):
+        # explicitly given names are used as given. Network.before_run (and
+        # the device in case of standalone) will check for name clashes later
+        return name
+
+    name = name[:-1]
+
     instances = set(Nameable.__instances__())
     allnames = set(obj().name for obj in instances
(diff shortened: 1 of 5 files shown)


thesamovar:
Looks great - happy for it to be merged when the tests pass.

owenmackwood:
I'm having problems with this, do you prefer discussion in brian-development rather than here?

mstimberg:
No I think it's fine to discuss it here if it is specifically about the change in this PR. If it is something that would be easier to sort out interactively, then you could also drop into our fancy (and more or less unused) chat: https://gitter.im/brian-team/brian2  (I'm off for lunch soon, though :) )

PR #530, opened 2015-08-05
[MRG] Special case the getting of all values of a state variable

### LINKED ISSUE #195: Special-case setting/getting of variables without indices
Currently, writing `G.v[:] = x` or `G.v = x` is basically equivalent to `G.v[np.arange(len(G))] = x`, i.e. an index array of all indices is created. This slows things down considerably for these simple cases, we should have a special code path for this "set/get all" use case.

mstimberg:
This is a trivial fix to make getting `group.var[:]` use `slice(None)` instead of `np.arange(len(group))` for indexing, it makes indexing in these cases (which will also be triggered by simple `group.var` usages) significantly faster (though it will not have any measurable impact on the performance most likely). I had a look at implementing an improvement for setting variables as well (as stated in #195) but decided against it: it's somewhat complicated and would mean changes to templates (with different code generated depending on the kind of indexing). All that for a questionable gain, setting a variable of a group of a million neurons only takes milliseconds to complete as it is...

Closes #195

### DIFF
--- a/brian2/core/variables.py
+++ b/brian2/core/variables.py
@@ -1143,14 +1143,14 @@ class VariableView(object):
             if not (isinstance(item, slice) and item == slice(None)):
                 raise IndexError(('Illegal index for variable %s, it is a '
                                   'scalar variable.') % self.name)
-            indices = np.array(0)
+            indices = 0
+        elif (isinstance(item, slice) and item == slice(None)
+              and self.index_var == '_idx'):
+            indices = slice(None)
         else:
             indices = self.indexing(item, self.index_var)
 
-        if variable.scalar:
-            return variable.get_value()[0]
-        else:
-            return variable.get_value()[indices]
+        return variable.get_value()[indices]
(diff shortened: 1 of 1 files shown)


thesamovar:
For setting (in runtime mode), can it not be done from Python directly rather than running via code generation? I think it's the most important optimisation to make because it will be used a lot in network operations, so even 1 ms adds up quickly if you several of them each time step.

mstimberg:
It does not go through code generation if there's no code involved, i.e. no string expressions for indexing or for the RHS value. But I did indeed forget that I could apply the same optimisation for this use case, I added this in a new commit (my reasoning above where for use cases where code generation is involved, i.e. with code for the right-hand-side).
With this, setting variables without indices takes more or less the same time as setting a raw array, the remaining overhead is basically all the checks whether this situation applies...

thesamovar:
Great! Good to merge when tests pass.

Issue #524, opened 2015-07-28
Some pre-synaptic updates give incorrect warnings


mariomulansky:
I'm using the following simple Synapse:
```
synapses = br2.Synapses(neurons, pre='v -= 0.2')
```
and I get a warning:
```
WARNING  brian2.codegen.generators.base: Came across an abstract code block that is not well-defined: the outcome may depend on the order of execution. 2 lines of abstract code, first line is: "v -= w (in-place)"
```

However, if I make the following trivial change:
```
synapses = br2.Synapses(neurons, pre='v += -0.2')
```
The warning disappears.

thesamovar:
Yes, this is a known limitation at the moment and will be fixed in a future release. As you say though, an easy workaround.

mstimberg:
Actually, since we don't have any other issue for this at the moment, I'll reopen it. It's not the only example of an incorrect warning, the following will generate one as well:
```Python
G = NeuronGroup(1, 'v : 1', threshold='v>1')
S = Synapses(G, G, pre='v_post += v_post')
```

thesamovar:
We will never be able to fix all incorrect warnings though (it's a hard mathematical problem). The best we can do is to cover the most common situations and change the wording of the warning to say that it 'may not be well-defined'. ``v_post+=v_post`` is a particular one in that it is well defined but it won't be possible to efficiently compute the class of all well-defined operations with numpy, and unless we explicitly cover the case ``v_post += v_post`` then this will be one of the cases where it won't be covered. In addition, I don't think it's a common use case, right?

thesamovar:
Also, we were supposed to have fixed this thing with ``-=`` and ``/=`` before but apparently we forgot. Not only that, we even had this same conversation before in #438. ;)

How about we just add ``-=`` and ``/=`` to the list of allowed operators and call that a day?

mstimberg:
The `-=` and `/=` is something we should definitely fix right away, before we forget about it again :)
In general, I'm less and less sure that our warning is useful -- I start to get the impression that we will more often have false warnings than useful ones.
The `v_post += v_post` example was of course contrived, but @btel came across this warning when simulating coupled oscillators, where the phase advance of the post-synaptic neuron depends on its current phase: `v_post += sin(-v_post)`.

thesamovar:
[...]

Issue #407, opened 2015-02-12
Make `INFO` the default log level and clean it up


mstimberg:
As we discussed, INFO should be the default log level and as a general rule it should output all information that is "guessed" by Brian, examples:
* Objects included in a magic network
* If time is reset or continued (in a magic network)
* Maybe the dt that is used?
* Automatically chosen integration method
* postsynaptic variables without suffix (#406)
* maybe: all "resolution conflict warnings", e.g. you have an `N` variable in your script but a string expression refers to the internal `N` (which has a different value)
* anything else...?

About the resolution conflict warnings: we currently have a mechanism to selectively suppress some warnings (even I don't think anyone ever used them) -- should we have the same for all log levels, maybe? With the above, what would remain to be a warning? Maybe everything that is almost certainly a mistake, e.g. an empty network or a synapses object without synapses?

thesamovar:
I would be wary of putting too much information on by default because we'll just train users to ignore it if we do that. I think the most important stuff that people should see by default is:

* Chosen integration method (unless it's exact or unless they specified it)
* Postsynaptic variables without suffix
* Resolution conflict warnings (unless there are too many of them in which case maybe just a short warning saying that there were resolutions conflicts and they should put a higher log level on to see them?)

The other stuff I would argue shouldn't be included by default. Maybe the times for the magic network runs, but even then I'd suggest against unless we have a reason to think that it might not be what is expected. So in the safe condition where e.g. it is the first run, or it is a second run with no change in objects, then we should not show it. In the unsafe conditions we could show it.

I would say the objects included in the magic network is not a good idea because in 90% of the cases, users won't give them names and so the names they see there will be just meaningless and confusing for them.

I don't think we need to list dt because by default we have a pretty safe value for that.

Are we going to need two different info levels now?! :)

mstimberg:
[...]

PR #531, opened 2015-08-05
[MRG] Improve order independence checks

thesamovar:
This is a partial solution to #524 (enough for the 2.0 release).

Note that the algorithm is described in https://github.com/brian-team/brian2/issues/435#issuecomment-86689930.


mstimberg:
Just one more thought: given that correct warnings are probably quite rare (and possibly occur less often than false positives), maybe we should downgrade this from a warning to an info message (which would still be displayed by default after implementing #407)?

thesamovar:
I would say not, because it's semantically a warning not just information, but I don't feel strongly about it so happy to change it if you'd prefer.

mstimberg:
I removed the condition that triggered the warning for `v_post += v_post` and similar statements -- I don't quite see how updating such a variable with an in-place modification can lead to an order-dependent update by itself (it "taints" the variable, though, and its use in other statements can lead to a warning). In particular, none of the "bad examples" triggered this error. If you don't have an example where this code path should trigger a warning (in which case this should be added as a test case), I'd be fine with merging this as it is. I think this small change significantly reduces the number of false positives and I'm therefore ok with keeping it as a warning.

thesamovar:
I'll need to check this very carefully I think because we can easily mess up here, and I also need to check that it's consistent with what happens in the ufunc.at code. Given that, might take me a little while to get round to merging this.

thesamovar:
I started looking into this and your modification wasn't good enough because it allowed ``v_post += v_pre; v_pre += v_post`` to pass, which shouldn't (now added to the test). I reverted that change and added a new test so that statements of the form ``var op (expr only involving var)`` now pass, which made your new tests pass. However, that doesn't really solve the problem. For example, ``v_post += w_syn*v_post`` should work but doesn't (also added to the test but not solved yet, which is why the tests are failing).

Still working on this.

I also added a new numerical check to the tests where it creates essentially a dummy pair of neuron groups and synapses between them and checks numerically if the order matters. This is useful as a test of the test (because it's quite hard to actually work out if a particular block of code should or shouldn't pass the test).

thesamovar:
OK I think this solution will probably have to do for now. The problem that I mentioned with the solution I developed in the previous post with the statement ``v_post += w_syn*v_post`` is actually more complicated than I thought. This statement is OK (order-independent, at least numerically). However, ``v_post += w_syn+v_post`` is NOT order-independent. So I think any approach that uses our current approach of just analysing which variables appear on the RHS of statements will always make a false-positive warning for this example. A more sophisticated approach is possible but would entail a lot more work, and more than is necessary for this edge case I think, given that it's only a warning and the text now only says it 'may be' order-dependent.

I did actually leave some (commented out) code for an alternative approach that just numerically evaluates it for some examples (same as in the tests), but this wouldn't be very clean and there would be plenty of issues with it (e.g. with user functions).

This last point is maybe something else to worry about. What do we do with user functions (which may not be stateless)?

Also, I haven't checked yet if we correctly handle the use of indices in statements (e.g. if the user were to write ``v_post = j`` what would happen?).

Finally, now that the order independence check is more permissive, does this potentially cause problems with the ufunc.at code? I haven't checked yet, but I think it might conceivably be the case that it could now produce incorrect results. Given all this, I'm changing this back from MRG to WIP for the moment.

@mstimberg: comments?

thesamovar:
Update: neither ``v_post=i`` nor ``v_post=j`` give any warning. It would be nice if this gave a warning, but the current version is at least an improvement on what came before.

thesamovar:
For the numpy ufunc.at vectorisation, we should add a new test that uses the same examples as the order independence check test (maybe combine the two tests into one). We should also check through the logic of it and be sure to raise an error if it might be wrong.

mstimberg:
This is really getting more and more complicated any minute... I start to wonder whether this is all really worth the effort. Are we doing all this to save the user from doing a mistake in their pre/post statement definition, or just to make them aware that their definition is not well-defined? And I don't quite remember: does ufunc.at give a completely bogus result in such cases or does it only give something that is not necessarily the same every time/ the same as weave, etc.?

thesamovar:
I agree, if it's just a matter of the warning it's OK if we miss some cases. The problem is I think that the ufunc.at vectorisation is giving bogus results in some cases. I was confident that with the previous version of the order independence check that ufunc.at wouldn't give bogus results, but with these changes it might (because it assumes that the old order independence check has passed). So for example, ``v_post += v_post`` is not currently raising any warning that vectorisation failed, but it definitely won't be giving correct results.

thesamovar:
One bit of good news though: it actually is handling the case ``v_post=i`` raises a warning and ``v_post=j`` passes, my test for this was wrong before.

thesamovar:
It's also giving a warning for ``v_post=rand()`` so perhaps we're ok there too.

mstimberg:
If we don't find a maintainable and understandable solution in a reasonable time, maybe another option: forget about the warning and the general case (since everything except numpy will give some reasonable solution for order-dependent formulations) and just set up a conservative check whether we can use `ufunc.at`, something that allows for the standard use cases but falls back to the pure loop (with the warning) whenever we are unsure. This should be considerable easier, and users don't do very complicated simulations with pure numpy anyway, I think.

thesamovar:
Yep, will do that.

thesamovar:
I think the situation is worse than we thought. I think the current numpy code is seriously broken for any cases except the ones where ufunc.at vectorisation is unproblematic (and we didn't notice this because it is unproblematic for all our examples). Firstly, we are using the ufunc.at vectorisation in cases where it is order dependent and I think this can't ever be right. Secondly, we are vectorising each line individually rather than considering them as a whole, which I think assumes something incorrect (that we can do statement1 for all indices then statement2 for all indices and this will always be the same as for all indices: statement1 then statement2).

In ``NumpyCodeGenerator.translate_one_statement_sequence`` we need to consider three cases:

1. Indices are all unique, in which case we do simple translation (as we do at the moment).
2. ufunc.at vectorisation is possible for the whole code block (not individual lines), in which case we do that. We don't even try to do this if it fails order-independence (I think, need to check this).
3. We fallback to a single loop over all indices, not one loop for each index.

As an example, consider ``v_post += 1; w_syn = v_post``. This is order-dependent. At the moment, this gets translated into ``add.at(v, _post_idx, 1); w = v[_post_idx]``. So w at the end will be getting the fully accumulated values of v, not the partially accumulated values that you would get it you looped. Maybe this doesn't matter so much because it is a pathological order-dependent case, but it illustrates the general problem that the assumption we're making isn't correct.

I'll work on this but it needs a fairly major rewrite of those two methods of ``NumpyCodeGenerator``, and some solid tests so that we don't fall in this trap again in the future.

thesamovar:
OK I think this is not going to be too hard to fix actually. I think the following rule for ufunc.at vectorisation works pretty well. It covers a lot of the easy and important cases, it doesn't rely on any of the details of the order-independence checking, and it's easy to see that it's robust (I think). The rule is simply this. Keep track of a list of ``used_variables`` which starts empty. Now for each statement ``var op expr`` check that the following rules are obeyed:

1. ``var not in used_variables``
2. ``var`` isn't used in ``expr``

If so, vectorise simply. If op is = then no change, if op is inplace replace with a ufunc.at. Now add all the variables in ``expr`` to ``used_variables``.

The logic of this is that for ``var op expr`` you want the variables on the RHS not to be modified later on, because if they are then the loops ``for statements: for indices:`` and ``for indices: for statements:`` will not give the same values (the value of the expression will change throughout the loop). This is guaranteed by the ``var not in used_variables`` condition. The same logic applies for ``var`` not in ``expr`` condition. 

I'll go ahead and implement this now (and write some tests), but I don't see any flaw in the logic, do you?

thesamovar:
OK I think that latest commit does it. It still needs some specific tests, and an improvement of the efficiency of the fallback loop (no need to read/write arrays for each statement any more).

In addition to the logic above, I realised that you only need to add main index variables to the ``used_variables`` set, because main index variables are only ever used once. This was necessary to pass one test, and for the best vectorisation.

@mstimberg it would be great if you could do a review of what I've done so far?

mstimberg:
I think the general logic sounds good, I'll have a closer look at it soon. 

> In addition to the logic above, I realised that you only need to add main index variables to the used_variables set, because main index variables are only ever used once.

I think you mean "only variables that *don't* use the main index", right? Actually, I think in addition you can exclude all variables that use a non-standard index, but one that has unique values. Otherwise, you'd unnecessarily treat subgroups differently.

I'm a bit confused where the new code is actually tested. This is not about the general order-independence tests but something specific to numpy, are there any tests for that?

thesamovar:
Agreed. New code is not yet tested.

thesamovar:
OK with that latest commit I think most of the problems are solved, but there are still a couple of things to resolve. Summary:

We now test ufunc.at vectorisation in cases where it should work. To do this I created a new class variable that allows you to turn off ufunc.at vectorisation (I don't think this should be a preference, no need to confuse users with it). I haven't written a test to check that in cases where ufunc.at vectorisation shouldn't work that it doesn't try to do so. In this case, all that happens is a warning is raised. Can we check for this in a test?

Improved efficiency of fallback code (reads/writes only done once). Also fixed a bug in here.

Improved order-dependence check to know about stateless/stateful functions. However, the main test suite is now raising a number of warnings for things like ``xi := (dt ** 0.5) * randn()`` which is indeed order-dependent but we shouldn't normally even be running the order-dependence check for cases like this, right? We need to look into this a bit more and decide what we want the required behaviour to be.

thesamovar:
This change fixes that last problem with too many warnings I think. It only does the order-dependence check if there are repeated indices. Is this OK do you think? If so, the only thing left to do here is to test that ufunc.at vectorisation fails when it should fail.

thesamovar:
Now we have the test that ufunc.at vectorisation fails when it should, I think this is ready to merge if you agree with what I did with the order-dependence warnings.

mstimberg:
I merged latest master into this branch and have a closer look at it now. One small thing I noticed: the `has_repeated_indices` function should be renamed, it returns the exact opposite, i.e. `False` if there are repeated indices.

mstimberg:
Ok, I removed some of the commented out/unused code and simply inverted the return value of the `has_repeated_indices` function. I have to admit I did not follow the logic of the order independence check in every detail, but the new test is very comprehensive and convincing. I am sure we will have way less false positives with this new check and even if we have some, the warning message is now much better. And of course most importantly, we should no longer generate wrong `ufunc.at` code...
Nice work! I am looking forward to not having to deal with these issues again...

Ready-to-merge from my side as soon as the test suites pass.

thesamovar:
Good stuff. Me too, I'll be very happy when this one is finished with! :)

thesamovar:
Tests passed, merging.

PR #532, opened 2015-08-05
[MRG] Improved weave functions and added tests for int(x) (fixes #325)

### LINKED ISSUE #325: Improve weave default function implementations
I'd like to improve the default implementations in ``cpp_generator.py`` of rand, clip, int, etc., to use inline functions and templates to be more generic.

thesamovar:


mstimberg:
The changes look good to me, I added a test that specifically tests for the bool->int conversion and updated the documentation (that implied that `int` is only to convert boolean values into integers). Good to merge after tests pass.

PR #534, opened 2015-08-05
[MRG] Loop-invariant optimisations for expressions with booleans

### LINKED ISSUE #394: Boolean flags are not optimised well by loop-invariant optimiser
For example, in the CUBA example we get the following abstract code:

```
_lio_const_1 = - dt
...
_ge = ge * exp((_lio_const_1 * int(not_refractory)) / taue)
...
```

We could have two scalar values depending on the value of ``not_refractory`` and use one or the other. We need to have something equivalent to an if statement to do this: so there's some work to do to figure how best to do it and whether or not it's justified in terms of efficiency. I think that in many cases it will be though: CUBA is typical in this.

thesamovar:
Started work on this (not ready to merge yet). There are going to be some issues to discuss.

Firstly, there's an issue with the current code in that it won't pull out, e.g. ``exp(y)`` where ``y`` is a scalar because the sympy op count for ``exp(y)`` is 1, but it's very much worth pulling out. I think the logic of this bit of code was not to pull out really trivial things like ``-y`` but we need a more sophisticated way of checking it. A related problem is that even an operation like ``y*y*y`` doesn't get pulled out because sympy simplifies it to ``y**3`` which is one operation, but then we use the unsimplified version ``y*y*y``.

Secondly, if you run the current example it transforms the vector code ``x += exp(y*b)`` where y is scalar and b is boolean to the following:

```
SCALAR
    _lio_const_1 := exp(y * 0.0) (constant)
    _lio_const_2 := exp(y * 1.0) (constant)
VECTOR
    x += (_lio_const_1*(1-b)+_lio_const_2*b) (in-place)
```

I wonder if it's worth simplifying the expressions for the LIO constants? Possibly not since they're scalar expressions so they're only evaluated once per call, compared to the vector operations. However, it might still be worth thinking about this because if one of the two has the value 0 then we could replace the vector operation by something simpler.

Things still to do include:
* Improve the op count checker
* Testing whether or not it's worth applying the boolean replacement LIO with the modified op count check
* Think about using sympy simplifications? (What are the issues here?)

Fixes #394


mstimberg:
I did not yet look at this in much detail, but I think we should certainly not overdo it, e.g. don't necessarily bother with simplifying, checking whether it is 0 etc.
About the op counts, a slightly more complex version of just using `count_ops` is to use the actual operations and assign some numerical values to them. The simple stuff will always be add, sub, etc. and everything else (e.g. exp, a function, etc.) is probably complex enough to pull out
```Python
>>> from brian2.parsing.sympytools import str_to_sympy
>>> expr1 = str_to_sympy('exp(-t/dt)')
>>> expr2 = str_to_sympy('-x')
>>> expr3 = str_to_sympy('myfunc(x)')
>>> subs = {'ADD': 1, 'SUB': 1, 'MUL': 1, 'NEG': 1, 'DIV': 1}
>>> expr1.count_ops(visual=True).evalf(subs=subs)  # complex
DIV + EXP + NEG
>>> expr2.count_ops(visual=True).evalf(subs=subs)  # simple
1.00000000000000
>>> expr3.count_ops(visual=True).evalf(subs=subs)  # complex
MYFUNC
```

thesamovar:
I believe that this should now be in a working state (although the tests aren't passing I think that's to do with something else).

For simple examples, this optimisation makes a very small improvement. However, for exactly solved examples that end up having a term like ``exp(-dt*not_refractory)`` it can make a huge difference. For reference, running CUBA for 10s using exact integration without this optimisation takes about 30s, and with this optimisation it takes 5.5s. By comparison, Euler integration takes about 7s for this example.

In my testing I noticed that the CUBA example doesn't use exact integration any more. I believe this is because the ``not_refractory`` variable messes things up. Quickly changing the linear integrator so that it treats ``not_refractory`` as constant allows it to work but I didn't commit this change because I wasn't sure what other effects it would have. I'll add a new issue for this.

In addition, there are possibly some minor improvements to make to this by looking for when the LIO constants are 0 or 1 and simplifying the expressions. Not sure it will make much difference though, because for weave or standalone it will already be able to simplify these.

Oh and last thing I made a new test module (``test_complex_examples``) because in an earlier iteration although all the tests were passing it was failing for the CUBA example. I think it's probably worth having a few 'real' examples in the test suite. I added CUBA running for just 10ms so it doesn't take too long to run.

thesamovar:
Oh, the Travis build error seems to be do with this issue and they're working on fixing it: https://github.com/conda/conda/issues/444

I have no idea what AppVeyor's problem is, it works fine on my Windows machine.

brian2/codegen/codeobject.py
```diff
@@ -286,11 +286,15 @@ def create_runner_codeobj(group, code, template_name,
...
+    user_defined = set()
     for v, u_v in zip(code.values(), user_code.values()):
-        _, uk, u = analyse_identifiers(v, all_variables, recursive=True)
+        d, uk, u = analyse_identifiers(v, all_variables, recursive=True)
         identifiers |= uk | u
-        _, uk, u = analyse_identifiers(u_v, all_variables, recursive=True)
+        defined |= d
+        d, uk, u = analyse_identifiers(u_v, all_variables, recursive=True)
         user_identifiers |= uk | u
+        user_defined |= d
```
mstimberg: If I'm not mistaken, neither `defined` nor `user_defined` are used anywhere, was this just for debugging maybe?

brian2/codegen/translation.py
```diff
@@ -227,13 +228,36 @@ def render_node(self, node):
...
+                        name_0 = self.optimisations[expr_0]
+                    if expr_1 not in self.optimisations:
+                        self.n += 1
+                        name_1 = '_lio_const_'+str(self.n)
+                        self.optimisations[expr_1] = name_1
+                    else:
+                        name_1 = self.optimisations[expr_1]
+                    newexpr = '({name_0}+{name_1}*{varname})'.format(name_0=name_0, name_1=name_1,
+                                                                     varname=varname)
+                    return newexpr
```
mstimberg: I think we should not use `*{varname}` directly but instead `*int({varname})` (actually, it would be good to enforce this in abstract code as well but I don't have a good idea how...). It's not guaranteed that direct multiplication with a boolean works for all languages (it works for Python and C but not for Java/RenderScript, for example).

brian2/tests/test_complex_examples.py
```diff
@@ -0,0 +1,40 @@
...
+    we = (60*0.27/10)*mV # excitatory synaptic weight (voltage)
+    wi = (-20*4.5/10)*mV # inhibitory synaptic weight
+    Ce = Synapses(P, P, pre='ge += we')
+    Ci = Synapses(P, P, pre='gi += wi')
+    Ce.connect('i<3200', p=0.02)
+    Ci.connect('i>=3200', p=0.02)
+
+    s_mon = SpikeMonitor(P)
+
+    run(10*ms)
```
mstimberg: It would be nice if we could assert something here, not sure what, though. Maybe at least stuff like `len(Ce)>0`?

dev/ideas/boolean_optimisations.py
```diff
@@ -0,0 +1,24 @@
+from brian2 import *
+from brian2.core.variables import *
+from brian2.codegen.translation import apply_loop_invariant_optimisations
+
+code = '''
+x += y
+x += y*y
+x += exp(y*b)
```
mstimberg: See my previous comment, I think this should be `x += exp(y*int(b))`

mstimberg:
About the travis build issue: it's now fixed in conda, but only for numpy 1.8.2 which is newer than the minimum version we previously used (1.8.0). I simply bumped up the minimum version, both Debian Stable and the latest long-term-stable release of Ubuntu have it, so it should be fine even for users relying on those distribution packages. No idea about the appveyor issue, the "Build failed" message is not very helpful, we should not hide compiler error messages like this.

brian2/codegen/codeobject.py
thesamovar: I thought I needed them but in the end didn't use them. Worth changing or just leave it?

brian2/codegen/translation.py
thesamovar: Yep, will make that change.

brian2/codegen/codeobject.py
mstimberg: I'd say it's worth changing to keep the diff as small as possible, unused local variables are also always a bit confusing.

thesamovar:
OK I think all the minor issues are fixed now, and in addition I changed standalone so that it will now always print stdout if there is a runtime or compile time error. This should help us debug appveyor.

mstimberg:
Well, appveyor does not seem to fail anymore... https://ci.appveyor.com/project/brianteam/brian2/build/665

I think after all the tests pass this is ready to merge. I'll run the examples to be sure, it is an important change.

mstimberg:
All the tests pass and the examples run fine as well. Anything you'd still like to add (tests/documentation/...)?

thesamovar:
I think it's good to go as far as I'm concerned.

mstimberg:
Ok, nice work, merging it now.

PR #537, opened 2015-08-07
[MRG] Fix OpenMP profiling

mstimberg:
This has two small fixes to the OpenMP support: in the synapses template, every code that used a scalar variable incorrectly triggered the use of the single-thread code (affecting the post code in STDP). The second fix is to use `omp_get_wtime` instead of `std::clock` to count runtimes, this avoid the weird profiling results and makes it unnecessary to divide by number of threads etc.

It's all fairly simple changes and it would be ready to merge but the OpenMP test now apparently no longer compiles on Windows. @thesamovar could you have a look at it? It does not give any helpful output in the log and trying to figure this out in my Windows VM is way too annoying for a Friday evening...

### DIFF
--- a/brian2/devices/cpp_standalone/codeobject.py
+++ b/brian2/devices/cpp_standalone/codeobject.py
@@ -35,6 +35,11 @@ def openmp_pragma(pragma_type):
             return '1'
         else:
             return '%d' %nb_threads
+    elif pragma_type == 'with_openmp':
+        # The returned value is a proper Python boolean, i.e. not something
+        # that should be included in the generated code but rather for use
+        # in {% if ... %} statements in the template
+        return openmp_on
 
     ## Then by default, if openmp is off, we do not return any pragma statement in the templates
     elif not openmp_on:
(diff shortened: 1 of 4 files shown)


thesamovar:
Will do!

thesamovar:
Ah, does this solve some of the issues that people have been telling us about recently?

mstimberg:
Only very partially (but the standard profiling_summary can now be used to look at issues in more detail): the synaptic post code in STDP was not parallelized and that is fixed with the first commit, so this will scale nicely now with more threads. For the pre pathway, it is still completely sequential, though. Same for the threshold, which can also take a fair amount of time.

thesamovar:
Runs fine here, and appears to be building fine on AppVeyor now...

thesamovar:
Ah no, it's not. OK, looking into it.

thesamovar:
OK the problem was just that in OpenMP 2.0 (the version with MSVC) you can't use unsigned ints, but in 3.0 (which gcc has) you can. Changed this and should be fine now.

thesamovar:
Also good to merge as far as I'm concerned.

mstimberg:
Thanks for figuring this out, I'm merging it now.

PR #540, opened 2015-08-12
Heun stateupdater

ttxtea:
Here we have added the stochastic Heun method as a StateUpdateMethod. A description can be found in [sde tutorial](http://infoscience.epfl.ch/record/143450/files/sde_tutorial.pdf). It was tested with the [sde test example](https://brian2.readthedocs.org/en/2.0b4/examples/advanced.stochastic_odes.html). The results are comparable to the Milstein method:

output_0_1

### DIFF
--- a/brian2/groups/neurongroup.py
+++ b/brian2/groups/neurongroup.py
@@ -370,7 +370,7 @@ class NeuronGroup(Group, SpikeSource):
     add_to_magic_network = True
 
     def __init__(self, N, model,
-                 method=('linear', 'euler', 'heun', 'milstein'),
+                 method=('linear', 'euler', 'heun'),
                  threshold=None,
                  reset=None,
                  refractory=False,
--- a/brian2/groups/neurongroup.py
+++ b/brian2/groups/neurongroup.py
@@ -370,7 +370,7 @@ class NeuronGroup(Group, SpikeSource):
     add_to_magic_network = True
 
     def __init__(self, N, model,
-                 method=('linear', 'euler', 'milstein'),
+                 method=('linear', 'euler', 'heun', 'milstein'),
                  threshold=None,
                  reset=None,
                  refractory=False,
--- a/brian2/spatialneuron/spatialneuron.py
+++ b/brian2/spatialneuron/spatialneuron.py
@@ -96,7 +96,7 @@ class SpatialNeuron(NeuronGroup):
                  threshold_location=None,
                  dt=None, clock=None, order=0, Cm=0.9 * uF / cm ** 2, Ri=150 * ohm * cm,
                  name='spatialneuron*', dtype=None, namespace=None,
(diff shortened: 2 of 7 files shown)


mstimberg:
Many thanks! I think it should replace milstein as the default integrator for equations with multiplicative noise since it also supports non-diagonal ones. For this, the default values for the `method` argument in `NeuronGroup`, `Synapses` and `SpatialNeuron` need to be adapted, as well as the documentation in `docs_sphinx/user/models.rst`. Finally, the tests in `brian2/tests/test_stateupdaters.py` should test `heun` as well -- I think basically it should be added everywhere where `milstein` is tested.
Would you mind doing those changes as well?

ttxtea:
On Thu, Aug 13, 2015 at 02:15:40AM -0700, Marcel Stimberg wrote:
> Many thanks! I think it should replace milstein as the default integrator for equations with multiplicative noise since it also supports non-diagonal ones. For this, the default values for the `method` argument in `NeuronGroup`, `Synapses` and `SpatialNeuron` need to be adapted, as well as the documentation in `docs_sphinx/user/models.rst`. Finally, the tests in `brian2/tests/test_stateupdaters.py` should test `heun` as well -- I think basically it should be added everywhere where `milstein` is tested.
> Would you mind doing those changes as well?

Yes sure - I'll do that!

ttxtea:
The heun state updater now works as a default for multiplicative SDEs, when not method is specified by the user. Thanks for pointing out where the changes had to be made!!

mstimberg:
Looks great! One small thing: you can replace `milstein` by `heun` in the default value for the method arguments instead of adding it. The way the automatic determination works is that it goes through the list and uses the first integrator where `can_integrate` returns `True`. Since `heun` can integrate everything that `milstein` can integrate, this will never go beyond `heun`, so there's no need for having `milstein` following it.

ttxtea:
Good point. Millstein really deals with a subset of heun - so not point. What is next - Crank-Nicolson for PDEs :D? Would that even be possible in brian2? Without vectors, one would have lots of odes for the spatial points.

mstimberg:
We do somewhat support PDEs, at least in the context of multicompartmental models :) But not Crank-Nicolson, since we can't have implicit methods currently.
About the change: it's almost perfect, the documentation still states that it will chose "..., heun or milstein", though. Pro tip: if you make a commit that only changes the documentation, include `[ci skip]` in the commit message (can be on its own line in the end). This way, the full test suite will not be run again for a commit that doesn't change any code.

ttxtea:
Thx - learning a lot here - and its great fun!

mstimberg:
Everything looks good to me and ok to merge. For some reason appveyor is stuck with the test suite, I will have another look tomorrow.

mstimberg:
Ok, I had to manually restart the appveyor test suite but all the test passes, examples also still work, so I'm merging it now. Thanks again for the contribution!

PR #541, opened 2015-08-13
[MRG] Improve stochastic parsing

mstimberg:
This branch contains a few improvements for dealing with complex, in particular stochastic, equations. In `Expression.split_stochastic`, the use of sympy's `expand` could result in very big strings that made everything slow and could even lead to memory errors (for the example provided on the mailing list https://groups.google.com/d/topic/brian-development/fL8wfggCMrw/discussion it resulted into a > 2MB string...). I removed the `expand` and also replaced the quite demanding `match` by a simple `collect`, it seems to work fine. The branch also contains a few smaller improvements that should avoid a few unnecessary sympy operations.

### DIFF
--- a/brian2/equations/codestrings.py
+++ b/brian2/equations/codestrings.py
@@ -68,16 +68,30 @@ class Expression(CodeString):
 
     Parameters
     ----------
-    code : str
+    code : str, optional
         The expression. Note that the expression has to be written in a form
-        that is parseable by sympy.
+        that is parseable by sympy. Alternatively, a sympy expression can be
+        provided (in the ``sympy_expression`` argument).
+    sympy_expression : sympy expression, optional
+        A sympy expression. Alternatively, a plain string expression can be
+        provided (in the ``code`` argument).
     '''
 
-    def __init__(self, code):
-        CodeString.__init__(self, code)
+    def __init__(self, code=None, sympy_expression=None):
+        if code is None and sympy_expression is None:
+            raise TypeError('Have to provide either a string or a sympy expression')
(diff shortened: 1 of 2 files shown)


ttxtea:
Hi,

Concerning the separation of the eq_expression into drift and diffusion part. I think this is a function of general interested. For example to extract the diffusion matrix from a brian2.Equations SDE once can just use:

Smat = [[sympy.S(j[1]).coeff(k)
               for k in sde.stochastic_variables] 
              for j in sde.eq_expressions]

this would not need a simplification of the expression. I am not sure how it is done at the moment. To extract the drift part I just substitute all stochastic_variables with 0 :D. Should one maybe add two function for that?

by the by: I worked with this branch a bit more - no hickups sofar.

mstimberg:
I'm not sure that this is something that many users need, after all you provide the equations yourself :) But in principle you can already use the function that Brian uses internally:
```Python
>>> eqs = Equations('dv/dt = -v/tau + xi*tau**-0.5 : 1')
>>> eqs['v'].expr.split_stochastic()
(-v/tau, {'xi': tau**(-0.5)})
```
It's not that different from `coeff`, but it will raise an error if the equation contains `1/xi` or `xi**2` or something like that, while `coeff` would just return you nothing for `xi`, I think.

thesamovar:
I'm happy with the changes if you're ready to merge.

mstimberg:
Ready to merge I am!

thesamovar:
Going Yodic are we?

mstimberg:
It's Friday night and all that :)

PR #543, opened 2015-08-20
[MRG] Make `BinomialFunction` extensible for other devices/codegen targets

### LINKED ISSUE #542: Make `BinomialFunction` extensible for other devices
Currently, all the code is hard-coded in `BinomialFunction.__init__`, i.e. new devices/codegen targets can't provide their own implementation. This should be made possible, the `BinomialFunction` class should store functions that will return the code in a dictionary that can then be extended by other targets.

mstimberg:
Closes #542


thesamovar:
I had 5 minutes and this looks like it might be useful genn so merging now.

Issue #251, opened 2014-04-25
Automatically build C++/Cython extension package from ``Network`` object


thesamovar:
One of the features I would like to have is that I could create a function to which I can pass a series of parameters and returns the result of a simulation, but that the function is automatically turned into a C++ or Cython extension package. This would work just the same as C++ standalone with a couple of additions:
- We would provide a simple interface for it, maybe a decorator or class
- We could specify a set of parameters that are passed to the function and wouldn't require a rebuild, including array data as well as constants
- You would specify what data should be returned by the function

This would enable a whole lot of incredibly useful features. Most importantly, you could easily control multiple runs, analysis code, multiprocessing, etc., from a Python script but have it run in optimised C++. I think if we write this, it might turn out to be the most useful way of running Brian.

thesamovar:
As a short term solution, we will allow reruns of a standalone project including changing parameters.

thesamovar:
I think we're not going to get to this before 2.0 so I'm dropping the milestone.

mstimberg:
Agreed...

thesamovar:
Just dropping a note to myself here. One example of freezing variables that could be avoided is for SpikeGeneratorGroup, we currently freeze the number of input spikes. This can be deduced from the file size so doesn't need to be frozen. This would be a quick change that would reduce some recompilation.

thesamovar:
OK I've picked off the lowest hanging fruit here with a couple of minor changes that reduce the number of recompilations and let you choose which variables to write out. I think to make any further progress, there needs to be a way to read arbitrary data from a file (or API call) into the simulation. At the moment, everything except static arrays is hard coded. We need a nice way of doing that. Any thoughts? If not, I'll just start trying a few things, but thought it was worth asking.

By the way, I've decided that this issue will be finished by the end of the summer. It has been long enough that I've been wanting this feature in Brian...

thesamovar:
I think the only thing that will be difficult will be how to not freeze parameters. The reason that'll be difficult is we might want to do something like this:

```
@encapsulate
def f(tau):
    v0 = 1.5
    eqs = '''
    dv/dt=(v0-v)/tau : 1
    '''
    G = NeuronGroup(1, eqs, threshold='v>1', reset=v=0)
    M = SpikeMonitor(G)
    run(100*ms)
    return M.t
```

How does Brian know that ``tau`` should be a modifiable parameter but not ``v0``? How do we represent that in our code since at the time that we build this network, there is no value for ``tau``?

[...]

PR #544, opened 2015-08-21
[MRG] Synapses to synapses connections

mstimberg:
This is quite a big change, necessary for the astrocyte project where we need `Synapses` that target other `Synapses`. There are a few minor tweaks and additions but the two main changes are the following:
* The variable `N` is now no longer an `AttributeVariable` that gets the length of the group via Python (or the `Synapses` class in the C++ standalone code) but instead either a constant or a scalar `ArrayVariable` (for synapses). In runtime code, it can still be simply accessed as `N`, as the value of constant scalar arrays is passed in the namespace. In standalone, two templates contain an `{% if ... %}` to generate the correct code depending on whether `N` is a constant or an array.
* The variables `N_incoming` and `N_outgoing` were previously set in Python code, they are now set in the `synapses_create` template. This is potentially redundant for multiple `Synapses.connect` calls, but the overhead should not be noticeable (and multiple connect calls are not that common, anyway).

With these changes the C++ `Synapses` class in C++ standalone became obsolete, so I removed it.

### DIFF
--- a/brian2/codegen/generators/cython_generator.py
+++ b/brian2/codegen/generators/cython_generator.py
@@ -187,6 +187,8 @@ class CythonCodeGenerator(CodeGenerator):
         handled_pointers = set()
         user_functions = []
         for varname, var in self.variables.items():
+            if isinstance(var, Variable) and not isinstance(var, (Subexpression, AuxiliaryVariable)):
+                load_namespace.append('_var_{0} = _namespace["_var_{1}"]'.format(varname, varname))
             if isinstance(var, AuxiliaryVariable):
                 line = "cdef {dtype} {varname}".format(
                                 dtype=get_cpp_dtype(var.dtype),
--- a/brian2/codegen/generators/cython_generator.py
+++ b/brian2/codegen/generators/cython_generator.py
@@ -235,7 +235,7 @@ class CythonCodeGenerator(CodeGenerator):
                 if not var.scalar:
                     newlines += ["cdef int _num{array_name} = len(_namespace['{array_name}'])"]
 
-                if var.array and var.scalar and var.constant:
+                if var.scalar and var.constant:
                     newlines += ['cdef {cpp_dtype} {varname} = _namespace["{varname}"]']
                 else:
                     newlines += ["cdef {cpp_dtype} {varname}"]
--- a/brian2/codegen/generators/cython_generator.py
+++ b/brian2/codegen/generators/cython_generator.py
@@ -235,7 +235,10 @@ class CythonCodeGenerator(CodeGenerator):
                 if not var.scalar:
                     newlines += ["cdef int _num{array_name} = len(_namespace['{array_name}'])"]
 
-                newlines += ["cdef {cpp_dtype} {varname}"]
+                if var.array and var.scalar and var.constant:
+                    newlines += ['cdef {cpp_dtype} {varname} = _namespace["{varname}"]']
+                else:
+                    newlines += ["cdef {cpp_dtype} {varname}"]
 
                 for line in newlines:
                     line = line.format(cpp_dtype=get_cpp_dtype(var.dtype),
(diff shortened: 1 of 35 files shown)


thesamovar:
Quick first thought. One of the tests failed due to an out of memory error. I've occasionally had this in the past when running the test suite too (although very rare). This makes me wonder if we have a memory leak?

thesamovar:
Apart from that issue, the code looks good (although I think I should also test on my machine before merging because it's a big change - will do that asap).

If we can do the same thing with the size of a ``NeuronGroup`` as with the ``N`` in ``Synapses`` then perhaps we are almost there in making it possible to build/compile C++ standalone once but run it for different group sizes? Or will that be substantially more work?

mstimberg:
> This makes me wonder if we have a memory leak?

There seems to be indeed some kind of leak. If I run the Cython test suite, memory usage goes to over 1GB in the end, this is not the case for numpy or weave. I'll look into this (not sure whether this was introduced in this branch but it needs to be fixed either way).

> If we can do the same thing with the size of a NeuronGroup as with the N in Synapses then perhaps we are almost there in making it possible to build/compile C++ standalone once but run it for different group sizes? Or will that be substantially more work?

No, this is still far off and beyond what we want to allow as a short-term solution for #251. The size of a `NeuronGroup` is something that is hard-coded into the generated code and to make this changeable without a recompile would mean a quite significant change. We'd have to special case `N` so that it is loaded before we even create any arrays. Or we'd have to change *all* arrays into dynamic arrays and add some code to initialize state variable arrays at some point. Either way, not something that we'll see before the 2.0 release I think.

mstimberg:
I think the memory leak might have been a red herring... I could reproduce increasing memory sometimes but not always, I think that might rather have been an internal Cython issue. The error in the test suite was most likely not triggered by this, it would be a big coincidence if it happened to trigger the `MemoryError` in the very test that was testing the new synapses to synapses connections... I think the issue was rather that `N_post` was not interpreted correctly (for a synapse connecting to another synapse, it should correspond to the number of target synapses but it referred to the `N_post` variable of the target group, i.e. to the number of neurons in the group that the target synapse was targeting...). In addition, scalar constants were not passed into Cython correctly and somehow that must have led to the use of an uninitialised value for `N_post` (hence the warning in the failed test). This value of this variable was used to resize a dynamic array, and I think this triggered the issue. I'm not 100% sure why this did not pose a problem all the time, but I'm reasonably confident that it works now (but of course, please run the tests on your machine as well).

thesamovar:
OK normal test runs fine on my machine, and it's now passing the CIs too. I will run a long test later before merging, but should be good to go.

For the memory, I watched memory consumption while running the tests on this branch and another, and in both cases it was almost identical. Gradually increasing during weave and cython (expected because loads of modules are staying in memory while the tests are run) from about 100MB at the beginning of the tests up to about 250MB by the end. Certainly nothing that should cause an out of memory exception.

Could it be something to do with the multiprocess tests?

Or possibly if there is an exception in the Cython code it doesn't clear up memory correctly - so even though you see a memory error at the end of the test run, it was actually caused by another unrelated exception earlier?

thesamovar:
OK long tests run fine on my machine. Happy for this to be merged - I would merge it but I wasn't entirely sure if you wanted to check a couple of things yet.

mstimberg:
About the `MemoryError`: I am pretty confident it had to do with the way the `N_post` variable was passed to the Cython code. The `MemoryError` was raised in a `DynamicArray.resize` which very likely got a sum involving `N_post` as an argument and there was a warning `local variable 'N_post' referenced before assignment` before. It's possible that this issue depended on something non-deterministic like the order of items in the `variables` dictionary, that's why we did not get it all the time. 

I don't have anything else to add right now but I'll run the long test suite and the examples on my machine as well. I'll merge if everything passes.

BTW: @Kwartke and @moritzaugustin I'm afraid this will again break your code, you'll have to do the equivalent changes that the C++ standalone templates did, I think.

mstimberg:
Everything passes and github seems to recover, merging now.

PR #546, opened 2015-08-25
[MRG] Resize variables correctly in `StateMonitor`

### LINKED ISSUE #518: `store` broken for `StateMonitor ` in weave and Cython
This problem was introduced with the performance optimisations for `StateMonitor` in PR #486. It triggers a test failure on my machine but since it is related to memory allocation, it does not necessarily trigger it all the time (it does not on travis/appveyor). The problem is that the `StateMonitor` templates do directly resize the dynamic arrays and do not update the `size` attribute of the `Variable` object.
It sounds like a trivial issue but it is actually a bit tricky. We can either update the `size` attribute from within the template, but this will bring down the performance again since we have to go through Python for that at every time step. The other option is to make everything consistent in the `after_run` method, but then we need to special case the `RuntimeDevice` or move all of this into a new template...

mstimberg:
This uses the `Variable` object itself to resize the dynamic array in state monitor. This is slightly slower than using `DynamicArray.resize` but it will update the `Variable.size` attribute as well, which is important, especially for the `store`/`restore` mechanism (which was broken for Cython and weave).

Fixes #518

### DIFF
--- a/brian2/codegen/runtime/cython_rt/templates/statemonitor.pyx
+++ b/brian2/codegen/runtime/cython_rt/templates/statemonitor.pyx
@@ -9,7 +9,7 @@
     cdef int _new_len = _curlen + 1
 
     # Resize the recorded times
-    {{_dynamic_t}}.resize(_new_len)
+    _var_t.resize(_new_len)
     {{_dynamic_t}}[_new_len-1] = _clock_t
 
     # scalar code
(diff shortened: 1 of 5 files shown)


thesamovar:
Code looks fine to me: happy for this to be merged when CIs pass it (appveyor needs to be restarted).

How much of a speed difference is it?

mstimberg:
> How much of a speed difference is it?

I did not run comprehensive tests, but it does not seem to change significantly. I added a `StateMonitor` to CUBA, it takes ~15.5% of the total simulation time before and something like 15.8% after the change...

mstimberg:
All checks passed, merging now.

PR #549, opened 2015-08-25
[MRG] Check that source/target indicies provided to `Synapses.connect` lie within the correct range

### LINKED ISSUE #548: Connection arrays for `Synapses.connect` are not checked for whether they make sense
The source and target index arrays are stored as they are given, which means that for incorrect arrays (too small or big values), weird problems (segmentation faults in weave, etc.) can occur later on that are hard to debug. Calculating the minimum and maximum of even very big arrays shouldn't take too long and we can check it against the size of the source/target group.

The only potential problem I see with this that it cannot work in standalone for connections to other Synapses, but I think this is a less important issue (you can still use string conditions). Potentially we could include a switch in `Synapses.connect` that allows to switch off the check for special circumstances.

mstimberg:
Fixes #548

### DIFF
--- a/brian2/synapses/synapses.py
+++ b/brian2/synapses/synapses.py
@@ -1156,6 +1156,18 @@ class Synapses(Group):
 
             sources = np.atleast_1d(sources).astype(np.int32)
             targets = np.atleast_1d(targets).astype(np.int32)
+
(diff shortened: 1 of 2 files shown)


thesamovar:
Happy to merge when CIs pass.

PR #550, opened 2015-08-27
[MRG] Better error message for `StateMonitor(synapses, ..., record=True)` in standalone

mstimberg:
This also updates the documentation and makes another error message slightly nicer (which occurs if you do something like `StateMonitor(synapses, ..., record=synapses['i==0'])` in standalone).


thesamovar:
Looks good to me: happy for it to be merged when tests pass.

Issue #551, opened 2015-08-28
Writes to shared variables not correctly handled with OpenMP


mstimberg:
Writes to shared variables will be executed by each thread, therefore things like `s += 1` will not give the expected result. Example:
```Python
from brian2 import *
set_device('cpp_standalone')
prefs.devices.cpp_standalone.openmp_threads = 4
G = NeuronGroup(10, 's : 1 (shared)')
G.run_regularly('s += 1')
run(defaultclock.dt)
device.build(run=True, compile=True)
print G.s
```
This should print 1.0, but it prints 4.0 (or sometimes 3.0 if two threads update `s` at the same time).

This is not trivial to fix, and I'm wondering whether we don't have to introduce a massive change to the templates to correct this. The problem is that we can't simply wrap the scalar code in the template with a `simple` or `master` pragma, because then all variable definitions within this block are not accessible outside, which would prevent vector code from accessing any of the scalar variables (e.g. the `lio` constants). So we'd have to split the `scalar_code` block into a part that declares the variables and one that contains the statements. I'd prefer to avoid this, since it entails low-level changes to the code generation pipeline and will also break the GPU devices that are currently worked on. I think a better option would be to not have the OpenMP `parallel` block in the network run, but in each code object individually. This way, we could have the scalar code executed serially before going into the parallel part. Another advantage would be that there would be no longer a need for the `IS_OPENMP_COMPATIBLE` annotation -- if a template does not know about OpenMP it does not open a parallel block and everything runs serially. It would be a change to all templates, but one that would not affect the GPU devices.
However, I think @yger did something like this when he started working on the OpenMP support and IIRC, this was too slow because of OpenMP overhead? Maybe I should just manually try this out with an example..

thesamovar:
I think the easiest solution would be to wrap the shared code with:

```
if(omp_get_thread_num()==0)
{
// scalar code
...
}
#pragma omp barrier
```

The if check will ensure that only one thread does it, and the barrier ensures that shared variables aren't used before they're defined.

mstimberg:
This is equivalent to the use of something lke `#pragma omp master` or `#pragma omp single`. This was actually the first thing I tried but as I mentioned above, we would have to split up the scalar block into declarations and statements for this to work. Otherwise, this will generate code like this:
```C++
#pragma omp single
{
// scalar code
const int _lio_const = exp(-0.01);
}

// vector code
... = ... _lio_const ... ;  //error: _lio_const not declared
```

[...]

Issue #390, opened 2015-01-23
Deal with preparation in templates


mstimberg:
Currently, the way to do `before_run` preparations so that they are properly supported both in runtime and standalone is to create a new `CodeObject` that is executed in `before_run`. This approach is used for initializing the `SpikeQueue` and for the preparation of `SpatialNeuron` (currently only in the `spatialneuron` branch). This adds a lot of boilerplate code to the main Python classes and also adds to our proliferation of templates. How about having them in the main templates as (optional) blocks? I.e. for every template, you could specify code that is executed before, during and after a run.
I don't see any use case for user-generated code in the before and after run part, currently, but we could always add this option later.
The only disadvantage of this that I'm seeing is that it would make the numpy templates slightly more complicated, since they don't use any blocks at all, currently. On the other hand, this would make them more similar to the other codegen target templates.

thesamovar:
I'm happy with this idea, and I don't think it's a problem to introduce blocks to the numpy templates.

I'm wondering if it would also be possible to have templates in different languages like this? e.g. preparation code in Python but executable code in C++. I think we do this at the moment, right?

mstimberg:
> I'm wondering if it would also be possible to have templates in different languages like this? e.g. preparation code in Python but executable code in C++. I think we do this at the moment, right?

We have a special `python_pre` block that we can use that will call Python code. I'm not sure this is really necessary, though, since all of our runtime targets can also call Python code directly. It'd be slightly involved for weave, if we were calling a function with arguments but currently our only use case is to call `_owner.initialise_queue()`, and this we could also write directly in weave directly (I didn't try it, but something like `_owner.call_method("initialise_queue");` I think.).

thesamovar:
OK I'm happy to go with whatever you think is the nicest solution here.

mstimberg:
I've started looking into this again after doing a few small changes to the templates to make them more consistent and easier to understand. I still think this is a good idea, but I realized that we have two options:
1. We include the preparatory code in the same file as the standard run code but execute it only the first time (with `if` or `try`/`except`).
2. We put the code into a separate file and call it explicitly.

I think I prefer 2 because it seems more direct and would also extend to `after_run` code, but it would add yet another source file and therefore increase compilation time (in particular for Cython). On the other hand, we use this feature very rarely (in most simulations, there will be one additional object/file for each synaptic pathway), so it probably does not have much of an effect in practice.

mstimberg:
[...]

PR #552, opened 2015-08-28
[MRG] Model clocks as `Group`

### LINKED ISSUE #389: Expose the integer time step of `Clock`
This could make things more straightforward for `TimedArray` and `SpikeGeneratorGroup` and avoid going from integer timesteps to floating point time and back.

### LINKED ISSUE #478: Changing dt's between runs does not currently work in standalone
In runtime mode, this code
```Python
from brian2 import *
G = NeuronGroup(1, 'v:1')
run(1*ms)
defaultclock.dt = 1*ms
run(3*ms)
print mon.t
```
correctly prints:
```
<statemonitor.t: array([ 0. ,  0.1,  0.2,  0.3,  0.4,  0.5,  0.6,  0.7,  0.8,  0.9,  1. ,
        2. ,  3. ]) * msecond>
```

whereas in standalone mode it prints
```
<statemonitor.t: array([ 0.,  1.,  2.,  3.]) * msecond>
```
i.e. the last `dt` is used for the full simulation.

mstimberg:
This is a proposal for a different approach to the internal implementation of clocks. 

The basic idea is that `Clock` is a `Group` with the state variables `timestep`, `t` and `dt`, which are all scalar arrays that are set/retrieved as other state variables (and therefore changing them between runs in standalone is not a problem anymore). In addition, it uses `run_regularly` to run abstract code that updates the clock, so this is now part of the generated code. As such, this is quite neat because it makes clocks less special (they are still special because they are used for scheduling, of course) and most importantly it allows us to get completely rid of `AttributeVariable`. Unfortunately, there are a few complications:
* We still want to have `t` and `dt` accessible from pure Python, because even in standalone mode you might want to write `run(defaultclock.dt)`, or use `defaultclock.t` after a run -- also some groups like `PoissonGroup` or `SpikeGeneratorGroup` need to do checks in Python that involve `t` and `dt`. So clocks now store `t` and `dt` twice, once as internal Python attributes, once as standard state variables. This needs some cleaning up and documentation
* To create state variables we need a device, but when the `defaultclock` is created we don't have a device yet... Also, you switch to the standalone device only *after* the `defaultclock` has been created, therefore its state variables are not stored with the device. Fixing this makes things slightly cumbersone, each device now has its own `defaultclock` and the standard `defaultclock` object is a proxy to the respective clock (but `dt` is carried over when you switch a device). Not sure, but this might even be an advantage for brian2genn where it would allow it to define a `GeNNClock` which could then be used for GeNN-specific clock code in the code generation process
* A minor annoyance: templates that directly (i.e. not in user code) refer to `dt` and `t` now have to use `{{dt}}[0]` and `{{t}}[0]` in standalone mode

So this all works now, except for standalone mode with OpenMP, because the code that updates the clock writes to shared variables and this does not work correctly (just opened this as #551).

I am more or less happy with it, because I never liked the whole `AttributeVariable` thing and I have the feeling that having `t` and `dt` stored in a standard way might simplify other things in the future (e.g. storing/restoring the state of a network). However, you (meaning: you as a developer, not as a user) now need to be careful whether you access the "state variable dt" or the "Python dt" and some other complications which I listed above. An alternative approach to fix #478 that would leave most things as they are would be to simply allow the setting of `AttributeVariable` values and implement this similar to the getting of values (i.e. assume that there's an object and a corresponding setter method in standalone). 

What do you think?

Fixes #478, also fixes #389


mstimberg:
Just to re-emphasize this: the test suite should pass, except for OpenMP standalone which is broken due to #551

thesamovar:
Just a quick note. Shared variables are scalars right? So why not use an atomic operation? Or check that we're in thread 0?

thesamovar:
Ignore that last thing about atomic writes. The point about thread 0 is now elaborated on in a new comment on #551.

I think you're right that the ``AttributeVariable`` mechanism is antiquated and that ``Group`` has evolved to the point where it's the natural place to put all data (now that we have the scalar mechanism). Some comments.

> In addition, it uses run_regularly to run abstract code that updates the clock, so this is now part of the generated code.

I'm not entirely sure about this. We would want to make sure that clocks are always updated as the last thing, so we still need some additional checks to make this work. I'm wondering if those additional checks are as much or even more effort than just having specific clock updating code?

> So clocks now store t and dt twice, once as internal Python attributes, once as standard state variables.

Can't they be properties so that the data is only stored once?

> To create state variables we need a device, but when the defaultclock is created we don't have a device yet

Could we just change the order of imports? I guess probably not since everything imports clock early.

Another option would be to make ``dt`` a preference and just get rid of ``defaultclock`` entirely. Now you'd write ``prefs.dt = ...`` at the beginning of your script, and each ``Network`` could have its own clock that would get created with that ``dt``. You'd lose ``defaultclock.t`` but maybe that's not so bad? We would presumably have an attribute ``Network.defaultclock`` so you could get access to it if you really needed to. But you can already do ``magic_network.t`` I think?

> {{dt}}[0]

We use this idiom a lot now in the templates and it's always quite ugly. I'm wondering if we could make this cleaner somehow? For example, maybe for scalar variables, ``{{dt}}`` doesn't give ``_clock_dt`` but directly ``_clock_dt[0]``? I guess there are some complications, but perhaps there's a way to introduce hooks into jinja that could make this part of the code generator objects. Another alternative, ``{{dt|scalar}}`` that would be just a filter that adds ``[0]`` afterwards but it would be at least more semantically satisfying.

mstimberg:
> I'm not entirely sure about this. We would want to make sure that clocks are always updated as the last thing, so we still need some additional checks to make this work. I'm wondering if those additional checks are as much or even more effort than just having specific clock updating code?

There's no specific code in place for this, it's simply using `when='after_end'` as its time slot, I don't think that any user object will ever use this. But I'm not completely sure either whether this is the right thing to do, having clocks as objects in the network needs a bit care sometimes, you have to do things like `if not isinstance(obj, Clock)` to avoid infinite recursions (the clock of a clock is the clock itself).


>> So clocks now store t and dt twice, once as internal Python attributes, once as standard state variables.
>> 
> Can't they be properties so that the data is only stored once?

They are already properties but that's not the issue. The problem is that state variables are not accessible from Python in standalone. For example, the `SpikeGeneratorGroup` does checks based on the `dt` of the clock, but if you try to access `dt` as a state variable in standalone, it will try to read it from disk and that obviously does not work.

> Could we just change the order of imports? I guess probably not since everything imports clock early.

It's not just the order of imports. After you did a `from brian2 import *`, the `defaultclock` does of course already exist. But state variables are created on the device that was active, i.e. on the runtime device. If you now switch to the standalone device, you will not have the state variables for the clock available in the standalone device.

> Another option would be to make dt a preference and just get rid of defaultclock entirely. Now you'd write prefs.dt = ... at the beginning of your script, and each Network could have its own clock that would get created with that dt

This would be an option but I'm a bit hesitant to do that. First of all, this would change the syntax and potentially make it more cumbersome (I think to be consistent, it would have to be something like `prefs.core.dt = ...`). But more importantly, in the straightforward implementation you'd have to set the preference before you create anything whereas now setting `defaultclock.dt` will change the time step for all objects using the default clock. Also, currently, you can do something like
```Python
run(...)
defaultclock.dt = ...
run(...)
```
If you do this with a preference, it's less clear whether this has any effect or not. An option (I think Victor suggested that a while ago), would be to always use `net.defaultclock.dt = ...` or `magic_network.defaultclock.dt = ...`, but I find it hard to argue that this is better for the user...

> We use this idiom a lot now in the templates and it's always quite ugly. I'm wondering if we could make this cleaner somehow? For example, maybe for scalar variables, {{dt}} doesn't give _clock_dt but directly _clock_dt[0]? I guess there are some complications, but perhaps there's a way to introduce hooks into jinja that could make this part of the code generator objects. Another alternative, {{dt|scalar}} that would be just a filter that adds [0] afterwards but it would be at least more semantically satisfying.

I don't think that `{{dt|scalar}}` helps a lot, but actually I don't think it's a big problem to change the jinja variable for `{{dt}}` to `arrayname[0]`. I think that makes sense and is pretty straightforward. In the rare situations where we need the pointer to the array itself, we can always use `{{get_arrayname(...)}}`. I'd be happy with that solution.

thesamovar:
> I don't think that any user object will ever use this

Dangerous words! :)

> For example, the SpikeGeneratorGroup does checks based on the dt of the clock, but if you try to access dt as a state variable in standalone, it will try to read it from disk and that obviously does not work.

Right, but that's a genuine problem. If those checks need to be done, and dt is really just another variable, then those checks out to be done on the device? Or are we thinking of those checks as being more like sanity checks that try to pick out errors (i.e. can be sacrificed)?

For defaultclock I'd be happy to consider a change to the syntax anyway. I was never entirely happy with the name defaultclock and now that you shouldn't really be using the value of t from that clock anyway, it's just a container for dt which seems excessive. I think the following semantics would not be too hard to explain: if there is no clock specified, it will be allocated a clock with the default dt at runtime. I also don't mind having ``prefs.dt`` rather than putting it in ``core``. It does feel slightly wrong to do this: like we're abusing a preference system to do something at runtime. On the other hand, it could be quite nice for users to have this in the preferences system because they could then use a different default value from our 0.1ms if they wanted to. Hmm, it also strikes me that it might be complicated for standalone. Having a defaultclock for each network doesn't quite work I think because when you create an object (and assign a clock to it) it's not yet attached to a Network.

Perhaps the conclusion should be that clocks are special? I do like the idea of having t and dt stored using the standard mechanisms but some of the complications are pretty ugly and maybe fixing them leaves us worse off overall. I think there might be a solution here that involves changing the semantics of defaultclock but I don't think we're quite there yet. Any more ideas?

mstimberg:
> I don't think that any user object will ever use this
> 
> Dangerous words! :)

I know what you mean :) But in this case, I don't see a problem: `after_end` is a somewhat special slot that you would not use normally, but even if you do, everything is still well-defined. In theory, there might even be a use-case for doing something after the clocks update but still in the same time step, some kind of very specific network operation. With the approach taken here, you could do that. 

> Right, but that's a genuine problem. If those checks need to be done, and dt is really just another variable, then those checks out to be done on the device? Or are we thinking of those checks as being more like sanity checks that try to pick out errors (i.e. can be sacrificed)?

In principle we could do those checks on a device, but that would first of all some kind of mechanism that I proposed in #390 (I actually started working on it, and it turned out to be more complicated than I though, not sure it is worth it...) and then we'd also have to raise the error in some meaningful way. Just doing it in Python seems to be a convenient solution and with the system in place where `t` and `dt` *are* standard state variables but at the same time they are also stored in the `Clock` object directly, this is possible. 

About `defaultclock`. I think there is a need for an explicit object, even though it is basically a container for `dt`. The reason is that all the objects that don't have a clock specified will have a reference to it, so changing `defaultclock.dt` will change the `dt` of all of those objects. In principle having a `network.dt` would be a nice alternative, but we do assume in many places that objects have a reference to their clock as soon as they are created and as you mentioned, at this time we don't know yet about the network.

> Perhaps the conclusion should be that clocks are special? I do like the idea of having t and dt stored using the standard mechanisms but some of the complications are pretty ugly and maybe fixing them leaves us worse off overall.

Well, either way they will remain special :) But please have a look at the code, all the issues are already fixed there. The code for the `Clock` class itself certainly needs cleaning up and these issues need to be documented but all in all it does not need more code than before (well, if you take into account that it allows to get rid of `AttributeVariable`).

mstimberg:
Oh, one more idea: maybe if the use of `_dt` and `_t` as internal "Python copies" of the state variable value is too ugly, we could build this in as a standard mechanism? E.g. we could have scalar arrays by default stored in memory as well for standalone, and only mark them as "not accessible" if they are set using a string? So if you have some parameter in an equation
```
group = NeuronGroup(..., '''
...
parameter : 1 (shared)
...''')
```
and later do:
```
group.parameter = 5
```
then accessing `group.parameter` later would not lead to an error, even in standalone mode.

thesamovar:
I'm almost convinced.

The main thing I'm worried about is safety, i.e. that you might end up with two different values (one in the array and one in the Python copy). If we have a general mechanism that we use for all variables and an explicit way to make it safe (i.e. your suggestion that you only keep the Python copy if you never set the value of the array in any generated code) then I'm comfortable with that.

So what we'd have then is some flag that says that the value of a scalar (or indeed, a vector - why not?) is known at Python-time. Any time you call a standalone ``CodeObject`` you set this flag to ``False`` for any variables that are written to and a subsequent attempt to read the Python copy of the data will raise an error. That would feel safe to me, and if we were using it for all scalars then it would get well tested.

The only nasty thing about this is that it introduces situations which the user couldn't be expected to know about in advance where standalone won't work but runtime will. That's already the case though, so maybe it's OK.

One thing I would like though is to allow for the possibility (in the future) that we could change dt before a run (e.g. calling standalone without a recompilation). It's not essential because we can make further changes later on to do that, but would this be possible? It would cause a problem for ``SpikeGeneratorGroup`` I guess because it precomputes the divide by dt?

mstimberg:
Ok, I'll look into turning this into a general mechanism and I do agree that there's no reason to not do it for vectors as well. Or maybe we don't want the extra memory cost? The situation where I could imaging this being meaningful is for pre-calculated connection arrays or weights.

> One thing I would like though is to allow for the possibility (in the future) that we could change dt before a run (e.g. calling standalone without a recompilation). It's not essential because we can make further changes later on to do that, but would this be possible? It would cause a problem for SpikeGeneratorGroup I guess because it precomputes the divide by dt?

For `SpikeGeneratorGroup` this is fine, it does not currently pre-compute the divide. Even if it did, this would still be done in pure Python code, only the static array stored on disk with the times would change. The only thing that would trigger a re-compilation at the moment is `PoissonInputGroup` because it pre-calculates the probability. Oh and of course constants and scalar arrays like dt are currently not read from disk but instead included in the source code, which I think is much nicer for the general use case but should be made optional as soon as we allow for re-running standalone without re-compilation.

thesamovar:
Great! For ``PoissonInputGroup`` we could easily not precompute the probability, it's unlikely to be a major component of the computational cost. But anyway, we can look at that later.

I'll wait until you've made more changes before reviewing the code.

mstimberg:
Um, I just realized that flagging variables that have been changed is not as easy as I thought... While we can analyse the abstract code statements and note down all the variables that are written, code in the templates itself can of course also change state variables. I don't see an easy way of figuring out such writes (I don't think we want to analyse the generated code), the most consistent way would be to add an `WRITES_VARIABLES` comment similar to the `USES_VARIABLES` comment, but maybe this becomes a bit annoying... Maybe it is worth it, though, it might be useful for other things as well to know what variables are changed by a template. In the long run (not something for the near future), this could even be useful to execute code objects in parallel if they don't operate on the same variables? Anyway, the question is, to we want to do such a change now? Or should we simply flag all variables as "dirty" after a run? After the build is done, we can always read them from disk, so the only disadvantage is that a script with multiple run statements cannot access state variables between the runs, even if they did not change. Actually, I think that's good enough for now, we will not mark "read-only" variables as dirty and these are the ones that are the most interesting to access for users (e.g. `len(group)` which uses `group.N`, or `defaultclock.dt`, or all other variables that the user is not allowed to write to). 

> For PoissonInputGroup we could easily not precompute the probability, it's unlikely to be a major component of the computational cost.

In principle yes, we'd have to pass in this information into the namespace of the function. Let's worry about this later, though.

thesamovar:
Flagging everything as dirty after a run and leaving ourselves the option of being more fine grained in the future seems like a reasonable compromise. Note that we don't have to flag as dirty for a runtime mode run, only for standalone mode.

mstimberg:
Ok, this should be it (apart from additional tests and documentation). The way it works is that the standalone device has an `array_cache` where it stores values that have been set explicitly and those values get invalidated when they are used in a code object. We assume that read-only constant variables are not changed in code objects. This is normally the case, user code cannot write to them in the first place. However, the synapse creation templates do write to read-only variables, so I added a `WRITES_READ_ONLY` comment to the templates. This prevents that `len(S)` or `S.N` returns 0 after synapses has been created (because `N` is constant and read-only without this mechanism it would be considered as not changing). I tried to rather err on the side of caution, but this needs more testing. Apart from that, I don't have anything fundamental to add to this branch if you are now fine with the general approach.

mstimberg:
Ok, I merged in the master branch and for convenience, `{{varname}}` in a template now translates into `arrayname[0]` directly (with the minor inconvenience that you need to use `{{get_array_name(var)}}` if you want the array name without the `[0]` at the end, but there was only a single place in the current templates where I had to make this change).
Still needing more tests and documentation but the main code should be final from my side.

mstimberg:
Um, there's an error in standalone on appveyor, where it can't find the stored value for the synaptic variable `N` on disk. If you can reproduce this on your machine, I'd be grateful if you could debug it...

thesamovar:
It works fine on my computer. The only thing I could think is that for some bizarre reason the hash function might not be returning the same value. I changed the hash function to use an MD5 hash rather than the built-in Python hash, but I'm not confident it will work.

thesamovar:
OK it didn't solve the problem on Windows and it introduced new problems on linux, so we could perhaps roll that change back.

thesamovar:
I'm not sure there's much more I can do on my computer, and I don't have any other ideas to test. Could there be something odd about the filesystem on AppVeyor?

mstimberg:
Yea, that did not seem to be it. The new problem is not really a Linux-one, it's a Python 3 issue, we are just not testing Python 3 with standalone on appveyor (maybe we should...). 
It's a long shot, but jinja2 has been updated not too long ago and appveyor is using the latest version. Could you check whether this makes any difference? I will give it a try in my virtual machine as well.

I could well imagine  that something is odd about appveyor's file system -- it is really weird that it consistently affects the same test and the same variable, though...

thesamovar:
Still works fine with latest jinja2.

mstimberg:
Fortunately I could reproduce it in my virtual machine -- it turned out to be a trivial bug, the test before did not use the correct teardown method to reinitialize the standalone device. No idea why the effect was specific to Windows, though...

brian2/core/base.py
```diff
@@ -100,9 +102,9 @@ def __init__(self, dt=None, clock=None, when='start', order=0, name='brianobject
         #: The scope key is used to determine which objects are collected by magic
         self._scope_key = self._scope_current_key
         
-        logger.debug("Created BrianObject with name {self.name}, "
-                     "clock={self._clock}, "
-                     "when={self.when}, order={self.order}".format(self=self))
+        # logger.debug("Created BrianObject with name {self.name}, "
```
thesamovar: Should this have just been temporarily commented out?

brian2/devices/cpp_standalone/templates/network.cpp
```diff
@@ -1,3 +1,4 @@
+{# IS_OPENMP_COMPATIBLE #}
```
thesamovar: I thought we didn't need this any more? Mistake?

brian2/devices/device.py
```diff
@@ -315,7 +336,9 @@ def reinit(self):
         Reinitialize the device. For standalone devices, clears all the internal
         state of the device.
         '''
-        pass
+        # Dummy call to set_device with the currently set device -- will trigger
+        # a new activate, important if we are reinitializing the current device
+        set_device(get_device())
```
thesamovar: There is something Zen about this line of code.

brian2/core/base.py
mstimberg: Ah, yes and no. The problem is that currently the `Clock` is a `BrianObject` itself. So during the creation of the `Clock`, the `BrianObject` initialiser wants to print this log message which mentions the clock of the object that has been constructed, but in the case of a `Clock` the `clock` attribute refers to itself, and the string representation of that will fail, because we are not finished constructing it yet...
It's a bit cumbersome to fix, but it's not difficult. That said, are we now sure that we want to keep the "advance the clock with a codeobject" approach? Because if not, then the `Clock` will no longer be a `BrianObject` and this problem disappears...

brian2/devices/cpp_standalone/templates/network.cpp
mstimberg: Indeed, this (and a few more of the same) must have slipped in during the merge.

brian2/devices/device.py
mstimberg: :-) Reading my own comment, it seems that `get_device().activate()` would make more sense here, actually.

thesamovar:
Great! We're almost there. The tests are now passing and I've had a look at the code and I think it's good except for one small and one slightly larger point. The small point is that maybe we should rename ``WRITES_READ_ONLY`` to ``WRITES_TO_READ_ONLY_VARS`` or something like that because it's a bit confusing at the moment.

The bigger point is about ticking the clock in abstract code rather than in ``Network.run``. This makes me feel a bit uncomfortable because it makes the functioning of ``Network.run`` rather difficult to understand without knowing that somewhere the clock is being updated. It also doesn't really save us much work writing a new device not to have to write code to update the clock (since we need to have Clock objects anyway, and scheduling code). So perhaps it would be better if ``Clock`` was a ``Group`` but not a ``CodeRunner``? This would also eliminate the problem of where to put it in the update schedule.

I'm not totally against having it be updated in abstract code if you really think it's better though. In this case, I think we should have a comment where the old clock tick code was to explain what is happening elsewhere.

brian2/devices/device.py
thesamovar: Nono! Keep it! :)

brian2/core/base.py
thesamovar: Got it. Let's see about that (latest comment is about this).

mstimberg:
> The small point is that maybe we should rename WRITES_READ_ONLY to WRITES_TO_READ_ONLY_VARS or something like that because it's a bit confusing at the moment.

Sure, there's also a few other details and docs that need to be updated. 

About the clock ticking: I'm not 100% sure whether this is a good idea myself... On the one hand, there's a certain elegance about it because all state variable changes are dealt with in abstract code, you could basically write out a long list of all the code blocks and you'd get all the statements that are executed during a timestep and exactly at which point. On the other hand, it introduces a few complications in the code, in particular because a clock uses itself as its clock. I think from the total amount of code, it would therefore be even less code with the special `Clock.tick` method. We even introduce some inconsistency because we are writing to read-only variables (`timestep` and `t`) in abstract code, which should not be allowed (we do not check this at the moment for some reason...).

So with all that and given that you rather seem to be in favour of keeping the old `tick` approach, I'm happy to change it back. In the long run, the clock might also be something that devices want to have fine control over, so having this not part of the standard codegen system would be an advantage for that as well.

I'll try to make all these changes soon and also make sure to update the docs where necessary (but I don't think we wrote much about this in the developer docs).

mstimberg:
I committed all the changes, so this is now finished from my side. There was one complication to make `Clock` no longer a `BrianObject`: because all the other class that derive from `Group` where also deriving from `BrianObject` , we decided to make `Group` derive from `BrianObject`. Now for `Clock` this is no longer true, so I separated `Group` into two classes: `BrianContainer` for the functionality regarding the access of state variables and `Group`, which derives from `BrianContainer` and adds the code for executing code in its context (i.e. `run_regularly` etc.). I think this is fine, I'm just not happy with the `BrianContainer` name. Any better idea? Maybe use `Group` for that class and something like `RunnableGroup` for what is now `Group`?

mstimberg:
I added two tests for the `array_cache` mechanism (and fixed a bug that the test revealed...), so appart from the naming of `BrianContainer` and `Group`, this is now really finished from my side.

thesamovar:
Looks like there are still a few problems to fix.

For the name, I thought about it and didn't come up with anything good. It's not something that any user will ever see I guess, so maybe it's fine? If not, some suggestions (none of which I'm happy with): ``VariableStore``, ``VariableContainer``, ``BrianData``, ``VariableRootObject`` (this is quite technically satisfying in some way), ``VariableRoot``, ...

mstimberg:
Yes, I'm fixing the error, it was unfortunately something that involved uninitialised memory and was not always caught by the test suite... About the names, your suggestion gave me an idea: `Variable.owner` refers to the object storing the `Variable`, so how about `VariableOwner`?

thesamovar:
Perfect!

mstimberg:
Hopefully this should do it. I tried to clean up a bit, there was a lot of redundant code for variable-setting in `CPPStandaloneDevice`. As a side-effect, the generated code is now more compact. E.g. for `defaultclock.dt = 0.1*ms` it now directly includes `_array_defaultclock_dt[0] = 0.0001;` in the main function, instead of reading an array of length 1 from disk and assigning it to the array in a loop... This leads to much more readable code that looks more similar to what you write in the Python script but it is somewhat in conflict with our goal to re-run code without re-compilation (#251). But I think for that we need a more principled approach anyway, and it will probably be optional. Either way, already before we did a mix of reading values from disk and including them directly in the code, so it does not change anything fundamentally.

thesamovar:
Good work. I'm happy with this and it can be merged. Will let you do that in case there are any last minute changes you wanted to make.

mstimberg:
I updated a few string/documentation references to `Group`, merging it now.

Issue #539, opened 2015-08-12
Improve OpenMP performance


mstimberg:
Whenever a synaptic code does not only update synaptic variables, we chose a conservative approach and run the loop sequentially. Similarly, all threshold code is run in sequence (because it updates the global spike queue). This is clearly not optimal, but improvements are non-trivial (and correctness is always more important than performance) and it is not clear what is the best approach (e.g. use atomic updates of shared variables, completely different data structures, different looping strategy, etc.). This is something we'd like to improve but it is not a high priority for the core developers, contributions are therefore welcome. Short of actual improvements to the generic Brian code, reports about manual alterations of the generated code that led to performance improvements are also welcome.
This will also very likely benefit from the work on the GPU backends, possibly some of their approaches can be "backported".

PR #553, opened 2015-08-31
[MRG] Fix scalar writes with OpenMP

mstimberg:
See discussion in #551, this puts the `parallel` pragma in individual templates instead of around the main run loop. This also means that by default everything will be executed serially, therefore there's no longer a need for the `IS_OPENMP_COMPATIBLE` flag in the templates (and much less `single` pragmas are necessary). I also updated the developer docs, mostly deleting things that are no longer applicable. I remember when @yger started implementing OpenMP support, we decided against doing it like this because it seemed to slow things down. However, I can't see any negative effect, possibly other things were responsible for the slow down.

Some quick-and-dirty benchmarks (median of three runs, everything normalized to a run without OpenMP (0.5 means 2x as fast), taking the output from the progress report, i.e. only runtime without synapse creation, compilation, etc.).

 | Before | after
-----------|----------|-------
IF_curve_Hodgkin_Huxley||
1 thread | 1.031 | 0.988
2 threads | 0.543 | 0.531
4 threads | 0.407 | 0.403
CUBA ||
1 thread | 1.000 | 1.000
2 threads | 0.607 | 0.629
4 threads | 0.652 | 0.652
STDP_standalone ||
1 thread | 1.524 | 1.439
2 threads | 1.788 | 1.814
4 threads | 2.140 | 2.137

So overall, OpenMP scaling is not great and is abyssimal for the STDP example where everything gets much slower instead of faster, but I don't see any significant differences before and after the changes from this pull request. Improving all of this is #539 

In the long run, we have even more flexibility with the new approach because you can use additional arguments to the `parallel` block to make variables thread-local, use reductions, etc.


thesamovar:
Indeed I don't see any real speed difference so let's go with this. Is OpenMP tested comprehensively in our test suite? If so, I'm happy for this to be merged once they pass, otherwise I'll review the code a bit more carefully.

mstimberg:
> Is OpenMP tested comprehensively in our test suite?

I would not say that... There's a single test `test_openmp_consistency` that uses a deterministic STDP example and looks for exact equivalence. And I added one more test for the specific scalar write issue. I'd say that our warning that OpenMP is not well tested still applies :-/

thesamovar:
Oh yeah I forgot the warning. OK, so that's fine then. I think this can be merged when tests pass.

mstimberg:
I did not quite like that OpenMP is not tested better so I added a new script `run_nose_tests_standalone_with_openmp` that runs the standalone test suite, but also runs all the standalone-compatible tests with OpenMP and 4 threads -- this should give us a bit more confidence that things are working correctly. The old `run_nose_tests_standalone` script will no longer run the two specific OpenMP tests. If all this passes, this should be good to merge (if you don't want to have a closer look at the code before).

thesamovar:
Great! Merging now.

PR #556, opened 2015-09-04
[MRG] Catch `OSError` instead of `WindowsError` when closing `std_silent` 

### LINKED ISSUE #555: Exception at the end of `std_silent`
Reported by @owenmackwood here:
https://groups.google.com/d/msg/brian-development/sc_Y0SuinvQ/gpggA4GWBAAJ

mstimberg:
Trivial fix, but without it any kind of exception that is raised in `std_silent.close` will lead to a `NameError`, because `WindowsError` does not exist on Linux/MacOS. We want to ignore those exceptions anyway, so just replacing it with `OSError` should be fine (`WindowsError` derives from `OSError`).

Fixes #555

### DIFF
--- a/brian2/utils/logger.py
+++ b/brian2/utils/logger.py
@@ -242,15 +242,15 @@ def clean_up_logging():
     '''
     logging.shutdown()
     if not BrianLogger.exception_occured and prefs['logging.delete_log_on_exit']:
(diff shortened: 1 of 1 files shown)


thesamovar:
Happy to merge when AppVeyor finishes.

PR #562, opened 2015-09-21
Spatialneuron efficiency/parallelism improvements for cpp_standalone

moritzaugustin:
spatialstateupdater: more efficient (linear instead of cubic) solution of the coupling matrix system (Hines1984) and generally more parallel version of the cpp standalone code

- new and required: list of children indices per branch
- separate arrays c1,c2,c3 instead of only c for temporary values in the 3x tridiagonal sytem solving loops (since the c arrays are written to concurrently)
- rewritten for exploiting more (and nested) parallelism in the openmp version (which is also more straight forward for cuda implementation and we should make them as similar as possible for maintenance reasons)
- the nested new openmp "pragma" are hopefully used correctly?! however, the openmp code does not run without errors (but this is just true for the master branch without my changes => c.f. rall.py with standalone_cpp and prefs.devices.cpp_standalone.openmp_threads = 4 gives runtime error segfault)

still missing: cython & weave & cuda_standalone implementations (I will take care of the latter, Marcel maybe after revising my improvements merge them into cython&weave)
still broken: openmp version of the spatialupdater (already present in old code, Marcel?)


moritzaugustin:
the O(n) linear system algorithm based on Hines84 and Koch98 (with reversed enumeration of the branches & excluding the tridiagonal parts which we have in the domain decomposition step1) and is based on four assumptions:
- technically: (I) _morph_i[k] = k+1, (II) _parent_i,  _children_i and _children_num_i are appropriately computed 
- structurally: (III) there is no cycle in the branch connectivity graph, (IV) the branches are always numbered (in _morph_i) such that parents have smaller indices than their children

remark: if assumptions (III) and (IV) are always fullfilled we should remove the part which solves the coupling system in cubic operations

further useful improvements which are not yet implemented include:
- use for the matrix P a sparse matrix format (e.g. CSR?) instead of a dense one (good for caching on cpu/coalescing on gpu/allowing to use shared memory on gpu) 
- optional: omit the upper diagonal elements in the matrix as they appear only in the elemination step (i.e. start with the eliminated lower triangular matrix) -- but this increases the complexity of the algori
- _children_i could consume less memory of num_branches x max_children_over_all_branches (rows x cols) but is currently num_branches**2

btw: why did travis CI build 2702.5 fail due to a missing shared (fortran) library?

moritzaugustin:
I have just recognized that the current pull-request is not functioning -- there are some issues. I have to the bugs first.  I am currently also trying to implement the sparse matrix version specific to the branch tree graph structure. please wait with the revision until the next commit is pushed (should come this week)

thesamovar:
Just a quick note to say that there is currently a bug with Travis and that's why it's failing. It's a known issue and will presumably be fixed soon.

moritzaugustin:
Before the last commit there were some issues and the new solver was broken as I did not test thoroughly enough. But now I have fixed the new version of the efficient solver part.

Additionally I introduced a sparse matrix version of the solver (instead of ```_P```, which is not used anymore, ```_P_diag```, ```_P_children``` and ```_P_parent``` contain all nonzero diagonal elements). For discussion/explanation purposes three solvers and debug outputs are still in the spatialstateupdate.cpp template. Of course the efficient solvers assume a non-cyclic branching connectivity graph. Can we assume that forever? Otherwise we should keep the dense and inefficient matrix version as fallback...

The ```examples/spatialneuron/*``` scripts[*] as well as all functions within ```tests/test_spatialneuron.py``` (in cpp_standalone mode) show the expected figures (if any) and complete without error. The efficiency of the new coupling system solver is demonstrated in the ```test_rallpack2()``` case (other examples do not show significant speed differences due to trivial branching structure): 
new (sparse and efficient) version: 0.25005 s (100%) simulated in 0.417276 s
old (dense and inefficient) version: 0.25005 s (100%) simulated in 27.454 s

[*] Note that I had to make some spatialneurons examples standalone compatible (not commited though) by e.g. hardcoding electronic length constants ```la``` or dendrite indexing (because of ```NotImplementedError: Cannot retrieve the values of state variables in standalone code before the simulation has been run.```). By the way, one example, ```bipolar_with_inputs.py```, ran in non-standalone mode only when enforcing ```'numpy'``` codegen instead of the default ```'weave'``` which gave a compile error in this particular example.

mstimberg:
Thanks for working on this! I did not look into your changes yet, but here are some general comments/answers:

* You can assume no cycles in the graph, the morphology has to be a tree. Actually, we don't explicitly check for this, we should. If you have cycles in the graph, the stage where we transform the morphology into flat arrays will run into an infinite recursion, so you'd never even get to the simulation
* Your other assumptions are also guaranteed to be true: `_morph_i[k] = k+1` -- I actually removed `_morph_i` for this reason in the (still work-in-progress) `morphology_refactoring` branch: 8c929238 and child compartments are guaranteed to have higher indices than their parents
* To get the travis build run again, merge with current master or cherry-pick this commit: d5be0cd6

About the problems you are experiencing in master:
* I can run `bipolar_with_inputs.py` without problems on weave on master, are you sure this is not something introduced in this branch?
* I also don't see a segfault for OpenMP with the `rall.py` example...

mstimberg:
@moritzaugustin not sure whether you've seen it, but the OpenMP problem in the spatialstateupdate template should be fixed now (#570). Half of the issue is already fixed in your code (concurrent write to `c` which you replaced by `c1`, `c2`, `c3`), but in addition you'll have to move `double ai,bi,_m;` into the `parallel` block with the sections, each section writes to these variables so they have to be thread-local.

moritzaugustin:
another aspect I would like to add is an explicit time discretization (which actually is independent of the coupling system). the explicit euler discretization of the time derivative in the cable equation has disadvantages (the timestep needs to be be small enough (\Delta V dependent!) because otherwise the numerical solution is unstable) but also advantages (much better in terms of parallelization since the tridiagonal matrix has just to be multiplied  with a vector [=O(n) independent operations] instead of solving the respective tridiag system [=O(n) sequential operations])

using the explicit euler in each timestep the (potential) violation of the numerical stability has to be checked and a warning or an error indicated

see, e.g., http://www.isn.ucsd.edu/courses/bggn260/2008/cablepde.pdf

mstimberg:
I still did not give this thorough review it deserves, but just two quick comments:
* Let's discuss the explicit Euler approach separately and not include it in this pull request -- I am not sure it is a good idea but from the GPU point of view, things might be different
* The nested parallelism seems to slow things down dramatically, e.g. in the `hodgkin_huxley_1952` example. But more importantly, I am surprised that it works at all! There are sequential dependencies in the loop, how can this work with a `parallel for`? E.g. this line (first loop of "system 2a"):
```C++
{{v_star}}[i]=({{v_star}}[i] - ai*{{v_star}}[i-1])*_m;
```
The result for `{{v_star}}[i]` will depend on whether `{{v_star}}[i-1]` has already been updated or not. Or do I miss something obvious here?

moritzaugustin:
I agree in making the exlicit time discretization separately and not discuss it in this issue (if at all). I will come back to this after I see the GPU performance with the implicit Euler for real world examples. 

 About the inefficiency issue:
I have no experience with OpenMP but if the nested parallelism is inefficient then we should either serialize the code block for `v_star`, `u_plus` and `u_minus` but parallelize within each of them the actual calculation of the independent tridiagonal systems (i.e., the for loop with index `_i`). Here the parallelism scales with the number of branches. Alternatively one could solve the systems for `v_star`, `u_plus` and `u_minus` in parallel but calculate the independent tridiagonal systems in a sequential manner. In the latter variant the parallelism is 3 and does not scale. 

Considering your question regarind correctness/possibility of a race condition:
What is declared parallel is the for loop ofer branch indices `_i` (and the tridiag system of two branches do not interfer). The sequential dependence of `v_star[i]` vs. `v_star[i-1]` is respected since the `i` loop (over compartments in fixed branch `_i`) is not declared parallel. However, I see that the usage of `_i` and `i` is very confusing and I will fix that.

mstimberg:
Thanks, I did indeed mix up `_i` and `i` here, so this does work correctly as it is. I agree that better index names would be good here (BTW: it's best to only use variable names starting with an underscore so that they can't clash with variable names in user code).

I merged the latest master into this branch (which switched on compiler optimizations by default among other things) and did very crude benchmarking (total run time (without analysis code) including preparation and compilation time). As a reminder: Rallpack 1 is a passive cable (many compartments in a single branch), Rallpack 2 is a passive tree (many branches with single compartments each) and Rallpack 3 is a cable with active conductances (again, many compartments in a single branch). Having one long cable vs. a tree of single compartments should be the two extremes and all real-world examples should lie somewhere between the two. Here are the benchmark results for 0-8 threads (master=current master branch, sections=calculate the three systems in parallel, inner=calculate the three systems sequentially but parallelize the loop over branches, both=do both with nested parallelism):

benchmark_results_openmp

Two conclusions: 1) the sparse efficient solver for the tree (Rallpack 2) works fine and much better than the naive code in master (as you already noted) and 2) using the nested parallelism makes everything scale terribly for >=4 threads).

From these results, I'd suggest to go for the "sections" approach -- the advantage that there is always something to parallelize seems to outweigh the disadvantage that we don't necessarily use all the cores.

moritzaugustin:
I will change the similar named variables `_i` and `i` soon when I finalize the code and remove the dense version of the solver etc.: https://github.com/moritzaugustin/brian2/issues/13

Thanks for merging and the nice runtime analysis of the different OpenMP configurations! 

About the conclusion: I think that the sections parallelism is faster compared to the inner one only due to the fact that Rallpack 1/3 do not offer inner parallelism (only one branch) and Rallpack 2 despite of parallelism (many branches) has almost no work per branch (only one compartment which is the size of the linear system). However, in a situation with a significant number of branches (e.g. more than three) AND a significant number of compartments within each branch I expect the inner parallelism version to win (maybe you can add such an example as third case in your sweet comparison script? and if so maybe add explicitly the number of threads = 3 since this should be the boundary from which the sections approach should lose vs. inner in the new condition). 
If the number of branches is available at code generation/compile time the optimal parallelism (either section or inner) could be set up.

mstimberg:
Right, here's another run of Rallpack 2 (i.e. the complex tree), but this time with 100, respectively 500 compartments per branch instead of just a single one:
spatialneuron_benchmark_results_2

So, we can still conclude that the nested loop approach does not work (I am not sure why, but well...). Here now, as you expected, the inner loop performs better for > 3 threads. The effect should be much less in most real-life examples, though, where the run time is often dominated by the integration of the per-compartment equations (as in Rallpack 3). I am kind of leaning towards sticking with the "sections" approach. We could analyse the morphology at compile time and switch the strategy accordingly, but I am not sure what a good heuristic would be. @thesamovar: Any opinion on this?

thesamovar:
Haven't really been following or looked at the code, so minimal comments from me. One thing is that it's really hard to get these things right automatically. Maybe better to make it an option for the user? It does seem like in small examples the sections approach is winning but in big examples the inner approach is winning: if so, I'd rather suggest going with inner rather than sections because the speed difference is more important in bigger examples, but I may have misread the discussion on this.

One of the things I've found that parallel/OpenMP performance depends on a lot is the architecture of the chip. So, for example if you have two cores sharing a cache that behaves very differently to two cores each with their own cache. Similarly, how they are wired into the bus makes a big difference (i.e. are they sharing bandwidth or not). Generally speaking, thinking about caches and memory bandwidth seems to be at least as important as thinking about the arithmetic aspects of the parallelisation. (Not as much as for GPUs, but even for multiple processors this seems to be more and more important.)

moritzaugustin:
I do not know what are the typical use cases but I would rather vote for model-dependent heuristic toggle. For example: if the number of branches is larger than 3 AND the number of compartments averaged over branches (i.e. the quotient of total number of compartments and number of branches) is larger than $avg_no_compartments use the inner approach. Otherwise take the sections way of parallizing. Maybe the easiest implementation would be to use the quotient within a `if(condition)` OpenMP expression, c.f. the solution of http://stackoverflow.com/questions/10589798/openmp-create-threads-only-once
The heuristic value of the user-defineable preference could be $avg_no_compartments=10

If not using a heuristic: sticking with the sections number it would be beneficial if the number of threads was upper limited by 3 because the thread management seems to be a real performance killer as seen from the "both" as well as the "sections" scaling behaviour for large number of threads...

mstimberg:
> It does seem like in small examples the sections approach is winning but in big examples the inner approach is winning: if so, I'd rather suggest going with inner rather than sections because the speed difference is more important in bigger examples, but I may have misread the discussion on this.

The situation is slightly more complicated: the "sections" approach can always parallelize a certain part of the code, not depending on the morphology but since it just executes three sections in parallel, it does not benefit from more threads. The "inner" approach will only be useful if there are many branches, i.e. it cannot do anything for a long cable, for example. Another option would be to actually change the internal morphology just for simulation purposes, i.e. instead of simulating one cylinder with 10000 compartments, simulate 100 cylinders with 100 compartments each. But coming up with a good heuristic for that might be even more complicated... Either way, as Dan said, it should be possible for the user to overwrite the heuristic. Maybe simply with a preference?

mstimberg:
I'd be fine with such a heuristic but let's try to keep it as simple as possible.

> If not using a heuristic: sticking with the sections number it would be beneficial if the number of threads was upper limited by 3 because the thread management seems to be a real performance killer as seen from the "both" as well as the "sections" scaling behaviour for large number of threads...

It depends, e.g. Rallpack 3 benefits so much from the multi-threading for the standard state update so that it still performs better with > 3 threads, even with the "sections" approach.

mstimberg:
Ok, I had another look at this -- would be great if we could finish this up fairly soon so it makes its way into the 2.0 release! I again merged latest master into the branch so that we don't diverge too far from it. I also made a copy of the `spatialstateupdate.cpp` template that you wrote into `dev/ideas`, so that we keep the full version with the dense implementation and debug output for future reference. For now, could you please strip down the template to just the "efficient, sparse" version and remove the debug output/superfluous comments, etc.? After this is done I will take care of merging the changes into cython and weave(should be straightforward) and numpy (might be a bit tricky).

About the parallelization strategy for the tridiagonal systems:
* I added a preference `devices.standalone.openmp_spatialneuron_strategy` that can be used to switch between `'branches'` (what we referred to as "inner" before) and `'systems'` (what we called "sections").
* If you are fine with this, I'd suggest we use the following heuristic if that preference is not set: Use `'systems'` for <= 3 threads, otherwise use `'branches'`. This seems to me quite straightforward, does not rely on any arbitrary measures of branching complexity and should give a good result in almost all cases, judging from the previous benchmarks.

To implement the different strategies in the template, I made the `prefs` object accessible in the template. I'd suggest to do the following in the beginning of the template:
```Django
{% set strategy = prefs.devices.cpp_standalone.openmp_spatialneuron_strategy %}
{% if strategy == None %}
    {% if prefs.devices.cpp_standalone.openmp_threads <= 3 %}
        {% set strategy="systems" %}
    {% else %}
        {% set strategy="branches" %}
    {% endif %}
{% endif %}
```

In the template, you can then use, e.g.
```Django
{% if strategy == 'systems' %}
{{ openmp_pragma('sections') }}
{
{% endif %}
```
to include the correct OpenMP pragmas depending on the strategy.

moritzaugustin:
Cleanup (removed dense parts, changed variable names, removed/optimized comments, removed debug parts) of the source code done except for getting rid of variable `_P` of `spatialneuron.py` (which should be done all target languages support exclusively the sparse efficient solution).

Missing: OpenMP strategy handling (see below).

I could not use the preference snippet you provided as even just the one defining `strategy` raised a TypeError during run() irrespective of whether a strategy was set or not.
```
File "/home/augustin/Projects/Brian2/github/moritzaugustin_brian2/brian2/codegen/templates.py", line 211, in __init__
s = autoindent_postfilter(str(f()))
TypeError: 'str' object is not callable
```

By the way and helpful for me finishing this would be the answer to: how can I run all the tests defined in `test_spatialneuron.py` using the `cpp_standalone` device directly after each other? Currently I can only set the device and run one test. Otherwise I get the following `RuntimeError`  suggesting that I have to build the device manually -- i.e. I have to change all the tests itself and complement the `run()` with a `device.build()`?
```
RuntimeError: The network has already been built and run before. Use set_device with build_on_run=False and an explicit device.build call to use multiple run statements with this device.
```

mstimberg:
Ok, great!

> I could not use the preference snippet you provided as even just the one defining strategy raised a TypeError during run() irrespective of whether a strategy was set or not.

I think this happens if you put he Jinja code outside of a block, it should be part of the "maincode" block.

> By the way and helpful for me finishing this would be the answer to: how can I run all the tests defined in test_spatialneuron.py using the cpp_standalone device directly after each other?

Unfortunately there is currently no way to do this (you can run `dev/tools/run_nose_tests_standalone_with_openmp` but this will of course not only run the `SpatialNeuron` test). Preparing each test module in a way that you can run all the tests that apply (i.e. here the "standalone-compatible" tests), with the correct settings for the device etc. would be quite a bit of work. Maybe we'll add a general filtering option to our test framework at some point.
For now, you could comment out all then "non-standalone-compatible" tests, set the device in the beginning and call `reinit_devices()` after each of the tests.

moritzaugustin:
Thanks for the suggestion to use `reinit_devices()`which works fine. 

The code is now final from my side unless you have particular concerns. The only thing which I would like to have changed is the default strategy: Maybe you could add it towards someething like (if possible): `if prefs.devices.cpp_standalone.openmp_threads <= 3 or total_no_of_branches_of_morph < 3 then 'systems'` to account for models where only one or two compartments exist and the `'system'` strategy might be the best (e.g. ball-and-stick model).

Do not hesitate to contact me in case for the numpy version some notes or inside knowledge is necessary.

mstimberg:
Thanks, I'll have a look at it and try to adapt the other codegen targets.

> The only thing which I would like to have changed is the default strategy: Maybe you could add it towards someething like (if possible): if prefs.devices.cpp_standalone.openmp_threads <= 3 or total_no_of_branches_of_morph < 3 then 'systems' to account for models where only one or two compartments exist and the 'system' strategy might be the best (e.g. ball-and-stick model).

Hmm, I am a bit hesitant to to do this as this means that we need to give the `spatialstateupdate` template an information that is only used in for OpenMP and ignored in all other cases... On the other hand, it might be useful for other optimisations n the future (e.g. for a simple cable, you could skip the complete stuff concerned with the branching structure). Ok, I'll add that check.

mstimberg:
Ok, I think that's it, the new algorithm is implemented for all codegen targets now and I deleted the dense version of the `P` matrix. The numpy version is still quite inefficient compared to the other targets (the Rallpack 2 test takes 45s compared to 2.5 s with weave on my machine...) but it is more than twice as fast as it was in the master branch, so that shouldn't stop us from merging this branch (I am also not too motivated to optimize this further...). I think this is now ready to merge. @thesamovar: I know this is the part of Brian 2 that you've been less involved in but could you give it a basic check and merge if you are ok with everything (e.g. run the long test suite and in particular the OpenMP tests -- if I am not mistaken both Moritz and me only tested everything on Linux so far)?

moritzaugustin:
Cool. And yes, I was testing on Linux only when implementing the efficient solver. 

By the way I had made some hand-written notes when I developed the sparse efficient solver and I am not sure whether I provided them to you already. Thus I attach them here (and do not hesitate to ask if you cannot decipher but need it). 
ma_notes_coupling_matrix_solution

mstimberg:
Hah, our first handwritten notes on github :) You've sent them to me privately before, but thanks for sharing them publicly.

thesamovar:
Happy to run this but my git-fu is lacking. How do I switch to a branch in a different remote repo? Can this be done from within PyCharm by any chance?

mstimberg:
I don't think you can add a new remote in PyCharm, but I might be mistaken. Run the following in the shell:
```
git remote add moritz https://github.com/moritzaugustin/brian2
git fetch moritz
```
Then the branches in his repository should turn up like the other branches in PyCharm (you can also manually do `git checkout moritz/spatialneuron_improved` of course).

thesamovar:
It runs fine on my win64/py2.7 machine. I had a brief look at the code, and it looks good as far as I'm concerned. The only thing I noticed was that there were a few range statements in the numpy codegen target, and I think there was no handling of python 2/3 here (i.e. we should use xrange instead of range for python 2). Not sure if it's necessary to fix this, and I'm happy for this to be merged.

mstimberg:
Hmm, yes, we never care about this in the numpy templates, in a quick test this only changed the runtime of the template by a few percent in the worst case so I am not sure this is worth the complication. The fastest solution would not to use `range`/`xrange` at all but instead pre-calculated indices (since they don't change), but again for me simplicity beats performance here (in particular for numpy which is more an emergency fallback).

mstimberg:
Just realized that we did not merge this yet. @thesamovar What do you think about my previous comment, do you think the `xrange/range` issue is important enough to complicate things?

thesamovar:
No I don't think it's essential - happy for this to be merged.

PR #570, opened 2015-10-06
[MRG] Fixes for OpenMP

### LINKED ISSUE #568: SpatialNeuron does not work correctly with OpenMP on current master
As @moritzaugustin noticed in PR #562, at least the `compartmental/rall.py` example is broken with multithreading on master. The regression has been introduced with PR #553 -- incidentally, there's even a `test_rall.py` in our test suite but it is marked as long so it is not run by default :-/

I'll have a look into this.

mstimberg:
This fixes #568, but I also realized two more issues along the road:
* The standalone-compatible tests that were supposed to run with OpenMP were actually not running with it (except for the first one)...
* Fixing the above revealed memory errors in some tests which had a single synapse

I also removed the long attribute from three multicompartmental neuron tests that aren't that time-consuming (they would have caught #568).

### DIFF
--- a/brian2/devices/cpp_standalone/templates/spatialstateupdate.cpp
+++ b/brian2/devices/cpp_standalone/templates/spatialstateupdate.cpp
@@ -11,8 +11,6 @@
 {% extends 'common_group.cpp' %}
 {% block maincode %}
 
-	double ai,bi,_m;
-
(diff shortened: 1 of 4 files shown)


thesamovar:
This looks good to me, happy for it to be merged when tests pass.

PR #571, opened 2015-10-08
[MRG] Fixes for numpy 0.10

mstimberg:
This should fix brian2 for use with the latest numpy release. Note that the test servers are not yet using numpy 0.10 because packages depending on numpy (scipy mostly, I think) are not updated yet on conda. I think this should change in the next days, though.

The test suite showed quite a number of errors with numpy 0.10, luckily most of them were problems in the test suite, brian itself should work mostly fine (except for `linspace`, see below). Most important changes:
* `assert_allclose` removed units in previous numpy versions, so some tests were sloppy about them (also, you cannot compare e.g. `np.arange(3)*mV` to `[0*mV, 1*mV, 2*mV]`, you'll have to use `[0, 1, 2]*mV` or `Quantity([0*mv, 1*mV, 2*mV])` for the latter).
* comparing `x == 'string'` raises a warning if `x` is a numpy array, I changed it into `(isinstance(x, basestring) and x == 'string')`
* Some ufuncs now raise a `TypeError` when used with illegal types instead of returning `NotImplemented`

The only major change that touches the main code is that `linspace(0*mV, 10*mV, 10)` worked "by accident" in previous numpy versions, we did not explicitly provide a unit-aware implementation. I think it can be useful, so I added explicit support for it and `arange` (i.e. you can do `arange(0*mV, 10*mV, 1*mV)` instead of `arange(0, 10, 10)*mV`).

 Good to merge from my side (note that tests on appveyor seem to get stuck or time out recently, we might have to re-launch them).

### DIFF
--- a/brian2/groups/group.py
+++ b/brian2/groups/group.py
@@ -197,7 +197,7 @@ class Indexing(object):
             raise IndexError(('Can only interpret 1-d indices, '
                               'got %d dimensions.') % len(item))
         else:
-            if item is None or item == 'True':
+            if item is None or (isinstance(item, basestring) and item == 'True'):
                 item = slice(None)
             if isinstance(item, slice):
                 if index_var == '0':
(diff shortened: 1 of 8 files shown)


thesamovar:
Looks great to me, happy for it to be merged once tests pass.

PR #572, opened 2015-10-12
[MRG] More fixes for numpy 0.10

mstimberg:
The previous merge left one test failure unfixed, this should now work (tests are now using numpy 0.10 on our test servers as well). I also fixed a warning in the Python spike queue.

There's one more issue, though. After removing the `codegen-independent` label from `test_cuda`, this now fails on Cython. The reason is a division by zero (that also raises a warning on numpy). I think this problem appears only now because the CUDA example can now use the exact integration and boolean variables are taken into account for the loop-invariant constants. The problem arises for the conductances `ge` and `gi` during the refractory period:
```Python
   # ...
    _lio_const_5 = (dt * (taue - taum)) * _int(0.0)
    _lio_const_6 = ((dt * (taue - taum)) * _int(1.0)) - ((dt * (taue - taum)) * _int(0.0))
   # ...
        _ge = (((((- ge) * taue) * taum) * (_lio_const_1 + (_lio_const_2 * _int(not_refractory)))) * (_lio_const_3 + (_lio_const_4 * _int(not_refractory)))) / (_lio_const_5 + (_lio_const_6 * _int(not_refractory)))
```
When `not_refractory` is `False`, then `int(not_refractory)` is 0 and therefore the whole divisor is zero (`_lio_const_5` is always zero). It does not matter in the end, since we don't write back the values during the refractory period, but we'll still get the error/warning. Is there a reason we have the division checks switched on for Cython, actually? The documentation says that it has an "up to 35% speed penalty" and given that there's no check with weave, this should probably simply switched off? This still leaves us with a warning in numpy, though.

### DIFF
--- a/brian2/__init__.py
+++ b/brian2/__init__.py
@@ -5,10 +5,11 @@ Brian 2.0
 # working weave/Cython on Windows with the Python for C++ compiler
 import setuptools as _setuptools
 
-# Hack to fix problem in numpy 1.10
+# Hack to fix a problem with weave on Windows, introduced in numpy 1.10
 try:
     import numpy.distutils.msvccompiler as _msvccompiler
-    _msvccompiler.MSVCCompiler.compiler_cxx = 'cl.exe'
+    if not hasattr(_msvccompiler.MSVCCompiler, 'compiler_cxx'):
+        _msvccompiler.MSVCCompiler.compiler_cxx = 'cl.exe'
 except ImportError:
     pass
 
--- a/brian2/__init__.py
+++ b/brian2/__init__.py
@@ -5,6 +5,13 @@ Brian 2.0
 # working weave/Cython on Windows with the Python for C++ compiler
 import setuptools as _setuptools
 
+# Hack to fix problem in numpy 1.10
+try:
+    import numpy.distutils.msvccompiler as _msvccompiler
+    _msvccompiler.MSVCCompiler.compiler_cxx = 'cl.exe'
+except ImportError:
+    pass
+
 # Check basic dependencies
 import sys
 from distutils.version import LooseVersion
--- a/brian2/codegen/generators/cython_generator.py
+++ b/brian2/codegen/generators/cython_generator.py
@@ -44,6 +44,13 @@ class CythonNodeRenderer(NodeRenderer):
         return {'True': '1',
                 'False': '0'}.get(node.id, node.id)
 
+    def render_BinOp(self, node):
+        if node.op.__class__.__name__=='Mod':
+            return '((({left})%({right}))+({right}))%({right})'.format(left=self.render_node(node.left),
+                                                                       right=self.render_node(node.right))
+        else:
+            return super(CythonNodeRenderer, self).render_BinOp(node)
+
 
 class CythonCodeGenerator(CodeGenerator):
     '''
--- a/brian2/codegen/runtime/cython_rt/extension_manager.py
+++ b/brian2/codegen/runtime/cython_rt/extension_manager.py
@@ -50,6 +50,8 @@ class CythonExtensionManager(object):
                          compiler=None,
                          ):
 
+        self._simplify_paths()
+
         if Cython is None:
             raise ImportError('Cython is not available')
 
@@ -206,8 +208,25 @@ class CythonExtensionManager(object):
         self._code_cache[key] = module
         return module
         #self._import_all(module)
-        
+
+    def _simplify_paths(self):
+        if 'lib' in os.environ:
+            os.environ['lib'] = simplify_path_env_var(os.environ['lib'])
+        if 'include' in os.environ:
+            os.environ['include'] = simplify_path_env_var(os.environ['include'])
+
 cython_extension_manager = CythonExtensionManager()
+
+
+def simplify_path_env_var(path):
+    allpaths = path.split(os.pathsep)
+    knownpaths = set()
+    uniquepaths = []
+    for p in allpaths:
+        if p not in knownpaths:
+            knownpaths.add(p)
+            uniquepaths.append(p)
+    return os.pathsep.join(uniquepaths)
 
 
 if __name__=='__main__':
--- a/brian2/codegen/runtime/cython_rt/templates/common.pyx
+++ b/brian2/codegen/runtime/cython_rt/templates/common.pyx
@@ -1,6 +1,6 @@
 #cython: boundscheck=False
 #cython: wraparound=False
-#cython: cdivision=False
+#cython: cdivision=True
 #cython: infer_types=True
 
 import numpy as _numpy
--- a/brian2/stateupdaters/exact.py
+++ b/brian2/stateupdaters/exact.py
@@ -182,7 +182,9 @@ class LinearStateUpdater(StateUpdateMethod):
         # Not very efficient but guaranteed to give the correct answer:
         # Just try to apply the integration method
         try:
-            self.__call__(equations, variables)
+            # Don't use sympy's simplify because it might take a very long time.
+            # We are not using the result, anyway.
+            self.__call__(equations, variables, simplify=False)
         except (ValueError, NotImplementedError, TypeError) as ex:
             logger.debug('Cannot use linear integration: %s' % ex)
             return False
@@ -190,7 +192,7 @@ class LinearStateUpdater(StateUpdateMethod):
         # It worked
         return True
 
-    def __call__(self, equations, variables=None):
+    def __call__(self, equations, variables=None, simplify=True):
 
         if variables is None:
             variables = {}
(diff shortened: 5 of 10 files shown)


thesamovar:
Yep, no problem with changing Cyton to not check for divide by zero. I guess I left it like that for debugging reasons and never changed it.

mstimberg:
I had to add a change to the handling of `%` in cython, because its behaviour also changes when we switch to `cdivision`. It now all works for me, but all appveyor tests now seem to be failing for cython. It's a bit of a weird error, this line fails:
```Python
os.environ['lib'] = environ_lib + os.environ['lib']
```
with the error that `the environment variable is longer than 32767 bytes`. I don't think this has something to do with my change, I rather think that something has changed in the configuration on appveyor, or maybe in Cython itself. Could you try the Cython tests on your Windows machine?

mstimberg:
Actually, the division-by-zero issue seems to be more severe than I though. If you remove the `unless-refractory` flags for `ge` and `gi` (it is kind to weird to have them their, actually...), then you'll get NaNs in the simulation even before the introduction of the loop-invariant constants with boolean variables. The problem is that the solutions for `ge` and `gi` are much more complicated then they should be (I guess because we removed the `simplify` that led to very long computation times for complicated equations) and it includes terms that cancel each other but that can include divisions by zero for `not_refractory` equal to `False`. I'll open a new issue for this, this is unrelated to the whole numpy 0.10 question.

thesamovar:
Will check later. I still think it's worth trying running simplify in a separate process with a timeout. With joblib cacheing, the timeout could even be quite long.

mstimberg:
Caching is certainly a nice thing (though unfortunately not absolutely trivial to implement if we want to do it right), but I think we should also be careful to have a good interactive workflow for models that were not simulated previously without too long pauses. Also, having long pauses can lead to the first impression that "Brian 2 is slow", even if it is only true for the first run.

mstimberg:
One more thought about time outs: From a user perspective, it might actually be nice to wait for a short time first and then give a message (maybe info-level) so that the user has some idea of what is happening. Something along these lines:
```Python
import multiprocessing
import time

def foo():
    while True:
        time.sleep(1)

if __name__ == '__main__':
    p = multiprocessing.Process(target=foo)
    print 'Starting potentially long-running command'
    p.start()
    # Wait a short time
    p.join(2)
    if p.is_alive:
        # Notify the user
        print 'this seems to be taking some time, waiting 10 more seconds...'
        p.join(10)
        if p.is_alive:
            print 'this is taking too long, terminating.'
            p.terminate()
            p.join()
        else:
            print 'finished, all good.'
```

thesamovar:
I like that idea. I would give no message at all for the first 1s or so though, because in 90% of cases it will be quickly resolved and there's no need to spam the user with text.

mstimberg:
> I like that idea. I would give no message at all for the first 1s or so though, because in 90% of cases it will be quickly resolved and there's no need to spam the user with text.

I agree -- that's why my code example was exactly doing that :)

mstimberg:
Um, actually there might be a misunderstanding -- the `print 'Starting potentially long-running command'` was only there for demonstration purposes, I would not include it in the actual code. And you have to check `is_alive()` and not just `is_alive` as I mistakenly wrote above...

thesamovar:
Ah OK - understood now! :)

mstimberg:
I implemented the short-term fix and call `simplify` in the linear state updater, but only when it is actually used, not during the determination of a suitable state updater. I think that should be good enough for now, the caching/timeout thing is a separate issue. Ready to merge from my side (modulo the appveyor issue with Cython).

mstimberg:
It seems Cython is still not working on appveyor.

thesamovar:
I ran the standard test suite on my machine and everything passed (including Cython). Still though, we should sort out the problem with AppVeyor before merging maybe? Also, the code looks good, happy to merge once we sort this issue out.

thesamovar:
Oh but I didn't upgrade my numpy though, should I do that and test again?

mstimberg:
Yes, please upgrade numpy. And I just realized that weave is failing on appveyor, as well... :-/

thesamovar:
I can confirm I'm getting the same problem on Cython as AppVeyor. I think I know what the problem is. The line ``os.environ['lib'] = environ_lib + os.environ['lib']`` is called in ``MSVCCompiler.initialize``. I'm not exactly sure where this method is getting called from, but my guess is that it gets called each time we build a Cython extension. So, what is happening is that ``os.environ['lib']`` is getting bigger each time we build an extension until it hits the 32k limit, and this bug hasn't been noticed before because few people are building as many extensions as we are! The same thing happens with ``os.environ['include']``. We can fix this in two ways: (1) directly patching MSVCCompiler, which feels a bit risky, or (2) before/after each extension build we analyse ``os.environ['lib']`` and remove duplicate entries. I think this should be safe and should solve the problem. I'll try it later.

thesamovar:
I didn't look into the weave errors, but my guess is that it will be the same issue.

thesamovar:
I just committed a fix for Cython that works on my machine for the path length problem based on what I said before. Didn't check weave yet.

mstimberg:
Great, that seems to fix the Cython problem. Weave is a different issue, though:
```
File "C:\Python27-conda32\lib\site-packages\brian2-2.0b4+git-py2.7-win32.egg\brian2\codegen\runtime\weave_rt\weave_rt.py", line 227, in run
    verbose=0)
  File "C:\Python27-conda32\lib\site-packages\scipy\weave\inline_tools.py", line 366, in inline
    **kw)
  File "C:\Python27-conda32\lib\site-packages\scipy\weave\inline_tools.py", line 496, in compile_function
    verbose=verbose, **kw)
  File "C:\Python27-conda32\lib\site-packages\scipy\weave\ext_tools.py", line 373, in compile
    verbose=verbose, **kw)
  File "C:\Python27-conda32\lib\site-packages\scipy\weave\build_tools.py", line 279, in build_extension
    setup(name=module_name, ext_modules=[ext],verbose=verb)
  File "C:\Python27-conda32\lib\site-packages\numpy\distutils\core.py", line 169, in setup
    return old_setup(**new_attr)
  File "C:\Python27-conda32\lib\distutils\core.py", line 151, in setup
    dist.run_commands()
  File "C:\Python27-conda32\lib\distutils\dist.py", line 953, in run_commands
    self.run_command(cmd)
  File "C:\Python27-conda32\lib\distutils\dist.py", line 972, in run_command
    cmd_obj.run()
  File "C:\Python27-conda32\lib\site-packages\numpy\distutils\command\build_ext.py", line 203, in run
    compiler.customize(self.distribution, need_cxx=need_cxx_compiler)
  File "C:\Python27-conda32\lib\site-packages\numpy\distutils\ccompiler.py", line 388, in CCompiler_customize
    elif not self.compiler_cxx:
AttributeError: MSVCCompiler instance has no attribute 'compiler_cxx'
```

thesamovar:
This appears to be a bug in numpy/scipy/something else. Easily reproduced with ``weave.inline('int x=2;', compiler='msvc')``. Looking into it now, but not found anything yet.

thesamovar:
Well it's pretty weird. It seems as though numpy 1.10 has changed its internal version of distutils and this has somehow fucked things up. I suspect that some other package is doing something like monkey patching MSVCCompiler in a way that isn't compatible with what numpy is doing or some weird combination. Not quite sure how to proceed. I think you looked into these issues at some point, do you remember what was going on here?

thesamovar:
Yeah so numpy 1.10 now has a ``numpy.distutils.msvccompiler.py`` file where previously it didn't, and it seems that this file is responsible for this problem and the Cython one. It appears they've added this file in order to support a few extra options like ``site.cfg`` files and parallel compilation (which are nice) but they've introduced new bugs in doing so. One option is that maybe we can monkey patch it back to the original version, although this doesn't feel like a great option...

thesamovar:
I couldn't track down why this problem was occurring, but I came up with a hack that makes the problem go away on my machine at least: simply create the attribute ``compiler_cxx`` that it can't find. With that, all tests pass no problem, and this check for this variable wasn't there in previous versions of numpy so hopefully it's not critical. Let's see if the tests pass, but assuming they do: what do you think? OK or too hacky?

mstimberg:
I think a little hack is fine here, hopefully this is something that gets sorted out in numpy soon (maybe we should ask on the mailing list/file an issue?). To be more safe, maybe wrap the setting of `compiler_cxx`?
 ```Python
if not hasattr(_msvccompiler.MSVCCompiler, 'compiler_cxx')`:
    _msvccompiler.MSVCCompiler.compiler_cxx = 'cl.exe'
```

thesamovar:
Good idea! If you're at your desktop can you make the change?

mstimberg:
Ok, done. If the tests pass, this is good to merge from my side.

thesamovar:
OK, me too - merging!

PR #576, opened 2015-10-15
[MRG] Fix Python spike queue

mstimberg:
I did not look into the issue in detail yet, but this branch includes an extended `test_transmission` test that fails with the Python spike queue.
(BTW: This test only tests transmission with one-to-one synapses, we should at least have tests for many-to-one and one-to-many connections as well).

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -65,9 +65,7 @@ matrix:
     - python: "2.7"
       env: PYTHON="2.7" STANDALONE=no CYTHON=no MINIMAL_VERSIONS=no REPORT_COVERAGE=no ARCHITECTURE="x86_64"
       os: linux
-    - python: "2.7"
-      env: PYTHON="2.7" STANDALONE=yes CYTHON=no MINIMAL_VERSIONS=no REPORT_COVERAGE=no ARCHITECTURE="x86_64"
-      os: linux
+
 
 # Use miniconda to install binary versions of numpy etc. from continuum
 # analytic's repository. Follows an approach described by Dan Blanchard:
--- a/.travis.yml
+++ b/.travis.yml
@@ -65,6 +65,9 @@ matrix:
     - python: "2.7"
       env: PYTHON="2.7" STANDALONE=no CYTHON=no MINIMAL_VERSIONS=no REPORT_COVERAGE=no ARCHITECTURE="x86_64"
       os: linux
+    - python: "3.4"
+      env: PYTHON="3.4" STANDALONE=no CYTHON=no MINIMAL_VERSIONS=no REPORT_COVERAGE=no ARCHITECTURE="x86_64"
+      os: linux
 
 
 # Use miniconda to install binary versions of numpy etc. from continuum
--- a/.travis.yml
+++ b/.travis.yml
@@ -65,7 +65,9 @@ matrix:
     - python: "2.7"
       env: PYTHON="2.7" STANDALONE=no CYTHON=no MINIMAL_VERSIONS=no REPORT_COVERAGE=no ARCHITECTURE="x86_64"
       os: linux
-
+    - python: "2.7"
+      env: PYTHON="2.7" STANDALONE=yes CYTHON=no MINIMAL_VERSIONS=no REPORT_COVERAGE=no ARCHITECTURE="x86_64"
+      os: linux
 
 # Use miniconda to install binary versions of numpy etc. from continuum
 # analytic's repository. Follows an approach described by Dan Blanchard:
--- a/brian2/devices/cpp_standalone/templates/summed_variable.cpp
+++ b/brian2/devices/cpp_standalone/templates/summed_variable.cpp
@@ -18,13 +18,14 @@
     const int _vectorisation_idx = -1;
     {{scalar_code|autoindent}}
 
-    {{ openmp_pragma('parallel-static') }}
+    {# The following could be parallelized with OpenMP, but this needs some care
+       to not write concurrently to the same post-synaptic variable. A general
+       critical block around the write slows things down too much, though. #}
     for(int _idx=0; _idx<_num_synaptic_post; _idx++)
     {
         // vector code
         const int _vectorisation_idx = _idx;
         {{vector_code|autoindent}}
-        {{ openmp_pragma('atomic') }}
         {{_target_var_array}}[{{_synaptic_post}}[_idx]] += _synaptic_var;
     }
 {% endblock %}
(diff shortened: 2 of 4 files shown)


mstimberg:
The pre-calculation of the offsets seems to be completely wrong (I'm wondering whether this not also affects Brian 1, IIRC we copied the Python spike queue mostly unchanged)! It pre-calculates the offsets for repeated delays for each source neuron independently, so in the following example, the offsets are all zero (because no source neuron has repeated delays):
```Python
G = NeuronGroup(3, 'v : 1', threshold='v>1')
G.v = 1.1
S = Synapses(G, G, pre='v+=0', connect='i==j')
S.delay = [0, 1, 1] * ms
```
Running this example for one time step with some debug output in the spike queue, leads to the following:
Spike queue with `precompute_offsets=True` (the default):
```
precalculated offsets [0 0 0]
After push, spike queue looks like this
[0] <-- current
[] 
[] 
[] 
[] 
[] 
[] 
[] 
[] 
[] 
[2]
```
With `precompute_offsets=False`:
```
calculated offsets during insert [0 0 1]
After push, spike queue looks like this
[0] <-- current
[] 
[] 
[] 
[] 
[] 
[] 
[] 
[] 
[] 
[1 2]
```

So with the pre-calculated offset, we are completely skipping one spike! To be honest I don't see a way how we can fix the pre-calculation approach that pre-calculates one offset per synapse -- their offset is not independent, a single synapse with whatever delay should always use offset 0 if it is the only synapse that is receiving a spike but have an offset >0 if there are other synapses with the same delay receiving a spike (if it is not the first synapse, obviously).

I'm surprised that our test suite did not catch this, but it seems all our synaptic propagation tests were too simple. The only really complex test that could have caught it is the `test_openmp_consitency`, but since this is a standalone test, we did not run it on travis without Cython and therefore with the Python spike queue (at the time my reasoning was that for standalone, having Cython or not should not make a difference)... I now included a test on travis that runs the standalone tests without Cython, this should hopefully fail and will be useful for similar consistency tests in the future.

This bug does certainly warrant a mail on the mailing list...

mstimberg:
> I now included a test on travis that runs the standalone tests without Cython, this should hopefully fail and will be useful for similar consistency tests in the future.

Forget what I said about the `test_openmp_consistency` test, it does not use delays so it does not trigger the issue. Still, I think it is good to run the standalone/Python spike queue combination, we'll hopefully have more consistency tests in the future.

mstimberg:
> I now included a test on travis that runs the standalone tests without Cython, this should hopefully fail and will be useful for similar consistency tests in the future.

Forget what I said about the `test_openmp_consistency` test, it does not use delays so it does not trigger the issue and in fact, standalone tests were always done without Cython, so the test did in fact compare to the Python Spike queue -- I reverted the change, it was a completely redundant test configuration.

For some reason, on appveyor now the `test_synapses_to_synapses_summed_variables` test now started failing on standalone -- this is really weird since this branch did not introduce any changes except for changes in `test_transmission`...

thesamovar:
:(

mstimberg:
Well, it seems that it was in fact me that introduced this bug in an "optimization" > 2 years ago :worried: 

The good news, the bug is not in Brian 1 and since we try to install the C version of the spikequeue during installation for Brian 2, everyone using the conda package and most users installing via easy_install or pip should not have been affected. Still, this is severe enough as it is... I wonder whether it might not even warrant yet another beta release (alternatively, we could give instructions how to find out whether you are affected and how to patch `spikequeue.py` by setting `precompute_offsets` to `False`).

Before the change (i.e. in Brian 1), insertion with pre-calculated offsets looped over all the spiking source neurons and then did one insert for each of them, therefore the "calculate offsets per neuron" logic made sense. In an ill-guided attempt at optimizing away the loop, I replaced this with a single insert for all synapses which introduced the problem.

The only thing I can say in my defense is that our code review process back at the time was not that thorough #78 ;)

mstimberg:
We have two options how to fix this: remove the precalculation of offsets completely or restore the "loop over source neurons" approach. Is looping+pre-calculated offsets really significant faster than calculating offsets on demand together with a vectorized insert? If the difference is not big, I'd prefer to remove the pre-calculation completely, after all the spike queue is only meant as a fallback and it would be nice to reduce its complexity.

thesamovar:
I guess we should check whether it makes a difference. Not having precomputations like that would be generally preferable I think if it doesn't make much difference.

thesamovar:
Or in other words, yes, I agree. ;)

mstimberg:
I did not do thorough benchmarking but testing with the CUDA example, I get (all times are for `synapses_pre` which includes the push but also the update itself):
Homogeneous delays: ~6.45s
Hetereogeneous delays without pre-computation (but vectorized): 7.63s
Hetereogeneous delays with pre-computation: 6.85s

For comparison with the C++ spike queue: 6.05s

So it is faster, but the speed advantage has a negligible influence on the total run time. Do you maybe have more thorough benchmarking code laying around from the time when you looked at optimizing the C++ spike queue?

thesamovar:
I don't have any more thorough benchmarking code. I think if the difference is just 11% then let's go for the slightly slower, simpler option. After all, people who need highly efficient code won't be using this version anyway (and hopefully hardly anyone will).

mstimberg:
Ok, then I'd say we remove it for now. I think in the future more and more users will be using the conda package anyway, so they will never come into contact with the Python spike queue. If we'll find a use case where the Python spike queue is unbearably slow without the pre-calculation, we can always re-introduce this feature (with better testing...).

I will add a few more synaptic transmission tests and then commit the change that removes the offset pre-calculation.

mstimberg:
Ready-to-merge if all tests pass. After the merge, I'll write a mail to the lists.

thesamovar:
Well, all this is a great relief. I'm glad we don't have to do another big bug announcement. Will review and merge soon.

mstimberg:
> Well, all this is a great relief. I'm glad we don't have to do another big bug announcement.

Well, we have to for Brian 2... But indeed it is a relief that it is not an issue for Brian 1, it would have been much more severe because there you had to manually install the C++ spike queue, so most people are using the Python version.

mstimberg:
There was still a bug open in the OpenMP implementation of the `summed_variable` template (that I probably introduced when we refactored it a bit), therefore tests on appveyor were failing. For now I removed most of the parallelization, I'm not sure that anything more complex (local sums, merging them later, critical sections, etc.) is actually worth the effort, the overhead from the synchronization did completely cancel any beneficial effect in some small tests I did. In the long run, it might be worth looking into it, but that can be done as part of #539 I think.

So if the tests now finally pass, this is really ready to merge.

thesamovar:
OK, code looks good to me so also happy for it to be merged when tests pass.

mstimberg:
All tests pass, merging now.

Issue #577, opened 2015-10-16
Improve weave's efficiency


mstimberg:
I noticed that simulations with weave were extremely slow on my machine and realized that weave is very inefficient on machines where you run a lot of different weave code (which we do of course, especially when running the test suite). 

On my machine, the database index in `~/.cache/scipy/python27_compiled/linux227compiled_catalog.dir` is ~400MB. This file is accessed using the "patched" `dumbdbm` from weave by reading in all lines (11639 in my case). Each line is of the form:
```
"some C code", (number, number)
```
And this will be converted into a tuple of string and two integers using Python's `eval` function (sic). This whole process is done 10 times for the CUBA simulation, I guess once for every code object... 

Actually, any benchmarking I did recently (e.g. #576) is probably mostly a reflection of these database operations. If I run the CUBA example with the 400MB database, it takes more than 1 minute. After deleting the database it only takes 26s even though it has to compile all the code! A run after that only takes 13s...

I wonder whether it makes sense to for example have one database directory per project directory, instead of a global one for the user (at least as an option). Granted that our situation is a bit extreme because we run the test suite often and we run it on different version of Brian. I can imagine something similar happening for other "power users" as well, and it can give a very bad impression for runtime mode -- in fact, I already started suspecting a harddisk problem as a reason for the slowdown...

thesamovar:
Oh wow! I'm surprised it's so bad. Can we access the database store for weave?

I think it shouldn't be so bad for Cython. If I remember correctly, we take a hash of the source code and use that as a filename, and then we just check if that filename is there or not. I thought that weave would be doing something similar.

mstimberg:
Actually, after a bit of investigation I found the situation may be specific to my machine or maybe anaconda installations. If I try `import dbhash`, then this fails with an error about `bsddb`. This is way weave falls back to the "dumb db" which is using the huge textfile + eval approach. After a `conda install bsddb` it now uses a Berkeley DB database file (`linux227compiled_catalog`) and I think this should be much faster and memory efficient. We should probably add this as a dependency, at least for the conda package.
Actually, maybe with the Berkeley DB issue #463 disappears and we can run weave in parallel?

mstimberg:
> Actually, maybe with the Berkeley DB issue #463 disappears and we can run weave in parallel?

Just tried it, this is still not working so the DB is not the issue. I think the problem is that several process might interfere while writing to the same file.

thesamovar:
[...]

PR #578, opened 2015-10-19
[MRG] Weave fixes

mstimberg:
I noticed that there were a few bugs which even made some examples no longer work with weave (all related to getting rid of `AttributeVariable` for `N`). None of these problems were caught by the test suite, mostly because they were in the templates that are used for expressions such as `G.v['i%2 == 0'] = -70*mV` and by default we execute them in numpy, since the compilation often takes longer than running the code in such simple cases. Now when the tests are run, also these simple statements are using the tested code generation target. I also added one small test and included `bsddb` in the run-time dependencies for the conda package which should make the issue with inefficient weave under anaconda go away (#577).

Nothing too spectacular in this branch, ready to merge from my side.


mstimberg:
Despite its name, this branch now also fixes an issue for Cython -- the `abs` function was not working for integer values on Windows. I hope this is now fixed.

thesamovar:
Happy for this to be merged when appveyor completes.

PR #579, opened 2015-10-20
[MRG] Store state to disk

mstimberg:
There's still a bit of documentation and testing missing, but the basic functionality is there. The `store/restore` mechanism now takes an optional `filename` argument, allowing you to store the state to disk instead of to memory. It internally uses a `_full_state` method that returns a dictionary of the values of all the state variables of the object (for `SynapticPathway`, it also contains a representation of the internal state of the spike queue). There is some redundancy with the `get_state` method here, but I'd prefer them to stay separate, the `get_state` method is something for the user which focuses on usability instead of completeness. 

The spike queue part is not quite working yet but the current test does not catch it.

### DIFF
--- a/brian2/core/magic.py
+++ b/brian2/core/magic.py
@@ -231,20 +231,20 @@ class MagicNetwork(Network):
         Network.run(self, duration, report=report, report_period=report_period,
                     namespace=namespace, profile=profile, level=level+1)
 
-    def store(self, name='default', level=0):
+    def store(self, name='default', filename=None, level=0):
         '''
         See `Network.store`.
         '''
         self._update_magic_objects(level=level+1)
-        super(MagicNetwork, self).store(name=name)
+        super(MagicNetwork, self).store(name=name, filename=filename)
         self.objects[:] = []
 
-    def restore(self, name='default', level=0):
+    def restore(self, name='default', filename=None, level=0):
         '''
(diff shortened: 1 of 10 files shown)


mstimberg:
I fixed a few issues and added tests and documentation, so this is now ready to merge from my side.

thesamovar:
So in runtime mode, all objects return a dictionary which is then stored in Network's own dictionary, is that right? Then this dictionary is just pickled to disk. In standalone mode, how does it work? (Since we have no pickling there.)

mstimberg:
The `store`/`restore` mechanism does not work for standalone, neither for storing to memory, nor for storing to disk:
```Python
    def network_store(self, net, name='default'):
        raise NotImplementedError(('The store/restore mechanism is not '
                                   'supported in the C++ standalone'))

    def network_restore(self, net, name='default'):
        raise NotImplementedError(('The store/restore mechanism is not '
                                   'supported in the C++ standalone'))
```
Getting it to work there would take some major work.

thesamovar:
Ah that's a shame. OK though, in that case I think I understand what's going on and happy for this to be merged when tests pass.

mstimberg:
github does not display the testing status for some reason, but all the tests passed on travis and appveyor. Merging it now.

PR #583, opened 2015-10-21
[MRG] allow Network.add to accept dicts (fixes #564)

### LINKED ISSUE #564: Support adding dictionaries of objects to Network
We support ``net.add(list)`` but not ``net.add(dict)``. This should be easy to implement and worthwhile I think.

thesamovar:


mstimberg:
Looks good of course, but maybe it would be slightly more general to use `isinstance(obj, collections.Mapping)` instead of `isinstance(obj, dict)`?

PR #586, opened 2015-10-22
[MRG] Better tracebacks

thesamovar:
We now have a ``BrianObject._creation_stack`` string that looks like this:

```
Object creation traceback:
  File "D:/programming/brian2/dev/temp_scripts/better_tracebacks.py", line 3, in <module>
    G = NeuronGroup(1, 'v:1')
```

Haven't used it anywhere yet. Any thoughts about how to use it?


mstimberg:
I think there are several places where it could be used:
* When a name cannot be resolved: https://github.com/brian-team/brian2/blob/master/brian2/groups/group.py#L769
* When the units of an equation are inconsistent (but I think it's a bit difficult to integrate into the current format of `DimensionMismatchError`): https://github.com/brian-team/brian2/blob/master/brian2/equations/equations.py#L858
* Basically anything that is checked in a `before_run` method, e.g. in `SpikeGeneratorGroup`: https://github.com/brian-team/brian2/blob/master/brian2/input/spikegeneratorgroup.py#L160

I wonder whether it wouldn't make sense to have a `BrianError` exception class that would deal with this somewhat automatically, instead of adding a bit of text manually to all the `ValueError`, `KeyError`, etc. exceptions we raise

thesamovar:
So how does this seem? In the following example:

```
from brian2 import *

tau = 1*second
eqs = '''
dv/dt = -v/tau : 1
'''

G = NeuronGroup(1, eqs)

tau = 1.0

run(1*ms)
```

You'll get the following error messages:

```
ERROR    brian2: Brian 2 encountered an unexpected error. If you think this is bug in Brian 2, please report this issue either to the mailing list at <http://groups.google.com/group/brian-development/>, or to the issue tracker at <https://github.com/brian-team/brian2/issues>. Please include this file with debug information in your report: c:\users\dgoodman\appdata\local\temp\brian_debug_z8ah6l.log  Additionally, you can also include a copy of the script that was run, available at: c:\users\dgoodman\appdata\local\temp\brian_script_96xgaq.py Thanks!
  File "C:/programming/brian2/dev/temp_scripts/better_tracebacks.py", line 12, in <module>
    run(1*ms)
  File "C:\programming\brian2\brian2\units\fundamentalunits.py", line 2364, in new_f
    result = f(*args, **kwds)
  File "C:\programming\brian2\brian2\core\magic.py", line 372, in run
  File "C:\programming\brian2\brian2\core\magic.py", line 232, in run
  File "C:\programming\brian2\brian2\core\base.py", line 269, in device_override_decorated_function
    return func(*args, **kwds)
  File "C:\programming\brian2\brian2\units\fundamentalunits.py", line 2364, in new_f
    result = f(*args, **kwds)
  File "C:\programming\brian2\brian2\core\network.py", line 770, in run
    self.before_run(namespace, level=level+3)
  File "C:\programming\brian2\brian2\core\base.py", line 269, in device_override_decorated_function
    return func(*args, **kwds)
  File "C:\programming\brian2\brian2\core\network.py", line 682, in before_run
    raise BrianObjectException("An error occurred when preparing an object.", obj, ex)
  File "C:\programming\brian2\brian2\core\network.py", line 680, in before_run
    obj.before_run(run_namespace, level=level+2)
  File "C:\programming\brian2\brian2\groups\neurongroup.py", line 781, in before_run
    level=level+1)
  File "C:\programming\brian2\brian2\equations\equations.py", line 863, in check_units
    *ex.dims)
DimensionMismatchError: Inconsistent units in differential equation defining variable v:
Expression -v/tau does not have the expected unit  / second (unit is 1).

Error encountered with object named "neurongroup".
Object was created here (most recent call last):
  File "C:/programming/brian2/dev/temp_scripts/better_tracebacks.py", line 8, in <module>
    G = NeuronGroup(1, eqs)

An error occurred when preparing an object. DimensionMismatchError: Inconsistent units in differential equation defining variable v:
Expression -v/tau does not have the expected unit  / second (unit is 1).
```

The idea was that the most user-relevant information should be at the end, working backwards.

mstimberg:
Hey, great! I'll look into it in more detail tomorrow, but this looks really good at a first sight. Are there any situations where we get wrong information about the place where the object was created? If yes, maybe change the wording to something like "Object was probably created here"? What happens in an interactive shell?

thesamovar:
I think it shouldn't ever create a problem with getting the place wrong. It just prints out the stack trace at the time that ``BrianObject.__init__`` is called, with all lines inside Brian removed. For an IPython console it looks like this:

```
Object was created here (most recent call last):
  File "C:\Anaconda\Scripts/ipython-script.py", line 5, in <module>
    sys.exit(start_ipython())
  File "C:\Anaconda\lib\site-packages\IPython\__init__.py", line 120, in start_i
python
    return launch_new_instance(argv=argv, **kwargs)
  File "C:\Anaconda\lib\site-packages\IPython\config\application.py", line 574,
in launch_instance
    app.start()
  File "C:\Anaconda\lib\site-packages\IPython\terminal\ipapp.py", line 371, in s
tart
    self.shell.mainloop()
  File "C:\Anaconda\lib\site-packages\IPython\terminal\interactiveshell.py", lin
e 383, in mainloop
    self.interact(display_banner=display_banner)
  File "C:\Anaconda\lib\site-packages\IPython\terminal\interactiveshell.py", lin
e 507, in interact
    self.run_cell(source_raw, store_history=True)
  File "C:\Anaconda\lib\site-packages\IPython\core\interactiveshell.py", line 28
71, in run_cell
    interactivity=interactivity, compiler=compiler, result=result)
  File "C:\Anaconda\lib\site-packages\IPython\core\interactiveshell.py", line 29
75, in run_ast_nodes
    if self.run_code(code, result):
  File "C:\Anaconda\lib\site-packages\IPython\core\interactiveshell.py", line 30
35, in run_code
    exec(code_obj, self.user_global_ns, self.user_ns)
  File "<ipython-input-5-9d6098464811>", line 1, in <module>
    G = NeuronGroup(1, 'v:1')
```

We could special case the IPython console by removing all lines inside the IPython package? Then you'd just get the last two lines.

thesamovar:
Or we could just show the last line always, but then you get a potential loss of information.

thesamovar:
I special cases IPython so it doesn't show any lines from any IPython module. I wonder if there are any other consoles we ought to consider?

thesamovar:
Called from the Python standard console you get this:

```
Object was created here (most recent call last):
  File "<stdin>", line 1, in <module>
    None
```

Maybe that's OK though?

thesamovar:
Hmm, it's not ideal. From an IPython notebook you get this:

```
Object was created here (most recent call last):
  File "C:\Anaconda\lib\runpy.py", line 162, in _run_module_as_main
    "__main__", fname, loader, pkg_name)
  File "C:\Anaconda\lib\runpy.py", line 72, in _run_code
    exec code in run_globals
  File "C:\Anaconda\lib\site-packages\zmq\eventloop\ioloop.py", line 151, in start
    super(ZMQIOLoop, self).start()
  File "C:\Anaconda\lib\site-packages\tornado\ioloop.py", line 840, in start
    handler_func(fd_obj, events)
  File "C:\Anaconda\lib\site-packages\tornado\stack_context.py", line 275, in null_wrapper
    return fn(*args, **kwargs)
  File "C:\Anaconda\lib\site-packages\zmq\eventloop\zmqstream.py", line 433, in _handle_events
    self._handle_recv()
  File "C:\Anaconda\lib\site-packages\zmq\eventloop\zmqstream.py", line 465, in _handle_recv
    self._run_callback(callback, msg)
  File "C:\Anaconda\lib\site-packages\zmq\eventloop\zmqstream.py", line 407, in _run_callback
    callback(*args, **kwargs)
  File "C:\Anaconda\lib\site-packages\tornado\stack_context.py", line 275, in null_wrapper
    return fn(*args, **kwargs)
  File "<ipython-input-1-a8d969e51554>", line 2, in <module>
    G=brian2.NeuronGroup(1,'v:1')
```

We could exclude all those modules (zmq, tornado and the runpy.py script), but it feels like a fragile solution. What about just showing the last file by default, and putting the full traceback of the object creation in the debug log?

thesamovar:
I implemented that last solution and I'm pretty happy with it. Here's the output from the code above:

```
C:\Anaconda\python.exe C:/programming/brian2/dev/temp_scripts/better_tracebacks.py
DEBUG    brian2.core.names: Created object of class NeuronGroup with name neurongroup
DEBUG    brian2.core.base: Created BrianObject with name neurongroup, clock=Clock(dt=100. * usecond, name='defaultclock'), when=start, order=0
DEBUG    brian2.groups.neurongroup: Creating NeuronGroup of size 1, equations dv/dt = -v/tau  : 1.
DEBUG    brian2.core.names: Created object of class StateUpdater with name neurongroup_stateupdater
DEBUG    brian2.core.base: Created BrianObject with name neurongroup_stateupdater, clock=Clock(dt=100. * usecond, name='defaultclock'), when=groups, order=0
INFO     brian2.core.magic.magic_objects: Updated MagicNetwork to include 2 objects with names neurongroup, neurongroup_stateupdater
DEBUG    brian2.core.clocks: Setting Clock defaultclock to t=<defaultclock.t: 0. * second>, dt=100. us
DEBUG    brian2.core.network.before_run: Preparing network magicnetwork with 2 objects: neurongroup, neurongroup_stateupdater
DEBUG    brian2.core.base: Error was encountered with object "neurongroup":
Object was created here:
  File "C:/programming/brian2/dev/temp_scripts/better_tracebacks.py", line 10, in <module>
    G = NeuronGroup(1, eqs)
ERROR    brian2: Brian 2 encountered an unexpected error. If you think this is bug in Brian 2, please report this issue either to the mailing list at <http://groups.google.com/group/brian-development/>, or to the issue tracker at <https://github.com/brian-team/brian2/issues>. Please include this file with debug information in your report: c:\users\dgoodman\appdata\local\temp\brian_debug_guu_rp.log  Additionally, you can also include a copy of the script that was run, available at: c:\users\dgoodman\appdata\local\temp\brian_script_z3k9fs.py Thanks!
  File "C:/programming/brian2/dev/temp_scripts/better_tracebacks.py", line 14, in <module>
    run(1*ms)
  File "C:\programming\brian2\brian2\units\fundamentalunits.py", line 2364, in new_f
    result = f(*args, **kwds)
  File "C:\programming\brian2\brian2\core\magic.py", line 372, in run
  File "C:\programming\brian2\brian2\core\magic.py", line 232, in run
  File "C:\programming\brian2\brian2\core\base.py", line 276, in device_override_decorated_function
    return func(*args, **kwds)
  File "C:\programming\brian2\brian2\units\fundamentalunits.py", line 2364, in new_f
    result = f(*args, **kwds)
  File "C:\programming\brian2\brian2\core\network.py", line 770, in run
    self.before_run(namespace, level=level+3)
  File "C:\programming\brian2\brian2\core\base.py", line 276, in device_override_decorated_function
    return func(*args, **kwds)
  File "C:\programming\brian2\brian2\core\network.py", line 682, in before_run
    raise BrianObjectException("An error occurred when preparing an object.", obj, ex)
  File "C:\programming\brian2\brian2\core\network.py", line 680, in before_run
    obj.before_run(run_namespace, level=level+2)
  File "C:\programming\brian2\brian2\groups\neurongroup.py", line 781, in before_run
    level=level+1)
  File "C:\programming\brian2\brian2\equations\equations.py", line 863, in check_units
    *ex.dims)
DimensionMismatchError: Inconsistent units in differential equation defining variable v:
Expression -v/tau does not have the expected unit  / second (unit is 1).

Error encountered with object named "neurongroup".
  File "C:/programming/brian2/dev/temp_scripts/better_tracebacks.py", line 10, in <module>
    G = NeuronGroup(1, eqs)

An error occurred when preparing an object. DimensionMismatchError: Inconsistent units in differential equation defining variable v:
Expression -v/tau does not have the expected unit  / second (unit is 1).

Process finished with exit code 1
```

So if we're happy with that solution, I'll raise that type of error in a few more places.

mstimberg:
Yes, I think this is a good solution. The only thing that looks a bit odd is that we see the main error message twice now, but I guess that's ok. I'll have a closer look at everything soon.

thesamovar:
This messes up a lot of tests (because they're expecting a NotImplementedError for example).

thesamovar:
OK, I solved the test failure problem. Now the function ``brian_object_exception`` raises a new class that is derived from the original exception class and ``BrianObjectException``, meaning that tests that looked for a particular exception class will still work. What do you think of this approach? It seems fine to me but using dynamically created classes always makes me mildly nervous.

mstimberg:
I agree, it makes me nervous, too :) I am not sure whether it has to be this complicated, maybe re-raising the exception would be more elegant in general (which is much nicer in Python 3: http://blog.ionelmc.ro/2014/08/03/the-most-underrated-feature-in-python-3/). For the tests, I wouldn't actually mind if we just switched the tests to test for a `BrianObjectException`. I think in most cases, we are raising exceptions such as `ValueError`, `TypeError`, `KeyError` simply for the lack of a brian-specific exception class...

thesamovar:
I quite like it because you might want to handle specific exception types as part of non-user library code that relies on Brian. Feels like turning them all into ``BrianObjectException`` throws away some information. I think I slightly prefer keeping it, but I'm happy not to and just modify the tests if you prefer? If so, is there a way of reraising exceptions that can work for Python 2 and 3? That would be nice.

thesamovar:
I was just looking at the three places where you suggested we could use this error, and actually they're all more or less covered by the one place I put it in (i.e. in ``Network.before_run``). If you do ``NeuronGroup(1, 'dv/dt=-v')`` you will get a ``DimensionMismatchError`` rather than a ``BrianObjectException`` but I think that's OK because in this case the traceback points to the line where the problem is. What we want to cover here is the case where the problem occurs with a traceback that doesn't point to the original object creation, which I think will always be coming from ``Network.before_run`` right? So perhaps this one place is enough to cover all the possibilities?

thesamovar:
I think on reflection, this solution is OK unless you have strong reservations about the dynamic exception thing? I believe it's safe and the code is not complicated (only 2 lines). The only way I can see it going wrong is that an attempt to inspect the source code of the class of the exception will fail because it's dynamically created, but how often do you want to do that?

Given that, and my previous comment, I think this is ready to merge.

mstimberg:
Sounds reasonable, `before_run` is where all the "delayed preparation" happens and this is where this kind of information is helpful. About the exception type: I was thinking of situations where e.g. the chosen stateupdater cannot integrate the equations where we raise a `ValueError` -- this is certainly not a very descriptive error type here. But on the other hand for stuff like the `NotImplementedError` we raise, we certainly do not want to cover it in a general Brian exception. And I don't think there's any syntax for re-raising that works nicely both on Python 2 and Python 3.

I'll have a closer look tomorrow, but I think I am fine with your approach, it is not terribly complicated either.

thesamovar:
AppVeyor seems to be having a couple of unrelated issues and I don't quite understand what's going wrong with it. Any ideas?

mstimberg:
Yes, I just saw this as well. On appveyor we test with the latest Python 3 version which is now 3.5 while on travis apparently we fixed it to 3.4. As with every Python version update so far, something changed in the AST representation and therefore we have an error in the rendering code. Also one of the jobs timed out, but that's just bad luck (there was some kind of network connectivity problem).

thesamovar:
Ah well, I guess it's good that we're covering all versions of Python somewhere! Should we fix in this branch?

mstimberg:
I think it would be good if the tests pass before merging, so fixing it here would be the easiest. I'll revisit our test settings on travis/appveyor later.

thesamovar:
OK great - would you mind doing it? I'd rather avoid getting into installing Python 3.5 for the moment. At some point, I ought to work out how to get multiple environments working on Windows, but maybe not right now.

mstimberg:
Sure, no problem. Setting up multiple environments is trivial with conda though :) http://conda.pydata.org/docs/using/envs.html

mstimberg:
Ok, I did two small changes and it runs fine on my machine with both Python 3 and Python 2, let's wait for the test suites to run through.

thesamovar:
Something weird went wrong...

mstimberg:
The default installation prefix for the Python 2 version of miniconda changed from `miniconda` to `miniconda2`... I now set the installation prefix explicitly, this should make it work again.

thesamovar:
Judging by the list of commits, you had a very boring and annoying afternoon! Thanks for looking into this.

mstimberg:
It is indeed a bit annoying to fix stuff for appveyor and travis, where you can't really test it on your own machine. It seems to be working fine now and all tests pass for Python 3.4 and 3.5 so I will merge it. I'll open a new issue to update the installation documentation for the release, a few things seem to have changed for windows (for Python 3.5 64 Bit, you no longer need to set any environment variables and anaconda now seems to include the msvc runtime, so possibly Windows user can use code generation without further downloads of msvc stuff now) but we need to check this in detail.

PR #587, opened 2015-10-23
[MRG] Store the "synapse number" of each synapse

mstimberg:
I realized that we never implemented supported for expressions referring to the index of the synapse, i.e. after you create multiple synapses:
```
syn.connect(True, n=2)
```
There was no way to e.g. set weights or delays differently for the first and the second synapse connecting `i` and `j` (with the exception that in run time mode you could use `syn.w[:, :, 0] = ...`). I just needed this myself so I implemented it. This is basically done and working but I have a few  questions before merging it:
* What should be the name of this attribute? In early drafts I think we proposed `k` (following `i` and `j`) but I have the feeling that the chance for clashes with user variables is too high, especially given that `k` in the sense of the synapse index will rarely be used. In the code I therefore opted for `synapse_number`, but now that I'm writing this, `synapse_index` would probably be a better choice?
* Do we want to store it in the first place? I think we have to if we want to allow users to use it; in principle we could possibly calculate it on demand, but this would mean we would have to special case this attribute in quite a few places in the code generation process. On the other hand, we are wasting quite a bit of memory for an array that is 1) not often used and 2) consists mostly of zeros. I could imagine two things to make this a bit better: 1) Use a smaller data type by default (e.g. something like `uint8`) -- but then I don't know how to elegantly handle the situation where a user wants to create many many synapses between the same source/targets. 2) don't use a standard array but some other data structure that does not store the zeros -- but then, this adds quite a bit of complication to it.
* The attribute is calculated during synapse creation for all synapses, this repeats a bit of work if you do multiple calls to `Synapses.connect`. I don't think that's very important, I think we should rarely have more than one or two calls (if we have, it's inefficient anyway).

Any thoughts?

### DIFF
--- a/brian2/codegen/codeobject.py
+++ b/brian2/codegen/codeobject.py
@@ -209,7 +209,7 @@ def create_runner_codeobj(group, code, template_name,
         Whether to check units in the statement. Defaults to ``True``.
     needed_variables: list of str, optional
         A list of variables that are neither present in the abstract code, nor
-        in the ``USES_VARIABLES`` statement in theqq template. This is only
+        in the ``USES_VARIABLES`` statement in the template. This is only
         rarely necessary, an example being a `StateMonitor` where the
         names of the variables are neither known to the template nor included
         in the abstract code statements.
--- a/brian2/codegen/codeobject.py
+++ b/brian2/codegen/codeobject.py
@@ -209,7 +209,7 @@ def create_runner_codeobj(group, code, template_name,
         Whether to check units in the statement. Defaults to ``True``.
     needed_variables: list of str, optional
         A list of variables that are neither present in the abstract code, nor
-        in the ``USES_VARIABLES`` statement in the template. This is only
+        in the ``USES_VARIABLES`` statement in theqq template. This is only
         rarely necessary, an example being a `StateMonitor` where the
         names of the variables are neither known to the template nor included
         in the abstract code statements.
--- a/brian2/codegen/runtime/weave_rt/templates/synapses_create.cpp
+++ b/brian2/codegen/runtime/weave_rt/templates/synapses_create.cpp
@@ -1,5 +1,5 @@
 {% extends 'common_group.cpp' %}
-#include<
+
 {% block maincode %}
     {#
     USES_VARIABLES { _synaptic_pre, _synaptic_post, rand,
--- a/brian2/codegen/runtime/weave_rt/templates/synapses_create.cpp
+++ b/brian2/codegen/runtime/weave_rt/templates/synapses_create.cpp
@@ -2,11 +2,11 @@
 #include<
 {% block maincode %}
     {#
-    USES_VARIABLES { _synaptic_pre, _synaptic_post, synapse_number, rand,
+    USES_VARIABLES { _synaptic_pre, _synaptic_post, rand,
                      N_incoming, N_outgoing, N,
                      N_pre, N_post, _source_offset, _target_offset}
     #}
-    {# WRITES_TO_READ_ONLY_VARIABLES { _synaptic_pre, _synaptic_post, , synapse_number,
+    {# WRITES_TO_READ_ONLY_VARIABLES { _synaptic_pre, _synaptic_post,
                       N_incoming, N_outgoing, N}
     #}
     srand((unsigned int)time(NULL));
--- a/brian2/codegen/runtime/weave_rt/templates/synapses_create.cpp
+++ b/brian2/codegen/runtime/weave_rt/templates/synapses_create.cpp
@@ -1,12 +1,12 @@
 {% extends 'common_group.cpp' %}
-
+#include<
 {% block maincode %}
     {#
-    USES_VARIABLES { _synaptic_pre, _synaptic_post, rand,
+    USES_VARIABLES { _synaptic_pre, _synaptic_post, synapse_number, rand,
                      N_incoming, N_outgoing, N,
                      N_pre, N_post, _source_offset, _target_offset}
     #}
-    {# WRITES_TO_READ_ONLY_VARIABLES { _synaptic_pre, _synaptic_post,
+    {# WRITES_TO_READ_ONLY_VARIABLES { _synaptic_pre, _synaptic_post, , synapse_number,
                       N_incoming, N_outgoing, N}
     #}
     srand((unsigned int)time(NULL));
@@ -15,12 +15,8 @@
     int *const _postbuf = new int[_buffer_size];
     int _curbuf = 0;
 
-    // Resize N_incoming and N_outgoing according to the size of the
-    // source/target groups
-    PyObject_CallMethod(_var_N_incoming, "resize", "i", {{constant_or_scalar('N_post', variables['N_post'])}} + _target_offset);
-    PyObject_CallMethod(_var_N_outgoing, "resize", "i", {{constant_or_scalar('N_pre', variables['N_pre'])}} + _source_offset);
-    int *_N_incoming = (int *)(((PyArrayObject*)(PyObject*){{_dynamic_N_incoming}}.attr("data"))->data);
-    int *_N_outgoing = (int *)(((PyArrayObject*)(PyObject*){{_dynamic_N_outgoing}}.attr("data"))->data);
+    const int oldsize = {{_dynamic__synaptic_pre}}.size();
+
     // scalar code
 	const int _vectorisation_idx = 1;
 	{{scalar_code|autoindent}}
@@ -51,8 +47,6 @@
                 }
 
                 for (int _repetition=0; _repetition<_n; _repetition++) {
-                    _N_outgoing[_pre_idx] += 1;
-                    _N_incoming[_post_idx] += 1;
                     _prebuf[_curbuf] = _pre_idx;
                     _postbuf[_curbuf] = _post_idx;
                     _curbuf++;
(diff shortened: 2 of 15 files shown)


thesamovar:
Yes, it's a shame to waste memory storing it when it will hardly ever be used. What about we have it off by default and you have to switch it on to use it? e.g. ``Synapses(..., multisynapses=True, ...)``.

I think ``synapse_number`` and ``synapse_index`` are both misleading because you might expect them to be the main index (which is called ``_synaptic_idx`` right?). I actually think ``k`` is OK. In principle we have the same problem with ``i`` and ``j`` if users don't know that they refer to synapse indices. Hopefully our namespace resolution should warn if there is an issue with this anyway, right?

mstimberg:
> Yes, it's a shame to waste memory storing it when it will hardly ever be used. What about we have it off by default and you have to switch it on to use it? e.g. Synapses(..., multisynapses=True, ...).

You mean we should switch off the possibility to have multiple synapses per source-target pair or only switch off the calculation of the synaptic index? Potentially we could have optimisations in a few places for the first case, but it would be somewhat hard to check/enforce since you can have multiple `connect` calls. Only calculation the synaptic indices when asked for should be relatively straightforward, though. A question is whether this would also switch off the option to use the three-argument indexing, i.e. something like `syn.w[:, :, 0] = ...`. Previously, this triggered a calculation of the synaptic index (and only worked in runtime mode) whereas in this branch it is based on the stored values.

> I think synapse_number and synapse_index are both misleading because you might expect them to be the main index (which is called _synaptic_idx right?).

I think our current approach is that the user should not care about the main index, nor refer to it in any way directly. There's no way to do this currently in code, there's no `_synaptic_idx` that the user could use. In the template, it might be referred to as `_spiking_synapse_idx` or `_idx`, but this is an implementation detail and user code cannot use such names.

> I actually think k is OK. In principle we have the same problem with i and j if users don't know that they refer to synapse indices. Hopefully our namespace resolution should warn if there is an issue with this anyway, right?

Actually, I'm not sure we are actually -- the `i`, `j` and `k` variables are not user-defined, so they don't go through the check. So if you connect to a `NeuronGroup` that has one of those variables, it will not trigger the error that you'd get otherwise. But apart from that, I think we should in general avoid to "block" common names, the exception for `i` and `j` is somewhat warranted (even though I wouldn't have minded to have something like `pre_idx` and `post_idx` instead) because in standard model descriptions you'd see something like `w_ij` -- I don't think there's such a standard meaning for `k`, though. But I don't have a strong opinion on this, either, if we make its calculation optional then probably `k` is ok. Alternatively, for full flexibility (but that's probably overdoing it), we could even allow a name for the argument to `Synapses`, i.e.  `calc_multi_synapse_index` (or whatever name) could not only be `True`, but also something like `'my_synapse_index'`.
Oh, just another idea, how about `nth_synapse`? Ok, maybe not great, either...

thesamovar:
I think switch off the option to have multiple synapses by default (and raise an error if you attempt to create them). It's presumably a fairly rare thing to do, right? And then we could have the ``k`` symbol not there by default. I'd still stick with ``k`` though because it follows naturally after ``i`` and ``j``, so that you index a synapse fully by ``(i, j)`` or ``(i, j, k)``.

I think that probably all our special variables, like i, j, k, xi ought to raise warnings/errors if they are shadowed by/shadowing user variables.

mstimberg:
> I think switch off the option to have multiple synapses by default (and raise an error if you attempt to create them).

Just quickly: this would mean for each synapse to go through all the existing synapses for every connect and see whether the source-target pair already exist, I think this would be prohibitively expensive for our standard unsorted/sparse data structure, no? 

Also, wouldn't it be a bit annoying and unnecessarily verbose if you do `synapses.connect(..., n=2)` and it complains that you did not switch multiple synapses on?

thesamovar:
In that case I'm not sure what the best behaviour is. I feel the best way would be to create it only when we need it, but how bad is the special casing in this case? Perhaps it doesn't need to be special cased but we could make that a general feature? For example, maybe it's useful for ``i``, and ``j`` as well? (Although I guess we've already created these so it's irrelevant in this case.)

I wouldn't be totally against having a method that creates ``k`` (or any other name). For example, ``syn.create_multisynapse_index('k')``.

mstimberg:
I agree that a general "on-demand" feature would be nice but I think it is a bit out-of-scope for now. In principle a solution could be as simple as including a subexpression `k = calc_synapse_index() : integer`, but we'd somehow have to give a reference to the `Synapses` object to the `calc_synapse_index` function, and this would have to work quite differently for runtime and standalone and the caching for effective computation would not be trivial and so on...

Just special casing it would be a bit cumbersome I think, we'd have to check at some quite low-level of the codegeneration system for the presence of `k` (because it could be used anywhere, in state variable assignments, in `pre` codes, in regular operations, ...), then calculate it and put a reference to the calculated value into the code objects namespace. I'd rather not do this.

Having "on-demand" variables in general sounds like a nice concept, but I'm not sure how useful it will be in practice, apart from this case. I can't think of many variables we provide that are not used in most simulations. Probably `N_incoming` and `N_outgoing`, but those are only stored once per neuron, not once per synapse. We don't actually store `i` and `j` per synapse explicitly, we only store the pre- and postsynaptic indices (which are not necessarily the same because of subgroups).

> I wouldn't be totally against having a method that creates k (or any other name). For example, syn.create_multisynapse_index('k').

I think the disadvantage of a method is that it is possible to call it at the wrong time. E.g. if it means that the "multisynapse index" will be re-calculated after every `connect`, what if you call it after the last `connect`? Or if it means that it calculates it at the point where you call it, then what if you call it before a `connect`? We could of course raise an error in this case, but it adds special handling for this in multiple places. I'd prefer to have it as an option in the `Synapses` constructor which then would trigger a calculation of it for every `connect`, so it cannot go wrong. And I'd say let `syn.w[:, :, 0]` fail with an error if it had not been switched on (even on runtime, where we could calculate it on the fly), this way the user will be explicitly notified what to do and we have less different behaviour between runtime and standalone.

What is a bit annoying in that case is that using e.g. `syn.w['k == 0']` will raise a general error about unknown `k` without a specific note about what to do. Maybe we could have a dummy `Variable` object that stands as a placeholder for something that the user might expect to work and that just contains a helpful error message? This might also be useful for stuff like incorrect function names, say `asin` instead of `arcsyn` or `normal` instead of `randn`).

thesamovar:
OK, let's do your version. As for syntax, I think having ``multisynapse_index='k'`` as a keyword argument to Synapses (with default value ``None``) would be good?

mstimberg:
> As for syntax, I think having multisynapse_index='k' as a keyword argument to Synapses (with default value None) would be good?

In general I like this syntax, the only disadvantage is that we can't give an error message if the user uses a variable that they expect to be the "multisynapse index", because there's not a single name that stands for it. Or should we just assume it is `k`? What I mean is if the user does this:
```Python
syn = Synapses(source, target, 'w : 1')
syn.connect(True, n=2)
syn.w = 'k'
```
raise an error that suggests to use the `multisynapse_index` argument? And what should we do about
```Python
syn.w[:, :, 1] = 1
```
I think I'd prefer to raise an error and refer to the `multisynapse_index` argument instead of making it work on runtime and fail on standalone.

thesamovar:
I think it's OK if we don't handle the case where the user expects 'k' to mean something because we never say that 'k' has a meaning (with this syntax). We could anyway do your idea of helping users out by catching some of the common mistakes in variable names, of course, but it's no longer so essential.

Agreed on ``syn.w[:,:,1]=1``, raise an error and say they have to declare a multisynapse index.

mstimberg:
Ok, I think this should now include everything we discussed. I don't know why the testing status is not reported correctly, the very last commit was just about documentation (`[ci skip]`) and the previous commit passed all tests. Ready to merge from my side.

brian2/codegen/codeobject.py
```diff
@@ -209,7 +209,7 @@ def create_runner_codeobj(group, code, template_name,
         Whether to check units in the statement. Defaults to ``True``.
     needed_variables: list of str, optional
         A list of variables that are neither present in the abstract code, nor
-        in the ``USES_VARIABLES`` statement in the template. This is only
+        in the ``USES_VARIABLES`` statement in theqq template. This is only
```
thesamovar: Typo!

brian2/codegen/runtime/weave_rt/templates/synapses_create.cpp
```diff
@@ -1,5 +1,5 @@
 {% extends 'common_group.cpp' %}
-
+#include<
```
thesamovar: Something wrong here

brian2/codegen/runtime/weave_rt/templates/synapses_create_array.cpp
```diff
@@ -1,29 +1,21 @@
 {% extends 'common_group.cpp' %}
 
 {% block maincode %}
-{# USES_VARIABLES { _synaptic_pre, _synaptic_post, sources, targets
-                    N_incoming, N_outgoing, N,
+{# USES_VARIABLES { _synaptic_pre, _synaptic_post, sources, targets. N,
```
thesamovar: Another typo!

brian2/devices/cpp_standalone/templates/synapses_create.cpp
```diff
@@ -31,7 +31,7 @@
 	        _post_idx = _all_post
 	        _cond = {user-specified condition}
 	        _n = {user-specified number of synapses}
-	        _p = {user-specified probability}
+	        _p qq= {user-specified probability}
```
thesamovar: Typooooo!

brian2/codegen/codeobject.py
mstimberg: Hah indeed, was working with two computers and two keyboards, pressing a key on one that was meant for the other one happened more than once :)

thesamovar:
There are a bunch of typos that need to be fixed, but otherwise this code is looking good to me!

mstimberg:
I am almost impressed how many typos I managed to sneak in without actually breaking any code... Thanks for catching them, I hope they should be all fixed now.

mstimberg:
So I guess we can merge this, right?

thesamovar:
Yep!

Issue #405, opened 2015-02-12
Refactor the generation of code objects to get error messages early


mstimberg:
Most of the code generation machinery is run at the time of the run and this includes checking code for syntactic correctness, matching units, etc. This can lead to error messages at this late stage, with useless tracebacks (related issue: #376). Instead, we'd prefer to have errors as early as possible, whenever the incorrect code is written by the user, for example. In a few places, we've written explicit code to that effect (which also duplicates some work until #124 is implemented), but this is more of a hack, see this example from the `Thresholder` class:
```Python
        # Check the abstract code for unit mismatches (only works if the
        # namespace is already complete)
        try:
            self.update_abstract_code(level=1)
            check_code_units(self.abstract_code, self.group)
        except KeyError:
            pass
```

We should have a clean process consisting of multiple stages, where as much as possible is done as early as possible.

thesamovar:
I'm actually wondering if it's worth considering a more substantial refactoring/rewrite of the codegen system. More and more it seems more complicated than it needs to be. Not sure if it's a good idea to be doing that again though.

mstimberg:
I don't think we should do another really *substantial* refactoring, maybe except for the generator code (e.g. `make_statements` is much too long and complicated) -- this code is also quite independent from everything else. There are a few places that could be significantly simplified, though, I also feel we are passing around way too many arguments. Probably we should create an object (e.g. `CodeObject`) earlier in the process and store arguments in its attributes.

Maybe it would be a good idea to create an overview over what is currently happening (similar what you did for https://brian2.readthedocs.org/en/latest/developer/codegen.html#code-path) and see where things could be improved. For example, related to the problems I described in this issue, we should make a clearer distinction between code-generation-independent abstract preparations (variable name resolving, scalar and vector statements, ...) and the translation of code into a target language.

thesamovar:
That sounds reasonable to me.

Yeah making an object we pass around rather than huge argument lists seems like a good idea. I thought that passing arguments was better at the beginning because I was hoping things could be modular. But it seems that actually it's very difficult to make them sufficiently modular so storing them in an object seems the better approach.

thesamovar:
High priority for 2.0?

thesamovar:
It seems unlikely we'll get this done in a reasonable time for 2.0, so let's just do #376 and postpone this.

PR #588, opened 2015-10-27
[MRG] Cython performance

### LINKED ISSUE #402: Do type inference for expressions to allow to pull more expressions out of loops
To fix #400, we are no longer pulling out expressions that involve boolean variables or integers. Instead, they should be pulled out but put into a temporary variable of the correct type.

For that to work we need to:
* [ ]  add the information about the return type to functions (similar to the current unit system, I presume)
* [ ] analyze the expression. We already have this for boolean variables (`is_boolean_expression` in `brian2.parsing.expressions`, the solution for integers would look quite similar.

### LINKED ISSUE #421: Check that Cython is running efficiently
In some recent benchmarks I found that the cython backend was running only slightly faster than the numpy backend in many cases. I guess something got messed up and it is falling back to Python. We should fix this, and also maybe add some tests that check that the generated Cython code doesn't fall back to Python. I think that this can be done programmatically?

thesamovar:
Still underway...

Fixes #421 
Fixes #402


mstimberg:
Cython and standalone are now using `bool` instead of `char` for boolean variables and profit from the same performance boost with gcc's `-ffast-math` in the CUBA example as weave. There were only a small minor workarounds necessary, the biggest change is that Cython no longer declares an `int` function in Cython, instead we directly import it from C++ code. This is because it does not seem to be possible to overload a function in Cython so that we can provide a different implementation for a bool argument. I put the code in `stdint_compat.h` because we were already importing this file and I did not want to create a new header file just for the `int` function -- it's not really about `stdint` compatibility but it is related to integers at least, so...

The tests all passed on my machine with those commits only but your simplification changes introduced a few test failures, I did not look into them.

thesamovar:
That's great - will try that out shortly on MSVC and see if it helps.

Yeah, simplification has a few problems. One is with units, but I think there are some others too. None of it should be too difficult to fix but I want to have a working solution before spending time on that.

thesamovar:
More (unexpected) progress!

So it turns out that the combination of your changes, and one tiny bizarre change are enough to make MSVC fast, even if all the simplification is switched off. The tiny bizarre change is to replace ``1/`` with ``1.0/``. In my latest commits I disabled this (along with the rest of the simplification stuff) so hopefully the tests should pass now. The reason it works is that part of the loop invariant optimisations code checks that a subexpression is not an integer subexpression partly by whether or not any integer literals are found in it. If there are any, it will never be pulled out into a loop-invariant. I did this just to ensure correctness and because of laziness, but it has a nasty side-effect. When the expression ``1/taum`` appears in the generated solution of the equation, this and any expression which has it as a subexpression cannot be pulled out by the loop invariant code. However, replacing it with ``1.0/taum`` allows that to happen.

So now I'm wondering if the time has come to do some proper type inference? It shouldn't be too hard actually, we have the types of all variables, and we just need to add some type information to functions with the fallback assumption that if the return type of a function isn't specified then it's float, and we're good.

thesamovar:
I just wrote some notes on how to do this. I'm just pasting them here for my own reference, but if you want to take a look and comment that's always helpful.

```
Analyse and rewrite parts of expression keeping track of:
	Datatype (bool, int, float)
		Assign directly for numbers, variables
		Type interactions, e.g. int+float->float, a<b->bool
		Rewrite certain expressions, e.g.
			integer_literal op float_expression -> float_literal op float_expression
			boolean * (int, float) -> int(boolean) * (int, float)
	Scalar or vector valued
		Literals and scalar variables are scalar
		Stateless function applications maintain scalarity
		Operators preserve scalarity
		Pull out maximal scalar valued subexpressions
	Simplifiability
		e.g. 0*x -> 0
		Be careful with dtype, 0*float->0.0 but 0*int->0
	Boolean subexpression simplifiability
		e.g. exp(not_refractory*dt/tau)->not_refractory*exp(1*dt/tau)+(1-not_refractory)*exp(0*dt/tau)
		Worth doing if the simplified replacement expression is less complex than the original expression
	Boolean statement simplifiability
		Generate equivalent if/then statements for a whole statement based on the value of the booleans
		New node type BooleanStatementNode with multiple subexpressions corresponding to the original and
		all of the simplified expressions coming from assuming particular values of the booleans
		Mark all nodes as dirty
		This process will generate a bunch of scalar statements which may not be used if the if/then
		replacement is discarded, but this can either be addressed at the end or just not worried
		about since scalar statements are O(1) to evaluate.
		
Process:
- Work on a group of ordered scalar and vector statements, all scalars evaluated first
- Scalar statements will be used as simplification assumptions
- Proceed through vector statements fully analysing and simplifying one by one
- First pass, generate AST with additional node data on datatype, scalarity
- Mark all nodes as dirty
- Attempt all dirty node rewrites, if a rewrite is done mark it as dirty, and all supernodes
```

This is more or less the process we're doing at the moment for loop invariant optimisation, but it's more explicitly worked out above, and the addition of type inference allows us to do some loop invariants we were not handling before (anything with integer literals for example), and we can add the information necessary for the if/then replacement for expressions with booleans.

I think this shouldn't be too much work to implement now that I've worked out the plan, and it will be more coherent than the current LIO code I hope.

mstimberg:
I did not look through your plan in detail, but I'm wondering whether it would not make sense to integrate all this in a more principled way, somewhat related to #405. I have the impression that we are currently duplicating a bit of work (and code), e.g. the general rendering for the languages, the determination of whether something is a boolean expression, of the value of expressions and of the unit of an expression. All of those are somewhat doing the same thing, going through the AST. Type inferencing would be very similar again. It would be nice if we had a single pass going from an abstract code block + variables to annotated abstract code which has the information about types, etc. What I mean is that we'd have one single point in the code where we go from:
```Python
x = not(b)  # x is a temporary variable
y = int(x)*y + int(not(x))*z
```
and `variables = {'y': ..., 'z': ..., 'b': ...}`
to a list of scalar and vector `Statement` objects (which would pull things out of the loop, do the two values for the boolean expression, etc.) and a new `variables` dictionary which now also includes `x: AuxiliaryVariable` with the information that `x` is boolean.

Also (but this would be an even bigger change, I guess), we are using strings in Python syntax as the main representation of code lines, but we also convert them to sympy for some manipulations and to AST objects for others. We could in principle only use the sympy representation and replace the current rendering by sympy printers

thesamovar:
Yes, it definitely would make sense to integrate this into #405. So, the question is what do we do given that we want to get Brian 2 out reasonably soon? We previously ruled out doing a major refactoring of codegen for 2.0 - do we want to revisit that? I see three options:

1. We do nothing beyond what's already in this branch. Performance will not be ideal for Windows, but that's not the end of the world. We put these issues as high priority for post-2.0.
2. I try to hack something together that will do what I said above but not necessarily in the most ideal way possible, and we will do it better for say Brian 2.1.
3. We try to do this and #405 for 2.0. I think the only way we can do this in a reasonable timeframe is probably to get together and do say a 2 day hackathon (could be in London or Paris).

I'm happy to try out (2) but can certainly be persuaded towards any of these options.

I would recommend against integrating more tightly with sympy given all the problems it causes us. If anything, I'd rather standardise on having our own AST-based representation than either strings or sympy expressions.

mstimberg:
I'd say either (1) or (2), and do a post-2.0 (3) :)


> I would recommend against integrating more tightly with sympy given all the problems it causes us. If anything, I'd rather standardise on having our own AST-based representation than either strings or sympy expressions.

I agree that sympy hasn't been very stable, but the same is true for the AST, it does change ever so slightly with each Python release. But the changes have always been small, so it is probably more reliable in general. Either way, a change like this is not something that I would consider any time soon.

thesamovar:
Just added some code that takes a Python AST representation of an expression and adds dtype information to it. It does so by using the ``NodeRenderer`` approach, but rather than returning a string representation it just returns the original AST nodes modified so that it has a new ``dtype`` attribute which can be one of ``boolean``, ``integer`` or ``float``. The only thing it doesn't handle at the moment is default constants and functions, e.g. ``abs(integer)`` will current have dtype ``float``. We could easily have a system where functions are assumed to return floats unless they have a ``returns_dtype`` attribute like the returns units system.

This wasn't too difficult to code, around 150 lines. Next I'll add scalarity information to the AST which I expect won't take long. Based on this, I think this approach can work without too much work. I think I'll go ahead with it, and at some point in the future hopefully we can use this as part of the refactoring.

thesamovar:
OK, I added the scalarity stuff and it was indeed simple. I also made a new ``brian2.codegen.optimisation`` module which will eventually replace all the simplification and LIO code. For the moment, it can only handle simplifications of an expression with multiplications by 0 or 1. However, it correctly handles all the dtypes involved, e.g. if x is a float and a an integer, then ``0*x`` gets simplified to ``0.0``, ``1.0*a`` is left as ``1.0*a`` while ``1*a`` is reduced to 1*a, etc. This takes about 30 lines of code compared to about 10 lines in the previous approach, but the previous simplification code was only 90 lines total, so it shouldn't bloat too much. Also, I think when I get round to adding the LIO stuff it will actually be shorter with the new approach than the old approach.

Any thoughts on the approach?

thesamovar:
Copying previous commit message here because I wrote some notes:

Now have LIO optimisations using Brian AST structure. Integrated with Brian. Works but is slower than before because I haven't implemented the boolean simplifications yet.

Some issues that were raised:
1. No idea if it'll work in Python 3 yet, this will probably cause a test failure.
2. Sometimes make_statements is called with an incomplete/inaccurate set of Variables. This is done when checking units. Maybe this is fine? To work around it, I don't do optimisations when calling from check units.
3. Not sure if external constants are being handled corectly yet: I'm assuming for the moment that they're floats, but we probably don't want to assume that in general.
4. Sometimes functions appear as Variables in variables rather than Functions (maybe also from check units?)

thesamovar:
Also, with this change the code to pull out loop invariants is now just 40 lines, so I think that validates this approach.

thesamovar:
OK there's still some work to do, but I think the basics are done. Could you give it a go on the CUBA example?

thesamovar:
OK one appveyor test failed but I think it's an AppVeyor problem not a problem with the code?

thesamovar:
Latest commit adds the ability to collect and cancel terms in the most efficient way. There are still some bugs in the logic though: the following expression from spatialneuron is simplified to 0 incorrectly: ``diameter / ((4 * Ri) * gtot__private)``. Apart from that though, it seems to be working well. On my Win/MSVC machine CUBA now runs super fast, and looking at the code shows that it is almost maximally simplified.

thesamovar:
OK, found the bug and now all tests are passing on my machine. I think it might be good to go. I do need to write some comprehensive tests though, but the fact that all our existing examples are going through it OK is a fairly good sign that it's working. The only thing I'm concerned about is handling of integers.

thesamovar:
And just to give an indication of the difference it makes. On my laptop with weave, switching on loop invariant optimisations and simplifications reduces the neurongroup stateupdater time from 7.2s to .55s, and on standalone from 6.2s to .17s. Well worth it I think.

I also realise I need to implement this for Cython as well as weave.

thesamovar:
Latest commit implements Cython support which is for the CUBA example now as fast as weave. Will run full speed tests. One thing that we might want to discuss is the type inference for functions. At the moment I'm just assuming all functions return floats, which is obviously problematic. How about we use the same system as for units?

mstimberg:
Great! Did not look at your code in detail yet, though.

> At the moment I'm just assuming all functions return floats, which is obviously problematic. How about we use the same system as for units?

Yes, even more so since any kind of unit implies floating point values, so if the return type is not a unit, we can use the same mechanism to define whether it is float/int/bool. Not sure if it is very nice if we keep the `@check_units` name, though. But if we use it, then we could either use `@check_units(result=int)` or `@check_units(result='int')` (and the same for bool). I think the former looks more consistent with `@check_units(result=volt)` etc. The question is if we really want to *check* it, as the name implies, or rather only annotate the function.
Thinking about it, I'd be fine with a simple `return_type` attribute of the `Function` object as well (given as a keyword argument to the initializer). Actually, maybe it is better to not use `@check_units` for it, since we were considering to take the unit system out into a separate package. This is fine for all the functions provided by Brian itself, but for user-defined functions we'd need something more straightforward. Maybe simply a `@returns_integer`/ `@returns_boolean` decorator that does nothing more than setting the attribute on the function object?

Do we also want to define the type of function arguments as well? This would be more difficult, since we want to allow for multiple argument types.

thesamovar:
It would be good to have input-dependent return types though:, e.g. abs(int)->int but abs(float)->float. Can we build that into the units system directly or do you think we should have an additional system?

mstimberg:
We have something like this already in place for units. For `foo(x)` you can specify `result=lambda x: x`, meaning that the unit of the return value is the same as the unit of the argument. Or something like `result=lambda x: x**2` for the square function, etc.

thesamovar:
That's what I mean: would that system be flexible enough to handle types as well as units?

mstimberg:
If we want to use the `@check_units` system for that, then I think we don't have to do much from the syntax side: Either one specifies a fixed unit/type for the return value, or a function that calculates the unit from the input arguments. I think in the latter case, we can use type inference on the unit expression to get the type as well? Except for the `lambda x: x` case, the type will always be float anyway if it's a functional specification, right?

thesamovar:
There are various possible behaviours for the function, e.g. ``bool``, ``int`` and ``float`` always return a fixed type, ``abs`` would return the same type as its argument, ``mod(x,y)`` returns the higher type of its two arguments (this would be the trickiest one I guess). Maybe we could have a specific argument for this actually (e.g. 'highest_type')?

By the way, there seems to be some problem with appveyor to do with conda. Any ideas? Will it maybe fix it if I merge master into this branch? (I did that quite recently though.)

mstimberg:
The "higher type" could be implemented already with `lambda x, y: x + y` -- type inference would use the higher type of the two arguments and for the unit it would not make a difference if `x` and `y` have the same unit. But of course this would not the clearest way of expressing it.

I don't quite know what's going on with anaconda there, there seems to be some packaging issue for the anaconda-client on win-32/Python 3.5. I expect this will be solved soon on their side.

thesamovar:
Ah, yes, good idea. Although, at the moment I do the type inference stuff not with actual types, but with general categories as strings ``'boolean'``, ``'integer'`` and ``'float'``. The reason for this is that this is what we allow for in Brian (you can't specify the type in a more detailed way than this in equations at least), and it simplifies the logic a lot. That's actually a design decision though: do we want to stick with that? From my point of view, it's good enough. We do allow for giving actual dtypes to Variable objects though, so maybe we ought to allow for more flexibility?

mstimberg:
I'm not sure I understood completely, but if it is about only doing a basic decision between float, int, bool instead of, say,  `mod(float32, float32)` returns `float32` whereas `mod(float64, float32)` returns `float64` then I agree that the basic decision (i.e. always use `float64` or the standard scalar type or something like that) is good enough. Apart from the fact that our dtype setting does not quite work anyway (#417), I think we can't really preserve types on that level with all the re-arranging of terms etc. we do.

thesamovar:
Here's a plot of the CUBA fixed connectivity speed test performance for numpy, weave and cython on the current version of this branch:

figure_1

Cython seems to have some strangely high fixed cost which is making it slower than numpy or weave for small examples but disappears for large ones. Need to investigate why that is (It doesn't show up on any of the other speed tests).

mstimberg:
On Linux this doesn't quite look the same, there's no difference between weave and Cython:


speed_test

It simply seems that weave is slower on my machine, though. I remember from my performance tests that performance increased quite a bit with longer runs, though -- maybe my processor is not using the full speed for such short runs? Not quite sure why this would affect weave/cython more than numpy, though...

thesamovar:
Nor why it would affect the CUBA example but not the others...

mstimberg:
I just ran it for 10s instead of 1s -- it looks the same, run times are all multiplied by 10.

thesamovar:
Do you see numpy being much faster than weave/cython for other examples or just this one? For me, it's just the CUBA example with Cython.

mstimberg:
I did not run any other speed tests on this branch yet, I'll start a couple of them now.

mstimberg:
So indeed that seems to be a common pattern on my machine, here a few other tests (not sure whether copy&pasting into this PR is the best way, but well...):
speed_test_hh

speed_test_synapses_sparse
 
speed_test_synapses_dense

thesamovar:
OK so for you it appears to be weave only, and only for certain examples. For me it's Cython only, and only for the one example. WTF?

mstimberg:
Now with all the optimizations switched on it is certainly harder to find the reason for such differences. It might simply be a matter of overhead for namespace conversion and checks, though, for < 1000 neurons that is probably significant.
Looking at the generated code, I find a few things that differ between numpy and weave. For example in the synapses code (that takes 0.3s and 0.25s in numpy vs. 0.69s and 0.65s in weave for the CUBA speed test with 100 neurons on my machine), the core part is this in numpy:
```Python
    _postsynaptic_idx = _array_synapses__synaptic_post[_idx]
    not_refractory = _array_neurongroup_not_refractory[_postsynaptic_idx]
    _numpy.add.at(_array_neurongroup_ge, _postsynaptic_idx[not_refractory], we)
    #  Abstract code:  lastupdate = t
    t = _array_defaultclock_t[0]
    lastupdate = t
    _array_synapses_lastupdate[_idx] = lastupdate
```
and this in weave
```C++
        	for(int _spiking_synapse_idx=0;
        		_spiking_synapse_idx<_numspiking_synapses;
        		_spiking_synapse_idx++)
        	{
        	    // vector code
        		const int _idx = _ptr_array_synapses_pre_spiking_synapses[_spiking_synapse_idx];
        		const int _vectorisation_idx = _idx;
                const int _postsynaptic_idx = _ptr_array_synapses__synaptic_post[_idx];
                double ge = _ptr_array_neurongroup_ge[_postsynaptic_idx];
                const double t = _ptr_array_defaultclock_t[0];
                const bool not_refractory = _ptr_array_neurongroup_not_refractory[_postsynaptic_idx];
                double lastupdate;
                if(not_refractory)
                    ge += we;
                lastupdate = t;
                _ptr_array_synapses_lastupdate[_idx] = lastupdate;
                _ptr_array_neurongroup_ge[_postsynaptic_idx] = ge;
        	}
```
Not sure whether this good explain this kind of difference, but we access `ge` and `we` and write back `ge` for neurons that are refractory in weave, whereas we don't in numpy, also we only access `_ptr_array_defaultclock_t[0]` once (but this the compiler should take care of I think).

thesamovar:
Another weird datapoint. On my laptop:

figure_1

thesamovar:
Hey could you try running it again with ``run_twice=True`` instead of ``run_twice=False``, just in case for some weird reason this is relevant (it really shouldn't be).

mstimberg:
Nope, did not change anything.

thesamovar:
OK well I still haven't worked out exactly what is going on, but here's a thought. We earlier had a problem where it depended on the version of Cython that we were using. Maybe even with these various improvements, that's still the issue? I think it might be that in Cython 0.23 there is a much higher call overhead (i.e. each Cython call has a fixed and relatively high overhead).

thesamovar:
OK, so I added some features to enable plotting the time taken by each code object. On the graph below (with Cython 0.23 on my desktop), I'm showing all code objects that take more than 15% of the total computation time:

figure_1

You can clearly see that the fixed costs for Cython (red) are much higher than the fixed costs for weave, and this pretty much accounts for the difference between them because the growth rate is about the same. I think we don't see this as much on the other plots simply because the CUBA example has more code objects. (The curve marked 'Overheads' is all the time spent not in a code object and as expected is roughly flat.)

So now the question is: has something changed between Cython 0.21 and 0.23 that hugely increases the costs of going into and out of Cython. If so, what is it and can we work around it?

thesamovar:
OK, I've verified that it's something to do with versions of packages, but not entirely straightforward. I updated both my laptop and my desktop to the latest versions of everything on conda. This downgraded numpy to 1.9 but then I reupgraded that to 1.10 to make Cython work. Now both have the same behaviour (cython time>numpy time>weave time for small n).

However, it's not just Cython version because downgrading Cython to 0.21 doesn't speed it up again. So another package is to blame, but what?

thesamovar:
Looks like numpy 1.10 is to blame. Using Cython 0.23 for the CUBA example n=10, with numpy 1.9 it takes 2.7s and with numpy 1.10 it takes 5.1s. The times for the numpy and weave codegen targets are the same (approx 3.3 and 2.8 respectively).

Any ideas as to what's causing the problem here?

thesamovar:
OK I think I know what the problem is. In numpy 1.10 they added a new distutils patch that changes the way the compiler is called. You remember we had to work around this in #572? I suspect that what is happening is that somehow the compilation flags are being lost and so it's compiling it without the optimisation flags, and hence it's slow. I will investigate if we can work around this problem again, but I guess it's something we're going to have to keep an eye on.

mstimberg:
But if this were the case, wouldn't it affect weave in the same way?

thesamovar:
Maybe they use distutils in a different way? I'll look into that.

thesamovar:
So you were right they call the compiler with the same arguments, which means I'm still mystified about what the cause of the difference in speed could be.

I can't reproduce it with ``cython.inline`` so that might give us a hint as to the speed difference (whatever they are doing doesn't seem to generate the problem).

I had a little look at the generated C++ code when using numpy 1.9 or 1.10 and at least based on a quick look they both seemed fine: no Python stuff had turned up in the inner loops for 1.10.

It could be the case that something about dtypes or strides has changed. We pass the namespace a dictionary to the generated cython function, and then create views on that data. Could it be that if it doesn't match our expectations then it is making copies of the data each iteration or something like that? I feel like that's unlikely because if it was then writes to that copied array wouldn't do anything and we'd see test failures. I'm running out of ideas though.

The following code reproduces the problem (on my laptop takes 0.19s on numpy 1.9 and 1.2s on numpy 1.10):

```
import cython
from pylab import *
import time
#import Cython.Build.Inline
from brian2.codegen.runtime.cython_rt.extension_manager import cython_extension_manager
N = 100
x = zeros(N)
ns = {'x': x}
gns = {}
# f = cython.inline('''
#     # test5
#     def f(x):
#         for i in range(len(x)):
#             x[i] += (x[i]);
#     return f
#     ''', locals=gns, extra_compile_args=['/Ox', '/fp:fast', '/arch:avx2'])
code = '''
#cython: boundscheck=False
#cython: wraparound=False
#cython: cdivision=True
#cython: infer_types=True

import numpy as _numpy
cimport numpy as _numpy

cpdef main(_namespace):
    # test
    cdef int i
    cdef _numpy.ndarray[double, ndim=1, mode='c'] x1 = _namespace['x'].view(dtype=_numpy.float64)
    cdef double * x = <double *> x1.data
    for i in range(len(x)):
        x[i] += x[i]
'''.replace('len(x)', str(N))
m=cython_extension_manager.create_extension(code, extra_compile_args=['/Ox', '/fp:fast', '/arch:avx2'])
ns={'x':x}
start = time.time()
for _ in xrange(100000):
    #f(x)
    m.main(ns)
print time.time()-start
```

thesamovar:
Actually you could run this on linux without the extra compile args bit - could you check to see if there is a numpy 1.9 versus 1.10 speed difference?

thesamovar:
Aha! Getting somewhere. This is the offending line:

```
cdef _numpy.ndarray[double, ndim=1, mode='c'] x1 = _namespace['x'].view(dtype=_numpy.float64)
```

If you comment out everything except that line, it still takes ages to run, but if you comment that line out it runs super fast.

thesamovar:
Right, and this works fast too:

```
cdef _numpy.ndarray[double, ndim=1, mode='c'] x1 = _namespace['x']#.view(dtype=_numpy.float64)
```

So something about ``ndarray.view`` must have changed in numpy 1.10, and that's what's causing the problem.

thesamovar:
Confirmed, the following code:

```
x = zeros(100)
%timeit x.view(dtype=float64)
```

Returns 11 microseconds per iteration on numpy 1.10 and 0.8 microsecond per iteration on numpy 1.9. This isn't mentioned in the changelog for numpy as far as I can tell.

So this leads to a surprisingly simple fix I think (we'll see if the tests pass). Simply deleting the ``.view`` method call makes Cython now faster than numpy and weave on the CUBA test, and so far the tests are passing on my machine. I don't know why it was there originally, but maybe it's needed on some older versions of Cython? Anyway, seems not now. Hopefully tomorrow I'll find that the problem is solved...

mstimberg:
Great, thanks for hunting this down. I can reproduce the issue with your script on Linux. I think it is fixed by numpy/numpy#6562 that will be included in 0.10.2 (which by the way should also make the `compiler_cxx` attribute workaround obsolete). Maybe that release will also improve things for weave?
Anyway, I don't quite remember why we needed the `.view` for Cython, my guess would have been the `boolean` --> `uint8` stuff, but it seems still to be working after your changes, so... I will run the long Cython test suite and the examples on this branch to make sure we don't have any regression.

thesamovar:
Ah OK - they're on to it already - good. Still, if the long tests pass I'm happy to just remove the view calls anyway. I think you're probably right about the boolean uint thing. If we can get rid of the ``compiler_cxx`` thing once 0.10.2 is out should we actually remove that code?

thesamovar:
Well that's interesting. I just ran the full speed tests and now Cython is clearly the fastest option (better than weave or numpy): in every test and for all N (including small N). Getting rid of the view call made a huge difference. Can you run the speed tests on your machine and verify that you get the same thing now?

thesamovar:
Also, what do you think about testing? Is it worth writing more extensive tests of the optimiser? For example, we could do something like the test that generates random expressions, but use a mix of bool, int, float, scalar, vector variables. I sort of feel like we ought to, but on the other hand I'm pretty confident now that it's working and we already have pretty extensive tests that are likely to cover most realistic use cases (and of course, good coverage because the optimiser is used in all the tests). What do you think?

mstimberg:
I can confirm that Cython is now faster in every test, also the long tests and the examples ran fine (apart from a small bug that I introduced in the `optimize_before_run` branch, I added a fix for that). There does not seem to be a similar change that we can do for weave, right? Maybe it will get faster with numpy 0.10.2, though.

About the testing: I don' think we need many more tests, but maybe test a few expected optimizations specifically (not randomly generated ones)? In general the optimizer code is not very easy to read as it is, maybe a bit of commenting/restructuring helps? Also, the LIO stuff in `codegen/translation.py` is now obsolete, no?

thesamovar:
Oh yes I totally didn't do any commenting of my new code. Will go correct that now, rip out the old code and write a couple of small tests. Then maybe we can finally be done with this one! When I opened this branch I thought it would be a one or two day job!

mstimberg:
Great, I'll do a more thorough review after you cleaned up the code.

thesamovar:
Oh actually I just remembered we needed a dtype system for functions like for units.

mstimberg:
Ah, right... I think I'd prefer it to be separate from the units systems (i.e. not included in the `@check_units` decorator). How about:
* we make it optional, i.e. by default we assume floating point arguments and return values
* In the `Function` class, we have `arg_types` and `result_type`, equivalent to the `arg_units` and `result_unit`. As for those, you can have a function of the argument types as the value for `return_type`. For things like `abs`, we can use `lambda arg1, arg2: np.result_type(a, b)`
* We create a `@check_type` decorator that allows the specification of non-float types for user functions

thesamovar:
That sounds great. Will do that.

thesamovar:
There's still one question we need to answer about the type system though: do we work with broad classes (bool/int/float) or dtypes (e.g. int32 different to int64).

If the former, then we probably shouldn't specify the ``result_type`` by ``lambda arg1, arg2: np.result_type(a, b)`` because this would assume a and b were dtypes, not strings representing broad classes. I suppose we could use representatives, but this gives the misleading impression that we'll respect the precise dtypes you specify here. But we could have a ``highest_type`` utility function that could work with strings.

If we do want to respect exact dtypes then we need to rewrite the optimisation logic substantially. This is potentially worth doing, but might be rather hard. Perhaps we can use numpy to help in this though?

thesamovar:
Here's a thought. We could specify either with strings or with dtypes. We try strings first, if that raises an error we try with dtypes. If dtypes works but strings doesn't, we log a warning saying that currently exact dtypes aren't respected, only broad categories, but like that we leave the option open to improve that in the future. We get forward compatibility, use the simplified system for the moment, but allow ourselves the option to switch to the more detailed system in the future.

mstimberg:
Sounds good to me. I think broad categories are good enough but as you said it might be useful to switch to another system in the future.

thesamovar:
OK I added this. You do something like this:

```
@declare_types(a='integer', result='highest')
def f(a, b):
    return a*b
```

The decorator works in either order with ``check_units``, but ``implementation`` has to be the last one applied because of the way it turns a function into a Function. What do you think of the solution I developed for allowing ``check_units`` and ``declare_types`` to be used interchangeably? It should be extensible to an arbitrary number of annotating decorators. Maybe we should redefine ``implementation`` to use this approach too? Then decorators could be used in any order.

On my machine, tests were passing (only checked codegen-independent, numpy and the first few in weave so far). The optimiser is using the type declaration system and it seems to be working, so I think this is done except for adding one or two more tests. What do you think?

thesamovar:
OK I added some more tests and I think this is ready to merge now as far as I'm concerned. We can hold off to improve the decorators so they can be called in any order if you want, but I think it's not high priority and I added a note in the documentation for ``implementation`` saying that it has to be the uppermost/last-applied decorator.

thesamovar:
Just realised there is a potential bug with the LIO code. An expression like ``y=rand()+rand()`` might get turned into ``x=rand(); y=x+x``. I think it won't actually happen but it needs to be checked.

mstimberg:
IIRC, the previous LIO code took care of this. This was why we introduced the `stateless` attribute for functions.

thesamovar:
I explained unclearly. It correctly knows that ``rand()`` is stateless and will consider it a non-scalar, so actually the LIO code itself will never pull it out. However, it might get simplified somewhere because I wrote some code where expressions are cached based on their string value, i.e. if you write ``rand()`` and that gets converted into another expression then the next time it sees a ``rand()`` it will get converted to the same expression. However, I think that I only did this for expressions that should be pulled out as LIO constants, and any expression with a ``rand()`` will never be pulled out, making this OK. But possibly I used it somewhere else.

mstimberg:
Ah ok, I will have a look at this when I review the PR.

thesamovar:
Thanks!

mstimberg:
It seems to be fine in most cases, but `v = rand() - rand()` *does* get incorrectly simplified to `v = 0.0`.

mstimberg:
I fixed that issue by transforming `rand() - rand()` into `rand(_stateful_1) - rand(_stateful_2)` during the cancellation process. I'll need more time to review this, though, it's not super straightforward.

I think much of our code generation needs a major refactoring post 2.0. IMO, there should be a clear processing pipeline were we parse user-provided statements once (say into our own annotated AST), then run transformations on this tree to simplify things (possibly using a `NodeTransformer`) and finally a single run of a `NodeVisitor` to render the final language-specific output. Currently, it is really difficult to follow what is going on.

thesamovar:
Maybe it would be better to add a ``stateless`` attribute to the Brian AST node type and only simplify terms which are stateless? Possibly this attribute could be useful in other places too.

Agree on the major refactoring.

mstimberg:
> Maybe it would be better to add a stateless attribute to the Brian AST node type and only simplify terms which are stateless?

Good idea, I reverted the previous change and implemented it this way. It is slightly less elegant than I hoped because nodes are generated in various places and we can therefore not guarantee that they have a stateless attribute, I'm therefore using something like `getattr(node, 'stateless', True)` everywhere.

thesamovar:
Huh, that's strange. It should always be possible to make the ``stateless`` attribute always be there on the AST node if you cover all the places where I set, e.g., ``node.scalar``.

mstimberg:
Right, I changed this now. I don't find it very elegant either because code outside of `BrianASTRenderer` has to know what attributes to create for newly created nodes. The same file has plenty of newly created `Num` nodes in the `collect` function where we don't need to create any new attributes. But then again, let's leave it for a major refactoring :)

thesamovar:
Not elegant but maybe safer because like that you quickly get errors if something isn't setting the attribute, which means you're confident that if you don't get any errors then the attribute is being computed correctly. The default value allows for problems to silently propagate making it harder to debug. But I agree that the code is very messy at the moment and could really do with a refactoring post-2.0.

Also looks like we might have a Mac build error. :/

mstimberg:
> Also looks like we might have a Mac build error. :/

We get this every once in a while, unfortunately :-( It's a segmentation fault *in the compiler*, i.e. not in our compiled code. I think this has to be considered a bug in the compiler by, so I did not care much about it so far -- hopefully it is something that gets fixed in the future. I restarted the build, let's see whether it works.

mstimberg:
I removed some code redundancy in a few places and removed some unused code (e.g. `create_assumptions_namespace` always only created the default namespace and `render_expr_with_additional_assumptions` was never called). I also added a few tests and now I managed to convince myself that it is doing something reasonable :) When looking at the test coverage I noticed that some lines in `ArithmeticSimplifier.render_BinOp` are never reached, they are all like this:
```Python
                # only simplify this if the type wouldn't be cast by the operation
                if dtype_hierarchy[right.dtype] <= dtype_hierarchy[left.dtype]:
                    return left
```
I can delete all lines like this, leaving only the Multiplication with 0 or 1 and the division 0/x in that file (i.e. delete lines 226-243 in the current version of the file), all the tests still pass. Are these things (addition/subtraction of 0, division by 1) maybe already handled as a special case by the `collect` function? If yes, we should remove them, if not I'd be grateful if you could add tests to `test_codegen.py` that cover this code.

Either way, I'll run the long tests and the examples but pending that, I'm happy to merge it (in case you agree with my latest changes of course).

thesamovar:
I would be happier to leave those lines in for safety purposes - I'll add some tests that cover them. I think you're right that at the moment ``collect`` will do everything that might call these but possibly we will use this code in other ways in the future (or make ``collect`` optional, or something else we can't foresee) and I'd be happier if it were left in a safe state with no assumption about what something else might do.

mstimberg:
Oh, actually I am not so sure anymore whether we should merge this already. For simple simulations like the CUBA example the additional computation slows down things considerably (as you mentioned in the Brian1 to 2 comparison, I think). On my machine, simulations take 4.8s (all included) for master, but 5.8s on this branch. Given that this was mostly about optimizing the performance, this is obviously not great...
I had a quick look in the profiler and the main time is spent in these lines (`ArithmeticSimplifier.render_node`):
```Python
        node = super(ArithmeticSimplifier, self).render_node(node)
        # can't evaluate vector expressions, so abandon in this case
        if not node.scalar:
            return node
        # try fully evaluating using assumptions
        expr = NodeRenderer().render_node(node)
        val, evaluated = evaluate_expr(expr, self.assumptions_ns)
```
Two quick thoughts:
* We are parsing the code, then rendering it to a string, then Python parses it for `eval` -- could we not potentially take the AST and compile it, skipping the rendering/parsing in between?
* I counted the number of times that `evaluate_expr` could successfully evaluate the expression in the "assumption namespace", it was about 10%. Can we maybe fail earlier, e.g. first extract all the identifiers and check them against the namespace dictionary. The time saved for the `eval` in 90% of the times might be worth it. Or even simpler: the `render_node` + `eval` is not necessary at all if the node is a `Name`/`NameConstant` or a `Num`, right?

thesamovar:
A few things to discuss it seems!

Firstly, I found a potential bug in the collect code. If you had something like ``1.0*int`` then it would be simplified to just ``int`` (which would be a mistake). That's now corrected. Maybe check the logic of my changes here, I'm not entirely certain because I was in a bit of a hurry when writing it. Actually, you had a test that checked for (I think) the incorrect behaviour here, so I changed that.

Second, I added some new tests that cover the code that you said wasn't being covered before. It still counts when using integer expressions because ``collect`` won't handle these cases. So e.g. ``int/1``, ``int+0`` and so on can all be simplified with those lines to just ``int`` (and that's added).

For the performance, you're right that it's very slow and maybe indeed we should address that here before merging. Your suggestions sound good to me. There are probably many opportunities to avoid computations here: I didn't think at all about optimising this code yet, only making sure that it's correct.

Another option is that if we are doing some joblib cacheing here maybe we don't need to worry if the one off costs are high? Still though, let's try to improve it a bit more before resorting to joblib.

Do you want to take a look at optimising the code or shall I?

mstimberg:
Right, `1.0*int` should indeed not be simplified. I am not sure whether my test should be considered incorrect, since both `v = 1.0 * int(...)` and `v = int(...)` are correct in general, but it is probably a good idea to always transform it into something that is explicit about the type.
The test coverage is now close to perfect, it is mostly missing the situation where it should not be simplified (i.e. the `1.0 * int` etc.), I added those. It still does not go into all the branches of those comparisons, but I am too lazy to figure this out in any more detail :)

For the performance, let's try to optimize first by short-cutting where possible. I'd prefer to avoid joblib caching here if it is not necessary. I'll have a look tomorrow morning.

mstimberg:
I fixed an error that I previously introduced (the expression `rand() - rand()` did correctly keep a `rand()` call, but "simplified" the expression incorrectly to `0.0 - rand()`...). I also did two small changes that brought down the optimisation time by half (the change to not evaluate numbers and constants that I mentioned previously, and a faster `brian_dtype_from_dtype` function). But then I realized that the rendering functions were simply being called way too often.
I am not 100% sure what is going on but I think it was something along the lines that `u + x + y + z` was recursively processed, but then `x + y + z` was recursively processed as well, and so on. So the time was growing quadratically with the complexity of the expression instead of linearly. A small change, similar to what you did for the `collect` calls, at the beginning of `ArithmeticSimplifier.render_node` drastically reduced the run time so that the optimisation step no longer has any measurable effect on the performance:
```Python
        if not hasattr(node, 'simplified'):
            node = super(ArithmeticSimplifier, self).render_node(node)
            node.simplified = True
```
All the tests still pass and I did not find any difference in result to the previous version, so...
I first thought it would be a tradeoff that we gain speed but we do less simplification, e.g. the following complicated way of writing `x = x`:
```Python
x = ((x - (1.0*x)) + x) + (0/x)
```
is only reduced to `x = (x + x) - x` but then I tested it with the previous version and it didn't do any further simplifcation either (which I think is completely ok since this is a very contrived example. If we'd be worried about something like this happening in real use cases we could call the simplification process on the top level repeatedly until there are no more changes to the code).
So if I am not missing something here (which would also mean that we are missing a test :) ), then this is ok this way and the performance is no longer an issue.

thesamovar:
That contrived example is brilliant. I thought it was actually doing all the stuff recursively that it ought to be doing but you found a way to defeat it! :)

I agree. This is doing probably 99% of the work we want it to, the efficiency problem has been solved - I think that's good enough for 2.0. I haven't yet looked at your changes or tested it out yet but I think this is looking like it's good to go.

thesamovar:
OK, long tests + standalone tests works fine on my machine, and code looks good. I'm happy for it to be merged if you are.

mstimberg:
I started long tests and all the example on my machine at work before I left, will have a look tomorrow morning and merge if all went well. That was quite a bit more work than expected but I am quite happy with the result, the optimisations should speed things up nicely and lead to better readable code as a nice side effect!

thesamovar:
Yeah I didn't realise we'd up writing our own mini-compiler for this issue, but indeed - well worth it.

mstimberg:
All examples passed but there was a small bug in two of the long standalone tests. A bit of a subtle issue that won't effect users normally, you'd have to set the `codegen.string_expression_target` to something else than `'numpy'` (as we do during testing) and then evaluate a subexpression in standalone mode. It was an easy fix, just making sure that runtime arrays had the dtype they promised to have. I don't think the bug was introduced in this branch actually. So if all the tests now run through on the servers, I'm happy to merge it.

thesamovar:
That's strange, the long standalone tests passed on my machine. Still, glad to know there is a fixed problem. :)

mstimberg:
It was probably Linux 64 Bit specific, it was a bout an `int` vs. `long` mismatch.

thesamovar:
OK all tests are passing and we're both happy with it so I'm going to finally merge this sucker.

PR #590, opened 2015-10-29
Frozen

Kwartke:
We need those small changes in order to refactor brian2cuda into a seperate project.


Kwartke:
Our CUDA version does not generate the profiler info file since most kernel calls are asynchronous, but still uses the run() method of CPPStandaloneDevice. So the file does not exists for every device.

mstimberg:
I don't see a problem with any of these changes, merging it now.

PR #592, opened 2015-11-06
[MRG] Use compiler settings for maximal speed by default

mstimberg:
I just enabled the fast compilation options in this branch, testing on Linux did not show any test failures but we don't yet have the results from appveyor. If there's no problem, then it seems we do not actually have to change anything (apart from the documentation). I included an auto-detection of the arch flag for Windows in the same way it was done in standalone when you specified `native=True`. This makes the latter obsolete, do you think this is important enough to have as a specific keyword argument to build instead of having it set simply via the compiler options? I think users will build for their own machine in 99% of the cases and if not, they probably have to set some compiler/linker settings anyway, no?

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -112,12 +112,12 @@ install:
        conda install --yes --quiet anaconda-client conda-build jinja2 pip setuptools;
     fi
   - if [[ $MINIMAL_VERSIONS == 'yes' ]]; then
-    conda create -n travis_conda --yes pip python=$PYTHON numpy==1.8.2 scipy==0.13.3 nose sphinx ipython sympy==0.7.6 jinja2==2.7 pyparsing setuptools coverage;
+    conda create -n travis_conda --yes --quiet pip python=$PYTHON numpy==1.8.2 scipy==0.13.3 nose sphinx ipython sympy==0.7.6 jinja2==2.7 pyparsing setuptools coverage;
     else
-    conda create -n travis_conda --yes pip python=$PYTHON numpy scipy nose sphinx ipython sympy pyparsing jinja2 setuptools coverage;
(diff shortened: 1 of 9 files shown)


thesamovar:
Code looks great to me: happy for this to be merged once tests pass.

mstimberg:
What do you think about getting rid of the `native` keyword argument?

thesamovar:
Yeah, now that we're computing the native architecture as the default I think it's fine to drop the native keyword.

mstimberg:
Ok, pending that all tests pass this is now ready to merge, I think. Bboth gcc and msvc now use the optimized compiler settings by default, and for msvc we use the `cpuinfo` package to make the decision about the correct `/arch` flag. Since it is now used also in runtime mode, I added the `py-cpuinfo` package to the list of requirements. It is a tiny package and it is Python-only, so I think it's easier to include it than to ask users to install it manually. We currently only make use of it on Windows, but it might be useful in the future also for Linux/OS X.
I also uploaded conda packages for `py-cpuinfo` to our `brian-team` channel, otherwise the installation via conda wouldn't work anymore. Of course in general we don't want to be responsible for providing conda packages of 3rd-party libraries but since it is Python-only, it was trivial to do here (build the conda package on your architecture with `conda skeleton pypi py-cpuinfo; conda build py-cpuinfo`, convert it to all other architectures with `conda convert -p all built_package_name.bz2` and repeat the process for other Python versions after setting `CONDA_PY` to `35`, etc. Finally, upload everything with `anaconda upload -c brian-team linux-32/* linux-64/* osx-64/* win-32/* win-64/*`).

thesamovar:
Yep, I'm happy with that.

mstimberg:
So all the tests pass (not quite sure what this appveyor thing "Waiting for status to be reported" is), this is good to merge, right?

thesamovar:
I think so, merging now.

Issue #124, opened 2013-09-20
Cache certain computations using joblib


thesamovar:
Some of the code generation stuff can end up actually taking a while. Solution: cache it!

thesamovar:
Could also use memoize (in Python 3 there is ``functools.lru_cache``). See also the way sympy uses it: https://github.com/sympy/sympy/blob/master/sympy/core/cache.py

mstimberg:
I think with all the improvements to the preparation time merged recently, maybe we should postpone this to post-2.0? A quick and dirty solution might do more harm then good...

thesamovar:
Yep!

PR #593, opened 2015-11-12
[MRG] Faster `before_run`

mstimberg:
As we discussed, instead of implementing joblib caching (#124) I first had a look at where the preparation time is spend. In this branch, I implemented a few small uncontroversial changes (e.g. `check_code_units` was called before `create_runner_codeobj`, but the latter called `check_code_units` again), but also some bigger changes that are at least worth discussing:
1. Instead of passing down the level argument down from `before_run` all the way until when namespaces are actually resolved, we get the local namespace in the `run` method and pass this namespace down. The advantage is that we don't have to get the local namespace several times, which is relatively expensive since we have to go through a lot of objects and call `isinstance` on them to see whether we are interested in them. Actually, now that I am writing this: we should probably not filter out "uninteresting" objects beforehand, because that means we will not get a warning if we have a clash between the name of such an object (say, a string) and a variable. We could rather always transmit the full namespace and then raise an error when such an object is actually used as a variable name (or a warning for a name clash). Either way, I think the change is a good one because it makes the code simpler (less `level=level+1` or `level=level+2`) and wastes less time in the mildly expensive namespace resolution. The only disadvantage is that we can't make a difference in warning messages anymore between clashes that occur with names in the local namespace vs. clashes with names in the explicitly provided `run(..., namespace={...})` namespace. But then, I don't think this leads to any confusion.
2. We no longer try to analyse certain strings (thresholds, resets, refractory condition) at creation time (where this check would only succeed if the expressions do not use external variables that are unknown at that time). This has clearer disadvantages: we don't get error messages for incorrect syntax / unit errors, etc. at creation time but only at the time of before run. The advantages: The code is clearer (no `try/except` around some checks) and we don't do things like unit checking twice. I am not 100% sure about this one, but with the new error messages that point to the creation statement, I think this is ok. It is mostly in interactive use that this could be annoying. But then, with some refactoring that better separate the various stages of code generation, we could at least deal with syntax changes directly where they occur again.
3. Another big change: Instead of two methods, a `can_integrate` function that returns a `bool` and a `__call__` function that creates the actual abstract code, state updaters now only have a single `__call__` function that will raise an `UnsupportedEquationsException` if it cannot integrate them. The reason for this is that for the more complicated state updaters (e.g. linear), the only robust way of determining whether it can integrate given equations is by actually trying it (this is what we did in `LinearStateUpdater.can_integrate`). Therefore, we actually generated the code twice for examples like CUBA, once for checking whether we can use `linear` and once for actually using it. I don't really see a disadvantage of the new approach, it is pretty straightforward. We had a feature before where you could tell Brian to use a state updater despite its claiming not to be able to integrate the given equations which we lose now, but I don't think that had any practical use.

Now, the cumulative effect of these changes is not great, but it is something. If I do a `run(0*ms)` and remove the connection code (which otherwise dominate the total runtime), then the total run time for CUBA goes down from 4.2s to 3.4s on my machine, and from 3.9s to 3.1s for COBAHH. If I use an artificial example that makes use of code generation and namespace resolution alot like this:
```Python
G = NeuronGroup(100, 'v : 1')
for x in xrange(100):
    G.v['i == x'] = 'x'
```
then the reduction is more significant, it goes down from 6.5s to 2.3s (but of course here the state updater stuff does not play a role).

Thoughts?

### DIFF
--- a/brian2/codegen/codeobject.py
+++ b/brian2/codegen/codeobject.py
@@ -255,18 +255,6 @@ def create_runner_codeobj(group, code, template_name,
     else:
         override_conditional_write = set(override_conditional_write)
 
-    if check_units:
-        for c in code.values():
-            # This is the first time that the code is parsed, catch errors
-            try:
-                check_code_units(c, group,
-                                 additional_variables=additional_variables,
-                                 level=level+1,
-                                 run_namespace=run_namespace)
-            except (SyntaxError, KeyError, ValueError) as ex:
-                error_msg = _error_msg(c, name)
-                raise ValueError(error_msg + str(ex))
-
     if codeobj_class is None:
         codeobj_class = device.code_object_class(group.codeobj_class)
     else:
@@ -326,6 +314,15 @@ def create_runner_codeobj(group, code, template_name,
                                        additional_variables=additional_variables,
                                        run_namespace=run_namespace,
                                        level=level+1))
+
+    if check_units:
+        for c in code.values():
+            # This is the first time that the code is parsed, catch errors
+            try:
+                check_units_statements(c, variables)
+            except (SyntaxError, KeyError, ValueError) as ex:
+                error_msg = _error_msg(c, name)
+                raise ValueError(error_msg + str(ex))
 
     all_variable_indices = copy.copy(group.variables.indices)
     if additional_variables is not None:
--- a/brian2/codegen/codeobject.py
+++ b/brian2/codegen/codeobject.py
@@ -261,11 +261,7 @@ def create_runner_codeobj(group, code, template_name,
         codeobj_class = device.code_object_class(codeobj_class)
 
     template = getattr(codeobj_class.templater, template_name, None)
-    if template is None:
-        codeobj_class_name = codeobj_class.class_name or codeobj_class.__name__
-        raise AttributeError(('"%s" does not provide a code generation '
-                              'template "%s"') % (codeobj_class_name,
-                                                  template_name))
+    template_variables = getattr(template, 'variables', None)
 
     all_variables = dict(group.variables)
     if additional_variables is not None:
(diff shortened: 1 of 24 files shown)


thesamovar:
This all looks pretty good to me. Some comments:

For (1) namespace clash errors, could we perhaps store alongside the namespace dict, a ``namespace_origin`` dict that has the same keys but the values are where the name came from, e.g. "local namespace", "specified in NeuronGroup call", etc. Like that, when a clash happens we can give a meaningful error message. 

(2) seems good to me, agreed that these early checks aren't so important with the new error reporting thing.

(3) is good too. In retrospect, integrating the equations twice is not very sensible! :)

I'm pretty happy for this to be merged once tests pass, but what do you think about the idea for namespace clash errors?

mstimberg:
> For (1) namespace clash errors, could we perhaps store alongside the namespace dict, a namespace_origin dict that has the same keys but the values are where the name came from, e.g. "local namespace", "specified in NeuronGroup call", etc. Like that, when a clash happens we can give a meaningful error message.

This is a good point (nothing that actually changed in this branch, though), we partly do this already, but only for clashes between variables in different namespaces, not between internal variables and variables in the namespace. What changed in this branch is that we cannot give different warnings for the following two situations anymore (which we did not do before either, though...):
```Python
# Situation 1
foo = 3
G = NeuronGroup(1, 'foo:1', threshold='foo>1')
run(0*ms)
# Situation 2
G = NeuronGroup(1, 'foo:1', threshold='foo>1')
run(0*ms, {'foo': 3})
```
But as I said, I think using the term "run namespace" in both cases should not lead to confusion. I will try to improve the error message in general, though.

> (3) is good too. In retrospect, integrating the equations twice is not very sensible! :)

Well, in the beginning we did some analysis in the first phase instead, e.g. tried to check whether the equations are linear etc. But it turned out that there are several things that can go wrong along the way, things that were to complex for sympy etc. so that the only reliable way to find out whether the non-explicit state updaters can integrate given equations is to actually do it...

mstimberg:
Ok, I pushed a change that should improve the error messages. For this lovely bit of code:
```Python
foo = 3
volt = 7
abs = 10
pi = 4
G = NeuronGroup(1, '''foo : 1
                      bar : 1''',
                namespace={'bar': 5, 'abs': 8},
                threshold='(foo + pi)*volt>abs(bar*volt)')
run(0*ms)
```
we now get these warnings:
```
WARNING  brian2.groups.group.Group.resolve.resolution_conflict: The name "volt" refers to different objects in different namespaces used for resolving names in the context of group "neurongroup". Will use the object from the units namespace with the value volt, but the name also refers to a variable in the run namespace with value 7.
WARNING  brian2.groups.group.Group.resolve.resolution_conflict: The name "pi" refers to different objects in different namespaces used for resolving names in the context of group "neurongroup". Will use the object from the constants namespace with the value 3.141592653589793, but the name also refers to a variable in the run namespace with value 4.
WARNING  brian2.groups.group.Group.resolve.resolution_conflict: "bar" is an internal variable of group "neurongroup", but also exists in the group-specific namespace with the value 5. The internal variable will be used.
WARNING  brian2.groups.group.Group.resolve.resolution_conflict: "foo" is an internal variable of group "neurongroup", but also exists in the run namespace with the value 3. The internal variable will be used.
WARNING  brian2.groups.group.Group.resolve.resolution_conflict: The name "abs" refers to different objects in different namespaces used for resolving names in the context of group "neurongroup". Will use the object from the functions namespace with the value <Function>, but the name also refers to a variable in the following namespaces: group-specific, run.
```
Maybe we could make the text more specific for the function, but I think it is already quite helpful like this.

thesamovar:
Great, those error messages are really useful. I think this is good to merge once AppVeyor finishes.

thesamovar:
Oh, it just did - merging now.

PR #594, opened 2015-11-12
[MRG] More checks for arguments of `SpikegeneratorGroup`

mstimberg:
A trivial PR, just adding checks and making sure that the type of `N` is `int` (two issues that came up on the mailing list recently)


thesamovar:
Good stuff - merging.

PR #595, opened 2015-11-16
[MRG] Don't check all objects in the local namespace for their type

mstimberg:
I mentioned this in #593 but forgot to actually change this. We are currently checking all of the objects in the local namespace for their type, and only include the objects we are generally interested in (numbers, functions, ...) in the namespace dictionary that we pass on. This is quite expensive, since the full local namespace can be big. With the small change here, we instead pass the full local namespace down to the resolve method and only do the type checks for names we are actually interested in.
This reduces the time that `before_run` takes by a very small amount for e.g. the CUBA example, but it can have a significant effect for scripts that make a use of string expressions a lot.

### DIFF
--- a/brian2/core/namespace.py
+++ b/brian2/core/namespace.py
@@ -41,20 +41,13 @@ def get_local_namespace(level):
         The locals and globals at the given depth of the stack frame.
     '''
     # Get the locals and globals from the stack frame
-    namespace = dict()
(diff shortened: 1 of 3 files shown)


thesamovar:
Happy for this to be merged once tests pass.

mstimberg:
All tests pass (still not sure why appveyor never reports the final status...), will merge now.

PR #601, opened 2015-11-27
[MRG] Treat `NotImplementedError` as a special test result

mstimberg:
This PR should be a simple one for a change :) Tests that raise `NotImplementedError` will now appear as "N" in the test results and by default not make the test suite fail (but you can call`brian2.test(..., fail_for_not_implemented=True)` to change that). Well, that's all, in principle it is very straightforward but there were a few tricky details to get right (in the long run we should just use `nose2`, but for now it is not in anaconda yet, so let's wait until this is the case).


thesamovar:
Sounds good to me: happy for this one to be merged as soon as tests pass.

Issue #446, opened 2015-04-12
Benchmark against Brian 1


thesamovar:
Seems like B2 might be slower than B1 in some important cases?

thesamovar:
This is still worth doing but at the moment it seems the most likely difference in terms of speed is that B2 has more Python overhead than B1 which makes it slower for small networks.

thesamovar:
CNS meeting update: let's not do this systematically but at least check a few examples.

thesamovar:
OK, the first plot is in, and it's not looking good for Brian 2:

![figure_1](https://cloud.githubusercontent.com/assets/2903870/11282841/655bce88-8ef9-11e5-8e7e-3c1b9dfecc73.png)

mstimberg:
That's not looking great indeed. Could you give some more details about what we are seeing here? I guess the x axis is number of neurons? And is this all with codegen target weave?

thesamovar:
Yeah it's just the same as the CUBA speed test example. The N is number of neurons, and there are 80 synapses per neuron. I haven't tried with different codegen targets yet so I don't actually know what they are using - will check that next. In the meantime, here is the same plot as above for a 10s run instead of a 1s run (showing that it's not just a startup time problem):

![figure_1](https://cloud.githubusercontent.com/assets/2903870/11285560/00538864-8f08-11e5-8364-707005e1bab0.png)

The script is in a new branch ``compare_to_brian_1``. The different times are as follows:

import brian: from the very start until ``from brian(2) import *`` has finished.
object creation: from the end of the previous step until just before run
initial run: a 1ms run
main run: a 1s/10s run
total: the whole time including all of the above

thesamovar:
So it turned out that my Brian 1 installation was pure Python with no options set, so I set Brian 2 to numpy mode and of course now it does even worse in the comparison. I also changed the script so that it only has the NeuronGroup, and same problem. Indeed, in this case for the 'main run' part there is a constant difference between them, which (since they're log plots) means that there's a constant time factor, i.e. Brian 2 is N times slower than Brian 1 for some N>1. Anyway, looks like this ought to be diagnosable at least. Maybe something to do with the way we load arrays and write them back? I think because we were more restricted in what could happen in Brian 1 it might have been more efficient.

thesamovar:
[...]

PR #602, opened 2015-11-27
[MRG] Fix various inefficiencies

mstimberg:
Ok, this one deserves slightly more discussion. It fixes a few "low-hanging fruits" for performance, hopefully it should now compare better against Brian 1. Nothing too controversial here, the changes are:
* As mentioned previously, the way the clock was accessed in `Network.run` was far from optimal. I changed it in a way that takes some logic out of the `Clock` class and puts direct access to the variables in `Network.run`. This makes the code somewhat less readable but I think it is justified here since `Clock.running` and `Clock.tick` were only ever used from within the runtime `Network.run` function, standalone had to do something else. So not having this in `Clock` but instead in `Network.run` seems like the sensible choice, instead of having very runtime specific code in the `Clock` class.
* During import, all the templates for all code generation targets were loaded into memory, which took a significant amount of time, I changed it so that they are now loaded on demand only.
* There was one left-over call where `NeuronGroup` tried to run `before_run` in its constructor, which only worked if the namespace was already complete. In line with the changes for the state updater/thresholder/resetter in the `optimize_before_run` branch we did earlier, I removed this.
* For every expression that went through code generation, we went through the variables dictionary, noted all the name translations (e.g. `acos` to `arccos`) and replaced them in the expression. We now create a dictionary of name translations when we create the `CodeGenerator` object.
* We were still using sympy's `simplify` function, which tries out all kind of simplifications and takes quite a long time. I replaced this by a few specific simplifications that for the linear equations of the CUBA example lead to basically the same result but take much less time.

Looking at the profiling output I don't see anything obvious to improve any more. The only significant improvements that I could think of would take more thought, e.g. things that optimize the whole run loop (special case a single clock or pre-calculate a completely fixed schedule, say. `clock1, clock2, clock1, clock2, ...`, instead of checking things on a per-timestep basis, merging code objects if there is nothing in between them, etc.).

If you can't think of any more obvious inefficiencies (the synapses discussion aside), then this is ready to merge from my side.

### DIFF
--- a/brian2/codegen/codeobject.py
+++ b/brian2/codegen/codeobject.py
@@ -260,7 +260,7 @@ def create_runner_codeobj(group, code, template_name,
     else:
         codeobj_class = device.code_object_class(codeobj_class)
 
-    template = getattr(codeobj_class.templater, template_name, None)
+    template = getattr(codeobj_class.templater, template_name)
     template_variables = getattr(template, 'variables', None)
 
     all_variables = dict(group.variables)
--- a/brian2/codegen/generators/base.py
+++ b/brian2/codegen/generators/base.py
@@ -2,8 +2,8 @@
 Base class for generating code in different programming languages, gives the
 methods which should be overridden to implement a new language.
 '''
-from brian2.core.preferences import prefs
 from brian2.core.variables import ArrayVariable
+from brian2.core.functions import Function
 from brian2.utils.stringtools import get_identifiers
 from brian2.utils.logger import get_logger
 from brian2.codegen.translation import make_statements
@@ -37,6 +37,13 @@ class CodeGenerator(object):
         self.device = get_device()
         self.variables = variables
         self.variable_indices = variable_indices
+        self.func_name_replacements = {}
+        for varname, var in variables.iteritems():
+            if isinstance(var, Function):
+                if codeobj_class in var.implementations:
+                    impl_name = var.implementations[codeobj_class].name
+                    if impl_name is not None:
+                        self.func_name_replacements[varname] = impl_name
         self.iterate_all = iterate_all
         self.codeobj_class = codeobj_class
         self.owner = owner
--- a/brian2/codegen/generators/cpp_generator.py
+++ b/brian2/codegen/generators/cpp_generator.py
@@ -158,11 +158,7 @@ class CPPCodeGenerator(CodeGenerator):
             return device.get_array_name(var, access_data=False)
 
     def translate_expression(self, expr):
-        for varname, var in self.variables.iteritems():
-            if isinstance(var, Function):
-                impl_name = var.implementations[self.codeobj_class].name
-                if impl_name is not None:
-                    expr = word_substitute(expr, {varname: impl_name})
+        expr = word_substitute(expr, self.func_name_replacements)
         return CPPNodeRenderer().render_expr(expr).strip()
 
     def translate_statement(self, statement):
--- a/brian2/codegen/generators/cython_generator.py
+++ b/brian2/codegen/generators/cython_generator.py
@@ -61,12 +61,7 @@ class CythonCodeGenerator(CodeGenerator):
     class_name = 'cython'
 
     def translate_expression(self, expr):
-        # numpy version
-        for varname, var in self.variables.iteritems():
-            if isinstance(var, Function):
-                impl_name = var.implementations[self.codeobj_class].name
-                if impl_name is not None:
-                    expr = word_substitute(expr, {varname: impl_name})
+        expr = word_substitute(expr, self.func_name_replacements)
         return CythonNodeRenderer().render_expr(expr, self.variables).strip()
 
     def translate_statement(self, statement):
--- a/brian2/codegen/generators/numpy_generator.py
+++ b/brian2/codegen/generators/numpy_generator.py
@@ -31,11 +31,7 @@ class NumpyCodeGenerator(CodeGenerator):
     _use_ufunc_at_vectorisation = True # allow this to be off for testing only
 
     def translate_expression(self, expr):
-        for varname, var in self.variables.iteritems():
-            if isinstance(var, Function):
-                impl_name = var.implementations[self.codeobj_class].name
-                if impl_name is not None:
-                    expr = word_substitute(expr, {varname: impl_name})
+        expr = word_substitute(expr, self.func_name_replacements)
         return NumpyNodeRenderer().render_expr(expr, self.variables).strip()
 
     def translate_statement(self, statement):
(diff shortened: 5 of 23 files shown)


thesamovar:
OK, great. In terms of performance for the numpy codegen target, Brian 2 is almost as good for low N and better for high N than Brian 1 (for the case of NeuronGroup only, no refractoriness, Euler integration forced). Here's the plot:

figure_1

And here's with refractoriness switched back on - also great.

figure_1

However, if I put ``SpikeMonitor`` into that simulation, the situation changes a lot. I guess this is because ``SpikeMonitor`` is accessing ``t`` in the old way? Perhaps it's worth considering a deeper change to the way state variables are accessed by templates in runtime? Not sure if that's the difference or not, but if it's really the case that each access to these variables is slowing things down this much then it's problematic.

figure_1

Next, here's a version with synapses included. Note that in Brian 1 we've got Connection and in Brian 2 we've got Synapses. Brian 2 is using the Cython SpikeQueue. Brian 1 is storing a weight matrix which Brian 2 is not doing, however Brian 2 is using a spike queue which B1 is not doing. Still, I think it may be the case that there are some inefficiencies there because it's uniformly slower in the second half run. You can also clearly see the N^2 construction cost hitting at large N (but let's deal with that separately).

Finally, here's NeuronGroup only but this time using exact integration:

figure_1

I'm not sure how fair this is given that with Anaconda you get an MKL-optimised matrix multiplication using all cores. On the other hand, this is the performance that a lot of users will see I guess.

Haven't looked at the code yet, but will do that later. What do you think about the ``SpikeMonitor`` and access to ``t`` issue?

Will also do more comprehensive comparisons using weave, cython, etc. and report on issue #446 but that doesn't affect whether or not this PR can merge.

thesamovar:
OK I messed that last comment up a bit. I misread the refractoriness one: actually performance is a lot worse with refractoriness on for B2 than B1. That might be specific to the numpy target but we'll need to check that. Possibly it's worth paying this cost in order to get refractoriness right.

Also, I missed out the plot with synapses. Here it is:

figure_1

mstimberg:
Maybe to clarify: I did not look at all at the performance of the code objects itself, but only at the preparation time and run loop overhead. So it might well be possible that there is some inefficiency in numpy specifically, so far we were always more concerned with the weave/cython performance (apart from the whole `ufunc.at` improvement).

mstimberg:
Oh, not sure whether this changes anything but if I am not mistaken, the correct equivalent to the Brian 1 code with refractoriness is to only mark `v` as `unless refractory`, and not `ge` and `gi` as well as in the `compare_to_brian1` script.

thesamovar:
With just one variable being refractory, here is the plot:

figure_1

So not quite as bad, but still slower. I think probably worth it for being correct.

thesamovar:
Just checked and number of threads that the MKL uses doesn't change the total runtime (verified that it does change the number of threads being used by monitoring CPUs).

mstimberg:
Just a quick remark about my last commit: there was a weird error in one of the tests when run with openmp. It happened on Mac OS with python 3, but the same problem arose at least once before and that time it was Linux with python 2... It is quite curious, since none of the relevant code objects is actually using openmp parallellization, so I have no idea what is going on there. I added some output to the test so that it will give more information when it fails again, but until then I don't really know what to do about it...

mstimberg:
I had a quick look at why the exact integration is so slow. There is of course still the issue that in Brian 1, we calculate everything for the numerical values of all constants whereas in Brian 2, we do it symbolically (even a bit to the extreme, where the generated code still contains `mvolt`, etc. -- this is nice for readability but probably replacing it with the numerical values would be faster). But apart from that, we are generating very inefficient code for the state updater (it is about 25 times slower than the respective code in weave...). Here's what we do for updating v:
```Python
_v = (El + ((((_lio_3 * (((ge * (_lio_4 + exp(_lio_5 * int(not_refractory)))) * int(not_refractory)) * exp(_lio_6 * int(not_refractory)))) / (_lio_7 + (taue * int(not_refractory)))) + ((_lio_8 * (((gi * (_lio_9 + exp(_lio_5 * int(not_refractory)))) * int(not_refractory)) * exp(_lio_6 * int(not_refractory)))) / (_lio_7 + (taui * int(not_refractory))))) + (v * exp(_lio_6 * int(not_refractory))))) - (El * exp(_lio_6 * int(not_refractory)))
```
This evaluates not only `int(not_refractory)` but also more complex expressions like `exp(_lio_6*int(not_refractory)` several times (we don't have this issue in C since we now handle boolean expressions differently). One relatively simple solution would be to use sympy's CSE identification. I tried this out and it does not seem to be too heavy in computation (it adds 0.2s to the code generation process on my machine) and it speeds up the code considerably, by a factor of 2. The new code looks like this:
```Python
_cse_v0 = int(not_refractory)
_cse_v1 = exp(_cse_v0 * _lio_6)
_cse_v2 = exp(_cse_v0 * _lio_5)
_cse_v3 = _cse_v0 * _cse_v1
_v = (((((- El) * _cse_v1) + El) + (_cse_v1 * v)) + ((((_cse_v3 * _lio_3) * ge) * (_cse_v2 + _lio_4)) / ((_cse_v0 * taue) + _lio_7))) + ((((_cse_v3 * _lio_8) * gi) * (_cse_v2 + _lio_9)) / ((_cse_v0 * taui) + _lio_7))
```
What do you think? It is worth including, right?

mstimberg:
I realized that there is one annoying issue that we did not take yet care of. In the work you did for the code optimization, we avoided incorrectly simplifying expressions that involved stateful functions. Unfortunately, we can't control this whenever we use sympy. In the current code, including `rand() - rand()` somewhere in equations will be simplified away directly. The changed I proposed above (extracting common subexpressions) would mean that we are using sympy for *all* statements, so this problem would also occur for code in `run_regularly` operations, in resets, etc. I guess we have to deal with this at the very first stage where we convert strings into sympy objects, but unfortunately this also means that we will need the `variables` dictionary there whereas currently we don't.

mstimberg:
What about merging this branch as it is (if you are fine with all the changes, of course)? I think the CSE extraction and the issue of the stateful functions is important enough to be discussed/merged on its own (and nothing with respect to it changed in this branch so far).

thesamovar:
Agreed - this is good to go now and we can discuss the other issues separately. Could you copy your comment about state updaters to the new issue/pr? I think that's actually a regression in my code: when I improved LIO optimisation stuff in the cython performance branch, I also removed the optimisation that simplifies boolean expressions because it wasn't necessary in weave/cython - but that meant that numpy got slower. Could re-enable that, but CSE is also a good idea.

PR #605, opened 2015-12-01
[MRG] Build devices by default at the first `run` call

### LINKED ISSUE #512: Devices build/runs by default
At the moment ``devicename_simple`` (by convention) builds/runs by default when ``run()`` is called. We should make this the default for ``devicename`` and have an argument to ``set_device`` that switches this behaviour off. For the initial implementation, we should make a second ``run()`` call an error, however this might be possible to fix later when we improve standalone so that it doesn't always have to rebuild/recompile.

mstimberg:
This fixes #512: `set_device` now has a `build_on_run` argument which defaults to `True` and the overwritten `network_run` function in the standalone device now calls `build` directly if it has not been set to `False`. For quite a few scripts, this now means that a single `set_device('cpp_standalone')` line is enough to switch to the standalone device. All other arguments that `set_device` gets will be stored in `Device.build_options` and will be propagated to `Device.build` when it is automatically called (so you can do something like `set_device('cpp_standalone', directory='my_dir', debug=True)`.

For testing, this also means that the `standalone-compatible` tests now no longer use a "simple" device (`CPPStandaloneSimpleDevice` does not exist anymore), but instead use this mechanism. For this to work, the device also has to support the new feature that you can set `directory=None` and it will create a temporary directory for the build. I also added a new function `reset_device`, which can be used without argument to reset to the previously used device. This is nice for testing and avoids explicitly storing `previous_device = get_device()` in the tests.
I tried to document everything a bit, from my side this is good to go.

### DIFF
--- a/brian2/devices/cpp_standalone/device.py
+++ b/brian2/devices/cpp_standalone/device.py
@@ -16,7 +16,7 @@ from cpuinfo import cpuinfo
 
 from brian2.codegen.cpp_prefs import get_compiler_and_args
 from brian2.core.network import Network
-from brian2.devices.device import Device, all_devices, get_device, set_device
+from brian2.devices.device import Device, all_devices, temporarily_switch_device, reset_device
 from brian2.core.variables import *
 from brian2.core.namespace import get_local_namespace
 from brian2.parsing.rendering import CPPNodeRenderer
@@ -109,6 +109,12 @@ class CPPStandaloneDevice(Device):
         #: Whether the simulation has been run
         self.has_been_run = False
 
+        #: Whether a run should trigger a build
+        self.build_on_run = False
+
+        #: build options
+        self.build_options = None
+
         #: Dict of all static saved arrays
         self.static_arrays = {}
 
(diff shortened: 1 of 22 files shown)


thesamovar:
This looks great to me. Is it the case that calling ``device.build()`` with no arguments works and is equivalent (with just a single run call) to what would happen with ``build_on_true=True``? If not, maybe we should do this rather than requiring the specification of output directory, compile and run arguments all to be set? Apart from this point, I think this is good to merge.

mstimberg:
> Is it the case that calling device.build() with no arguments works and is equivalent (with just a single run call) to what would happen with build_on_true=True?

Yes, `build_on_run` basically calls `device.build` without arguments.

thesamovar:
Great - in that case good to go when tests pass.

thesamovar:
Tests now passing (I guess the large number of test failures before were just due to the failure to clean up one test?). I'll go ahead and merge.

mstimberg:
> Tests now passing (I guess the large number of test failures before were just due to the failure to clean up one test?).

Yepp, and whether it was a problem depended on the order of the tests, that's why I did not see it on my machine.

PR #606, opened 2015-12-01
[WIP] More overheads/efficiency improvements when comparing to Brian 1

thesamovar:


thesamovar:
See #446 for background.

mstimberg:
So I did a small change to the spike monitor logic for subgroups that probably does not improve things a lot, except for a very high number of spikes but I found it to be more reasonable (if you are not using subgroups, it now only checks the first and the last spike instead of all of them).

I had a more general look at what is taking all the time in the spike monitor but it is not straightforward. I shaved of ~15% by declaring the arrays that are storing the recorded data as "constant size" -- this is obviously not correct in a strict sense, but it prevents that we'll pass in a new reference to the underlying arrays at every time step. We don't use this reference anyway, since the templates do the resizing themselves and later have to make sure to access the correct data manually. I did the same change for the other monitors and added a lengthy comment explaining it.

Now, there seems to be a very large overhead that is added by weave/Cython boilerplate code. If I remove *everything* in the `spikemonitor.cpp` template that actually does anything (i.e. everything except the "pointer lines" and the `USES_VARIABLES`) then I still see a runtime of ~80ms for the `spikemonitor` codeobject, compared to ~170ms with it actually doing something. This cannot be all general overhead because for example the `neurongroup_resetter` template only takes ~40ms when it actually is doing its job... I am not quite sure what is going on there, I think it has to be all the namespace conversion code at the beginning (you can see the same for Cython). Maybe we have to be a bit more careful what we pass in the namespace, e.g. we could only pass in the length of arrays if they are asked for explicitly or things like that.

thesamovar:
I just ran the tests on your previous commits and I got this compile error with weave

```
c:\users\dgoodman\appdata\local\temp\dgoodman\python27_compiled\sc_33dd70a20f61c48204480c72b874f8a10.cpp(2797) : error C2440: 'initializing' : cannot convert from 'int *' to 'const npy_int32 *'
        Types pointed to are unrelated; conversion requires reinterpret_cast, C-style cast or function-style cast
```

I think the problem line is in the comment I just made above.

thesamovar:
For the changes in SpikeMonitor - wouldn't it be better to do a bisection algorithm here actually? We know that the events are in sorted order, right? This would be particularly useful for the numpy version which is doing a lot of unnecessary work, but also for weave/cython too.

mstimberg:
I think I tried using bisecting a while ago, but it wasn't really worth it. You are right about numpy, though, that is far from efficient. I am not sure it makes any measurable difference, you'd need the combination of small subgroups and many spikes for it to have any impact on the performance. On the other hand, there's no good reason to not do it properly...

thesamovar:
Maybe we should rename ``constant_size`` to something else so that we don't need the long comments explaining the hack?

mstimberg:
> Maybe we should rename constant_size to something else so that we don't need the long comments explaining the hack?

Sounds reasonable :) How about something like `needs_reference_update`?

thesamovar:
For the boilerplate code, could we just delete anything from the namespace that isn't in ``get_identifiers(snippet)``?

thesamovar:
Yep, that sounds good (``needs_reference_update``)!

mstimberg:
> For the boilerplate code, could we just delete anything from the namespace that isn't in get_identifiers(snippet)?

We are already only passing in the variables that are used in the code + the variables that the template needs + variables manually requested via `needed_variables`. What I meant was that for array variables we are automatically also adding the length as `_num...` to the namespace and for dynamic variables we add both the dynamic variable object and the underlying data array. For the `_num...` variable I am wondering whether we actually need it, normally one should be able to use an explicit variable such as `N` instead.

mstimberg:
I started taking out a few of the variables from the namespace, it did not have much effect so far, though... Since it broke a few things I am not commiting it for now, I'll continue working on it tomorrow.

thesamovar:
I wonder if the issue is that certain variables (e.g. dynamic arrays) lead to more expensive boiler plate than others, or if it's just the total number of variables is large. In any case, I think a modification of the idea I mentioned should work. At the moment we know all the base variables, and we always add ``_num...`` and ``_dynamic...`` but maybe we aren't always using them. We could check this explicitly by looking at the snippet + template and pulling out the identifiers that are actually used? Like that we keep the flexiblity of always having access to them but we only actually process the ones we need.

mstimberg:
Ah, sorry, I misunderstood earlier, you want to look at the generated code, not at the abstract code? Hmm, not sure whether this will work nicely -- we don't want to parse the code but simple functions like `get_identifiers` are very primitive (they don't care about strings, comments, etc.). But maybe it is a better approach than to extensively annotate each and every object.
Maybe we should make sure to not optimize the wrong thing first, though? Maybe run an example like `CUBA` with a handwritten list of things that should go into the namespace and see what difference it makes? It is possible that the things that are taking the most time are the objects the code actually needs...

thesamovar:
Yes, exactly: it would be a safe solution but not perfect (it would occasionally handle objects it didn't need to, although that not that often because they have weird names). I agree though that the first thing is to find out why the boilerplate is so slow.

Tests for this branch are now running fine on my machine btw.

thesamovar:
Latest plots.

compare_to_brian1
compare_to_brian1 do_monitor
compare_to_brian1 do_refractory
compare_to_brian1 do_synapses
compare_to_brian1 exact_method

Overall, B2 is looking better and better.

Refractory is still a little slower for B2/numpy but that might partly be solved by putting back the boolean optimisation, and is anyway worth having to get refractoriness right.

Exact integration is still very slow for B2/numpy and has a high initial cost. I'm wondering: maybe instead of adding ``*int(not_refractory)`` to the equations, we should instead solve the equations twice, for refractory on and off, and then do ``refractory*sol1+(1-refractory)*sol2`` in the state updaters. The advantage of this is that it gives sympy a much easier job to not have the boolean variable there, and it generates better/simpler code too. Worth considering?

mstimberg:
I had a look at the effect of the number of namespace variables (I had some issues measuring them since it is not taken into account for the weave/cython caching system, i.e. make sure to delete the cache directory!). I implemented your suggestion and it works fine, for the spike monitor the number of objects in the namespace goes down from 33 to 14 and the time (for a 10s run) from 1.65s to 1.35s. I could manually shave of another 0.03s or so by removing the references to the underlying arrays of the dynamic arrays where the values are stored (we resize these arrays from within the template and then get new references manually). Implementing this would be a bit more complicated, though, the array names *are* used in the code, but only in the lines defining the pointers... But changing this would mean to mess with the `CPPCodeGenerator` and I don't think this complication is worth it. I am happy with this for the namespace part now. I'll have a quick look at whether optimizing the subgroup check is worth it.

mstimberg:
It does not seem as if using bisection for the subgroup check in the spike monitor improves things... I modified the CUBA example so that it uses separate e and i neuron populations (with four `Synapses` objects between them) and one spike monitor for each of them. The difference between using actually two `NeuronGroup` objects for the two populations or using one group with two subgroups should show the time spent in the "subgroup filtering" code. For weave, I did not see any measurable difference (at least for 4000 neurons). For numpy, there was a 10% penalty for the smaller inhibitory population (~2s vs. ~2.2s).
Now I tried using numpy's `searchsorted` function to use bisection, but this did not work very well. If I use a single call like this:
```Python
_start_idx, _stop_idx = _numpy.searchsorted(_events, [_source_start, _source_stop])
_events = _events[_source_start: _source_stop]
```
then both of the monitors are *slower* than before (~5% for the [0, 3200[ monitor and ~10% for the [3200, 4000[ monitor). Since often we start subgroups at the start or end them at the end (as in my example), I also tested the following variant:
```Python
    if _source_start !=0:
        _start_idx = _numpy.searchsorted(_events, _source_start)
        _events = _events[_start_idx:]
    if _source_stop != {{N}}:
        _stop_idx = _numpy.searchsorted(_events, _source_stop)
        _events = _events[:_stop_idx]
```
This now has a positive effect for the bigger subgroup (~2-3% faster) but still a negative effect (~10% slower) for the smaller group... I guess the overhead of the function call (checking of argument types, etc.) outweighs the benefit of the logarithmic vs. linear complexity, unless we have a large number of spikes. I guess we could invoke the bisecting code only for a certain minimum number of events, but I am not sure this is worth all the complication.

I think many simulations are like the CUBA example and don't have many spikes in a single time bin (in my run, there are between 0 and 12 spikes, with a mean of 2.2...)

mstimberg:
> Exact integration is still very slow for B2/numpy and has a high initial cost. I'm wondering: maybe instead of adding `*int(not_refractory)` to the equations, we should instead solve the equations twice, for refractory on and off, and then do `refractory*sol1+(1-refractory)*sol2` in the state updaters.

This would probably be a better approach, I agree. The only problem with it is that we'll have to make this change in the stateupdater code itself, i.e. every stateupdater has to be updated to support this. It is probably not crazily hard to do, but still a significant amount of work. Maybe rather post-2.0?

thesamovar:
I tried the boolean optimisation in numpy using array indices and it works and the code is quite short, but it's not clear if it's worth doing because it makes it faster in some cases and slower in others. I've committed it but the code is commented out: just so that we have a copy in case we decide to look into this in more detail in the future. I'll try common subexpression next (but maybe using AST analysis rather than sympy because of the issues with using sympy), but perhaps we shouldn't get too stuck on numpy performance.

thesamovar:
For reference, by directly writing the state updater code in the template, I managed to get the time for that codeobject only down from 1.58 to 0.72, so we're actually not doing too badly compared to (hopefully) optimal code. I think we could get some nice gains by not doing so many copies of the data. We do that for safety at the moment, but it might be worth thinking again about that and only making copies when needed.

thesamovar:
Only one test failed in the above, and so it's tempting to think it's a bug in appveyor, but actually I also got a segfault when running the latest version (before I made my most recent commit), so there might actually be some sort of out-of-bounds pointer access going on or something like that that only manifests intermittently in the code you introduced recently.

mstimberg:
Yes, I've seen this as well. On appveyor, it only affects standalone on 64 Bit Windows. You are running things with 32 Bit on your machine, right? I'll see whether I can reproduce the issue in my Windows VM.

thesamovar:
These days I'm on 64 bit thanks to Anaconda. I actually ran the test suite fine, but when I was running my script to do benchmarking against Brian 1 it crashed on one of the runs (which makes me think it's an intermittent bug rather than something that'll be easily reproducible).

mstimberg:
Great... Hopefully we'll manage to reproduce it on one of our machines -- debugging via appveyor is not fun. I'll quickly create a test branch, though, that only runs the 64 Bit standalone test suite and prints out the test names -- maybe this gives us a hint.

thesamovar:
Hypothesis: could it be that in the weave synapses, we now call peek, store a pointer, and then call advance. For C++, calling advance calls ``queue[offset].clear();``, which means that we have cleared the vector storing the data if the length of queue is 1 (which will be the case when there are no delays). There's a note saying:

```
        // empty the current queue, note that for most compilers this shouldn't deallocate the memory,
        // although VC<=7.1 will, so it will be less efficient with that compiler
```

VC 7.1 is 2003 so we ought to always be using a higher version: but perhaps this note is wrong and sometimes this memory will get deallocated even on higher versions? In any case, we're relying on an undocumented and (strictly) unsupported behaviour of vector it seems, so we probably ought to change that. Can we just call ``advance`` at the end of the method?

thesamovar:
Another possibility: can a single SpikeQueue serve multiple code objects? Or is there a unique mapping between spike queues and code objects that call them? If not, then maybe something weird is going on with peek and advance being called too many times?

mstimberg:
Ah, I wasn't aware that `advance` clears the queue at the offset. I don't see any problem moving the `advance` to the end of the `synapses` template, this should be safer. About the `SpikeQueue` access: There is one `synapses_push` codeobject that pushes to the queue (well, it just calls the Python code to do this actually in runtime) and one `synapses` object that does `peek` and `advance`. Each pathway has its own queue, so there should not be more than one pair of them accessing the spike queue.

But all of this is about weave, right? We were seeing the crash on appveyor in the standalone tests.

thesamovar:
Good point, although standalone also works in the same way so maybe it has a similar problem? Wouldn't explain why we're seeing it now though. FWIW, the problem I saw was not on standalone but on weave.

mstimberg:
> Good point, although standalone also works in the same way so maybe it has a similar problem?

Not quite, in standalone we do the `advance` directly before the push, so it does not have the potential `peek`+`advance`+"access vector previously returned by peek" issue.

The error on appveyor seems to be reproducible in the `test_openmp_consistency` test:
https://ci.appveyor.com/project/brianteam/brian2/build/970

Did you run the test suite on your machine with OpenMP as well? It does not necessarily have to do anything with OpenMP, it's just that this test is not run if you run the standard standalone test suite,

mstimberg:
So the problem does indeed to be in the weave code, it crashes during the comparison run for runtime:
https://ci.appveyor.com/project/brianteam/brian2/build/971

What is curious that it seems to stop after the run was finished, so after the final `report_callback`, but before my `After run...` debug output. There should be not much more there than clearing `Network.objects` and calling garbage collection... I'll try to add some more debug output, hopefully that can clear it up.

mstimberg:
I think the above was a red herring, it can crash during the simulation as well: https://ci.appveyor.com/project/brianteam/brian2/build/973 Would be great if you could reproduce this on your machine...

thesamovar:
I can consistently generate the error now with my test script: it fails with weave and SpikeMonitor on, but is fine in other configurations (including synapses), so I guess it's something to do with the changes related to SpikeMonitor but not sure what yet.

thesamovar:
Oh, can also reproduce with OpenMP but not with the standard test suite.

thesamovar:
Maybe we should move advance to the end even if it's not causing this bug? Now that I know it could be a problem it makes me uncomfortable. ;)

thesamovar:
When I run the OpenMP standalone tests, it crashes python.exe with error code 0xC0000374, which according to this MSDN blog is a heap corruption exception. http://blogs.msdn.com/b/calvin_hsia/archive/2015/01/30/10590143.aspx Not sure if that helps much.

mstimberg:
Something strange is going on, now that I moved the `advance` to the end of the runtime templates (not changing anything else), standalone tests are failing all over the place...

thesamovar:
Well if it is a corruption of the heap, then we might not see the error at the place that caused it. Can we compile with debug mode and maybe keep track of whether or not we're doing any out of bounds errors, or dereferencing pointers to memory that has been deallocated.

thesamovar:
Oh I see the errors are compile errors. This is really odd.

thesamovar:
It seems to be running fine on my machine, so I restarted the build to see if it was just a weird bug on appveyor's side.

thesamovar:
On the other hand, the ``compare_to_brian1`` script is still crashing with SpikeMonitor on and Synapses off (for weave).

mstimberg:
Standalone is still failing on appveyor, no idea what happens there... I just commited a potential fix for the `weave` issue, though, but I am not sure at all whether it actually does fix anything. We were resizing everything as part of the template and that included setting `N`. But `N` is accessed in the template via a restricted pointer, so we were breaking the promise of changing it only through this pointer. I could imagine that this screwed things up in some subtle way (we don't access `N` after the resize anymore, but still...)

thesamovar:
Very good catch! For what it's worth, that makes the compare to brian1 script I mentioned above work without crashing. Will run the OpenMP tests next. The normal long and standalone tests were already running without error on my machine.

thesamovar:
If this is the fix, should we have a new keyword that allows or disallows the ``__restrict__`` keyword?

thesamovar:
I mean, should we make certain variables not be allowed to use ``__restrict__``?

mstimberg:
This might be a useful feature. For now, I think it is ok to set `N` manually in the monitors (I only changed it for `SpikeMonitor`, the other monitors might have the same issue), we are already doing a bunch of stuff related to data access manually there...

mstimberg:
I *think* this fixed the weave issue -- at least it no longer seems to be crashing on appveyor. The standalone issue still persists, I don't really know what is going on there. I will put the last commit where it still worked into a new branch to see whether it has anything to do with the changes afterwords or whether it is something else that changed (dependency version, etc.).

mstimberg:
Ah wait, there was one commit before it started failing that might potentially affect standalone: 5e2ab792 I don't really see what harm that could have done, though, and it is still running on our machines... Either way, last commit without problems is testing again here: https://ci.appveyor.com/project/brianteam/brian2/build/981

mstimberg:
So the problem is not in any of the recent commits, the test I linked above is failing as well. I noticed that the failing test is now run on Python 2.7.11 while previously it was run on 2.7.10 -- could you update your Python version and see whether it makes a difference (it does not on Linux, but might be something OS-specific)?

thesamovar:
I updated just Python and it worked fine, and then I updated all packages to the most recent version: still fine. I think it must be something to do with Appveyor's setup again.

mstimberg:
Ah well, actually when it failed earlier it was still 2.7.10. But it seems that only OpenMP tests fail, those are working fine for you as well?

thesamovar:
I hadn't actually checked OpenMP but just did and it passed.

One difference I'm noting is that they are using Visual Studio 2015 whereas I have 2010 as the most recent version on this laptop. It could be the case that there was a breaking change that has caused this. I had a look at the list for 2015 and 2013 but couldn't see anything that might immediately obviously have caused this problem to occur. There have been some changes to the way library headers are organised, and maybe this is somehow messing up our code? Can you check if Appveyor recently upgraded their version of VS to see if this is likely to be the problem?

https://msdn.microsoft.com/en-gb/library/bb531344%28v=vs.120%29.aspx
https://msdn.microsoft.com/library/bb531344%28v=vs.140%29.aspx

thesamovar:
Hmm, the errors all seem to be in ``objects.cpp`` and only occur with OpenMP on. Something to do with the brian namespace maybe? Some weird interaction between the two?

thesamovar:
On the off chance that the new MSVC treats namespaces differently I just committed a change to the way I declare variables in objects.cpp - no idea if it will work though.

thesamovar:
Actually maybe the best thing to do if my attempted fix above doesn't work (likely not to) is to set AppVeyor to use an older version of visual studio, and open a new issue for compilation problems with visual studio 2015?

mstimberg:
I have to admit that all this Windows compiler configuration is less than non-obvious to me. We have an elaborate system (copied from elsewhere, of course) in place so that depending on the Python version and the architecture, we are choosing the right compiler from the SDK. If I understand it correctly, we should always either use the MSVC 2010 from SDK 7.1 or the MSVC 2008 compiler from SDK 7.0, i.e. never a compiler from visual studio directly (which are installed on appveyor as well). Now for standalone we manually search for the `vcvarsall.bat` with the highest version number, and this does indeed seem to select MSVC 2015 now (we should probably add some diagnostic information about this to the appveyor test run). It would probably be more consistent if we default to the same compiler as used for weave/cython? But still of course, we should make it work on VS 2015, as you said this could be a new issue though.

thesamovar:
I think maybe you can specify in the appveyor.yml file which version of visual studio you want installed?

mstimberg:
No, there are several of them installed

mstimberg:
http://www.appveyor.com/docs/installed-software#sdks

thesamovar:
Ah, so the short term fix would be we look for the highest version that isn't 2015? In that case, let's wait to see if the change I made fixes the problem, and if not we do that and open a new issue.

mstimberg:
My proposal would be that the short-term fix would be to use the same version that distutils would chose, i.e. the version that will be used with weave/cython. This should be always 2008 or 2010, I think (except for Python 3.5 maybe).

thesamovar:
The problem with that is you lose out on the option to compile for more recent hardware (eg. AVX, AVX2).

mstimberg:
Ah, I see. Did we ever check whether it actually makes any measurable difference? Either way, a simpler fix would indeed to look for the `vcvarsall.bat` only in the range 2008-2013 (starting at 2013).

thesamovar:
Well we'll know in a minute: it's almost at the point where the errors kick in. :)

thesamovar:
Nope, no good. All errors again. Will do the 2013 version or lower idea next.

mstimberg:
Pro tip: Maybe move the standalone tests to the top in `appveyor.yml` :)

mstimberg:
This seems to have worked (tests are failing because of some unrelated problem with appveyor's storage which seems to be a temporary problem we can't do anything about) -- should we maybe revert the previous commit then and deal with the VS 2015 error in a new issue?

thesamovar:
Great! Actually we can leave the previous commit, I think it's slightly tidier like that than it was before (one single ``namespace brian`` rather than multiple ``brian::``.

We should just add a note in the docs that VS2008-20013 are the only ones supported, and add an issue, and then I think we can be done with this problem (finally!).

Once appveyor sorts itself out, should we merge the changes in this branch but keep the PR open for a few more improvements?

thesamovar:
OK, made those two changes, but actually let's not merge yet - will do another round of speed testing with the latest changes now that it's working.

mstimberg:
I just merged most recent master into this branch (which shouldn't change anything speed-wise, though). On my machine, I got the failure in the state monitor test (`test_state_monitor` in `test_monitor.py`) that we'd seen before with OpenMP -- but of course I could not reproduce it afterwards anymore :-/ The additional output that I added a while ago to the assertion showed that for some seemingly random time steps (and not the same for the two neurons: [8, 16, 52, 63, 82] for first neuron, [6, 57] for second), `not_refractory` equals `False`, even though the cells does not spike at all in this example (`v` starts at 1 and only decays). Now there are two places were `not_refractory` could be set to `false`: either if the cell spikes in the `thresholder` (which as mentioned before is not the case) or in the state updater:
```C++
	#pragma omp parallel for schedule(static)
	for(int _idx=0; _idx<_N; _idx++)
	{
	    // vector code
		const int _vectorisation_idx = _idx;
                
        const double lastspike = _ptr_array_neurongroup_f2d700cb_5344_4d35_8459_2a5f196cd7f8_lastspike[_idx];
        const double t = _ptr_array_defaultclock_t[0];
        double v = _ptr_array_neurongroup_f2d700cb_5344_4d35_8459_2a5f196cd7f8_v[_idx];
        bool not_refractory;
        not_refractory = (t - lastspike) > 0.002;
        const double _v = _lio_1 * v;
        v = _v;
        _ptr_array_neurongroup_f2d700cb_5344_4d35_8459_2a5f196cd7f8_not_refractory[_idx] = not_refractory;
        _ptr_array_neurongroup_f2d700cb_5344_4d35_8459_2a5f196cd7f8_v[_idx] = v;

	}
```
Alternatively, something could go wrong with the recording (I removed the code recording all the other variables we are not interested in):
```C++
    #pragma omp parallel for schedule(static)
    for (int _i = 0; _i < _num_indices; _i++)
    {
        // vector code
        const int _idx = _ptr_array_statemonitor_6__indices[_i];
        const int _vectorisation_idx = _idx;
                                        
                    const bool _source_not_refractory = _ptr_array_neurongroup_226bfc52_32b3_41cb_ab15_240ffe255d2b_not_refractory[_idx];

                    bool _to_record_not_refractory;
                    if(_source_not_refractory)
                        _to_record_not_refractory = true;
                    else 
                        _to_record_not_refractory = false;

            _dynamic_array_statemonitor_6_not_refractory(_new_size-1, _i) = _to_record_not_refractory;

    }
```

We've seen this test failure only with OpenMP switched on, but I can't really see anything that could go wrong in the code...

mstimberg:
Ah, finally I figured it out! The values of `not_refractory` recorded for each time step are stored in a `std::vector<bool>` and writing to it concurrently is of course not safe... It's an easy fix, should I commit it or should I wait for your VS 2015/OpenMP fix?

thesamovar:
Great catch! Go ahead and commit I'm still dealing with a couple of things that are taking longer than expected so I haven't started on the VS/OMP fix.

mstimberg:
Ok, I just pushed it.

thesamovar:
OK, and I just pushed the slightly less restrictive check. I think with that and your changes, we now solved all the problems here? I think we could still do with some more optimisations/comparisons to B1, but maybe now is the time to merge this (but leave the branch and PR open)?

thesamovar:
I just reran the comparisons to Brian 1 (below). I think that it basically looks good except for the slow synapses creation and the fact that the exact method is very slow for B2/numpy compared to B1. I think that's enough for this branch to be closed actually (which is not to say that there aren't any more optimisations to do).

compare_to_brian1 do_synapses
compare_to_brian1 exact_method
compare_to_brian1
compare_to_brian1 do_monitor
compare_to_brian1 do_refractory

mstimberg:
Looks great to me, I agree that this can be merged now. We already have issues open for synapse creation and the numpy inefficiency, so they can be handled in new branches/PRs. Let's wait for the tests to run through first, though. I don't know why "All checks have passed" is already displayed, the tests are still running...

thesamovar:
I just pushed a new change, an alternative approach to using the boolean simplification information for numpy. It works much better! The state update code object for the CUBA example dropped from taking 1.7s to 1.1s with Euler integration, and from 6.9s to 1.5s for exact integration. It's still much slower than B1 though for exact integration.

thesamovar:
Any idea what the test fail is about?

mstimberg:
> Any idea what the test fail is about?

I could not reproduce it on my machine with Python 3.5, but I think it might be related to the fact that `Network.t` was set to a one-element array instead of a scalar value and therefore the `completed` value also was an array. Python 3.5 introduced a `__round__` special method that is invoked when you call `round` and numpy arrays do not support it (yet). Let's see whether this fixes it.

mstimberg:
Great work on the alternative implementation for numpy, by the way -- this looks much more straightforward and if it is faster, even better! All the tests now pass so I guess this is good to merge?

thesamovar:
Yes! Finally. :) I'll just go ahead and do that.

Issue #412, opened 2015-02-25
Better handle subexpressions in state updates


mstimberg:
Currently, subexpressions are handled quite elaborately handled in statements (re-calculation is avoided) but in state updates, they are simply replaced so they don't appear in the generated code. This makes the code quite complicated and hard to read (especially for typical HH-like equations). It would be nicer if state updater would treat them in a smarter way, but it is a non-trivial problem. The best solution I could think of:
* In the symbolic manipulations that we do with sympy, treat subexpressions as functions of the variables they depend on (complicated by the fact that they can refer to other subexpressions).
* For the generated code statements, enumerate all the different ways a subexpression is used (e.g. in an RK2 algorithm it might be called once as `alpha(v)` and once as `alpha(v + k/2)`) and generate the respective statements (for, say, `alpha_1` and `alpha_2`) and replace the subexpression calls accordingly.

I think this can get quite complicated, but it should be doable. Not sure it is worth the effort, though.

thesamovar:
Isn't the problem that the end result can be simplified in a way that the subexpressions don't appear any more? If so, maybe a more useful thing to do (but that won't help readability) would be to carry out common subexpression analysis on the final generated abstract code in a similar way to the LIO stuff?

mstimberg:
We could try common subexpression analysis (at least in some simple way -- using sympy will take too long I think), but I'm a bit worried that we end up with a huge list of newly defined expressions that don't have much of a relation to the original equations (e.g. you don't find anything resembling alpha in it). On the other hand, maybe that's ok and we should think of the subexpressions as something that only helps with the readability of the equations but has nothing to do with the final code? Final question: does it actually buy us much, at least with `-Ofast`, gcc will most likely eliminate common subexpressions better than we do. Hmm, not sure whether it actually optimizes expressions including function calls (e.g. does it know that `exp(x)` will always return the same value?). Oh and couldn't our creation of temporary variables make the use of registers impossible in some cases?

Maybe we should do some clearer analysis of the potential benefits before starting such a fundamental change.

thesamovar:
It's definitely not a high priority. Maybe a good student project at some point?

mstimberg:
> Maybe a good student project at some point?

Good idea, it might be nice for a computer science student since it doesn't need any background in computational neuroscience. I moved the milestone to a vague "2.0", but I don't think it's even essential for the 2.0 release.

PR #611, opened 2015-12-14
[MRG] Fix #608

### LINKED ISSUE #608: Incorrect results for stochastic equations with `(unless refractory)`
I don't have time to look at it right now, but there is a severe problem for stochastic differential equations with `(unless refractory)`. Test code:
```Python
np.random.seed(1234)
G = NeuronGroup(1, 'dx/dt = -x/second + xi/sqrt(second) : 1 (unless refractory)', refractory=1*ms)

mon = StateMonitor(G, 'x', record=0)

run(1*ms)
print mon.x[0]
print G.state_updater.abstract_code
```
For Brian 2.0b4, the output without and with `unless refractory` is:
```
[ 0.          0.00471435 -0.00719588  0.00713191  0.00400468 -0.00320161
  0.00567034  0.01426566  0.007899    0.00805517]
not_refractory = (t - lastspike) > 0.001000
xi = dt**.5 * randn()
_x = -dt*x/second + x + 1.0*xi/sqrt(second)
x = _x
```
and
```
[ 0.          0.00471435 -0.00719588  0.00713191  0.00400468 -0.00320161
  0.00567034  0.01426566  0.007899    0.00805517]
not_refractory = (t - lastspike) > 0.001000
xi = dt**.5 * randn()
_x = -dt*x*int(not_refractory)/second + x + xi*int(not_refractory)/sqrt(second)
x = _x
```
i.e. the values stay the same as they should. With current master, however, we get:
```
[ 0.          0.00471435 -0.00719588  0.00713191  0.00400468 -0.00320161
  0.00567034  0.01426566  0.007899    0.00805517]
not_refractory = (t - lastspike) > 0.001000
xi = dt**.5 * randn()
_x = -dt*x/second + x + 1.0*xi/sqrt(second)
x = _x
```
and
```
[  0.00000000e+00   4.71435164e-07  -7.19587674e-07   7.13191253e-07
   4.00468037e-07  -3.20160743e-07   5.67034214e-07   1.42656592e-06
   7.89899763e-07   8.05517145e-07]
not_refractory = (t - lastspike) > 0.001000
xi = dt**.5 * randn()
_x = dt*(xi/sqrt(second) - x/second)*int(not_refractory) + x
x = _x
```

I suspect I introduced the bug somewhere around or in this commit: adafd63e7

mstimberg:
This fixes the incorrect integration of stochastic differential equations with refractory period. The problem was that the `collect` function did not pull the symbol out of the equations if it was nested (as by the multiplication with `int(not_refractory)`). I fixed this by expanding the equations first and I also added an assertion into the function itself so that it will raise an error if the "non-stochastic part" still has a stochastic symbol in it.
A new test that I added also triggered a bug with Cython that was not triggered before: the boolean simplification creates code like this:
```Cython
        if not_refractory:
            v *= _lio_2
        else:
            v *= 1.0
```
and then because v is a "conditional write" write variable, this is wrapped in another `if not_refractory` which expects the following code to be just a single line so it ends up as:
```Cython
        if not_refractory:
            if not_refractory:
            v *= _lio_2
        else:
            v *= 1.0
```
I don't know what a good way is to fix this for real, but I did a minimal fix that at least makes the code compile, even though it is of course not optimal:
```Cython
        if not_refractory:
            if not_refractory:
                v *= _lio_2
            else:
                v *= 1.0
```

I'd be happy to merge it like this and to fix it properly later.


thesamovar:
This looks good to me. One question, are we only expanding out for stochastic equations? Also, we wanted to avoid exanding everything out originally for performance reasons (subexpressions are typically reused) - will this change hurt that?

Appveyor is presumably failing for the standard reason? Should we fix that before merging here (just about to do that in the compare to brian 1 branch) or just assume that it's OK?

mstimberg:
> One question, are we only expanding out for stochastic equations?

Yes, if there is no stochastic symbol in the equations, then we return directly. 

> Also, we wanted to avoid exanding everything out originally for performance reasons (subexpressions are typically reused) - will this change hurt that?

We don't deal with this yet, we are always substituting subexpressions at the moment (#412) and I don't think the change in this branch changes anything in that regard.

> Appveyor is presumably failing for the standard reason? Should we fix that before merging here (just about to do that in the compare to brian 1 branch) or just assume that it's OK?

Ah yes, I forgot that we did not merge that fix yet. I'm fine to wait until the merge of `compare_to_brian_1` -- but give me a moment before merging that branch, I now manage to at least sometimes reproduce the OpenMP problem I mention in my last comment.

thesamovar:
Ah yes, OK in that case I'm happy for this to be merged whenever.

For the compare to Brian 1 branch, I still have to make the change so that only OpenMP+MSVS2015 is disallowed. Will do that in a few minutes.

thesamovar:
Do you want to go ahead and merge master into this and then hopefully it should pass tests and we can merge?

mstimberg:
Yes, sure. Will do this in a minute.

mstimberg:
We both agreed to merge this and all tests passed --> merging now.

PR #612, opened 2015-12-17
[MRG] Make clip not use fused types (fixes #609)

### LINKED ISSUE #609: Cython problem with clip function
See discussion here: https://groups.google.com/forum/#!topic/brian-development/wqpJc_6l9kg

thesamovar:
For some reason this code wasn't working, but we never actually use floats anywhere so for the moment it's fine to just have doubles I think.

Tiny change ready to merge as soon as tests pass I think.


mstimberg:
I am fine with merging this as soon as tests pass, we can come back to this when we deal with #417

thesamovar:
Indeed!

PR #613, opened 2015-12-17
[MRG] Customise Cython cache dir (fixes #599)

### LINKED ISSUE #599: Put Cython code into temp directory rather than user directory?
At the moment, Cython files go into ``~/.brian/cython_extensions``. Perhaps this directory should be a Brian option (with the default being to put it in a temp directory, like weave does)?

thesamovar:
Changed it so that by default it uses the same cache directory that Cython's inline uses, but you can override with your own preference.

### DIFF
--- a/brian2/codegen/runtime/cython_rt/cython_rt.py
+++ b/brian2/codegen/runtime/cython_rt/cython_rt.py
@@ -32,7 +32,17 @@ prefs.register_preferences(
         Whether to use a lock file to prevent simultaneous write access
         to cython .pyx and .so files.
         '''
-        )
+        ),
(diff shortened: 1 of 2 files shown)


mstimberg:
Looks good to me. Maybe the error message when the dir creation fails could mention the preference? I am not sure whether cython is ok in general with a non-writeable cache dir, but I could imagine a situation where the default cython cache dir is not writeable for the user (e.g. on a network drive) and therefore it would be useful for them to know how to change it.

thesamovar:
Good idea! Done.

mstimberg:
All looking good to me, merging now.

PR #614, opened 2015-12-18
[MRG] Fix problem with sympy 0.7.7 

mstimberg:
As noted by Kees de Leeuw on the mailing list, our test suite fails with sympy 0.7.7 (not released yet). This is because the default sympy printing changed, equalities will be printed as `Eq(a, b)` instead of `a == b` -- this commit manually overrides this behaviour in our own custom printer (we already do a similar thing to have `a and b` instead of `And(a, b)`).


thesamovar:
Yep, go ahead and merge as soon as tests pass.

Issue #119, opened 2013-09-17
event-driven should be automatic for linear synaptic equations?


thesamovar:
We could have a ``timestep-driven`` flag if users want to insist on not having event-driven updates, but the default is to use the most sensible one?

thesamovar:
New option: we force users to specify event-driven or timestep-driven (maybe clock-driven). Like this we don't have to guess, and it seems that both event-driven and timestep-driven are options that users would expect to be true, but they can't both be true. @romainbrette - any thoughts?

mstimberg:
Some feedback on this: raising an error was judged to be too drastic, instead raise a warning if an equation does not have a flag, stating that it will be simulated as clock-driven which is slow and possible not what the user wanted/needed. The user can easily get rid of the warning by explicitly specifying `(clock-driven)`. I'd be happy with this solution, it also has the advantage of being fully backwards-compatible.

thesamovar:
OK!

PR #615, opened 2015-12-22
[MRG] Emit a warning for synaptic ODEs without flag

mstimberg:
As discussed in #119, synaptic differential equations can now have a `clock-driven` flag that explicitly state that the equation should be integrated every timestep. If neither `clock-driven` nor `even-driven` is specified, we default to `clock-driven` but emit a (lengthy) warning, telling the user why this might be a bad idea.


thesamovar:
Looks good to me. Happy for this to be merged. Is that really the only example that uses differential equations for Synapses? I guess it's not that common except for STDP.

mstimberg:
> Is that really the only example that uses differential equations for Synapses?

Oh actually I only looked at the tests, not at the examples. There were three instances in examples and one in the tutorial where this warning would have been triggered, I fixed them with the last two commits. I'm currently running all the examples to make sure I did not miss one, I'll then merge.

mstimberg:
Actually, the standalone examples are failing on my machine. This is clearly unrelated to this branch, though, so I'll merge it anyway.

PR #616, opened 2015-12-23
[MRG] Standalone error for manual build with `build_on_run`

mstimberg:
The reason the standalone examples were failing was that they used the new `build_on_run` mechanism (which we switched on by default) as well as an explicit `device.build` call. I fixed this (also in one test) and made `CPPStandaloneDevice.build` raise an error for this situation (which can occur when copying older examples or upgrading code to the new version) -- this should make it *much* clearer what the actual problem is.

### DIFF
--- a/brian2/devices/cpp_standalone/device.py
+++ b/brian2/devices/cpp_standalone/device.py
@@ -839,7 +839,7 @@ class CPPStandaloneDevice(Device):
         direct_call : bool, optional
             Whether this function was called directly. Is used internally to
             distinguish an automatic build due to the ``build_on_run`` option
-            from a manual ``device.buil`` call.
+            from a manual ``device.build`` call.
         '''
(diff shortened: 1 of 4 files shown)


brian2/devices/cpp_standalone/device.py
```diff
@@ -836,7 +836,19 @@ def build(self, directory='output',
             ``True``.
         additional_source_files : list of str, optional
             A list of additional ``.cpp`` files to include in the build.
+        direct_call : bool, optional
+            Whether this function was called directly. Is used internally to
+            distinguish an automatic build due to the ``build_on_run`` option
+            from a manual ``device.buil`` call.
```
thesamovar: Minor typo here (should be device.build)

thesamovar:
Good to merge. I'll just fix the typo above and then merge directly since that won't change the tests.

PR #617, opened 2016-01-06
[WIP] Plasticity example

mstimberg:
This adds the Clopath et al. 2010 example contributed by Jacopo Bono. It also fixes small issues with encodings (you can now use "ö" etc. in the doc string) and allows for more than one comment line in the beginning of the example file.

I slightly changed one of Jacopo's example to use a `run_regularly` operation instead of a loop over runs, but I left the nested loop for the second example in place. Replacing this with a single run is certainly feasible, but it does not take a crazily long time as it is, so I think it is not really urgent and might make the code harder to read.

Ready to merge from my side.

### DIFF
--- a/brian2/sphinxext/generate_examples.py
+++ b/brian2/sphinxext/generate_examples.py
@@ -1,10 +1,9 @@
 import os
 import fnmatch
 import shutil
-import unicodedata
 from collections import defaultdict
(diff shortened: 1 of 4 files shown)


thesamovar:
Happy for this to be merged.

PR #619, opened 2016-01-07
[MRG] Make examples run under Python3

mstimberg:
Not much to review here, I just noticed that the `run_examples` script did not work under Python 3, and the same was true for a few individual examples. Since Python 3 is becoming more popular now and we never said that the examples should be run with Python 2, this needed fixing.

### DIFF
--- a/examples/advanced/stochastic_odes.py
+++ b/examples/advanced/stochastic_odes.py
@@ -59,7 +59,7 @@ rows = floor(sqrt(len(dts)))
 cols = ceil(1.0 * len(dts) / rows)
 errors = dict([(method, zeros(len(dts))) for method in methods])
 for dt_idx, dt in enumerate(dts):
-    print 'dt: ', dt
+    print('dt:', dt)
     trajectories = {}
     # Test the numerical methods
     for method in methods:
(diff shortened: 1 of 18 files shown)


examples/advanced/stochastic_odes.py
```diff
@@ -59,7 +59,7 @@ def exact_solution(t, dt):
 cols = ceil(1.0 * len(dts) / rows)
 errors = dict([(method, zeros(len(dts))) for method in methods])
 for dt_idx, dt in enumerate(dts):
-    print 'dt: ', dt
+    print('dt:', dt)
```
thesamovar: It's not terribly important, but this will look odd in Python 2 because it'll print ``('dt', 0.01*ms)`` rather than ``dt: 0.01*ms``. Same for the other examples.

thesamovar:
Looks good to me (although see line comment).

examples/advanced/stochastic_odes.py
mstimberg: Ah indeed, I forgot about this. What do you think would be the best solution?

1. Add `from __future__ import print_function` to each file
2. Use `'dt: ' + str(dt)`
3. Use `'dt: %s' % dt
4. Use `'dt: {0}'.format(str(dt))` (or a variant of this)

Some examples already use variants of 2 and 3, actually...

thesamovar: My preference would probably be (3), but (2) is nice too when it's a simple example like that (maybe (2) when simple and (3) when more complicated?).

mstimberg: Ok, I think I'll just go for (3) everywhere

mstimberg:
Ok I did the changes and also avoided a few "resolution conflict" warnings. If you are fine with everything as it is, feel free to merge it right away (nothing in this pull request will ever be run by the test suite anyway).

PR #621, opened 2016-01-07
[MRG] Fix "order dependence" warnings for statements involving constants

### LINKED ISSUE #620: Constants not handled correctly in the "order dependence" warning
The example `frompapers/Brette_2012/Fig5A.py` contains the synaptic `pre` statement `v = EL` which currently triggers the warning about an order dependence. This would be correct if `EL` was a synaptic variable but it is a scalar constant, so the order of execution does not matter. This should be straightforward to fix, I'll also add a test for this situation.

mstimberg:
Most of this pull request consists of changes to the tests to support constants in synaptic statements...

Fixes #620

### DIFF
--- a/brian2/codegen/permutation_analysis.py
+++ b/brian2/codegen/permutation_analysis.py
@@ -3,6 +3,7 @@ Module for analysing synaptic pre and post code for synapse order independence.
 '''
 from brian2.utils.stringtools import get_identifiers
 from brian2.core.functions import Function
+from brian2.core.variables import Constant
 
 __all__ = ['OrderDependenceError', 'check_for_order_independence']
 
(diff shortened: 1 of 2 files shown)


brian2/codegen/permutation_analysis.py
```diff
@@ -33,10 +34,11 @@ def check_for_order_independence(statements, variables, indices):
...
     main_index_variables = set([v for v in variables
-                                if (indices[v] in ('_idx', '0')
-                                    or getattr(variables[indices[v]],
-                                               'unique',
-                                               False))])
+                                if (not isinstance(variables[v], Constant) and
+                                    (indices[v] in ('_idx', '0')
+                                     or getattr(variables[indices[v]],
+                                                'unique',
+                                                False)))])
```
thesamovar: Could you explain the logic of this change a bit to help me review it? Won't this change mean that constants will just be moved into ``different_index_variables`` which feels like the wrong behaviour?

mstimberg: I agree this is not really clear. This all works because the `different_index_variables` are considered as permutation-independent (which is correct for a constant) and a constant can never get into the permutation-dependent collection because it does never appear on the LHS. But it would be probably better to have constants in a third set that is neither `main_index_variables` nor `different_index_variables`. That said, looking at the code I think scalar variables are still handled incorrectly, e.g. a `pre` statement `v_post = scalar_syn` will unnecessarily trigger a warning. I'll try to fix (and test) that tomorrow.

mstimberg:
I changed the logic slightly for the lines you commented on previously and also dealt with scalar variables. I initially wanted to make sure that pre statements such as `'x_shared = w_syn; v_post = x_shared'` also raised a warning but then I realized that we don't allow writing to shared variables in such contexts in the first place, so the `check_for_order_independence` does not have to care about this.

thesamovar:
This looks good to me! Happy for you to merge it when tests pass.

PR #630, opened 2016-01-15
[MRG] Raise error for multiple stateless functions call in an expression

### LINKED ISSUE #604: Treat stateful functions correctly when using sympy (e.g. in equations)
Quoting myself from https://github.com/brian-team/brian2/pull/602#issuecomment-160972614 :

I realized that there is one annoying issue that we did not take yet care of. In the work you did for the code optimization, we avoided incorrectly simplifying expressions that involved stateful functions. Unfortunately, we can't control this whenever we use sympy. In the current code, including `rand() - rand()` somewhere in equations will be simplified away directly. The changed I propose in #603 (extracting common subexpressions) would mean that we are using sympy for all statements, so this problem would also occur for code in `run_regularly` operations, in resets, etc. I guess we have to deal with this at the very first stage where we convert strings into sympy objects, but unfortunately this also means that we will need the variables dictionary there whereas currently we don't.

mstimberg:
It is less elegant than I'd like it to be but to make sure we catch this problem everywhere where we use sympy, I had to include it in `str_to_sympy` and change this function so that it optionally takes a `variables` dictionary (so it knows which functions are stateful). In a way this makes perfect sense because an abstract code block should always be accompanied by a `variables` dictionary, otherwise it is not meaningful. There are still a few places where the function is used without this argument, though, but those are about printing (using `str_to_sympy` to then use sympy's pretty printing facilities), for parsing state updater descriptions or in one somewhat ugly case, to just find out whether there are implicitly vectorized functions (i.e. `func()` calls) in an expression.

One disadvantage of this change is that we can no longer do the conversion to sympy once for each expression when it is created (because at that time the variables are not yet known) so we do some extra work. This should not be noticeable in normal use cases, though.

Fixes #604

### DIFF
--- a/brian2/codegen/translation.py
+++ b/brian2/codegen/translation.py
@@ -28,7 +28,8 @@ from brian2.utils.stringtools import (deindent, strip_empty_lines,
 from brian2.utils.topsort import topsort
 from brian2.units.fundamentalunits import Unit
 from brian2.parsing.statements import parse_statement
-from brian2.parsing.sympytools import str_to_sympy, sympy_to_str
+from brian2.parsing.sympytools import (str_to_sympy, sympy_to_str,
+                                       check_expression_for_multiple_stateful_functions)
 
 from .statements import Statement
 from .optimisation import optimise_statements
@@ -230,7 +231,7 @@ def make_statements(code, variables, dtype, optimise=True):
                                                 dtype=dtype, scalar=is_scalar)
                     variables[var] = new_var
             elif not variables[var].is_boolean:
-                sympy_expr = str_to_sympy(expr)
+                sympy_expr = str_to_sympy(expr, variables)
                 sympy_var = sympy.Symbol(var, real=True)
                 try:
                     collected = sympy.collect(sympy_expr, sympy_var,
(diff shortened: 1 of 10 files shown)


thesamovar:
Yes that is all quite annoying, but anyway this code looks good to me and I'm happy for it to be merged when the tests pass.

I think in the long term we should try to reduce our dependency on sympy and switch to doing as much as we can with ASTs.

mstimberg:
> I think in the long term we should try to reduce our dependency on sympy and switch to doing as much as we can with ASTs.

I am still not sure what the best approach is. On the one hand, sympy is regularly introducing problems and by construction it does not play well with functions such as `rand()`. On the other hand, we end up creating tons of code that is basically a re-write of parts of sympy (e.g. the simplification code) and thereby possibly introduces bugs that would be quickly discovered in a widely used project such as sympy. In some way, the problems with the stateful functions are our own fault because we introduced the non-mathematical `rand()` syntax.

Sympy's performance is definitely a cause for concern, though, I hope the cSympy project advances quickly... On the other hand, our own AST processing is heavily based on recursion and not super-fast, either.

Either way, nothing for right now ;)

PR #634, opened 2016-01-25
[MRG] Log levels

### LINKED ISSUE #407: Make `INFO` the default log level and clean it up

mstimberg:
This should basically implement everything we discussed in #407, the commit messages should give an overview over what changed. I made the commits quite small so it should be straightforward to revert single changes if we don't agree on them. The only thing missing from my side before merging is to update the developer documentation.

Closes #407


brian2/core/network.py
```diff
@@ -853,27 +859,31 @@ def run(self, duration, report=None, report_period=10*second,
...
+        logger.debug(("Finished simulating network '%s' "
+                      "(took %.2fs)") % (self.name, end_time-start_time),
+                     'run')
         # Store profiling info (or erase old info to avoid confusion)
         if profile:
             self._profiling_info = [(name, t*second)
                                     for name, t in profiling_info.iteritems()]
             # Dump a profiling summary to the log
-            logger.debug('\n' + str(profiling_summary(self)))
+            logger.diagnostic('\n' + str(profiling_summary(self)))
```
thesamovar: Possibly a candidate to stay in debug? If users are trying to understand why their code is running slow, this might be helpful?

mstimberg: Yeah, I wouldn't mind. Another potential candidate is the information about which `SpikeQueue` is used (maybe only in case it is the slower Python queue?).

brian2/devices/device.py
```diff
@@ -412,10 +414,10 @@ def spike_queue(self, source_start, source_end):
         # Use the C++ version of the SpikeQueue when available
         try:
             from brian2.synapses.cythonspikequeue import SpikeQueue
-            logger.info('Using the C++ SpikeQueue', once=True)
+            logger.diagnostic('Using the C++ SpikeQueue', once=True)
         except ImportError:
             from brian2.synapses.spikequeue import SpikeQueue
-            logger.info('Using the Python SpikeQueue', once=True)
+            logger.diagnostic('Using the Python SpikeQueue', once=True)
```
thesamovar: I agree with making the C++ SpikeQueue be at diagnostic level, but maybe if they are using the Python one that should be at debug level because it will be problematic for performance. I'm also wondering if we could have a common bit of text for performance information, either in the message or in the log item name)? Like that you could quickly grep/search the log for performance related information.

thesamovar: Actually using the python spike queue is sufficiently serious a problem for performance that it might be worth warning? Same for codegen maybe? (Just ideas at the moment.)

thesamovar: We could potentially have a "performance checklist" that is either displayed by default at some time or can be requested. For example, if you run a simulation that is expected to take more than 10m and there are performance related problems we detected, we show the checklist.

brian2/stateupdaters/base.py
```diff
@@ -127,10 +138,16 @@ def apply_stateupdater(equations, variables, method):
...
                                   'given equations has been found.'))
-            logger.info('Using state updater: %r' % the_method)
+            if group_name is not None:
+                msg_text = ("No numerical integration method specified for group "
+                            "'{group_name}', choosing method '{method}'.")
+            else:
+                msg_text = ("No numerical integration method specified, "
+                            "choosing method '{method}'.")
+            logger.info(msg_text.format(group_name=group_name,
+                                        method=the_method), 'method_choice')
```
thesamovar: Maybe don't put this at debug level if the 'exact' integration method was chosen?

brian2/synapses/synapses.py
```diff
@@ -993,12 +994,12 @@ def _create_variables(self, equations, user_dtype=None):
...
-                              'in the pre-synaptic group '
-                              '{source}.').format(var=name,
-                                                  synapses=self.name,
-                                                  source=self.source.name))
+                logger.diagnostic(('Cannot include a reference to {var} in '
+                                   '{synapses}, {var} uses a non-standard indexing '
+                                   'in the pre-synaptic group '
+                                   '{source}.').format(var=name,
+                                                       synapses=self.name,
+                                                       source=self.source.name))
```
thesamovar: What is the potential consequence to the user of this happening? I guess not much which is why it's in diagnostic?

brian2/stateupdaters/base.py
mstimberg: I'd vote for keeping it. After all, "exact" (which is called "linear", isn't it?) is problematic quite often. If you notice that the preparation time is taking forever and you see that it chooses the linear method after a long wait in the debug messages you know that it might be a good idea to explicitly choose a method.
But maybe the debug output should be rather like this:
```
DEBUG No numerical integration method specified for group 'neurongroup', trying the methods 'linear', 'exponential_euler', 'euler', 'heun' in order until one method can be applied to the equations ...
DEBUG ... selecting method 'linear'
```

thesamovar:
I think it's great. I made a few optional suggestions that we might discuss, but I think this already does pretty much everything we wanted.

brian2/stateupdaters/base.py
thesamovar: Oops, I meant don't keep it at info level - it should be at debug level. Either way, I'm happy to keep it if you think it's better like this. Your idea about the linear method taking a long time also raises another potential item for my idea of the performance checklist: if the choosing of the state updater takes more than a few seconds we could add that to the checklist? (Or just have an info/debug message about it.)

mstimberg: Ah, ok. Well, I think either all the state updater messages should go to debug or all should stay at info, most of the time the user does not care about the integration method and we have this other issue open about warnings for instabilities. Actually, if we move them to debug, then nothing remains at the info level and we wouldn't need a new log level in the first place... We could decide to lower some messages from warning to info to make them sound less severe (e.g. everything related to performance?).

brian2/synapses/synapses.py
mstimberg: I am not 100% sure, but I think it covers situations like this:
```Python
group1 = NeuronGroup(5, 'x : 1')
group2 = NeuronGroup(5, 'x_l : 1 (linked)')
group2.x_l = linked_var(group1, 'x', [4, 1, 2, 3, 0])
S = Synapses(group2, group2, ...)
```
In this situation, you cannot use `x_l_pre` or `x_l_post` in the synaptic group, as it would require double indexing. I think in the long run, the best solution would be to include special `UnusableVariable` objects in this case which would allow to give a meaningful error message when the user actually tries to use one of these variables (currently they'll simply get a general "unknown identifier" error message).

brian2/devices/device.py
mstimberg: Something like this good be a good idea (maybe just collect information about this in a documentation page and refer to that in the warning?) -- but I'd say let's open a new (post 2.0) issue about that. Somewhat related, I think it would be good to have a function that allows something like `print brian_configuration()` and would give you the codegen target, the spike queue, the version of Brian and its dependencies, etc. (similar to what we have in the log), for quick inspection.

mstimberg: About the Python spike queue being a warning: I don't think this is a good idea, a warning sounds very serious and this is only about *potential* performance issues. For simulation that are fast anyway (say, CUBA) or simulations that are not dominated by the synaptic propagation code, using the Python spike queue should be perfectly fine.

brian2/stateupdaters/base.py
thesamovar: OK, in that case I think we should keep the integration method at info level. I would also agree on moving all performance warnings to info level too - as you say, this seems more appropriate actually.

brian2/synapses/synapses.py
thesamovar: I see. And we don't want to display this message at a higher level because most of the time the user isn't trying to access this variable so it would just be noise. I think that's a fair compromise for now.

brian2/devices/device.py
thesamovar: OK to making this a post-2.0 issue (I'll open it now) and also to having Python spike queue being info level and not warning level.

thesamovar:
With the changes we discussed above (and I think you wanted to add some documentation) I think this is good to go. Happy for it to be merged once you think that's ready.

brian2/stateupdaters/base.py
mstimberg: What about two messages as I proposed above? On the one hand, it would be nice to give an indication about what takes so much time in the preparation phase, on the other hand it will be a bit verbose for those cases where it is quick. It will also be obsolete as soon as we introduce a time-dependent warning, but this not something we'll implement in the short term.

brian2/synapses/synapses.py
mstimberg: Ok, agreed.

mstimberg:
Ok, so to summarize: I'll move all performance warnings to `INFO` level, move the profiling summary from `DIAGNOSTIC` to `DEBUG` and update the documentation.

brian2/stateupdaters/base.py
thesamovar: I don't quite see the idea of the two messages? Is the point to see how long it's taking to select a method? I think it's too verbose for info level, and I'm not sure how useful it would be at debug level (but I have no objection to it there, it may be useful in some cases). I think more useful would be a message at info level saying that choosing the integration method took X seconds if X>some threshold. We can do both of course.

mstimberg: Yes, the idea was that it is a hint about the time it takes. Just adding the time is a good idea (I would probably not bother with a threshold). There is a subtle issue that the full time this takes is a mix of two things: discarding non-applicable methods and applying the methods. I think both informations are interesting, but it is probably a bit too verbose for `INFO` -- so I think I'll just add it to `DEBUG`?

thesamovar: So you'd go for an info level message like "Selected method ... took ... seconds" and more detailed information in debug? That sounds reasonable.

mstimberg: Hmm, no I was thinking about moving all of the timing information to the `DEBUG` log -- the thing is that the time that the method itself took is not necessarily interesting in the cases with complicated equations where checking the methods takes a long time but then applying Euler is quick. But maybe adding both times to the `INFO` log is ok actually.

thesamovar: As long as the info log doesn't get too overburdened that's OK (otherwise just the total time is fine I think). Certainly it should only be one line at most.

mstimberg:
So this should be it, I think. This is the about you now get for the `CUBA` example:
```
INFO       No numerical integration method specified for group 'neurongroup', using method 'linear' (took 0.53s). [brian2.stateupdaters.base.method_choice]
```
Most other examples explicitly chose an integration method (so there is no info output), but if I remove it for example from `Rothman_Manis_2003.py`, you'll get:
```
INFO       No numerical integration method specified for group 'neurongroup', using method 'euler' (took 0.38s, trying other methods took 0.52s). [brian2.stateupdaters.base.method_choice]
```

So from my side this is only missing the update of the docs, I'll do this next.

thesamovar:
Agreed!

mstimberg:
All tests pass, feel free to merge if you are happy with everything.

PR #638, opened 2016-01-26
Update spatialneuron.py

rcaze:
Adding the event keyword argument. Hey this is my first pull request ;)

### DIFF
--- a/brian2/spatialneuron/spatialneuron.py
+++ b/brian2/spatialneuron/spatialneuron.py
@@ -90,7 +90,7 @@ class SpatialNeuron(NeuronGroup):
     '''
 
     def __init__(self, morphology=None, model=None, threshold=None,
-                 refractory=False, reset=None,
+                 refractory=False, reset=None, event=None,
                  threshold_location=None,
                  dt=None, clock=None, order=0, Cm=0.9 * uF / cm ** 2, Ri=150 * ohm * cm,
                  name='spatialneuron*', dtype=None, namespace=None,
@@ -223,6 +223,7 @@ class SpatialNeuron(NeuronGroup):
 
         NeuronGroup.__init__(self, len(morphology), model=model + eqs_constants,
                              threshold=threshold, refractory=refractory,
+                             event=event,
                              reset=reset,
                              method=method, dt=dt, clock=clock, order=order,
                              namespace=namespace, dtype=dtype, name=name)


mstimberg:
Thanks :) I'd like to change two small and one potentially bigger thing before merging:
1. Could you also copy the documentation for the `events` keyword into the docstring
2. Please change the formatting of the `NeuronGroup` to get normal linebreaks instead of having `event` and `reset` on their own line
3. (the bigger one) Could you add a test to `brian2/tests/test_spatialneuron.py` that tests whether this actually works? It can be very simple, something like this but using `SpatialNeuron` instead of `NeuronGroup`:  https://github.com/brian-team/brian2/blob/master/brian2/tests/test_neurongroup.py#L725

mstimberg:
Ah, I just realized that you used github's internal editor to make the change. Can you make further changes in this way? You'll probably have to clone your repository and commit the changes to your branch...

rcaze:
Doing it know cloning in git. I wanted to start easy! I am going to
implement the three things!

2016-01-26 17:51 GMT+00:00 Marcel Stimberg <notifications@github.com>:

> Ah, I just realized that you used github's internal editor to make the
> change. Can you make further changes in this way? You'll probably have to
> clone your repository and commit the changes to your branch...
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/brian-team/brian2/pull/638#issuecomment-175140020>.
>



-- 
Dr R D Caze
http://romaincaze.wordpress.com/
Marie Curie Initial Training post-doctoral Fellow
@ Department of Bioengineering
Imperial College London
Tel: +44 (0) 7840596166

rcaze:
I managed to push and pull and to work locally on the files. It took time but I managed. It seems that the check are not OK. Sorry I need a bit to be "pris par la main". My next pull request will be better! ;)

mstimberg:
No worries, it takes some time to get used to it (we have some developer documentation as part of the docs that might be helpful). Make sure to use an environment (e.g. with conda) that is independent of your standard environment and use `pip install -e .` or `python setup.py develop` in your cloned repository so that you can directly use the version from your source directory for testing. You can run the full test suite with `import brian2; brian2.test()` but for now I'd only run the `test_spatialneuron.py` file. This file has some long running tests, I'd suggest to disable them for now by commenting them out in the `if __name__ == '__main__'` part in the end. Start with one of the simple tests (like `test_construction`) first and make sure that it works.
As you might have already figured out, the tests are currently failing (hint: "event"-->"events" :-) ).

rcaze:
I know have the environment. Things seems to work locally and set things up with python setup.py to have my local version! Do I have something else to do than just push?

mstimberg:
You opened the pull request for the `patch-1` branch (or rather, github created this branch for you), but you pushed your changes into the `master` branch. There are various ways of fixing this, I think the simplest would be to create a new branch in your local repository:
```
$ git checkout -b spatialneuron_events
```
And then push the new branch to github
```
$ git push origin spatialneuron_events
```
You'll then have open a new pull request for this branch.

This leaves your master branch in an inconsistent state with our master, though. There are of course solutions to this, but I think the easiest would be to simply completely delete your brian2 fork from github (and your local machine) -- after(!) we've merged the change into the main brian repository. For your next contribution, you then simply fork the repository again.

PR #639, opened 2016-01-27
SpatialNeuron Events

rcaze:
No sure If I had to but this is just to add events arg.

### DIFF
--- a/brian2/.gitignore
+++ b/brian2/.gitignore
@@ -1,1 +1,4 @@
 /default_preferences
+# Do not take into account RC temporary files
+*.swp
+.rope*
--- a/brian2/spatialneuron/spatialneuron.py
+++ b/brian2/spatialneuron/spatialneuron.py
@@ -60,6 +60,10 @@ class SpatialNeuron(NeuronGroup):
         It is uniform across the neuron.
     reset : str, optional
         The (possibly multi-line) string with the code to execute on reset.
+    events : dict, optional
+        User-defined events in addition to the "spike" event defined by the
+        ``threshold``. Has to be a mapping of strings (the event name) to
+         strings (the condition) that will be checked.
     refractory : {str, `Quantity`}, optional
         Either the length of the refractory period (e.g. ``2*ms``), a string
         expression that evaluates to the length of the refractory period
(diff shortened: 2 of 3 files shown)


mstimberg:
Looking good, thanks for resubmitting. I'll merge as soon as the test suite passed on our test servers.

mstimberg:
Just curious: what are "RC temporary files" ? And do they only occur in the `brian2` package, not in the main directory?

rcaze:
I updated the .gitignore file to not include the .swp produced by vim and the .rope folders that are used in code refactoring. This is just me being picky though. Looking forward to complete my first pull request !!

mstimberg:
Hmm, there seems to be an error, and I can reproduce it on my machine. It is very weird, though, I don't quite understand what it is about. @rcaze Maybe you'll get it as well if you either run the full test suite or just the `test_spatialneuron.py` file but with `prefs.codegen.target = 'numpy'`.
Actually, the `@attr` for the test is wrong, but it is wrong for `test_custom_events` in `test_neurongroup.py` as well, so definitely not your fault. It should not be `codegen-independent` (which we only test with numpy) but instead `standalone-compatible` (which will be tested with numpy, weave, cython and the standalone mode).
I'll look into the error tomorrow.

mstimberg:
Ah, actually it might have to do with the fact that you use a single `Soma` compartment for the morphology, I don't think we have a test like that, it might be that the numpy code does not handle this edge case (it should, of course).

mstimberg:
So it turns out that this is actually neither your nor our fault. The `solve_banded` function is not working for 1x1 matrices in scipy 0.16.0, a regression from earlier versions (scipy/scipy#5127). This has been fixed with scipy release 0.16.1, but there seem to be no Windows builds of this release in Anaconda at the moment. I opened a bug on the Anaconda bug tracker for this (ContinuumIO/anaconda-issues#642).

Given that this issue is not really central to this pull request, could you for now maybe just change the test so that the morphology uses more than one compartment? You can still restrict the event to one compartment by appending something like `and i==0` to its condition. Then the test suite should pass, and we can merge this pull request.

rcaze:
Look a bit technical. I am going for the change and add a dendrite!

mstimberg:
Ok, great, all tests passed -- merging it now. Congratulations to your first contribution to Brian 2 :)

PR #643, opened 2016-01-31
[MRG] Rename SpatialNeuron variables

mstimberg:
Not a particularly interesting pull request... Renames all the internal variables used in the `spatialstateupdate` template so that they start with `_`. Also pulls them out of the `SpatialNeuron` equations (they are nothing the user cares about). To make this easier, introduces a new `add_arrays` method that allows to add several `ArrayVariable` objects at once.


thesamovar:
OK to merge when tests pass.

PR #652, opened 2016-02-26
[MRG] Morphology revisited

mstimberg:
This ended up being a more or less complete rewrite of the morphology module,
but I am quite happy with it and I think it is more straightforward and less
error-prone now (it is also now extensively covered by tests).

From the user point of view, the documentation should be more or less complete,
I let it upload to readthedocs for easier reading:
http://brian2.readthedocs.org/en/morphology_revisited/

Some of the new principles (I'll try to document this in more detail in some
developer docs soon):
* Individual sections are either `Soma`, `Cylinder` or `Section` (for the name,
  see below) objects, all deriving from the abstract class `Morphology`
* All the attributes of a morphology are set when creating the section (apart
  from the parent/child structure), the geometrical properties are constant from
  then on. Functions that want to change properties (e.g. to add coordinates for
  visualization, or to resegment a section [not implemented yet]), return a 
  new `Morphology` instead of changing things in place
* There is an extensive list of attributes (e.g. `start_diameter`, `diameter`,
  `end_diameter`, `area`, `volume`, ...) but they are all implemented as
  properties and all data is stored only once without redundancy.
* Thanks to the properties, setting the `parent` attribute of a `Morphology`
  will return correct parent-dependent attributes (such as distance) whenever
  requested, without the need to change the internal values.
* The result of indexing with the position on the cable is now more predictable
  (for several use cases, it was actually completely broken) and avoids rounding
  issues at the boundaries between compartments (e.g. previously, for a section
  with compartments of length 10 um each, `morpho[0*um:10*um]` would select
  nothing whereas `morpho[40*um:50*um]` would return a single compartment).
* The `SpatialNeuron` class does not use the diameter of the section to
  calculate its resistance, but instead an `r_length` (again, see below for the
  name) attribute that encapsulates the geometry-specific calculation. For now,
  I did not do a similar thing yet for the `space_constant`, it directly uses
  `start_diameter` and `end_diameter`.

## Open questions

*Names*:

* `Section` is the generalization of `Cylinder` (truncated cones instead of
  cylinders) -- is this a good name though? `TruncatedCone` (or
  `TruncatedCones`) seems to be too long and I don't think many user would
  understand the word `Frustum` or `Frusta`...
* `r_length` is the value (in meter) that is multiplied with the specific
  axial conductance to give the total conductance of a compartment. It is
  proportional to `d**2/l` for a cylinder or `d_1*d_2/l` for a frustum. There
  does not seem to be a standard name in the literature for this value (but
  it is not a big issue, since this value should be only used internally, the
  user does not care)
* We could extract a similar constant to make the `space_constant` calculation
  in `SpatialNeuron` independent of geometry. The factor would be
  `sqrt(d_1) + sqrt(d_2)` for a truncated cone and `2*sqrt(d)` for a cylinder.
  But then, I am not sure that 1) we will add other geometries in the near
  future and 2) that those will be expressable just with a different factor

*Undefined attributes of Soma*

* A somewhat big issue: what should be the `r_length` of the soma? In the previous
  code, the axial resistance was made "negligible" by setting its length (just
  for this calculation) to diameter/100. The equivalent with the current
  approach would be to set its `r_length` to `d*100`. But "100" is of course
  arbitrary, and setting it to a different value does (slightly) change the
  result. I think the only "correct" way would be to use a value like `inf`,
  but this does not play nicely with the numerical methods for solving the
  equations.
* All sections need a `start_diameter`, `end_diameter`, and `length` but this
  does not make sense for a spherical soma. Those are all set to the sphere's
  diameter for now -- is there anything else that makes sense?
* Related to the above: the calculated space constant for a `Soma` does not
  make sense (it is the space constant of an equivalent cylinder)
* The "distance to the root" at the start of the first compartment that connects
  to the soma is 0 um. This makes sense when interpreting it as "electrical
  distance" but something like `diameter/2` might be more intuitive?

I'll try to discuss all this with @romainbrette soon.


mstimberg:
I just discussed the PR with @romainbrette, we agreed on a few changes, mostly simplifications (e.g. `Cylinder` will only be used for the simple cylinder, everything else uses `Section` where you are forced to specify everything, including start diameter/coordinates). I'll report back here when this is done.

thesamovar:
Did you want any comments from me? I haven't really been involved in the discussions on this and I'm pretty happy for you guys to work it out. Are there any core changes that I should know about?

mstimberg:
I applied all the changes that I discussed with Romain, so this is ready to merge from my side now. @thesamovar: I'd be grateful if you could have a look, no need to get into the details of the computation (which is now quite extensively tested and gives results that are *very* close to NEURON) , but some feedback on the following questions would be appreciated:
* Is all this reasonably easy to understand from a user point of view: http://brian2.readthedocs.org/en/morphology_revisited/user/multicompartmental.html ?
* Is the general structure in `morphology.py` and `spatialneuron.py` reasonably clear for a developer?

I don't think it needs a very thorough review, as it is certainly an improvement over the previous code (no internal state change after calling `Morphology.compress`, etc.) and it is much better tested (with some fixes to the way the coverage was calculated on coveralls.io, we are now approaching 95% test coverage overall!).

romainbrette:
Just having a quick look at the docs now:
* In the first example for Section, units are missing for diameter.
* Terminology: I think "branch" and "section" need a little 
clarification, in the context of a full neuron (ie by section you mean 
the main section, to which other sections are attached).
* I'd be nice to have a little explanation on the algorithm for 
generating coordinates (eg what section_randomness means).
* It's written but I think we could stress a bit more the fact that 
attributes x,y,z,diameter do not match x,y,z,diameter given at construction.

mstimberg:
Thanks for the feedback @romainbrette, I tried to address all points in the latest version (http://brian2.readthedocs.org/en/morphology_revisited/user/multicompartmental.html). There is one syntax choice that we did not really decide yet, is this interpretation of the "length" of the morphology really the most intuitive one or should we change something about this (and possibly the other names):

> The number of compartments in a section can be accessed with `morpho.n` (or `morpho.L.n`, etc.), the number of total sections and compartments in a subtree can be accessed with `morpho.n_sections` and `len(morpho)` respectively.

romainbrette:
Ah that's right. The discrepancy between morpho.n and len(morpho) can be 
troubling.
I think we should check how len(morpho) is used in the rest of the code.

thesamovar:
Comments on docs:

* I found the ``morpho.L``, ``.L1`` etc. notation a bit difficult to follow. Maybe a diagram illustrating this would solve the problem?
* I think you should mention right at the top that you can load morphologies from .swc files.

I had a brief look at the code in those two files and from what I can see it looks very good and clear.

mstimberg:
> I found the morpho.L, .L1 etc. notation a bit difficult to follow. Maybe a diagram illustrating this would solve the problem?
> I think you should mention right at the top that you can load morphologies from .swc files.

@thesamovar : please have a look at the updated docs, I moved the morphologies to the top of the "Complex morphologies" section and used the `.L1` etc. notation to introduce the `topology` method which hopefully makes it clearer.

> Ah that's right. The discrepancy between morpho.n and len(morpho) can be troubling. I think we should check how len(morpho) is used in the rest of the code.

@romainbrette : I am not sure what you mean by "how len(morpho) is used in the rest of the code.". Do you mean the examples? We don't use it in the core code anywhere.

After having thought about it, I think having `len(morpho)` correspond to `morpho.n` is more consistent with the rest of Brian2. With `NeuronGroup` and `SpatialNeuron`, you can do things like this:
```Python
G.attribute[:len(G)/2] = value
G.attribute[len(G)/2:] = other_value
```
I.e. the length of an attribute is the length of the group. It is less of an issue for `Morphology` since we don't set any attributes of it after creation, but something like `morpho.diameter[:len(morpho)/2]` would not work with the current semantics. Now, `Morphology` is a bit special anyway, since attributes can not only be attributes of the current section but also complete child sections, but well...
There's another inconsistency: for `NeuronGroup` etc. we use `N` as the attribute, for `Morphology` we use `n`. One option would be to use `n` for the number of compartments in a section and `N` for the total number of compartments in a `Morphology` (which has the advantage that `Morphology.N` would be consistent with `SpatialNeuron.N`), but I am not sure that a distinction just based on lower vs. upper case is a good idea...
Probably explicit names are better, how about:
```Python
morpho.n  # number of compartments in section
len(morpho) # the same
morpho.total_compartments  # total number of compartments in subtree
morpho.total_sections  # total number of sections
```
Finally, we could of course simply raise an error if `len(morpho)` is used (pointing the user to one of the attributes). Actually, maybe this *is* the most explicit and non-ambiguous option?

thesamovar:
Would ``.L1=x, .L2=y`` be the same as ``.LL=x, .LR=y``? If so, maybe state that explicitly, and if not, then there at least to me it's still not obvious what it means.

mstimberg:
> Would .L1=x, .L2=y be the same as .LL=x, .LR=y? If so, maybe state that explicitly, and if not, then there at least to me it's still not obvious what it means.

Ok, I see, I'll try to make this clearer in the documentation: when you create new sections you can give them arbitrary names: `morph.left = ...`, `morph.L = ...`, `morph['1'] = ...` (you cannot use the attribute syntax with a number). There is nothing special about the names `L`, `R`, `1`, etc. here. But if you use an attribute name that only consists of `L`, `R` and numbers, say `L1LR123` it will be interpreted as `['L']['1']['R']['1']['2']['3']`, i.e. each letter/digit refers to a sub-section.
I'm not extremely fond of this syntax, but it has one main use: for morphologies loaded from files, the sections most of the times do not have any useful names and we therefore give them standard names. To refer to a specific section, say at the end of a specific dendrite, you'll likely have to use a fairly long chain of attributes and then something like `morph.LLL3RL` is nicer than `morph.L.L.L['3'].R.L`. I am not entirely convinced that the few saved dots are worth the reduction of explicitness, but well. We never really discussed this syntax (it's not new in this branch), and there are a few issues around this naming (some details are different from the previous approach, I am sure @romainbrette has an opinion on this :-) ):
* Loaded morphologies do not have unique names for sections, e.g. `morph.dend` and `morph.L` might refer to the same thing
* Loading morphologies (or creating them with `Morphology.from_points`) is not the same as creating morphologies "by hand": in the first case, you will get automatic names such as `L`, `R`, `1`, etc., in the second case you won't
* A minor issue: the automatic naming/numbering does not update after manually adding/deleting sections after loading a morphology

thesamovar:
Ah OK, now I understand. I didn't get that from the documentation but this explanation was very clear so if you put something like that in the docs I think it would be great. I don't mind the notation actually. It sort of makes sense to have something like this because there's no particular reason to give branches meaningful 'names' in the majority of cases, I guess.

mstimberg:
> Ah OK, now I understand. I didn't get that from the documentation but this explanation was very clear so if you put something like that in the docs I think it would be great.

I updated the docs, hopefully that makes things clearer.
About the remaining syntax question, what do you (@thesamovar and @romainbrette) think about my last suggestion?
```Python
morpho.n  # number of compartments in section
morpho.total_compartments  # total number of compartments in subtree
morpho.total_sections  # total number of sections in subtree
len(morpho)  # raises an error, something like:
             # the "length of a Morphology is not well defined, use ... instead"
```

thesamovar:
Docs good - no strong feelings on the n/len issue, but seems fine to me.

romainbrette:
Yes good idea about n/len(morpho).

mstimberg:
I did a last refactoring in the creation of morphologies from a list of points (used to load SWC files), it was a bit fragile and did not deal well with incorrect files. The new version should also be a bit clearer. That's all I wanted to do for this branch, if all the tests pass then this is good to be merged.

thesamovar:
Also happy for you to merge this, of course!

thesamovar:
Tests have now all passed so merging!

PR #654, opened 2016-03-08
[MRG] Implement `PopulationRateMonitor.smooth_rate`

### LINKED ISSUE #430: Make `PopulationRateMonitor` more useful
Currently, `PopulationRateMonitor` records the population rate per time bin. This is rarely what users want, e.g. plotting it won't give a smooth PSTH. There are two things to do:

- [ ] allow for binning during the run (which will save memory, even though a single value for a population per time step should not be really of any relevance)
- [ ] have a convenient binning/smoothing option. For this, we could simply convert the `smooth_rate` function from Brian1: https://github.com/brian-team/brian/blob/master/brian/monitor.py#L634

The second part should be trivial, we should do this rather sooner than later, I guess.

mstimberg:
Closes #430

### DIFF
--- a/brian2/monitors/ratemonitor.py
+++ b/brian2/monitors/ratemonitor.py
@@ -1,11 +1,15 @@
 import numpy as np
 
+from brian2.utils.logger import get_logger
 from brian2.core.variables import Variables
 from brian2.units.allunits import second, hertz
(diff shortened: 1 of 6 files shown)


thesamovar:
Small change and all looks good, so merging now.

PR #655, opened 2016-03-09
[MRG] Unit performance improvements

mstimberg:
In general, we do not care much about the performance of `Quantity` operations, since they are not used during the actual simulation. However, the `Morphology` object does use them, and constructing/analyzing/plotting such objects can therefore benefit from performance improvements. This PR includes a few optimisations, most of them are just different ways of doing the same thing (e.g. replacing an `isinstance` test by a try/except), if I am not mistaken there's only one semantical difference: previously, calling `Quantity` with an existing quantity *and* a dimension argument would raise an error if the two were inconsistent (e.g. `Quantity(3*mV, dim=second.dim)`). Now, this will just force the dimension given as the `dim` argument. Raising an error does make more sense, I think, but I don't see any real use case where a user would do this. Small performance gains like this in `Quantity.__new__` have a lot of impact, since this function is called very often (e.g. a simple `0*mV` is more or less equivalent to `Quantity(np.array(0) * np.array(mV), dim=mV.dim)`).

Ready-to-merge if you are happy with the changes. As a practical data point for the benefits: generating random coordinates for the example morphology with ~1000 sections that we use in the "Rall" tests and for illustration in the docs takes ~18s instead of ~44s. This is still slow, but it is a somewhat extreme case in the first place.

### DIFF
--- a/brian2/tests/test_units.py
+++ b/brian2/tests/test_units.py
@@ -130,11 +130,10 @@ def test_get_dimensions():
     assert is_scalar_type(np.float32(5.0))
     assert is_scalar_type(np.float64(5.0))
     assert_raises(TypeError, lambda: get_dimensions('a string'))
-    
     # wrong number of indices
-    assert_raises(ValueError, lambda: get_or_create_dimension([1, 2, 3, 4, 5, 6]))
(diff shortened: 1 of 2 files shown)


thesamovar:
I'm happy with this. Indeed, I doubt any users are using or have ever used the dim argument of Quantity!

PR #668, opened 2016-04-07
[WIP] Various changes for 2.0RC release

thesamovar:
This branch and PR for any little changes to e.g. docs etc. that we want to make before releasing 2.0rc.


thesamovar:
Do you think we should put a warning about the poor RNG for cpp_standalone? For example, we could make it so that if ``rand()`` is called during a ``cpp_standalone`` run it logs a warning. (We could this by making some attribute of the ``rand`` ``Function`` implementation for cpp_standalone be a computed attribute that raises a warning.) We could also put a warning in the docs.

thesamovar:
Added a fix and improvements for #670.

mstimberg:
I fixed a few small things in the docs and added some release notes -- feel free to add to them or correct things you are not happy with!

thesamovar:
Looks good to me!

mstimberg:
I think we agreed about all the changes in this PR and all tests pass, I'll go ahead with the merge.

Issue #554, opened 2015-09-02
Warn about unstable integration


thesamovar:
One of the things we see a lot on the support list is people who are seeing weird plots or getting warnings about undefined values and not understanding where they come from. Usually the problem is that the numerical integrator was unstable at that timestep and for those equations. We could actually detect this fairly easily by running a check on the variables once every second or so of simulated time: if we see values larger than 1e100 say, it's most likely due to an instability and we can fire off a warning with a slightly more detailed explanation and suggestions.

mstimberg:
It seems you mentioned this already as part of #499

thesamovar:
Hah, yes. Oops. :)

Actually I wonder if it's worth having a separate issue for it though: #499 will be quite a bit of work whereas this could quickly and easily be done. What do you think? Happy for it to be closed if you think it's not necessary.

mstimberg:
Ok, let's keep them separate. The unstable integration issue is something that can be implemented with a network operation, i.e. it is something that is independent of code generation etc. (except if we want to have working for standalone as well), whereas the indexing checks and so on are quite specific and need to be included in the template.

thesamovar:
The other option is that we do the check only at the end of a run or (for standalone) when variables are loaded from disk. Like this the same code works for both standalone and runtime.

Issue #626, opened 2016-01-13
simple dynamical systems weirdness


epaxon:
Hi,

I was trying to make some examples of simple 2D linear dynamical systems using brian2. Found some strange behavior/bugs. Here is the code:

```python
from brian2 import *

# dynamic parameters
a = 0.0 / second
b = 1.0 / second
c = -0.5 / second
d = -0.1 / second

eqs_2D = Equations("""
    dx/dt = a * x + b * y : 1
    dy/dt = c * x + d * y : 1
""")

defaultclock.dt = 0.1 * ms
run_time = 40 * second

# This is just to run several simulation in parallel
n_runs = 10
system = NeuronGroup(n_runs, model=eqs_2D, threshold='x>100', reset='x=0; y=0', refractory=0*ms)

# set the initial conditions
system.x = np.random.randn(n_runs)
system.y = np.random.randn(n_runs)

system_monitor = StateMonitor(system, True, record=True)

simple2D = Network()

simple2D.add(system)
simple2D.add(system_monitor)
simple2D.store('init')
simple2D.restore('init')
simple2D.run(run_time)
```

So, this works fine so long as I set [a,b,c,d] such that the system does not have any spirals (stable or unstable). When I set [a,b,c,d] to have spirals (i.e. like it is set right now, which is just damped harmonic oscillator), then the simulation does not work -- every x and y value after the initial condition in the system_monitor is nan. Saddle points and stable stars run fine.

I've tried a ton of stuff to fix this (changing the dt, changing the parameters), but whenever I get a dynamical system that has a spiral it returns only nan values.

I finally did something completely trivial to the equations:

```python
eqs_2D = Equations("""
    dx/dt = a * x + b * y : 1
    dy/dt = c * x + d * y : 1
    dz/dt = 0 / ms : 1
""")
```

and then it works perfectly!

```python
figure('x and y vs t')
subplot(211)
plot(system_monitor.t / second, system_monitor.x.T)
subplot(212)
plot(system_monitor.t / second, system_monitor.y.T)

figure('x vs y')
plot(system_monitor.x.T, system_monitor.y.T)
```

![image](https://cloud.githubusercontent.com/assets/6740515/12283227/7b613256-b959-11e5-8ef0-a66b8efa02b7.png)
![image](https://cloud.githubusercontent.com/assets/6740515/12283231/813978c8-b959-11e5-9a42-50592b85ede5.png)

mstimberg:
[...]

PR #673, opened 2016-04-13
[MRG] Don't info the method choice if linear and time<2s (fixes #663)

### LINKED ISSUE #663: Update examples to specify method to avoid info notices?
Or don't show method choice for method=linear?

thesamovar:
We can still discuss this but writing this only took a moment so thought I should just throw it out there as an option.

### DIFF
--- a/brian2/stateupdaters/base.py
+++ b/brian2/stateupdaters/base.py
@@ -139,9 +139,14 @@ class StateUpdateMethod(object):
             else:
                 msg_text = ("No numerical integration method specified, "
                             "using method '{method}' ({timing}).")
-            logger.info(msg_text.format(group_name=group_name,
-                                        method=the_method,
-                                        timing=timing), 'method_choice')
+            logfunc = logger.info
+            # Don't need to inform the user about method choice if it took a short amount of time to compute
+            # and the method is linear as this is the best possible method.
+            if the_method=='linear' and one_method_time<2.0:
+                logfunc = logger.debug
+            logfunc(msg_text.format(group_name=group_name,
+                                    method=the_method,
+                                    timing=timing), 'method_choice')
             return code
         else:
             if hasattr(method, '__call__'):
--- a/brian2/tests/test_stateupdaters.py
+++ b/brian2/tests/test_stateupdaters.py
@@ -487,9 +487,8 @@ def test_determination():
     # included in this list
     all_methods = ['linear', 'exponential_euler', 'euler', 'heun', 'milstein']
     eqs = Equations('dv/dt = -v / (10*ms) : 1')
-    with catch_logs(log_level=logging.INFO) as logs:
+    with catch_logs(log_level=logging.DEBUG) as logs:
         apply_stateupdater(eqs, variables, all_methods)
-        assert len(logs) == 1
         assert 'linear' in logs[0][2]
     
     # This is conditionally linear


thesamovar:
Oh it fails the test designed to check that the log entry is there. :)

mstimberg:
I am still not sure about this... The reason why I think it makes sense to have the message on the screen is that we want to be explicit and do not hide an important decision like the numerical integration method from the user. It might well be that some changes in Brian (or sympy) will change what method is selected, and this should be obvious for the user. I know that you argue that `linear` is the best choice and it is therefore not necessary to mention it, but we had issues like the code in #626 which fail with linear integration and work fine with euler.

But just visually it makes the output look noisy and I guess users quickly start to ignore it, therefore they won't notice if it changes, anyway. On the other hand, it might motivate users (and us, for the examples) to explicitly specify the integration method. The advantage would be that 1) results should be reproducible across Brian/sympy versions and 2) if you change something in your model so that it is no longer linear, you'll have to make an explicit choice what method to use instead of just falling back to euler.

@romainbrette what do you think about this?
As a quick background in case you did not follow this dicussion: in current master, creating a `NeuronGroup` without explicitly setting `method=...`, you'll get an INFO message like this:
```
INFO       No numerical integration method specified for group 'neurongroup', using method 'linear' (took 0.52s). [brian2.stateupdaters.base.method_choice]
```
(the time is mentioned because the linear method can sometimes take very long for complex equations and it is not always obvious to the user what took so long). This pull request is about removing the message if the method is "linear" (since it is the best method possible in general) and if it took less than 2s (i.e. it is not prohibitively costly to do the symbolic solution).

thesamovar:
Even taking #626 into account I'd still argue that linear is still the in-principle best solution and so we don't need to notify the user. I don't think anyone would guess for example that if it said the integration method was linear, that that might mean it was wrong! That's a bug and not a fault with the way we log information, I'd say.

We should really try to find some way to address #626 by the way. I'd completely forgotten about that.

mstimberg:
> Even taking #626 into account I'd still argue that linear is still the in-principle best solution and so we don't need to notify the user. I don't think anyone would guess for example that if it said the integration method was linear, that that might mean it was wrong! That's a bug and not a fault with the way we log information, I'd say.

I'm not sure. If you get NaNs in your integration and you have a message that tells you that linear integration was chosen for you automatically, you might be more likely to try out a different integration method, no? Maybe we should "train" our users to always switch on the debug mode if something is going wrong, though...

However: if we fix #554 and the warning for an unstable integration includes the chosen method, than probably all this would be less of an issue. I'd probably then have the info message only appear with a time limit, independent of the chosen integration method.

thesamovar:
Last suggestion sounds perfect to me! Seems like the best way to transmit information like that to the user is if we notice something has gone wrong.

mstimberg:
I had a discussion about this with @romainbrette and our consensus was:
* having these messages all the time is not a big deal since you can switch them off, either by specifying the integration method manually, or by using `BrianLogger.suppress_name('method_choice')` (if there's demand for this, we could also turn this into a preference)
* it seems to be justified to show these messages, given that this is the only choice in Brian that we make for the user (apart from the codegen target, but this is more technical)
* switching the messages off for certain methods and/or timing feels a bit arbitrary (why is 1.5s fine, but 2.1s isn't) and will make the output non-deterministic

I know that with the last point I am somewhat contradicting my previous comment, but I feel that it is most in the "spirit of Brian" to never hide an important detail like the integration method.

So here's a proposal: let's leave the output on for all methods and without the time limit (as it is in current master) for now, and re-evaluate before the final release of 2.0. What do you think?

thesamovar:
OK if you both agree to leave the messages I'm happy to be outvoted. ;)

We should probably update the examples though? Maybe not necessary (and definitely not for RC).

mstimberg:
> We should probably update the examples though? Maybe not necessary (and definitely not for RC).

Right, I'll just do it -- it does not trigger the test suite so it will not delay anything.

mstimberg:
I think we decided to actually not do this, and we've adapted the examples so that they don't show the INFO message anymore. I'll therefore go ahead and close this PR.

PR #675, opened 2016-04-14
[MRG] Provide an actual `Unit` (and not a `Quantity`) for the `Variable` object in `check_units_statements`

### LINKED ISSUE #674: Unit check for statements uses `Quantity` instead of `Unit` when creating a `Variable` object
That leads to problems later because the `Variable` object assumes that its argument is a `Unit` or `1` (and not a `Quantity`), leading to a difficult-to-interpret dimension mismatch error.

mstimberg:
In addition, make the `Variable` constructor handle `Quantity` objects as well.
Change one test for the summed variable mechanism to expose this error.
Fixes #674


thesamovar:
Happy for this to be merged as soon as tests pass.

PR #682, opened 2016-04-19
[MRG] Check that the source provided for `SpikeMonitor` is a `SpikeSource`

### LINKED ISSUE #647: SpikeMonitor should check that the source is actually a `SpikeSource`
In particular, you can currently create a `SpikeMonitor` of a `Synapses` object, which will later fail with a hard-to-interpret error message. I think this only works at all "by accident", because in the `Synapses` object you can access the post-synaptic `spikespace` by its name (which you shouldn't). This combined with luck in the indexing (e.g. one-to-one connections) might actually lead to this not raising any error, but it is certainly not doing what the user expects it to do.

mstimberg:
Fixes #647 (see its discussion for details).

### DIFF
--- a/brian2/monitors/spikemonitor.py
+++ b/brian2/monitors/spikemonitor.py
@@ -4,6 +4,7 @@ import numpy as np
 
 from brian2.core.variables import Variables
 from brian2.core.names import Nameable
+from brian2.core.spikesource import SpikeSource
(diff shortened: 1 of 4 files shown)


thesamovar:
Happy for this to be merged once tests pass.

PR #683, opened 2016-04-21
[MRG] Do not alter the stored state of a network when restoring it

### LINKED ISSUE #681: Repeated `Network.restore` calls fail
The following code will fail at the second restore:
```Python
In [1]: from brian2 import *

In [2]: G = NeuronGroup(1, 'v: 1')

In [3]: store()

In [4]: restore()

In [5]: restore()
KeyError                                  Traceback (most recent call last)
<ipython-input-5-ad7bb0bcaaad> in <module>()
----> 1 restore()

/home/marcel/programming/brian2/brian2/core/magic.pyc in restore(name, filename)
    411     Network.restore
    412     '''
--> 413     magic_network.restore(name=name, filename=filename, level=1)
    414 
    415 

/home/marcel/programming/brian2/brian2/core/magic.pyc in restore(self, name, filename, level)
    244         '''
    245         self._update_magic_objects(level=level+1)
--> 246         super(MagicNetwork, self).restore(name=name, filename=filename)
    247         self.objects[:] = []
    248 

/home/marcel/programming/brian2/brian2/core/base.pyc in device_override_decorated_function(*args, **kwds)
    276                 return getattr(curdev, name)(*args, **kwds)
    277             else:
--> 278                 return func(*args, **kwds)
    279 
    280         device_override_decorated_function.__doc__ = func.__doc__

/home/marcel/programming/brian2/brian2/core/network.pyc in restore(self, name, filename)
    431             with open(filename, 'rb') as f:
    432                 state = pickle.load(f)[name]
--> 433         self.t_ = state.pop('0_t')
    434         clocks = set([obj.clock for obj in self.objects])
    435         restored_objects = set()

KeyError: '0_t'
```

The problem seems to be the `state.pop` which pops the `0_t` value from the stored state (instead of e.g. popping it from a copy).

mstimberg:
Fixes #681 

As a side note: I added a commit that mentions the bug (and the users that reported it) already in the next release notes. I think it is a good idea that PRs that introduce a major new feature or fix important bugs do that as part of the PR, that way we'll have less work at release time and it is less likely that we forget something.

### DIFF
--- a/brian2/core/network.py
+++ b/brian2/core/network.py
@@ -430,7 +430,7 @@ class Network(Nameable):
         else:
             with open(filename, 'rb') as f:
                 state = pickle.load(f)[name]
-        self.t_ = state.pop('0_t')
+        self.t_ = state['0_t']
         clocks = set([obj.clock for obj in self.objects])
         restored_objects = set()
         for obj in self.objects:
@@ -446,7 +446,7 @@ class Network(Nameable):
             clock._restore_from_full_state(state[clock.name])
         clock_names = {c.name for c in clocks}
 
-        unnused = set(state.keys()) - restored_objects - clock_names
+        unnused = set(state.keys()) - restored_objects - clock_names - {'0_t'}
         if len(unnused):
             raise KeyError('The stored state contains the state of the '
                            'following objects which were not present in the '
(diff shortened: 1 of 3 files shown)


thesamovar:
OK to merge!

thesamovar:
And, agreed on new PR policy.

mstimberg:
Great, merging it now (the github issue does not show that the test suites successfully ran through because the last commit is `[ci skip]` :-/).

PR #687, opened 2016-04-26
[MRG] Fix the handling of empty non-stochastic parts in equations

### LINKED ISSUE #686: Incorrect integration of SDEs without non-stochastic term
@romainbrette found a quite severe bug in the integration of stochastic equations, which turned out to be a regression (i.e. it worked correctly in 2.0b4). It might be worth doing a 2.0rc2 release that includes the fix.

In current master:
```Python
In [2]: eqs = Equations('dx/dt = xi/sqrt(ms) : 1')
In [3]: print(euler(eqs))
xi = dt**.5 * randn()
_x = 1.0*xi/sqrt(ms)
x = _x
```
Note how `x` is set to a new value every time step.
The same code in 2.0b4 gives the correct result:
```Python
In [2]: eqs = Equations('dx/dt = xi/sqrt(ms) : 1')

In [3]: print(euler(eqs))
xi = dt**.5 * randn()
_x = x + 1.0*xi/sqrt(ms)
x = _x
```

The problem was introduced in commit adafd63, I'll have a look into this.

mstimberg:
Fixes #686 

There were two issues that together led to the bug:
1. The explicit state updater code does not apply the stochastic part `g(x,t) * dW` of a state updater description to an expression if that expression does not have a `...*xi` term. This is fine, the problem is that the same logic was applied to the *non-stochastic* part, but that part is (e.g. for Euler) `x + dt*f(x, t)`, i.e. even if there is no `f(x, t)` in the expression, it should still include the `x`.
2. The above approach was present in the code for a long time but it did not lead to any issue because our code to split an expression into stochastic and non-stochastic part did return `0.0` (and not `None`) for an empty non-stochastic part, meaning that the non-stochastic part of the state updater description was *always* applied.

The second point changed as part of a performance optimisation (replacing the use of sympy's `match` by `collect`) and therefore the problem was introduced.

In this PR, I changed both: the non-stochastic expression will be `0.0` instead of `None` and just to be sure (though this should not be necessary) we'll always apply the non-stochastic part of the state updater description. I updated the test suite as well to guard us from a future regression.

### DIFF
--- a/brian2/equations/codestrings.py
+++ b/brian2/equations/codestrings.py
@@ -153,7 +153,10 @@ class Expression(CodeString):
                                   'stochastic and non-stochastic '
                                   'term') % self.code)
 
-        return (f_expr, stochastic_expressions)
+        if f_expr is None:
+            f_expr = Expression('0.0')
+
+        return f_expr, stochastic_expressions
 
     def _repr_pretty_(self, p, cycle):
         '''
--- a/brian2/stateupdaters/explicit.py
+++ b/brian2/stateupdaters/explicit.py
@@ -513,9 +513,8 @@ class ExplicitStateUpdater(StateUpdateMethod):
         
         non_stochastic, stochastic = expr.split_stochastic()
 
-        if not (non_stochastic is None or non_stochastic_expr is None):
-            # We do have a non-stochastic part in our equation and in the state
-            # updater description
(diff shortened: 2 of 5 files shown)


thesamovar:
Great! Happy for this to be merged.

PR #688, opened 2016-04-27
[MRG] Change the stored scalar value of `dt` immediately when the clock is changed …

### LINKED ISSUE #684: The `dt` attribute referenced by groups can be out-of-sync with the actual dt
If you change the `dt` of a clock, then there is a bit of complicated machinery delaying the actual update until the start of the run (this is to deal with changes in `dt` between runs and the update of the internal clock based on integers). If you ask the clock for its `dt`, then this internal change is hidden from the user, but for the objects using the clock this is not the case. This can be confusing for a user checking the `dt` of a group, but more importantly it also screws up the `PoissonInput` class, which uses the group's `dt` for an internal check at the start of a run (see report on the mailing list: https://groups.google.com/d/topic/brian-development/AcmUIWURBoE/discussion)

Note that there is nothing wrong during the simulation, the groups use the correct clock/dt.

```Python
In [2]: defaultclock.dt = 0.5*ms

In [3]: G = NeuronGroup(1, '')  # uses defaultclock

In [4]: defaultclock.dt
Out[4]: 0.5 * msecond

In [5]: G.clock.dt
Out[5]: 0.5 * msecond

In [6]: G.dt  # <-- the problem
Out[6]: <neurongroup.dt: 100. * usecond>

In [7]: run(0*ms)  # This will consolidate the clock information

In [8]: G.dt  # <-- now all is good
Out[8]: <neurongroup.dt: 0.5 * msecond>
```

mstimberg:
… instead of at the start of the run

For technical reasons, when `dt` is changed we delay the change of the internal integer clock until the start of the run (only then we know whether we start a new simulation at t=0s or continue a previous run). For that we need both the old and the new dt. Previously, we just stored the new `dt` in an internal variable and the `dt` getter returned that value so the user would not notice. Still, this meant that the actual value used during simulation (stored in a 1-element array) was not updated until the start of the run. This gives an inconsistent result when you ask an object directly for its `dt` (instead of asking its a clock), which made a check in `PoissonInput` fail. Now, the approach is the inverse, the new value is immediately set but we remember the previous value. This seems to be safer and fixes the problem with `PoissonInput`.

Fixes #684

### DIFF
--- a/brian2/core/clocks.py
+++ b/brian2/core/clocks.py
@@ -42,10 +42,7 @@ class Clock(VariableOwner):
         # We need a name right away because some devices (e.g. cpp_standalone)
         # need a name for the object when creating the variables
         Nameable.__init__(self, name=name)
-        #: Note that right after a change of dt, this
-        #: will not equal the new dt (which is stored in `Clock._new_dt`). Call
-        #: `Clock._set_t_update_t` to update the internal clock representation.
-        self._new_dt = None
+        self._old_dt = None
         self.variables = Variables(self)
         self.variables.add_array('timestep', unit=Unit(1), size=1,
                                  dtype=np.uint64, read_only=True, scalar=True)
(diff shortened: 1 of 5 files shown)


thesamovar:
Yep, happy for this to be merged.

PR #689, opened 2016-05-02
[MRG] Avoid empty state updaters

mstimberg:
This is a small optimization but something that bothered me for a while. With this PR, groups that do not need a state updater (say `G = NeuronGroup(10, 'v:1')` ) will:
* no longer search for and apply an integration algorithm (i.e. no INFO message either)
* no longer generate "do-nothing" state update code that will be executed at every time step

### DIFF
--- a/brian2/devices/cpp_standalone/device.py
+++ b/brian2/devices/cpp_standalone/device.py
@@ -1036,9 +1036,22 @@ class CPPStandaloneDevice(Device):
 
         # Generate the updaters
         run_lines = ['{net.name}.clear();'.format(net=net)]
+        all_clocks = set()
         for clock, codeobj in code_objects:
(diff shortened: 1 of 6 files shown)


thesamovar:
Oh, yes please! Let's merge this as soon as it passes. :)

mstimberg:
The test suite revealed a quite subtle problem in standalone mode: we do collect the clocks based on the code objects and under some artificial use case (used in one of the tests) this meant that a clock was not advanced but its value used. The last commit should fix this, maybe have a quick look whether this makes sense to you.

thesamovar:
Yep, all good as far as I'm concerned!

mstimberg:
I'll take the liberty and merge this even though the second appveyor test has not yet run through. In PRs like the present one, we are running both test suites twice unnecessarily: one is running the test suite on the branch, the other one is running the test suite after merging the branch into master. But since the branch is on top of master and master did not have any other commits in the meantime, both test suites will be run on the exact same code.

PR #690, opened 2016-05-03
[MRG] Interpret string expressions for the `rates` argument of `PoissonGroup` at every timestep

### LINKED ISSUE #660: PoissonGroup's interpretation of string expressions is very confusing
`PoissonGroup` allows a string expression for the rates, but it evaluates this expression when it is constructed, not at every time step. This is not what the user expects (see the mailing list: https://groups.google.com/d/topic/briansupport/uWXO31OIwDU/discussion), especially for something like `rates='timed_array(t)'`, which will be evaluated at the current time and then used as a constant rate...

It should be very easy to replace the parameter `rates : Hz` used in the `PoissonGroup` by a subexpression `rates = user_expression : Hz` that does what the user expects. At the absolute minimum, the documentation should mention the fact that it does not work as one might expect.

mstimberg:
Following our discussion at #660 I did not add any warning or additional mechanism to use the previous behaviour (but I adapted the documentation).

Closes #660

### DIFF
--- a/brian2/input/poissongroup.py
+++ b/brian2/input/poissongroup.py
@@ -25,7 +25,9 @@ class PoissonGroup(Group, SpikeSource):
         Number of neurons
     rates : `Quantity`, str
         Single rate, array of rates of length N, or a string expression
-        evaluating to a rate
+        evaluating to a rate. This string expression will be evaluated at every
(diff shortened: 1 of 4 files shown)


thesamovar:
Yep, looks good to me.

Issue #306, opened 2014-08-08
Exporting/importing Brian data from/to other formats


mstimberg:
[Following up to my comment on #298.] I think it would be nice if we had a general mechanism to export data from Brian to other formats, so that it is very easy for users to connect Brian with existing analysis tools or store results. I'm not thinking of anything complicated here, data would still be internally stored as numpy arrays (respectively dynamic arrays). For example, each `Group` (i.e. every object that has state variables, including monitors) could have an `export` method:
```Python
syn = Synapses(...)
s_mon = StateMonitor(...)
spike_mon = SpikeMonitor(...)
...
s_mon_data = s_mon.export()  # export all state variables to a format set by a preference
synaptic_weights = syn.export(['w'])  # export a subset of state variables
spikes_neo = spikes.export(format='neo')  # export to a specific format
```
I think the approach above with a general method and a `format` keyword would be a nice approach, we could implement this as  a plugin mechanism, allowing a user (or an external library) to register itself for use in Brian without having to change Brian code itself.

mstimberg:
To add some more details about what I have in mind:
* there should be a simple registration mechanism that allows you to register a new import/export class (look for function names called `register...` in the current Brian2 code base to see examples for such mechanisms)
* The `Group.get_states` and `Group.set_states` method should delegate to a registered class for the given format (or raise an error if it doesn't exist)
* The current implementation of the above methods that has hard-coded support for the `'dict'` target should mostly move into a `DictImportExport` class
* `ImportExport` classes should raise appropriate error messages if they can't fullfill the request, e.g. if they can't deal with physical units but the `units` argument was set to `True`, or if they need an additional library that isn't available.

It would be nice to have at least one additional target in Brian itself, a `pandas` dataframe for example should be both useful and straightforward to implement. Other targets should propably rather go into the brian2 cookbook.

Finally, it would be nice to have the same for a complete network (at least the export part), this could be a simple wrapper around the above mechanism, creating a dictionary with the object names (e.g. `neurongroup`, `statemonitor`, etc.) as keys and the result of `Group.get_state` (i.e., depending on the `format` argument either a dictinonary, or a pandas dataframe, etc.) as the values.

thesamovar:
This all sounds great to me.

ankitkmr:
Hi Marcel

[...]

PR #693, opened 2016-05-07
Issue306 importexport 

dokato:
#306 I implemented submodule importexport with possibility to register new method and rewrote an export to dict functionality from Group.get_states.
There are two files:
* importexport with abstract class 
* explicit with importexport concrete implementation of static classes 

I added also pandas support, but because pandas is not brain2 dependency I decided to import it directly in method - not sure if that's the best solution.


mstimberg:
Thanks, that looks great. I'll make a few nit-picky comments on the code itself, but there are two more general issues that I'd like to have fixed before merging:
* [ ]  There were some small additions to the master branch since you started your work, please merge latest master into your branch
* [ ] Even though pandas is not a standard dependency, we should still have a test for `get_states` with pandas. You can base it on the test `test_neurongroup.test_get_states` and add this at the start:
```Python
try:
    import pandas
except ImportError:
    raise SkipTest('Cannot test export to Pandas data frame, Pandas is not installed.')
```
This way the test will run when pandas is installed and if not it will be skipped.

This said, I just realized we don't actually have a test for `Group.set_states`... If you feel like it, having a test for that would be very much appreciated :-)

brian2/groups/group.py
```diff
@@ -32,6 +32,8 @@
 from brian2.utils.logger import get_logger
 from brian2.utils.stringtools import get_identifiers, SpellChecker
```
mstimberg: remove the empty line here to group all `brian2...` imports

brian2/groups/group.py
```diff
@@ -511,18 +513,14 @@ def get_states(self, vars=None, units=True, format='dict',
         '''
         # For the moment, 'dict' is the only supported format -- later this will
         # be made into an extensible system, see github issue #306
-        if format != 'dict':
+        if format not in ImportExport.methods.keys():
```
mstimberg: a bit shorter (and more idiomatic): `if format not in ImportExport.methods:`

brian2/groups/group.py
```diff
@@ -543,10 +541,9 @@ def set_states(self, values, units=True, format='dict', level=0):
...
```
mstimberg: same as above

brian2/importexport/explicit.py
```diff
@@ -0,0 +1,53 @@
...
+    '''
+    An importer/exporter for variables in format of dict of numpy arrays.
+    '''
+    @property
+    def name(self):
+        return "dict"
+
+    @staticmethod
+    def export_data(group, variables, units=True, level=0):
+        # taken from get_states
```
mstimberg: this comment will be hard to understand in the future, I'd say just remove it

brian2/importexport/explicit.py
```diff
@@ -0,0 +1,53 @@
...
+        data = {}
+        for var in variables:
+            data[var] = np.array(group.state(var, use_units=units,
+                                            level=level+1),
+                                 copy=True, subok=True)
+        return data
+
+    @staticmethod
+    def import_data(group, data, units=True, level=0):
+        # taken from set_states
```
mstimberg: same here

brian2/importexport/explicit.py
```diff
@@ -0,0 +1,53 @@
...
+    An importer/exporter for variables in pandas DataFrame format.
+    '''
+
+    @property
+    def name(self):
+        return "pandas"
+
+    @staticmethod
+    def export_data(group, variables, units=True, level=0):
+        # as pandas is not a default brian2 dependency we import it only in that namespace
```
mstimberg: Yes, this is exactly what we'd like to do, but please wrap it for a nicer error message:
```Python
try:
    import pandas
except ImportError as ex:
    raise ImportError('Exporting to pandas needs a working installation of pandas. '
                      'Importing pandas failed: ' + str(ex))
```

brian2/importexport/explicit.py
```diff
@@ -0,0 +1,53 @@
...
+        import pandas as pd
+        # we take adventage of already implemented exporter
+        data = DictImportExport.export_data(group, variables,
+                                            units=units, level=level)
+        return pd.DataFrame(data)
+
+    @staticmethod
+    def import_data(group, data, units=True, level=0):
+        # as pandas is not a default brian2 dependency we import it only in that namespace
+        import pandas as pd
```
mstimberg: see above

brian2/importexport/explicit.py
```diff
@@ -0,0 +1,53 @@
...
+        colnames = data.columns
+        array_data = data.as_matrix()
+        for e, colname in enumerate(colnames):
+            group.state(colname, use_units=units, level=level+1)[:] = array_data[:,e]
```
mstimberg: really nit-picking now: `array_data[:, e]` should have a space after the comma :-)

brian2/importexport/importexport.py
```diff
@@ -0,0 +1,74 @@
...
+    @staticmethod    
+    def register(name, importerexporter):
+        '''
+        Register a import/export method. Registered importerexporter can be referred to
+        via their name.
+        
+        Parameters
+        ----------
+        name : str
+            A short name for the import/export class (e.g. `'dict'`)
```
mstimberg: Need double-backtick quotes here in RST for `'dict'` (single backticks are to refer to existing classes/functions)

brian2/importexport/importexport.py
```diff
@@ -0,0 +1,74 @@
...
+            The importerexporter object, e.g. an `DictImportExport`.
+        '''
+        
+        name = name.lower()
+        if name in ImportExport.methods:
+            raise ValueError(("An import/export methods with the name {}" +\
+                             "has already been registered").format(name))
+
+        if not isinstance(importerexporter, ImportExport):
+            raise ValueError(('Given importerexporter of type {} does not seem to ' +\
```
mstimberg: same here for the line continuation. Also for consistency (we are not very consistent over the Brian 2 code base, though...) use either single or double quotes for strings in the function.

mstimberg:
[It might not be obvious, given that I added so many comments, but I am really happy with the quality of the code :+1:  ]

dokato:
Thanks for feedback!

> You can base it on the test test_neurongroup.test_get_states and add this at the start:

Are you sure that it should be in `test_get_states`? Because I think that we don't want to omit all other  tests only because of pandas. Maybe I should define some dedicated `test_set_get_states_pandas` case just for pandas functionality with the try/except statement you mentioned?

mstimberg:
> Are you sure that it should be in test_get_states? Because I think that we don't want to omit all other tests only because of pandas.

No, I did not mean it to be in it, just "based on it", in the sense that it could be almost identical. So ideally I'd like to have four tests: `test_get_states`, `test_set_states`, `test_get_states_pandas`, `test_set_states_pandas`. Hope that makes thinks clearer.

dokato:
> Hope that makes thinks clearer.

Yes, thank you!

BTW while implementing test cases I noticed that `get_states` and `set_states` are not symmetric. What I mean is shown in that example:
```
G = NeuronGroup(10, '''v : volt
                       x : 1
                       subexpr = x + v/volt : 1
                       subexpr2 = x*volt + v : volt''')
G.v = 'i*volt'
G.x = '10*i'
q = G.get_states(['v','N'])
G.set_states(q)
```
which gives `TypeError: 'int' object does not support item assignment`

So basically we can export some constant variables to a dict, but then cannot import it. I don't know if it actually makes sense, so I would be grateful to hear some comments on that.

mstimberg:
> So basically we can export some constant variables to a dict, but then cannot import it. I don't know if it actually makes sense, so I would be grateful to hear some comments on that.

There are certainly a few rough edges around those two functions. At least the error message should be clearer... In this case, setting `N` will not be possible, since `N` is a read-only variable (the number of neurons in the group), you can't do `G.N = ...` either.  We do have the `read_only_variables` argument to `get_states`, so we kind of guarantee that you can do `G.set_states(G.get_states(read_only_variables=False))` (which would exclude `N`):
```
        read_only_variables : bool, optional
            Whether to return read-only variables (e.g. the number of neurons,
            the time, etc.). Setting it to ``False`` will assure that the
            returned state can later be used with `set_states`. Defaults to
            ``True``.
```
It is entirely possible that this fails for scalar variables (marked as `(shared)` in the equations), though, despite this promise...

brian2/tests/test_neurongroup.py
```diff
@@ -1206,6 +1206,95 @@ def test_get_states():
...
+    G = NeuronGroup(10, '''v : volt
+                           x : 1
+                           subexpr = x + v/volt : 1
+                           subexpr2 = x*volt + v : volt''')
+    G.v = 'i*volt'
+    G.x = '10*i'
+    states_units = G.get_states(['v', 'x', 'subexpr', 'subexpr2'], units=True, format='pandas')
+    states = G.get_states(['v', 'x', 'subexpr', 'subexpr2'], units=False, format='pandas')
+    assert len(states_units.columns) == len(states.columns) == 4
+    assert_equal(states_units['v'].as_matrix(), np.arange(10))
```
mstimberg: A minor issue: I think the use of `.as_matrix()` is deprecated (it is confusing because it returns a numpy array, not a numpy matrix...), you can use `.values` instead.

mstimberg:
Hey, sorry I did not see that you made changes (in case of non-trivial additions to a PR it is the best to make a quick comment which will trigger a mail). The tests are looking good (I made a small comment), but I would prefer if you take the unit-support out again. I appreciate the idea behind it, but I think the solution is too fragile (it depends on the order of columns, it will be lost after any transformation of the data frame, e.g. sorting it, etc.). I cannot think of any better solution either, so I think it would be best to be clear about the fact that we don't support units when exporting with pandas. The most straightforward way would be to raise a `NotImplementedError` when `units` equals `True`, I guess.

mstimberg:
Ok, this looks all good to me. I'll do some more testing and will then merge it.

mstimberg:
All the tests pass, all looking good --> merging now.

PR #697, opened 2016-05-11
[MRG] Fix references to synaptic variables when setting delays on synaptic pathways

### LINKED ISSUE #694: Synaptic pathway delays cannot use synaptic/neuronal variables when set with a string
[See: https://groups.google.com/d/topic/brian-development/xYRoHNRNE3E/discussion]
When you set the delay variable with a string, you can refer to synaptic and pre-/post-synaptic neuronal parameters, as for all other synaptic variables. This is because such statements are executed in the context of the `Synapses` object. The same thing does however not work if you explicitly set the delay of the pathway, e.g. with `synapses.pre.delay = '...'`. If there is only one pre-synaptic pathway, then you can use `synapses.delay = '...'` instead, but there is no alternative for the post-synaptic pathway or when using several pre-/post-synaptic pathways.

mstimberg:
This needed a few iterations to get things working correctly for subgroups, but now it should be fine and ready to merge.

Fixes #694

### DIFF
--- a/brian2/core/variables.py
+++ b/brian2/core/variables.py
@@ -1682,8 +1682,7 @@ class Variables(collections.Mapping):
         self._add_variable(name, var)
 
 
-    def add_referred_subexpression(self, name, group, subexpr, index,
-                                   ignore_errors=False):
+    def add_referred_subexpression(self, name, group, subexpr, index):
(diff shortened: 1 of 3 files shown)


thesamovar:
Good to go I think.

PR #698, opened 2016-05-13
[MRG] Do not lose precision when inserting floating point values into standalone code

### LINKED ISSUE #695: 1-timestep difference when running certain simulations in standalone mode
The following code will run for 250 time steps in runtime mode but for 251 steps in standalone:
```Python
from brian2 import *
#set_device('cpp_standalone')
defaultclock.dt = 20.000000000020002*us
G = NeuronGroup(1, 'v:1')
run(defaultclock.dt*250)
print len(mon.t)
```

The example may look contrived, but @romainbrette stumbled across this in a real use case, the weird dt value was of course not specified manually but taken from the difference between two time stamps in a file.

The reason for this issue is actually quite simple: we insert the desired runtime as a string into the standalone code, using default string formatting which does not preserve the full precision. We should instead force the use of the `repr` representation:
```Python
dt_float = float(20.000000000020002*us)
print('{t}'.format(t=250*dt_float))
print('{t!r}'.format(t=250*dt_float))
```
```
0.00500000000001
0.0050000000000050004
```

I had a quick check and it seems for other constants we are already using the conversion via `repr`, so they should be fine.

mstimberg:
As discussed in #695, using the "str" instead of "repr" representation of the floating point value of the total runtime led to a one-timestep difference under certain (quite specific) circumstances. This should be fixed with this PR.

Fixes #695

### DIFF
--- a/brian2/devices/cpp_standalone/device.py
+++ b/brian2/devices/cpp_standalone/device.py
@@ -1052,7 +1052,7 @@ class CPPStandaloneDevice(Device):
             if clock not in all_clocks:
                 run_lines.append('{net.name}.add(&{clock.name}, NULL);'.format(clock=clock, net=net))
 
-        run_lines.append('{net.name}.run({duration}, {report_call}, {report_period});'.format(net=net,
+        run_lines.append('{net.name}.run({duration!r}, {report_call}, {report_period!r});'.format(net=net,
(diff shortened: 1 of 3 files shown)


thesamovar:
Good - merging!

PR #699, opened 2016-05-17
[MRG] Improve memory usage

mstimberg:
As reported in this [mailing list thread](https://groups.google.com/d/topic/brian-development/CK76RuDHRbg/discussion), there is a memory leak in Brian 2.0rc1. It turned out that there were actually two issues: unit objects were unnecessarily stored over an over again in the unit registry (this is what you see in the output from the `pympler` module). However, this was only a minor memory leak (and one that existed, albeit less severe, in earlier versions). The main memory leak was in the spike propagation template for weave, where we received a reference to a numpy array from the spike queue and did not `PY_DECREF` it in the end. This memory leak would not matter in most use cases, but by using a loop over runs and storing/restoring the network, many spike queues were generated and the memory leak became severe.

### DIFF
--- a/brian2/codegen/runtime/weave_rt/templates/synapses.cpp
+++ b/brian2/codegen/runtime/weave_rt/templates/synapses.cpp
@@ -24,4 +24,5 @@
 
     // Advance the spike queue
     PyObject_CallMethod(_queue, "advance", "");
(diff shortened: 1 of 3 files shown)


thesamovar:
Good to merge.

thesamovar:
And nice work tracking all that down!

PR #702, opened 2016-05-25
[MRG] Use `std::make_pair` without template types

### LINKED ISSUE #701: cpp_standalone compilation fails with C++14 (default on gcc 6)
Using an up2date Arch Linux the following error occurs during (unsuccessful) compilation of the STDP_standalone example with brian2.0rc1 installed via pip (but the same error happens also with brian2.0b4). Any idea/further information required?
```
network.cpp: In member function ‘void Network::add(Clock*, codeobj_func)’:
network.cpp:29:71: error: no matching function for call to ‘make_pair(Clock*&, void (*&)())’                        
     objects.push_back(std::make_pair<Clock*, codeobj_func>(clock, func));                                          
                                                                       ^                                            
In file included from /usr/include/c++/6.1.1/bits/stl_algobase.h:64:0,                                              
                 from /usr/include/c++/6.1.1/vector:60,                                                             
                 from network.h:5,                                                                                  
                 from network.cpp:2:                                                                                
/usr/include/c++/6.1.1/bits/stl_pair.h:425:5: note: candidate: template<class _T1, class _T2> constexpr std::pair<typename std::__decay_and_strip<_Tp>::__type, typename std::__decay_and_strip<_T2>::__type> std::make_pair(_T1&&, _T2&&)                                                                                                                  
     make_pair(_T1&& __x, _T2&& __y)                                                                                
     ^~~~~~~~~                                                                                                      
/usr/include/c++/6.1.1/bits/stl_pair.h:425:5: note:   template argument deduction/substitution failed:              
network.cpp:29:71: note:   cannot convert ‘clock’ (type ‘Clock*’) to type ‘Clock*&&’                                
     objects.push_back(std::make_pair<Clock*, codeobj_func>(clock, func));                                          
                                                                       ^                                            
make: *** [network.o] Error 1
```

See also: [brian_debug_2az2_E.log.txt](https://github.com/brian-team/brian2/files/279391/brian_debug_2az2_E.log.txt)

mstimberg:
I think the general idea of `make_pair` is that it infers its types automatically from the arguments (otherwise you would create the pair explicitly with `std::pair<T1, T2>`). Apparently leaving away the types works on gcc and msvc, while having them does not work with gcc 6 (in its default C++14 mode).

Fixes #701

### DIFF
--- a/brian2/devices/cpp_standalone/templates/network.cpp
+++ b/brian2/devices/cpp_standalone/templates/network.cpp
@@ -25,9 +25,9 @@ void Network::clear()
 void Network::add(Clock* clock, codeobj_func func)
 {
 #if defined(_MSC_VER) && (_MSC_VER>=1700)
-    objects.push_back(std::make_pair<Clock*, codeobj_func>(std::move(clock), std::move(func)));
+    objects.push_back(std::make_pair(std::move(clock), std::move(func)));
 #else
-    objects.push_back(std::make_pair<Clock*, codeobj_func>(clock, func));
+    objects.push_back(std::make_pair(clock, func));
 #endif
 }
 


thesamovar:
Sometimes I hate C++ templates.

This change seems absolutely fine to me if everything passes.

mstimberg:
> Sometimes I hate C++ templates.

They still remain a mystery to me in all but the simple situations...

Anyway, tests pass on all OS, merging it now.

Issue #33, opened 2013-04-22
Run automatic benchmarks


mstimberg:
We should port Brian's mechanism for automatically running benchmarks and come up with a couple of benchmarks, both micro-benchmarks and realistic model simulations (e.g. the classical CUBA, COBA sims). This should include comparing the different code generation targets and maybe comparisons to as a reference.

It would also be good to include memory usage in addition to speed, but I'm not sure how feasible that is.

This issue is not a high priority, though, first we need to make the simulations run, then we can worry about performance :)

mstimberg:
The tool "airspeed velocity" (http://asv.readthedocs.io/en/latest/) seems to be what the cool kids are using nowadays (see e.g. https://pv.github.io/numpy-bench/).

thesamovar:
Looks like a nice tool indeed.

mstimberg:
I've setup a basic framework for benchmarking in the `brian-team/brian2_benchmarks` repository. It seems to work fine so far. I added a few "full" benchmarks (examples like CUBA and COBAHH, but nothing yet for multicompartmental modeling), and just an example for a "microbenchmark". I think it is good to have a few complex examples that track real-world performance and can show regressions/improvements that are not covered by a microbenchmark. On the other hand, we shouldn't add too many of them, as they generally take quite a while to execute. Also, microbenchmarks of course give a much clearer idea of where a performance improvement/regression is coming from.

I'll run a few example benchmarks and publish the results on a fancy github page (the `asv` tool takes care of that).

thesamovar:
Would it be useful to have a dedicated (older) machine for this sort of thing?

mstimberg:
In principle it could be useful, we could have it run the benchmarks for each and every commit. But I'm not 100% sure that it is worth the effort.  Either way, the tool keeps track of the machine that was used for running the benchmarks -- it might be useful to at least run things on a Windows and a Linux machine, maybe?

mstimberg:
> Well I could probably find an unused older machine lying around here somewhere that could be dedicated to this task, but only if it's worth the effort and it sounds like maybe not?

We'd probably have a machine that could be used for this as well... From my experience, configuring a machine to continuously run these things has a bit of maintenance (not to speak of electricity ;) ) cost, there are updates, etc. I think at least for now I'd prepare to run benchmarks on my machine every now and then, e.g. when we plan to do a new release, similar to what we do for the examples.

[...]

PR #703, opened 2016-05-27
[MRG] Faster before run

mstimberg:
While looking into #699 I noticed that the code in 2.0rc1 not only had a memory leak but was also running somewhat slower than in 2.0b4. The test code was certainly a bit extreme (very short run repeated many times), but similar use cases are common. While profiling I noticed that the namespace resolution took a lot of time, because we'll ask for the stack frame again and again, whenever we resolve names in code (even if we resolve the names with internal names, we still check for name clashes with external names to issue a warning). This pull request changes this and gets the local namespace once at the earliest point of time (most importantly in `run()`, but also when setting a variable with a string expression or when connecting synapses). The namespace dictionary is then passed down (instead of passing down the level argument, increasing it by 1 at every step and getting the namespace in the end). I also added a few smaller optimisations.

I did not run extensive benchmarking on this, but it is certainly faster than current master. Compared to 2.0b4 I am not that sure, I got some contradicting results (quite a bit changed since then, you have to re-compile the Cython spike queue, etc.), but I think it is not worth spending more time on this right now.
In general it would of course be nice to set up some automatic benchmarking (#33) so we have a clearer picture of where we stand.

### DIFF
--- a/brian2/codegen/codeobject.py
+++ b/brian2/codegen/codeobject.py
@@ -103,60 +103,6 @@ class CodeObject(Nameable):
         raise NotImplementedError()
 
 
-def check_code_units(code, group, user_code=None, additional_variables=None,
-                     level=0, run_namespace=None,):
-    '''
-    Check statements for correct units.
-
-    Parameters
-    ----------
-    code : str
-        The series of statements to check
-    group : `Group`
-        The context for the code execution
-    user_code : str, optional
-        The code that was provided by the user. Used to determine whether to
-        emit warnings and for better error messages. If not specified, assumed
-        to be equal to ``code``.
-    additional_variables : dict-like, optional
-        A mapping of names to `Variable` objects, used in addition to the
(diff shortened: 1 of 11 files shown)


thesamovar:
This is good and happy for this to be merged once tests pass.

Is there any potential breaking of projects like GeNN that use our internal API?

mstimberg:
> Is there any potential breaking of projects like GeNN that use our internal API?

Indeed, good point. I had a quick look at the code of @brian-team/brian2genn  and @brian-team/brian2cuda and their code does not seem to be effected, everything should be taken care of by inheriting from `CPPStandaloneDevice`. If anything, the necessary changes should be comparable to https://github.com/brian-team/brian2/pull/703/files#diff-e5dac9a6ab63eac91d799ae9988f052e, i.e. to a removal of the `level` argument. 

I'll therefore go ahead with the merge, all tests passed.

PR #705, opened 2016-06-01
 Remove hard-coded assumption that summed variable is post-synaptic

### LINKED ISSUE #704: `(summed)` flag does not work correctly for pre-synaptic variables
In principle, our syntax also allows the `(summed)` flag for pre-synaptic variables. This should set this variable for each pre-synaptic neuron to the sum of the expression on the right-hand-side, evaluated for all the synapses originating from the respective neuron. However, it seems we (meaning: I :-( ) have hard-coded the indexing over post-synaptic variables into the templates. This means this will most likely do something completely wrong...

The easiest way to see this is when connecting two groups of differing size. The following will raise an error with the numpy target and abort/segfault in weave/cython:
```Python
G1 = NeuronGroup(1, 'syn_sum : 1')
G2 = NeuronGroup(10, '')

synapses = Synapses(G1, G2, '''syn_var : 1
                               syn_sum_pre = syn_var : 1 (summed)''')
synapses.connect(True)

run(defaultclock.dt)
```

mstimberg:
Remove hard-coded assumption that summed variable is post-synaptic
Fixes #704

### DIFF
--- a/brian2/codegen/runtime/cython_rt/templates/summed_variable.pyx
+++ b/brian2/codegen/runtime/cython_rt/templates/summed_variable.pyx
@@ -2,23 +2,23 @@
 
 {% block maincode %}
 
-    {# USES_VARIABLES { _synaptic_post, N_post } #}
+    {# USES_VARIABLES { N } #}
     {% set _target_var_array = get_array_name(_target_var) %}
-
+    {% set _index_array = get_array_name(_index_var) %}
     cdef int _target_idx
 
     # Set all the target variable values to zero
-    for _target_idx in range(N_post):
+    for _target_idx in range({{_target_size_name}}):
         {{_target_var_array}}[_target_idx] = 0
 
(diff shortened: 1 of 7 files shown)


thesamovar:
This all looks good to me and ready to merge if tests pass.

Could break GeNN / brian2cuda? Should we notify them on merge?

mstimberg:
> Could break GeNN / brian2cuda? Should we notify them on merge?

Well, I think none of them work with current master anyway (they are both using a "frozen" version of Brian 2). Having said that, this PR should not change anything for them (it will only provide additional keywords to the template, all other changes are within the templates). I did not look into it but I suspect that the `summed_variable` template in the @brian-team/brian2cuda repository suffers from the same bug that I fixed here (given that it was modelled after the C++ standalone template). brian2genn does not support summed variables in the first place.

All tests pass, merging it now.

PR #712, opened 2016-06-09
[MRG] Fix pathway variables and other small issues

mstimberg:
This PR collects fixes for a few issues that were discovered in the context of synaptic pathways (in recent mails to the mailing list):
* nested subexpressions led synaptic pathways to fail because of some unnecessarily complicated attempts to link to them (references to subexpressions are a bit tricky in general because they refer to other variables in the source group, not in the group where they are linked from -- this means we cannot simply use the subexpression literally).
* not related to the synaptic pathway, but it manifested for a user on the mailing list in this context: our conditional write mechanism is a bit fragile, since we include the conditional write variable (i.e. `not_refractory` for `v`) under this name in the variables dictionary. This can clash with an already defined variable of the same name (e.g. in a `StateMonitor` that records both `v` and `not_refractory`). Simply renaming the variable (e.g. use `_cond_write_not_refractory`) does not work, because then the state update does not work correctly. The first thing it does is to set `not_refractory` and if we have included the conditional write variable under a different name, then that one is not updated. I did not find a straight-forward fix, so now the problem is basically ignored (with a DIAGNOSTIC message). I only found situations where this is a false positive (e.g. in the `StateMonitor`, the conditional write is irrelevant since we do not write to `v`) and none where it actually poses a problem. A better fix would be to make the relevant code path aware of what variables are written to and which are not, but this needs some major refactoring.

During test-writing I stumbled across another bug that I fixed here, even though it is not directly related to synaptic pathways:
* There was a problem in the loop-invariant optimisation where the variables dictionary was not shared between the different renderers. This mean that an LIO variable was always considered to be float (the default), which led to errors when it was actually boolean and used in a boolean context.

The two tests that I added cover these three issues.

### DIFF
--- a/brian2/parsing/bast.py
+++ b/brian2/parsing/bast.py
@@ -207,7 +207,6 @@ class BrianASTRenderer(object):
         node.dtype = 'boolean'
         for subnode in node.values:
             if subnode.dtype!='boolean':
-                print subnode
                 raise TypeError("Boolean operator acting on non-booleans")
         node.scalar = logical_all(subnode.scalar for subnode in node.values)
         node.complexity = 1+sum(subnode.complexity for subnode in node.values)
--- a/brian2/parsing/bast.py
+++ b/brian2/parsing/bast.py
@@ -109,8 +109,11 @@ class BrianASTRenderer(object):
     '''
     This class is modelled after `NodeRenderer` - see there for details.
     '''
-    def __init__(self, variables):
-        self.variables = variables.copy()
+    def __init__(self, variables, copy_variables=True):
+        if copy_variables:
+            self.variables = variables.copy()
+        else:
+            self.variables = variables
 
     def render_node(self, node):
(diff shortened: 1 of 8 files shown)


brian2/parsing/bast.py
```diff
@@ -204,6 +207,7 @@ def render_BoolOp(self, node):
         node.dtype = 'boolean'
         for subnode in node.values:
             if subnode.dtype!='boolean':
+                print subnode
```
thesamovar: Remove this debugging code

thesamovar:
Ah, I'm glad you fixed that boolean problem!

All good to merge except one tiny change above.

brian2/parsing/bast.py
mstimberg: Debugging with print statements, who would do such a thing ;) Thanks for spotting this.

Issue #692, opened 2016-05-04
Some logging preferences are not taken into account


mstimberg:
Basic logging preferences (such as whether to log to a file or not) are completely ignored (i.e. their defaults are always used) because they are checked during the import of the logger module and this happens before the parsing of the preferences file. We should probably not try to a complex setup like this directly in the module, but instead have an explicit `initialize_logger()` call  and call that after the import of all Brian packages.

Also see #691

Issue #691, opened 2016-05-04
Improve logging to files (and its documentation)


mstimberg:
A user noted very big log files, we should probably switch the default to `DEBUG` instead of `DIAGNOSTIC`. Other possible improvements:
* The documentation for the logging should be easier to find (it is hidden in the developer documentation)
* We have to make clear that some preferences are only checked at initialization time (or we try to deal with these preferences more elegantly)
* We need a preference to switch off the writing of stdout/stderr to a file (currently you can only switch on/off whether they are displayed on screen)
* For the stderr/stdout files, the temporary file names should be prefixed with `brian_stdout` and `brian_stderr` or something like that to make clear that they are generated by Brian.

PR #714, opened 2016-06-10
[MRG] Logger improvements

mstimberg:
This fixes/improves the issues mentioned in #692 and #691. Most importantly, changing the logging preferences via a file now works (previously, the logging system was initialized before preferences files were loaded). Additional changes:
* new preference to switch off writing stdout/stderr to a file `logging.std_redirection_to_file`
* the stdout/stderr filenames are prefixed with `brian_stdout`/`brian_stderr` to be more consistent with `brian_debug...` and `brian_script...`
* The documentation of the logging system is now split into a part for users and a part for developers, I also added a warning about setting basic logging preferences (which only works via a file, not in the script itself)

### DIFF
--- a/brian2/__init__.py
+++ b/brian2/__init__.py
@@ -102,3 +102,6 @@ for name, version in [('numpy',  '1.8.2'),
                       ('sympy',  '0.7.6'),
                       ('jinja2', '2.7')]:
     _check_dependency_version(name, version)
+
+# Initialize the logging system
+BrianLogger.initialize()
--- a/brian2/tests/__init__.py
+++ b/brian2/tests/__init__.py
@@ -145,9 +145,9 @@ def run(codegen_targets=None, long_tests=False, test_codegen_independent=True,
         prefs.read_preference_file(StringIO(prefs.defaults_as_file))
 
     # Suppress INFO log messages during testing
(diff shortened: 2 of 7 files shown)


thesamovar:
Looks good to me. One possible change, maybe add something in the 'user guide' section as well as the 'advanced' section. It could be just one paragraph mentioning in particular the info level and the method choice. Other than that, good to merge.

mstimberg:
> One possible change, maybe add something in the 'user guide' section as well as the 'advanced' section. It could be just one paragraph mentioning in particular the info level and the method choice.

Good idea, I'll add a note in the section about numerical integration (http://brian2.readthedocs.io/en/2.0rc1/user/models.html#numerical-integration) -- I don't expect that will stop anyone from asking "why do I get this warning" on the mailing list, though ;)

thesamovar:
I agree, but we have to try! :)

Actually that makes me think, we should have a frequently asked questions section in the documentation, and make it so that to find your way to the support mailing list, you have to go to the FAQ section first.

mstimberg:
> Actually that makes me think, we should have a frequently asked questions section in the documentation, and make it so that to find your way to the support mailing list, you have to go to the FAQ section first.

In some way we take good care of this in the documentation -- we never mention the mailing list (apart from remarks in the release notes...)!

PR #715, opened 2016-06-13
[MRG] Synapse to synapse fix

mstimberg:
I realized that our changes to synapse creation did actually break the "synapse-to-synapse" connections we need for the astrocytes project. We had a test for that, but that test used 1-to-1 connections everywhere so the use of incorrect indices went unnoticed. The change is to not rely on the pre-defined arrays of `arange(N)` that every `NeuronGroup` has but instead create them on the fly -- performance-wise it should not change anything (addition of two variables instead of an array lookup), except maybe for numpy (because we now do an explicit check instead of relying on an automatic `IndexError` -- however, this has the advantage that the error message is now clearer and consistent with weave/cython). I did not notice anything significant in some quick testing.

The names `all_pre` and `all_post` are not ideal, maybe we should also use the opportunity to rename them to something more understandable...?

### DIFF
--- a/brian2/codegen/runtime/cython_rt/templates/synapses_create_generator.pyx
+++ b/brian2/codegen/runtime/cython_rt/templates/synapses_create_generator.pyx
@@ -14,6 +14,8 @@ cdef int _buffer_size = 1024
 cdef int[:] _prebuf = _numpy.zeros(_buffer_size, dtype=_numpy.int32)
 cdef int[:] _postbuf = _numpy.zeros(_buffer_size, dtype=_numpy.int32)
 cdef int _curbuf = 0
+cdef int _all_pre
+cdef int _all_post
 
 cdef void _flush_buffer(buf, dynarr, int buf_len):
     _curlen = dynarr.shape[0]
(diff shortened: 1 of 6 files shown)


thesamovar:
Looks good to me. What would you prefer to ``all_pre`` and ``all_post``? I think they're but but I don't mind changing them unless that might break other packages code (e.g. genn).

mstimberg:
> What would you prefer to all_pre and all_post?

I don't have a really good idea, the difference in the code is that `i` and `j` refer to the variables as used by a user (e.g. when connecting to from `group[5:]`, `i == 0` would correspond to `group[5]`), where `all_pre` and `all_post` refer to the actual indices into the array (i.e. `5` for the above example). Maybe `real_pre_idx`/`actual_pre_idx`/`full_pre_ix` or something like that?

> I think they're but but I don't mind changing them unless that might break other packages code (e.g. genn).

I think it can't break anything that we did not already break with the switch to the generator-based approach...

thesamovar:
Ah I see. I don't really like real/actual/full in this case, but not quite sure why. How about ``raw_idx_pre``, etc.? Or ``raw_offset_pre``?

mstimberg:
"Raw" sounds good to me, I would rather call it index and not offset. I'll make this change.

mstimberg:
I went for `_raw_pre_idx` instead of `_raw_idx_pre` because this was more consistent with other variables ending in `_idx`.

thesamovar:
Great - good to merge.

PR #716, opened 2016-06-13
SpikeGeneratorGroup: period and np.max(times) comparison error

martinosorb:
I don't know if I'm being very naive here and I'm missing something; but whenever and however I use SpikeGeneratorGroup, it returns an DimensionsMismatchError.
The code imposes "period" to be a quantity in seconds. However, np.max(times) outputs a unitless quantity (numpy converts everything to numpy.array), so brian itself complains it cannot compare them.
I couldn't find any workaround at all, except changing the code.
Am I missing something?
Thanks!

### DIFF
--- a/brian2/input/spikegeneratorgroup.py
+++ b/brian2/input/spikegeneratorgroup.py
@@ -14,7 +14,10 @@ __all__ = ['SpikeGeneratorGroup']
 
 class SpikeGeneratorGroup(Group, CodeRunner, SpikeSource):
     '''
-    SpikeGeneratorGroup(N, indices, times, dt=None, clock=None, period=1e100*second, when='thresholds', order=0, sorted=False, name='spikegeneratorgroup*', codeobj_class=None)
+    SpikeGeneratorGroup(N, indices, times, dt=None, clock=None,
+                        period=1e100*second, when='thresholds', order=0,
+                        sorted=False, name='spikegeneratorgroup*',
+                        codeobj_class=None)
 
     A group emitting spikes at given times.
 
@@ -71,6 +74,7 @@ class SpikeGeneratorGroup(Group, CodeRunner, SpikeSource):
 
         self.codeobj_class = codeobj_class
 
+        times = Quantity(times)
         if N < 1 or int(N) != N:
             raise TypeError('N has to be an integer >=1.')
         N = int(N)  # Make sure that it is an integer, values such as 10.0 would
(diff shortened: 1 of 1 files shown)


martinosorb:
The errors in appveyor seem to have nothing to do with my changes.

```Fetching package metadata: .......Using Anaconda Cloud api site https://api.anaconda.org
Error: HTTPError: 504 Server Error: GATEWAY_TIMEOUT for url: https://conda.anaconda.org/brian-team/noarch/repodata.json.bz2: https://conda.anaconda.org/brian-team/noarch/```

```conda create --yes --quiet -n appveyor_test python=%PYTHON_VERSION%
Fetching package metadata: .... 
Solving package specifications: .....Command exited with code -1073741819```

**Comment by mstimberg** [comment]
> I don't know if I'm being very naive here and I'm missing something; but whenever and however I use SpikeGeneratorGroup, it returns an DimensionsMismatchError.

Could you give some details about what version of Brian and other libraries you are using (e.g. copy&paste the start of the debug log, its location in the filesystem will be printed when your script raises an error)? Numpy's max function should not convert quantities to simple numpy arrays:
```Python
>>> from brian2 import *
>>> np.max([1, 2, 3]*mV)
3. * mvolt
```

PS: Don't worry about appveyor, unfortunately the anaconda repository sometimes is not responsive and we rely on it in the test suite.

martinosorb:
`In [1]: from brian2 import *

In [2]: import numpy as np

In [3]: np.max([1, 2, 3]*mV)
Out[3]: 3. * mvolt

In [4]: np.max([1*mV, 2*mV, 3*mV])
Out[4]: 0.0030000000000000001`

brian2.__version__ is 2.0rc1

mstimberg:
Ahh, thanks that explains everything -- we should probably make it clearer in the documentation: `[1*mV, 2*mV, 3*mV]` is a list of `Quantity`, not a `Quantity` object itself -- we do not have any control over what numpy does with something like this. You can convert such a list to a `Quantity` object by calling `Quantity` explicitly, i.e.: `Quantity([1*mV, 2*mV, 3*mV])`. We should probably do this at the start of `SpikeGeneratorGroup`, your list will then work fine.

A general remark: operations on such lists are much slower than on a `Quantity` object:
```Python
>>> ar = [x*mV for x in np.arange(100000)]  #bad
>>> ar2 = np.arange(100000)*mV  # good
>>> %timeit np.max(ar)
100 loops, best of 3: 5.6 ms per loop
>>> %timeit np.max(ar2)
The slowest run took 6.04 times longer than the fastest. This could mean that an intermediate result is being cached.
10000 loops, best of 3: 33.8 µs per loop
```

martinosorb:
Right, so I reverted to np.max, but added a `times = Quantity(times)` line.
Interestingly, there was an unused import of `Quantity` already there, so maybe that check was present in a previous version of the function, and was lost at some point.

mstimberg:
Thanks -- in the long run it'd make sense to include this conversion in the `check_units`, so that not every function has to re-do it. But for now it is certainly an improvement, I will merge it into master.

martinosorb:
Great, thanks.

PR #719, opened 2016-06-17
[MRG] Respect the `active` flag in C++ standalone

### LINKED ISSUE #718: C++ standalone ignores the active property
In runtime mode, you can switch off an object by setting its `active` attribute to `False`. This can be useful to switch inputs/monitors on/off between runs or when using the `store`/`restore` mechanism (because you are not allowed to add/remove objects in that case). This does not work in standalone mode, e.g. the following continues to integrate `v` while in runtime it would remain constant after the first run:
```Python
from brian2 import *
set_device('cpp_standalone', build_on_run=False)
group = NeuronGroup(1, 'dv/dt = -v/(10*ms) : 1')
group.v = 1
mon = StateMonitor(group, 'v', record=0)
run(5*ms)
group.active = False
run(5*ms)
device.build()

plot(mon.t/ms, mon[0].v)
show()
```

mstimberg:
A fairly uncontroversial PR, I think :)
Fixes #718

### DIFF
--- a/brian2/devices/cpp_standalone/device.py
+++ b/brian2/devices/cpp_standalone/device.py
@@ -1029,8 +1029,9 @@ class CPPStandaloneDevice(Device):
         # running order, assuming that there is only one clock
         code_objects = []
         for obj in net.objects:
-            for codeobj in obj._code_objects:
-                code_objects.append((obj.clock, codeobj))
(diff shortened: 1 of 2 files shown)


thesamovar:
Indeed - no issues with merging this. :)

Issue #646, opened 2016-02-02
Better error message when code execution fails


mstimberg:
During a run, we run `exec self.compiled_code in self.namespace` every time step. If this fails with an exception (which is almost always a bug in Brian), then the user does not get a helpful error message, but only something like: 
```
File "(string)", line 23, in <module>
IndexError: index 5 is out of bounds for axis 0 with size 5
```

i.e. no information about the code object/code that failed to execute. We should probably wrap this in a `try/except` (according to the Python FAQ: "A try/except block is extremely efficient if no exceptions are raised.") -- after all, debugging is probably now one of the main reasons to use numpy as the code generation target.

I don't think we can do anything meaningful for weave/cython.

PR #722, opened 2016-06-21
[MRG] Error message improvements

mstimberg:
This fixes some less than useful error messages that I had on my list for quite a while, see explanations in the commit logs. The only somewhat debatable change is the fix for #646 -- the `exec` used to execute numpy code objects is now wrapped in `try/exec` and will give the actual code line that raised the error. This comes at the cost of a slightly slower code execution, in one example (in some very unsystematic benchmarking) I saw a runtime increase of 10%. I still think this is worth doing, numpy would somewhat become the "debug device". If you see for example a segmentation fault when running your simulation, it is very much possible that switching to numpy will give us a helpful error message (e.g. an `IndexError` when we go beyond array bounds). I actually did this change a few times manually when I was debugging such issues.

### DIFF
--- a/brian2/codegen/runtime/numpy_rt/numpy_rt.py
+++ b/brian2/codegen/runtime/numpy_rt/numpy_rt.py
@@ -1,8 +1,11 @@
 '''
 Module providing `NumpyCodeObject`.
 '''
+import sys
+
 import numpy as np
 
+from brian2.core.base import brian_object_exception
 from brian2.core.preferences import prefs, BrianPreference
 from brian2.core.variables import (DynamicArrayVariable, ArrayVariable,
                                    AuxiliaryVariable, Subexpression)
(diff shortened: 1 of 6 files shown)


thesamovar:
I'm fine with merging all this. Adding a try/except must surely only be an additional fixed cost. My guess is that your benchmarking was with a very small example?

mstimberg:
> Adding a try/except must surely only be an additional fixed cost.

Sure, it's a fixed cost (per time step). I think the performance penalty is nothing to worry about, the example where I saw the biggest performance drop (`IF_curve_Hodgkin_Huxley.py`) also happens to be one where switching to weave/cython brings a dramatic (~10-fold) increase in performance.

I'll go ahead and merge this now (two tests failed in the end when building the conda package because the conda infrastructure was at capacity -- unfortunately this happens more and more often lately...)

Issue #724, opened 2016-06-24
Cython SpikeQueue build failures on Windows


mstimberg:
Some builds started failing on appveyor lately, I don't quite see the pattern, it seem to be 64-bit builds, but not those that use Python 3.5...?
One example is here:
https://ci.appveyor.com/project/brianteam/brian2/build/1501/job/bs679qmyb2off0mq

The error seems to be that `link.exe` is called with an extra parameter:
```
C:\Program Files\Git\usr\bin\link.exe /DLL /nologo /INCREMENTAL:NO /LIBPATH:C:\Miniconda3-x64\envs\_build\libs /LIBPATH:C:\Miniconda3-x64\envs\_build\PCbuild\amd64 /EXPORT:PyInit_cythonspikequeue build\temp.win-amd64-3.4\Release\brian2\synapses\cythonspikequeue.obj build\temp.win-amd64-3.4\Release\brian2/synapses/cspikequeue.obj /OUT:build\lib.win-amd64-3.4\brian2\synapses\cythonspikequeue.pyd /IMPLIB:build\temp.win-amd64-3.4\Release\brian2\synapses\cythonspikequeue.lib /MANIFESTFILE:build\temp.win-amd64-3.4\Release\brian2\synapses\cythonspikequeue.pyd.manifest
link: extra operand '/INCREMENTAL:NO'
Try 'link --help' for more information.
error: command 'C:\\Program Files\\Git\\usr\\bin\\link.exe' failed with exit status 1
```
Quick googling did not turn up anything obvious, @thesamovar any idea?

thesamovar:
I think the problem might be to do with path order, it is trying to use the GNU linker.

mstimberg:
> I think the problem might be to do with path order, it is trying to use the GNU linker.

Hmm, but isn't distutils taking care of this stuff (or rather the call to `vcvarsall.bat`)? Do you think there is anything we can do about this from Brian's side or is it something in the appveyor configuration and only needs to be fixed there?

thesamovar:
I think it's probably in the appveyor configuration but if that was the case then I'd expect to see bug reports on their end and a quick fix.

mstimberg:
This does not seem to be an issue anymore, let's close it for now and re-open it in case we encounter the problem again.

Issue #591, opened 2015-11-06
Update installation docs for Windows


mstimberg:
Recently, a few things seem to have improved for C++ code generation under Windows:
* Python 3.5 does no longer need the SDK environment variables under 64 Bit
* Anaconda now comes with a msvc_runtime package that might make a manual download of the "C++ compiler for Python" obsolete

We should check this and update the installation docs accordingly before the release.

thesamovar:
Interesting, that would be great if true.

PR #726, opened 2016-07-01
[MRG] Remove MSVC workarounds

mstimberg:
This removes the workarounds for MSVC that are no longer necessary: VS 2015 works fine with OpenMP now and numpy 0.10.2 fixed the `compiler_cxx` issue.
I also updated the installation instructions for code generation on Windows with Python 3.5. Unfortunately we still need a manual installation of compilers, the runtime environment that I mentioned in issue #591 does not include the compilers.
I also changed a small thing in the appveyor builds: conda packages are now only built for pull requests or on the master branch, that makes the tests slightly faster in development branches (and the conda builds do sometimes fail anyway just because of server issues).

It *seems* that the issue reported in #724 has fixed itself, but let's wait for the tests to run through.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -120,6 +120,10 @@ before_install:
 # command to install dependencies
 install:
   - conda update --yes conda
+  # For faster tests, only build conda packages for the master branch or pull requests
+  - if [[ $TRAVIS_PULL_REQUEST == 'false' ]] && [[ $TRAVIS_BRANCH != 'master' ]]; then
+       CONDA_BUILD='no';
+    fi
   - if [[ $CONDA_BUILD == 'yes' ]]; then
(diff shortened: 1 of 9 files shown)


thesamovar:
Looks good to me, and seems like it is likely to pass all tests. Ready to merge when the last one passes I'd say.

mstimberg:
The build of the conda packages failed on appveyor. With a newer version of `conda-build`, it blocks with a "overwrite file yes/no/all" question while running our custom `bld.bat`. It turned out that the `bld.bat` is not necessary at all, so I got rid of the custom build scripts (`bld.bat` and `build.sh`), everything is now in `meta.yaml` and seems to working fine (still waiting for a few last tests to run through).

mstimberg:
The tests all passed, and since the commits I added since you LGTM'ed only concerned the conda build, I'll go ahead with the merge.

Issue #721, opened 2016-06-20
Check monitored expressions for stateful functions


mstimberg:
The following is a not uncommon mistake:
```Python
group = NeuronGroup(1, '''dv/dt = (-v + I) / (10*ms) : 1
                          I = rand() : 1''')
mon = StateMonitor(group, 'I', record=True)
```
The user expects that the `StateMonitor` records the same `I` that is used in the differential equations, but instead `rand()` will be evaluated twice, giving different values. We should detect this situation (the most general check seems to be to check for subexpressions involving "stateful functions" ) and raise a warning (or even error?). Alternatively, we could completely forbid to record subexpressions, but that would be quite annoying for all the cases where it is unproblematic.

The warning/error message should direct the user to the use of a parameter and `run_regularly`, or some new syntax as discussed in #720.

thesamovar:
Could there be a case where the user would expect ``I`` to be the same value in two different places without recording? e.g. with a linked variable or some other mechanism? If not, I'd be happy either to raise a warning or to ban recording of subexpressions with stateful functions in them. I also wouldn't forbid recording subexpressions: seems like overkill.

thesamovar:
How does ``I=rand():1`` work with multi-step integration methods btw? Does it use multiple ``rand()`` calls or just one? If we haven't already documented that - it might be worth doing too.

mstimberg:
Good points, you are right that there is nothing monitor-specific here, it is really about the use of stateful functions in a subexpression. I think our current behaviour related to this is inconsistent, IIRC (I did not double check) we substitute all subexpressions for the state update step but only evaluate them once in general code blocks (e.g. resets), with some logic of re-evaluating them if some of the variables they used change. Not sure we added anything to re-evaluate them unconditionally if they involve stateless functions.

I think one fundamental problem in our syntax is that subexpressions look like statements but they are interpreted as functions of their variables.

mstimberg:
One more related remark: if we come up with some syntax for #720, then I think the cleanest would be to raise an error if the subexpression involves stateful functions and force the user to make an explicit choice for that expression, somewhat similar to `clock-driven` vs. `event-driven` for synapses. This way, you cannot simply ignore/overlook the warning, but you also have a way to make the warning disappear.
Also needs some though what that means for subexpressions referring to other subexpressions.

Issue #720, opened 2016-06-20
New syntax for "constant over timestep" variables


mstimberg:
In some situations, e.g. when using certain state updaters or to avoid re-evaluations of expressions involving random numbers, it would be useful to have a way of declaring a subexpression that is only evaluated once per timestep. The current workaround is to declare the variable as a parameter and update it with `run_regularly` -- this is a bit clumsy. I think we could introduce a new flag for subexpressions, something like
```
x = rand() : 1 (static)
```
with hopefully a clearer name than "static".

thesamovar:
The problem I have with this syntax is: when is it evaluated? It doesn't make a difference for ``rand()`` but would if there were any state variables in there. Feels like asking for trouble in some way, although I agree that people use ``x = rand() : 1`` and expect it to behave differently than it does. At the very least we should warn when people use a stateful function in a subexpression (which I see you've just opened a new issue for so I'll stop writing here!).

mstimberg:
One general thought (related to what I just wrote in #721): if we add a syntax, it might be good to have an explicit syntax for both cases, i.e. something like `(evaluate-once)` vs. `(evaluate-everytime)`.

> The problem I have with this syntax is: when is it evaluated?

I see at least three options:
1. Make it part of the flag, something like `(evaluate: start)`
2. The update for these "static subexpressions" happens in its own object, which uses our default scheduling system (e.g. it defaults to `when='before_groups'` but you can change it by assigning something else to its `when` attribute).
3. It is evaluated when it is used the first time during a time step.

I think 3 is a bit intransparent, I'd therefore prefer 1 or 2. I have a slight tendency towards 2 because it would be more consistent with what we do in other places (e.g. for `(summed)` in synapses), it would be straightforward to implement and it does not add a lot of syntax.

thesamovar:
OK, agree with having an explicit syntax for both cases. Should we make that only necessary when there are stateful functions or always necessary? I'd say only when there are stateful functions, with a default of the current behaviour of evaluate-everytime when there are no stateful functions. This does allow for some slight problems in that you might write x=v and hope it means the value of v at the beginning of the timestep and actually it wouldn't always mean that, but this seems relatively unlikely and the saving in unnecessary syntax in the majority of cases seems worth it.

Also agree with having before groups be the default if it matches what we're doing elsewhere (and in any case, this does seem fairly logical). Is there any use case for allowing the user to specify something else? I don't really see any. In which case, it's just a question of the syntax.

[...]

Issue #678, opened 2016-04-15
Expose the network simulation schedule to the user


mstimberg:
While we have a nice system to control the execution schedule of all objects in a network, the final schedule is not necessarily very obvious to the user. A method like `Network.object_schedule` (`schedule` is already taken for the general schedule, but we can probably come up with a better name) with an equivalent for the magic network should print something like this:

```
dt     | slot       | order | name
------------------------------------------------------
0.1 ms  start         0       statemonitor
0.1 ms  groups        0       neurongroup_stateupdater
0.1 ms  thresholds    0       neurongroup_thresholder
0.1 ms  resets        0       neurongroup_resets
```

or ideally it would rather return an object that can be converted to a string like the above but also allows for convenience methods to access certain details (similar to `ProfilingSummary`). Yet another "ideally": it would be great if instead of collecting this information for the user on demand, the object would also be internally used as part of the scheduling so that we can sure that the reported schedule and the actually used schedule are the same.

Obviously, the above output is an example and I'm open to a different formatting... Oh and one detail that we should deal with: `NeuronGroup`, `Synapses`, etc. will appear in the object list even though they do not do anything in their `run` method, all computation is performed by the state updater, thresholder, etc. objects. This will certainly not be very obvious to the average user... (I myself stumble over it every time ;-) )

thesamovar:
Sounds like a good idea to me!

PR #727, opened 2016-07-05
[MRG] "constant over timestep" subexpressions

mstimberg:
This is a start for what we discussed in #720 and #721. I went for a simple approach where the "constant over one timestep" subexpressions are pulled out in the beginning and converted into the "parameter + run_regularly" approach. The advantage of this is that nothing "downstream" (codegen, etc.) does have to be aware of the difference -- I think it's too late in the 2.0 process for any other solution (e.g. one that would break other devices).

Still, there are a few open questions:
1. What names for the flags to use -- we did not finish this discussion in #720. For now, I went for "constant over dt" and "variable over dt" but I am still happy to be persuaded by better options. @romainbrette any opinion on that?
2. The "constant over dt" subexpressions are currently updated in the `before_groups` slot. For most uses that makes sense, but maybe `start` is a better choice? In a little test I used a `run_regularly` operation and realized that subexpressions are updated only afterwards which would be quite surprising. On the other hand, even with `start`, it would compete with the `run_regularly` operation. Maybe `before_start`? It sounds silly, but we do allow it.
3. Currently, we only evaluate subexpressions once per code block if the variables they refer to are not written to. This means that even when subexpressions are marked as "variable over dt" and use to time-dependent/stateful expressions, they are not updated. We could either update the code in `make_statements` and take time-/state-dependence into account or simply get rid of the caching. I'd prefer the latter as the code is already very complicated and I think the performance benefits from the caching are minimal.
4. I added the same mechanism also to `Synapses`, but there is a problem here: if you define a subexpression as "constant over dt", then it will be updated every time step, even though it may only be used in `on_pre`/`on_post` statements and/or in `event-driven` equations. On the other hand, I am not sure it is worth complicating things with explicit checks for that, since "constant over dt" is an explicit choice and you have workarounds (not use a subexpression at all...). It should be mentioned in the documentation, at least.

### DIFF
--- a/brian2/codegen/translation.py
+++ b/brian2/codegen/translation.py
@@ -299,13 +299,6 @@ def make_statements(code, variables, dtype, optimise=True, blockname=''):
         line.will_write = will_write.copy()
         will_write.add(line.write)
 
-    # generate cacheing statements for common subexpressions
-    # cached subexpressions need to be recomputed whenever they are to be used
-    # on the next line, and currently invalid (meaning that the current value
-    # stored in the subexpression variable is no longer accurate because one
-    # of the variables appearing in it has changed). All subexpressions start
-    # as invalid, and are invalidated whenever one of the variables appearing
-    # in the RHS changes value.
     subexpressions = dict((name, val) for name, val in variables.items() if isinstance(val, Subexpression))
     # sort subexpressions into an order so that subexpressions that don't depend
     # on other subexpressions are first
@@ -314,8 +307,7 @@ def make_statements(code, variables, dtype, optimise=True, blockname=''):
     sorted_subexpr_vars = topsort(subexpr_deps)
 
     statements = []
-    # all start as invalid
-    valid = dict((name, False) for name in subexpressions.keys())
+
     # none are yet defined (or declared)
     subdefined = dict((name, False) for name in subexpressions.keys())
     for line in lines:
@@ -324,53 +316,38 @@ def make_statements(code, variables, dtype, optimise=True, blockname=''):
         write = line.write
         will_read = line.will_read
         will_write = line.will_write
-        # check that all subexpressions in expr are valid, and if not
-        # add a definition/set its value, and set it to be valid
-        # scan through in sorted order so that recursive subexpression dependencies
-        # are handled in the right order
+        # update/define all subexpressions needed by this statement
         for var in sorted_subexpr_vars:
             if var not in read:
                 continue
-            # if subexpression, and invalid
-            if not valid.get(var, True): # all non-subexpressions are valid
-                subexpression = subexpressions[var]
-                # if already defined/declared
-                if subdefined[var]:
-                    op = '='
-                    constant = False
-                else:
-                    op = ':='
-                    subdefined[var] = True
-                    # set to constant only if we will not write to it again
-                    constant = var not in will_write
-                    # check all subvariables are not written to again as well
-                    if constant:
-                        ids = subexpression.identifiers
-                        constant = all(v not in will_write for v in ids)
-                valid[var] = True
-                statement = Statement(var, op, subexpression.expr, comment='',
-                                      dtype=variables[var].dtype,
-                                      constant=constant,
-                                      subexpression=True,
-                                      scalar=variables[var].scalar)
-                statements.append(statement)
+
+            subexpression = subexpressions[var]
+            # if already defined/declared
+            if subdefined[var]:
+                op = '='
+                constant = False
+            else:
+                op = ':='
+                subdefined[var] = True
+                # set to constant only if we will not write to it again
+                constant = var not in will_write
+                # check all subvariables are not written to again as well
+                if constant:
+                    ids = subexpression.identifiers
+                    constant = all(v not in will_write for v in ids)
+
+            statement = Statement(var, op, subexpression.expr, comment='',
+                                  dtype=variables[var].dtype,
+                                  constant=constant,
+                                  subexpression=True,
+                                  scalar=variables[var].scalar)
+            statements.append(statement)
+
         var, op, expr, comment = stmt.var, stmt.op, stmt.expr, stmt.comment
-        # invalidate any subexpressions including var, recursively
-        # we do this by having a set of variables that are invalid that we
-        # start with the changed var and increase by any subexpressions we
-        # find that have a dependency on something in the invalid set. We
-        # go through in sorted subexpression order so that the invalid set
-        # is increased in the right order
-        invalid = {var}
-        for subvar in sorted_subexpr_vars:
-            spec = subexpressions[subvar]
-            spec_ids = set(spec.identifiers)
-            if spec_ids.intersection(invalid):
-                valid[subvar] = False
-                invalid.add(subvar)
+
         # constant only if we are declaring a new variable and we will not
         # write to it again
-        constant = op==':=' and var not in will_write
+        constant = op == ':=' and var not in will_write
         statement = Statement(var, op, expr, comment,
                               dtype=variables[var].dtype,
                               constant=constant,
--- a/brian2/equations/codestrings.py
+++ b/brian2/equations/codestrings.py
@@ -166,3 +166,47 @@ class Expression(CodeString):
             raise AssertionError('Cyclical call of CodeString._repr_pretty')
         # Make use of sympy's pretty printing
         p.pretty(str_to_sympy(self.code))
+
+
+def is_constant_over_dt(expression, variables, dt_value):
+    '''
+    Check whether an expression can be considered as constant over a time step.
+    This is *not* the case when the expression either:
+
+    1. contains the variable ``t`` (except as the argument of a function that
+    can be considered as constant over a time step, e.g. a `TimedArray` with a
+    dt equal to or greater than the dt used to evaluate this expression)
+    2. refers to a stateful function such as ``rand()``.
+
+    Parameters
+    ----------
+    expression : `sympy.Expr`
+        The (sympy) expression to analyze
+    variables : dict
+        The variables dictionary.
+    dt_value : float or None
+        The length of a timestep (without units), can be ``None`` if the
+        time step is not yet known.
+
+    Returns
+    -------
+    is_constant : bool
+        Whether the expression can be considered to be constant over a time
+        step.
+    '''
+    t_symbol = sympy.Symbol('t', real=True, positive=True)
+    func_name = str(expression.func)
+    func_variable = variables.get(func_name, None)
+    if func_variable is not None and not func_variable.stateless:
+        return False
+    for arg in expression.args:
+        if arg is t_symbol and dt_value is not None:
+            # We found "t" -- if it is not the only argument of a locally
+            # constant function we bail out
+            if not (func_variable is not None and
+                        func_variable.is_locally_constant(dt_value)):
+                return False
+        else:
+            if not is_constant_over_dt(arg, variables, dt_value):
+                return False
+    return True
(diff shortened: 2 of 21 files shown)


thesamovar:
For the names, I agree with your reasoning in the previous post on #720. I like the general idea of "constant/variable over dt" but how about "timestep" instead of dt? It feels more semantically correct. We could also go with "during" instead of "over" but it's more to type and I think "over" is clear enough. I think that "variable" is a good enough antonym to constant in this case (no chance of confusing it with "a variable").

For the time slot, I think if our name "constant over timestep" it would be strange if it actually changed during the timestep. In that case, maybe "constant for dt" or "constant over dt" would be more sensible indeed. Perhaps this is an argument for using the "before_start" slot though. In general, it fits what I'd guess the semantics should be better. I don't have really strong feelings about this, but this (and several other things) make me think more and more that we should prioritise having an automatically generated "simulation description" (or "methods section") in plain text that describes precisely what was simulated. Another thing we might think about doing here is having a check that if a constant-over-dt variable is accessed in a timestep before it is initialised for that time step then we have a warning. I could actually imagine that we could have quite a useful/interesting system of checking which variables are accessed when, and testing for and warning about potential mistakes.

I'd be happy to get rid of the caching if indeed it doesn't make much of a difference to performance (but we should really check this). If it does make a difference, we can reconsider whether or not it's worth it anyway or if we need to complicate.

For Synapses, what about we give an info log statement or warning to the effect that it may be unnecessarily slowing down the simulation? I guess it's a very rare case anyway.

mstimberg:
> For the names, I agree with your reasoning in the previous post on #720. I like the general idea of "constant/variable over dt" but how about "timestep" instead of dt? It feels more semantically correct. We could also go with "during" instead of "over" but it's more to type and I think "over" is clear enough. I think that "variable" is a good enough antonym to constant in this case (no chance of confusing it with "a variable").

Ok, I'll get some feedback about the name from the others during our team meeting this Friday. I agree that "timestep" is somewhat less technical but I am a bit worried that the whole flag becomes really long...

> For the time slot, I think if our name "constant over timestep" it would be strange if it actually changed during the timestep. In that case, maybe "constant for dt" or "constant over dt" would be more sensible indeed. Perhaps this is an argument for using the "before_start" slot though. In general, it fits what I'd guess the semantics should be better.

I am still not sure about this (again, I'll ask for feedback on Friday). I think `before_start` is the most "principled" choice, though. We could go for an alternative (and much more difficult to implement) choice to update the value at the first use during a time step but that does not quite feel right, either.

> I don't have really strong feelings about this, but this (and several other things) make me think more and more that we should prioritise having an automatically generated "simulation description" (or "methods section") in plain text that describes precisely what was simulated.

That would indeed be useful, a first step would be to give easy access to the schedule (#678).

> Another thing we might think about doing here is having a check that if a constant-over-dt variable is accessed in a timestep before it is initialised for that time step then we have a warning. I could actually imagine that we could have quite a useful/interesting system of checking which variables are accessed when, and testing for and warning about potential mistakes.

Again agreed, but I think this is some major (and therefore definitely post-2.0) work. This is more or less #126 I guess.

> I'd be happy to get rid of the caching if indeed it doesn't make much of a difference to performance (but we should really check this). If it does make a difference, we can reconsider whether or not it's worth it anyway or if we need to complicate.

Ok, I'll try to run some benchmarking for this.

> For Synapses, what about we give an info log statement or warning to the effect that it may be unnecessarily slowing down the simulation? I guess it's a very rare case anyway.

Yes, makes sense.

mstimberg:
>> I'd be happy to get rid of the caching if indeed it doesn't make much of a difference to performance (but we should really check this). If it does make a difference, we can reconsider whether or not it's worth it anyway or if we need to complicate.
> 
> Ok, I'll try to run some benchmarking for this.

Um, I'm having some trouble coming up with non-contrived example for this. For all the examples we have, none is actually re-using a subexpression in a way that would make use of the caching (I raised an exception in the "subexpression can be re-used" code path and it was never triggered...).

thesamovar:
It's true that the name is long, but on the other hand it'll be rarely used so I think it's OK.

For the example, don't some of the HH examples use subexpressions? I know they did in Brian 1 but maybe we rewrote them not to use this for B2?

mstimberg:
> For the example, don't some of the HH examples use subexpressions? I know they did in Brian 1 but maybe we rewrote them not to use this for B2?

They do but all subexpressions are substituted when the state updater is applied, so the state update step does not deal with subexpressions at all. Even if it did (which is not trivial to get right: #412), the HH equations use each subexpression only once, so they wouldn't be affected by the caching.

thesamovar:
OK, so let's drop the subexpression caching then!

mstimberg:
> OK, so let's drop the subexpression caching then!

Ok, I'll do that.

Here's the feedback I got from @romainbrette and @yger during our meeting:

The `constant over dt` or `constant over timestep` is clear, no strong preference for either.
The opposite `variable over dt` is not clear -- we did not find any better alternative but eventually agreed on the following: given that we have a hard time thinking of a use case where the expression refers to a stateful function *and* you'd like it to be re-evaluated for each use, how about not offering an alternative at all? This means: 1) standard subexpressions are by default re-evaluated for each use but can be made fixed by using `(constant over dt)`. 2) subexpressions referring to stateful functions raise an error unless they use `(constant over dt)`.

What do you think about that solution? I'd say we should give it a try, if at some point we come across a convincing use case for "variable over dt" stateful subexpressions, we can always add a flag later. There's a straightforward workaround (define one subexpression for each use or do not use a subexpression in the first place) anyway.

Otherwise, updating in `before_start` makes sense and no clear opinion on whether we need a warning in the case of synapses. The example is quite contrived, maybe we don't actually have to bother.

mstimberg:
Oh and not related to today's discussion but I realized that I currently I check whether the subexpression is "stateful or time-dependent" -- I don't think we actually need to care about time-dependence, though.

mstimberg:
Sorry for the noise, I just realized another thing: if we go with the proposal that subexpressions that depend on stateful functions *have to be* "constant over dt", then the whole issue of updating the caching mechanism to handle "variable over dt" yet stateful subexpressions go away. We could therefore leave the caching code in place (I still wouldn't mind deleting it, but maybe leaving it untouched reduces the probability of breaking something with this PR).

thesamovar:
OK I'm happy with that solution. On reflection, "constant over dt" might actually make more sense than "constant over timestep" if we were to allow setting it in slots other than before_start. Since it's also shorter, let's stick with that. In other words, what you're already doing seems perfect to me.

mstimberg:
Ok, great. I made a few last changes, ready-to-merge from my side.

thesamovar:
OK, I haven't reviewed the code as thoroughly as I'd like because it's rather complex and I'm a bit busy with other stuff at the moment, but I had a quick scan through. A couple of comments:

1. It looks like subexpression cacheing has been removed, right? (At least, there seems to be a whole bunch of deleted code about subexpression cacheing.) I'm actually happy with that because it removes a potential source of errors in difficult-to-foresee scenarios.
2. The documentation mentions that declaring a variable as constant over dt might be useful in a differential equation to force the use of the linear integrator. Is that correct? Are we actually using that information? Do we have an example of when this might happen?

mstimberg:
> It looks like subexpression cacheing has been removed, right? (At least, there seems to be a whole bunch of deleted code about subexpression cacheing.) I'm actually happy with that because it removes a potential source of errors in difficult-to-foresee scenarios.

Yes, I removed it and I agree that it is better to leave it out until we see a clear need for it.

> The documentation mentions that declaring a variable as constant over dt might be useful in a differential equation to force the use of the linear integrator. Is that correct? Are we actually using that information? Do we have an example of when this might happen?

Yes, that's correct. It is actually documenting something that I mentioned a few times as a somewhat clumsy workaround in the mailing list. If `x` is preventing you from using a linear integration, then you can replace
```Python
group = NeuronGroup(..., '''
x = non_lin_fun(...) : 1
''', ...)
```
by
```Python
group = NeuronGroup(..., '''
x : 1
''', ...)
group.run_regularly('x = non_lin_fun(...)')
```

This is exactly how the `(constant over dt)` flag is implemented right now, i.e. the parameter + `run_regularly` approach is equivalent to:
```Python
group = NeuronGroup(..., '''
x = non_lin_fun(...) : 1 (constant over dt)
''', ...)
```

Of course it might well be that it would be simpler to just use e.g. Euler integration instead of this exact integration + stepwise approximation approach. There's however at least one use case where you don't have the choice: in `SpatialNeuronGroup`, the membrane current *has* to depend linearly on `v` -- the `constant over dt` mechanism can be useful there.

This approach has disadvantages for the readability of the generated code (in the state update code, you don't see what `x` actually stands for) but it makes it super-simple for e.g. the stateupdater to support those new subexpressions -- it simply does not know about them and therefore does not care.

thesamovar:
OK I think this is good to merge. I'll let you do the honours, just in case there were any last minute changes you wanted to make.

mstimberg:
Great. No last-minute changes, merging it now.

ShanqMa:
Hey there,

I am currently working on a spatialneuron. And i want to add **independent** current noise **in different compartments**. But i find there is only "constant over dt" option if I use randn() state function to generate white noise, which means the noise in every timestep will be global then.
Below i showed my eqs and the I_noise is the variable that i want to evaluate everytime when it appears (i think the exact case of "variable over dt" in your previous discussion), I am wondering what can i do to make it variable over dt?

`eqs = '''
Im =I_noise + (EM-v)/rm + gNa * m**3 * h * (ENa-v) + (gK  * n**4 + gKA * z **3 * j + gKCa ) * (EK-v) + gCa * c**3 * (ECa-v)  : amp/meter**2 
I_noise = sigma*randn()*dt**-0.5 : amp/meter**2  (variable over dt)`

Thank you a lot in advance!

Best,
Shan

mstimberg:
Hi @ShanqMa . The `constant over dt` flag means that the value of the subexpression stays constant over the duration of the time step. For `rand()` calls, it means they are only generated once and not repeatedly if multiple places reference the variable – but this is not the case in your equations, it only gets referenced once. There is therefore not a global value for all compartments (this would require the `shared` flag): with your above code (and using `constant over dt`), you will already get independent noise across compartments. Here's a simple example showing this:
```Python
from brian2 import *

eqs = ''' Im = I_noise : amp/meter**2
          I_noise = 1*nA/cm**2 * randn(): amp/meter**2 (constant over dt)'''
morpho = Cylinder(length=100*um, diameter=10*um, n=10)
# Huge Ri to make the compartments "unconnected"
neuro = SpatialNeuron(morpho, eqs, Ri=1*Mohm*cm)
mon = StateMonitor(neuro, 'v', record=True)
run(100*ms)
plt.plot(mon.t/ms, mon.v.T/mV)
plt.show()
```
noise_compartments

ShanqMa:
@mstimberg Thanks a lot for your reply! You are right, I misunderstood the documentation there. and I should have checked the results by myself firstly. 

Actually I still have another question, in the below equation I want to add independent noise for different channels:
eqs = '''
Im = (EM-v)/rm + INa + IK + IKA + IKCa + ICa  : amp/meter**2 
INa= gNa * (m**3 * h+sigma***randn()***dt**-0.5) * (ENa-v) : amp/meter**2
IK= gK * (n**4+sigma***randn()***dt**-0.5) * (EK-v) : amp/meter**2
IKA= gKA * (z**3 * j +sigma***randn()***dt**-0.5)* (EK-v) : amp/meter**2
IKCa= gKCa * (EK-v) : amp/meter**2
ICa= gCa * (c**3+sigma***randn()***dt**-0.5) * (ECa-v): amp/meter**2'''
But I got the errors saying 

> contains more than one call of randn, this is currently not supported since randn is a stateful function and its multiple calls might be treated incorrectly (e.g."rand() - rand()" could be  simplified to "0.0").
How can I overcome this setting and use multiple stateful function?

Thank you so much again!

Best,
Shan

PR #729, opened 2016-07-08
Numba

zhenruiliao:
Implemented numba codegenerator and numba runtime. There may still be some issues dealing with the synapses but all the basic and neuron code works.


thesamovar:
Hi - thanks for submitting this. We'll probably have a few more comments as we try this out and test it more thoroughly, but a couple of immediate things.

There are a few unnecessary changes to various files where you've just added or removed some newlines (e.g. in Cython). Could you revert those files? If you look at the "Files Changed" tab on this PR page you should be able to see easily which files those are.

Did you upload all the templates? There don't seem to be very many of them?

What is the current status of what works and what doesn't? Do synapses work at all or just neuron groups and monitors?

thesamovar:
I chatted with @mstimberg, what we decided is that for the moment, the best thing to do is to create a new repository for this. Users will access it by doing ``import brian2numba`` and then ``prefs.codegen.target = 'numba'``. I've created a new repository and given you direct access to it. Could you copy across your files and work on it there?

zhenruiliao:
Oops, didn't mean to change those. I was probably playing with them early on when trying to get an idea of how brian worked with Cython and Numpy. The version history is not very "clean" either for the same reason. 

Yeah, I'll migrate everything over to the new repo

zhenruiliao:
So there are two branches in existence. One is based on a slightly outdated version of brian (downloaded a month or two ago) and everything works. The one I just pushed is based on the newest version of brian (downloaded about a week ago), and I've just been moving everything over - all of the neuron groups and monitors work and synapses work up to the last example in the tutorial. I'll try to see what issues remain tonight.

mstimberg:
I'll close this PR for now, as the development happens in a separate repository: https://github.com/brian-team/brian2numba
Let's continue the discussion over there (e.g. open issues about missing functionality). The only change necessary in the main brian2 repository should eventually be to mention the project in the documentation.

Issue #730, opened 2016-07-11
Clock.epsilon too small?


thesamovar:
The following was reported on the mailing list:

```
When I restore a network model from last run and run the simulation, I encountered such error, ValueError: Cannot set dt from 100. us to 10. us, the time 112. s is not a multiple of 10. Us.

This error will not happen if the last simulation time is very short such as 10. S. Don’t know where is the problem.
```

I just checked and ``(112/10e-6)*10e-6 - 112 = 1.4e-14``, and ``Clock.epsilon=1e-14``. At the moment we have an absolute check for whether a time is a multiple of dt, i.e. ``abs(t_new-t_old)<Clock.epsilon`` and so this would be failing this check.

Two options, either we increase ``Clock.epsilon`` or we use a relative measure. Any thoughts?

mstimberg:
We do need something more robust here (and possibly in other places), I think just lowering the absolute epsilon is not a good idea. How about using `numpy.allclose` with both an absolute and a relative error (http://docs.scipy.org/doc/numpy-dev/reference/generated/numpy.allclose.html#numpy.allclose)?

thesamovar:
For background: I had a feeling we discussed this before, so I went back to check when this was added (in https://github.com/brian-team/brian2/pull/318) and it seems that we didn't discuss this point in detail.

Yes, I think having an absolute and relative error check makes sense. 

One option would be to look at the fractional part of ``(i_old*dt_old)/dt_new`` and raise an error if it is too far from 0. So for the example above we get a value of ``1.5e-8``. If we increase the time to be a day long simulation at a dt of 100us changing to 10us we get a value of 1.5e-5. Given that seems to be on the extreme side, I think I'd be happy with a cutoff of 1e-4 here, i.e. we'd tolerate an error of up to 0.01% of a timestep. That seems large enough to catch the extreme case and small enough that it wouldn't introduce much error. What do you think?

mstimberg:
In general I agree, but just from the point of documenting it, the "fractional part of `(i_old*dt_old)/dt_new`" is the same as `abs(t_new - t_old)/dt_new`, no? Other than that, 0.01% of a timestep sounds very reasonable.

thesamovar:
From the point of view of documenting it, yes, but I think the calculation is better without the subtraction? Maybe it's also better to document the actual calculation too? Not sure.

[...]

PR #735, opened 2016-07-28
[MRG] dt and t checks

mstimberg:
This should implement what we discussed in #730. Let me know if I forgot anything or misunderstood something.

### DIFF
--- a/brian2/core/network.py
+++ b/brian2/core/network.py
@@ -722,7 +722,7 @@ class Network(Nameable):
         minclock, min_time, minclock_dt = min(clocks_times_dt, key=lambda k: k[1])
         curclocks = set(clock for clock, time, dt in clocks_times_dt if
                         (time == min_time or
-                         abs(time - min_time)/min([minclock_dt, dt]) < Clock.epsilon_dt))
+                         abs(time - min_time)/min(minclock_dt, dt) < Clock.epsilon_dt))
         return minclock, curclocks
 
     @device_override('network_run')
--- a/brian2/core/network.py
+++ b/brian2/core/network.py
@@ -720,7 +720,7 @@ class Network(Nameable):
         minclock, min_time = min(clocks_times, key=lambda k: k[1])
         curclocks = set(clock for clock, time in clocks_times if
                         (time == min_time or
-                         abs(time - min_time) < Clock.epsilon))
+                         abs(time - min_time) < 1e-14))
         return minclock, curclocks
 
     @device_override('network_run')
--- a/brian2/core/network.py
+++ b/brian2/core/network.py
@@ -715,12 +715,14 @@ class Network(Nameable):
                 obj.after_run()
         
     def _nextclocks(self):
-        clocks_times = [(c, self._clock_variables[c][1][0])
+        clocks_times_dt = [(c,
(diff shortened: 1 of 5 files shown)


brian2/core/network.py
```diff
@@ -715,12 +715,14 @@ def after_run(self):
...
+                            self._clock_variables[c][1][0],
+                            self._clock_variables[c][2][0])
                         for c in self._clocks]
-        minclock, min_time = min(clocks_times, key=lambda k: k[1])
-        curclocks = set(clock for clock, time in clocks_times if
+        minclock, min_time, minclock_dt = min(clocks_times_dt, key=lambda k: k[1])
+        curclocks = set(clock for clock, time, dt in clocks_times_dt if
                         (time == min_time or
-                         abs(time - min_time) < Clock.epsilon))
+                         abs(time - min_time)/min([minclock_dt, dt]) < Clock.epsilon_dt))
```
thesamovar: Mildly more efficient not to use a list for ``[minclock_dt, dt]`` but just pass them as arguments directly? Not sure actually.

thesamovar:
The ``check_dt`` function is a bit strange because there's an ``old_dt`` and ``new_dt`` but also ``target_t``. If it's changing from one dt to another dt, why not just directly pass the ``old_i`` value rather than the ``target_t`` value? That seems safer to me, rather than inferring the ``old_i`` value from ``target_t``. Maybe it's OK because any time the epsilon check is passed, ``round`` would give that same integer value.

brian2/core/network.py
mstimberg: You are right, it is a bit more efficient (a tuple would be fine but a list takes more time) -- I think I had numpy's min in mind where you cannot use that syntax.

mstimberg:
> The check_dt function is a bit strange because there's an old_dt and new_dt but also target_t. If it's changing from one dt to another dt, why not just directly pass the old_i value rather than the target_t value? That seems safer to me, rather than inferring the old_i value from target_t. Maybe it's OK because any time the epsilon check is passed, round would give that same integer value.

The whole thing is not very straightforward, the function where `check_dt` is called does not only set `dt`, but also `t`. We delay the actual change of `dt` until the start of the simulation, this is because otherwise code like this would throw an error:
```Python
group = ...
net1 = Network(group)
net1.run(0.3*ms)
defaultclock.dt = 0.5*ms  # this is not ok if we continue the run, but...
group = ...
net2 = Network(group)
net2.run(1*ms)  #... we actually start at 0 again
```
For the second run, we set both `t` and `dt` at once, we don't have a previous time step. I am sure one could write this differently, but the whole system is rather fragile and I'd prefer not to make changes if not necessary.

thesamovar:
Ah OK, I understand now. In that case, I think it's good to go! Will you make the minor change to ``max``? Probably doesn't make much difference but may as well have it I guess.

mstimberg:
I did the change and also added a warning for extremely long simulations which I forgot. All good to go from my side now.

thesamovar:
Great! Good to merge when it passes then.

mstimberg:
Oh, forgot to merge this. Merging it now

Issue #737, opened 2016-08-15
Setting delay before connect is confusing to users


thesamovar:
If you set delay before you call connect, it doesn't have the intended effect for some users. I think maybe the best thing to do is to make this clearer in the docs (repeating ourselves if necessary) and raise a warning if you set delay on 0 synapses?

rat-h:
I believe it should be  an error message and the exit from the script totally. Otherwise user will still run a simulation and will get a result. It is quite easy to miss or ignore a warning message in output.

mstimberg:
> I believe it should be an error message and the exit from the script totally. Otherwise user will still run a simulation and will get a result. It is quite easy to miss or ignore a warning message in output.

It is often a difficult choice between a warning and an error and I agree that it is easy to overlook a warning. However, we also should not take raising an error lightly, in particular for code that is correct albeit probably a mistake (as in the case of setting a synaptic variable on 0 synapses). One use case I am seeing where an error would be really annoying is a network with probabilistic synaptic connections. Especially if you have many different types of connections in a sparsely connected network it *is* possible that you might have a run where 0 synapses are created for a connection type. This will probably not happen during a normal run (and you'll get a warning about simulating a `Synapses` without any actual synapses for a good reason) but it might come up for example when you quickly scale down the size of your network for testing. In that case, wrapping every `synapses.delay = ...` in a `if len(synapses) > 0:` just to avoid this error would be very annoying. Even more problematic, it will not work in standalone mode as we don't know the number of synapses before the code is actually executed. I therefore think that a warning is a better choice here.

A final remark: if you set the delay to a constant value for all synapses (as was the case in the original report on the mailing list), then there is a preferable syntax which cannot go wrong no matter how many synapses: you can set the delay as an argument of the `Synapses` call itself:
```Python
synapses = Synapses(source, target, model='...', on_pre='...', delay=const_delay)
```

thesamovar:
I agree it shouldn't be an error.

A related point, perhaps we ought to allow a syntax like this?

```
S = Synapses(G, G, 'w:1', on_pre='v_post += w')
S.connect(p=0.1, delay=3*ms, w=0.1)
S.connect(p=0.2, delay=5*ms, w=0.2)
```

I'm not actually sure there is a way to do this at the moment, and it makes it clear that the delay applies just to that connect call, etc. What do you think?

mstimberg:
[...]

Issue #263, opened 2014-06-11
Synapse.connect returns created synapse indices


victorbenichoux:
Hi guys,

For now, calling Synapses.connect only creates the connections -- which does not sound so bad -- but does not return anything.

I think that it would be very convenient to have this method return the array of indices of the synapses that were just created. This would make setting other attributes very easy.

An example is when I want to sequentially add Synapses within a loop that calls S.connect:

```
S = Synapses(..., 'w:1')

for k in some_iterable:
    indices = S.connect(something(k), something(k))
    S.w[indices] = some_weights(k)
    S.delays[indices] = some_delays(k)
```

What do you think?

PR #739, opened 2016-08-17
[MRG] Error for missing connect call

mstimberg:
This implements the error messages as dicussed in #737 -- you get an error when you write to a synaptic variable without a preceding `connect` or when you include a `Synapses` object without a `connect` in a simulation and run it.

I realized that a warning for a `connect` that results in 0 synapses is actually not that trivial to implement. The synapse creation is done in the template and on the Python level, we do not know about the number of synapses created. We could make a change to the templates so that they return the number of created synapses but this might be a bit of a waste since we probably want to change this again when we implement #263. Also, we need to deal with standalone (same for the simple solution where we check the number of synapses before and after the synapse creation).

I am tempted to postpone this warning to post-2.0 and maybe deal with it at the same time as #263. What do you think?

### DIFF
--- a/brian2/core/variables.py
+++ b/brian2/core/variables.py
@@ -844,6 +844,15 @@ class VariableView(object):
         if variable.read_only:
             raise TypeError('Variable %s is read-only.' % self.name)
 
+        # Check whether the group allows writing to the variable (e.g. for
+        # synaptic variables, writing is only allowed after a connect)
+        try:
+            self.group.check_variable_write(variable)
+        except ReferenceError:
+            # Ignore problems with weakly referenced groups that don't exist
+            # anymore at this time (e.g. when doing neuron.axon.var = ...)
+            pass
+
         # The second part is equivalent to item == slice(None) but formulating
         # it this way prevents a FutureWarning if one of the elements is a
         # numpy array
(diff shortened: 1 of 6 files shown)


brian2/core/variables.py
```diff
@@ -844,6 +844,15 @@ def set_item(self, item, value, level=0, namespace=None):
...
             raise TypeError('Variable %s is read-only.' % self.name)
 
+        # Check whether the group allows writing to the variable (e.g. for
+        # synaptic variables, writing is only allowed after a connect)
+        try:
+            self.group.check_variable_write(variable)
+        except ReferenceError:
+            # Ignore problems with weakly referenced groups that don't exist
+            # anymore at this time (e.g. when doing neuron.axon.var = ...)
+            pass
```
thesamovar: Is this not a problem that should result in an error? Or does it come up somewhere else?

mstimberg: In what sense should it result in an error? Writing `neuron.axon.var = ...` works fine, but the `VariableView` for "var" does only hold a weak reference to the `SpatialNeuronSubgroup` that is generated on-the-fly for `neuron.axon`. We had a long discussions about `VariableView` holding a strong or weak reference to its source at some point (discussed mostly in the context of a monitors, though), and we decided to go with weak references, so I don't see anything obvious to change here.

thesamovar: You're right - was my misunderstanding!

PR #742, opened 2016-09-01
[MRG] Improve morphology performance

### LINKED ISSUE #741: Coordinates for `Morphology` are calculated inefficiently
`Morphology` sections only store relative coordinates internally, when you access their coordinates it has therefore to access the coordinates of the parents section (which in turn has to access its parent section, and so on) to calculate the absolute coordinates. This approach has many advantages over storing and updating absolute coordinates (it is much less error prone, composing sub-morphologies is easy, etc.) but it is obviously less efficient computationally.
The current version is really inefficient and does the several calculations more often than necessary. It also stores everything as `Quantity` objects, which makes simple operations such as additions or getting single elements relatively costly. For small morphologies (only the number of sections matters) this is not noticeable, but it can make for example plotting of reconstructed morphologies so slow that it becomes unusable (see brian-team/brian2tools#13).

mstimberg:
This fixes certain inefficiencies in the `Morphology` class. Coordinates are now internally stored without units (you can access the unitless coordinates using names with a trailing underscore) and a lot of unnecessary evaluations of (calculated) coordinates are removed. 

Plotting the morphology mentioned in brian-team/brian2tools#13 now takes 0.5s in 2D and 3.5s in 3D on my machine -- there's still room for improvement (also on the `brian2tools` side), but this is much better than "forever" :)

Closes #741

### DIFF
--- a/brian2/spatialneuron/morphology.py
+++ b/brian2/spatialneuron/morphology.py
@@ -738,75 +738,189 @@ class Morphology(object):
         '''
         raise NotImplementedError()
 
-    @abc.abstractproperty
+    @property
     def start_x(self):
         '''
         The x coordinate at the beginning of each compartment. Returns ``None``
         for morphologies without coordinates.
         '''
-        raise NotImplementedError()
+        start_x = self.start_x_
+        if start_x is None:
+            return None
+        else:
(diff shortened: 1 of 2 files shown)


thesamovar:
Great! Merging.

PR #745, opened 2016-09-06
[MRG] Support user-defined functions with boolean argument/return value

mstimberg:
It's a bit of a messy addition (we should do a bit of refactoring at some point, `parse_expression_unit` and `is_boolean_expression` could potentially be merged, treating "boolean" as a special unit as we do in other places), but I stumbled across this missing feature when documenting how to translate Brian 1's custom thresholds (defined by functions) to Brian 2. This was very cumbersome before this change since you could not define a boolean-returning function that actually gets recognized as such by the parsing system.

### DIFF
--- a/brian2/core/functions.py
+++ b/brian2/core/functions.py
@@ -120,6 +120,10 @@ class Function(object):
         self.sympy_func = sympy_func
         self._arg_units = arg_units
         self._return_unit = return_unit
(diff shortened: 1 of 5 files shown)


thesamovar:
Agreed. We'll probably want to refactor a few things later on, but at this point let's just get it out there. Merging and closing.

PR #746, opened 2016-09-07
[MRG] Correct linear dependency check in SpatialNeuron

### LINKED ISSUE #743: SpatialNeuron does not always detect non-linear dependencies on v
See discussion in https://groups.google.com/d/topic/brian-development/QNaudHGckQY/discussion

mstimberg:
This fixes #743, where `SpatialNeuron` did not detect a non-linear dependency on ``v`` when it was "hidden" in a point current. The code in `SpatialNeuron.__init__` is far from optimal, it mixes creating new equations from `SingleEquation` objects and from strings, this should be rewritten at some point... Still, this fixes the bug for now. I also enabled the `(constant over dt)`  flag for `SpatialNeuron`. This is quite important, since it is possible that some users cannot run their models with `SpatialNeuron` anymore now that it correctly detects a non-linearity in their equations. By marking the non-linear part as "constant over dt", everything should work again.

### DIFF
--- a/brian2/equations/equations.py
+++ b/brian2/equations/equations.py
@@ -639,16 +639,20 @@ class Equations(collections.Mapping):
         for name in self.names:
             Equations.check_identifier(name)
 
-    def get_substituted_expressions(self, variables=None):
+    def get_substituted_expressions(self, variables=None,
+                                    include_subexpressions=False):
         '''
         Return a list of ``(varname, expr)`` tuples, containing all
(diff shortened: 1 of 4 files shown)


thesamovar:
Looks good to me, happy for it to be merged.

mstimberg:
> Looks good to me, happy for it to be merged.

Ok, thanks -- tests passed, merging it now.

Issue #713, opened 2016-06-09
Update/fix documentation about running the standalone device repeatedly


mstimberg:
You cannot directly convert code that runs a simulation multiple times to standalone because the standalone device gathers all objects that have been created since it was activated (therefore, it would duplicate and not replace earlier objects). Some documentation suggests using `device.reinit()` which is actually no longer enough, due to changes around the clock system. What we do for testing is to call `reinit_devices()` which works, but is doing a bit more than necessary. There are also certain caveats, e.g. changes in `defaultclock.dt` are not preserved.

Not sure yet whether we need a new convenience function for the use case, but at least it should be clearly documented (and false information removed/updated).

Issue #708, opened 2016-06-07
Document stochastic equations (xi term) in user manual


thesamovar:
At the moment it shows up in the tutorial and a few other places, but it's not explained clearly in the user manual.

mstimberg:
Closed via #748

PR #748, opened 2016-09-09
[MRG] Docs misc for final

### LINKED ISSUE #747: Final cleanup before 2.0 final
Just look through the docs, etc. and make sure everything looks good.

thesamovar:
Just look through the docs, etc. and make sure everything looks good.

Code changes:

- [x] Fix the documentation build process so that there are no warnings/errors (maybe some errors in the rst files, or maybe in the build process, not sure)
- [x] ``DimensionMismatchError`` for a function call should not say ``dimensions were (1) (s)`` but rather ``dimension given was (1), expected (s)``. Also, maybe use longer names rather than short names in these types of errors? e.g. ``expected (second)``?

Documentation changes:

- [x] Update installation documentation "other useful packages" to point to jupyter instead of ipython? Also add brian2tools?
- [x] Installation documentation can have ``--pre`` options removed.
- [x] Update Known Issues (integer division, linear integration restrictions)
- [x] Add table of all known special symbols? (i, j, k, N, xi, etc.)
- [x] Document how writeable variables works, e.g. when explaining refractoriness?
- [x] Add more in "how to read this documentation" talking about navigating examples, reference documentation, etc.
- [x] Document stochastic variables (#708)

Closes #747, #708


thesamovar:
So far I've updated the user docs, and I think they're a lot friendlier now for a first time user. Each doc has basic information at the beginning, and an "advanced topics" section with more detailed information at the end. Would be good for someone else to look this over and check that it all makes sense.

Above are some suggestions of changes that we might want to make (to be discussed).

mstimberg:
> So far I've updated the user docs, and I think they're a lot friendlier now for a first time user. Each doc has basic information at the beginning, and an "advanced topics" section with more detailed information at the end. Would be good for someone else to look this over and check that it all makes sense.

I agree that re-arranging things a bit has made things clearer. I am not that convinced of the "Advanced" section, though. What is "advanced" or not depends a lot on the task at hand. I think for some stuff we can make the distinction quite clearly (e.g. "custom events", "user-defined functions", "interfacing with external code"), but others are really arbitrary (e.g. in the current state of this branch, "flags" are advanced in the equations docs, but "event-driven updates" are a standard topic in the synapses docs. `PoissonInput` is advanced but `TimedArray` is not). I certainly agree with putting the basic stuff first and more advanced stuff later, but I would not make a "Advanced" break somewhere in the middle (also it becomes a bit weird to have "advanced topics" in the "User's guide", but to also have an "Advanced guide"). What I think is more important is to have a quick overview from the start of the page (as we all know, no one will read a full page of documentation...). What about adding some:
```rst
.. content::
    :local:
```
to the start of all (long) pages (the only page where we currently use it are the "Known Issues")? The user can then quickly jump to the stuff they are interested in, regardless of whether it counts as "advanced" or not. For example, I don't think anyone will think of going down to "Advanced Topics" on the "Recording during a simulation" page, to find out about `get_states()`.

Other random remarks:
* I think it makes sense to move the user-defined functions to the "Advanced guide", but many beginners ask about the functions that can be used in equations. Maybe include that part in the equations documentation?
* Maybe the "list of special symbols" could also go into the equation docs?
* I think the whole "Computational methods and efficiency" page could move into the "Advanced guide"

thesamovar:
I see what you mean. What I was trying to get was to make the users first go through the docs a bit less heavy. At the moment in the user guide there are definitely a few things that you don't need to read the first time. The idea of "advanced topics" break was just that you can come back to it later, but it's still in the most logical section. We could call it "further topics" rather than "advanced", but I think you want a more substantial change that that? What does this content local thing do actually?

Agree on moving special symbols to equations, and also adding list of built-in functions there too. Also happy to move computational methods and efficiency to advanced guide now that it's mostly automatic to get codegen working.

thesamovar:
Notes from meeting:
- For longer pages, add a table of contents
- Add a note to divide between advanced topics rather than a section (http://www.sphinx-doc.org/en/stable/rest.html#substitutions)
- Use clear section names for discoverability

thesamovar:
I just pushed a change - what do you think of how that looks? If you're happy I'll apply it everywhere.

mstimberg:
Looks very good to me!

thesamovar:
I just added some explanation of xi and SDEs but I'm not entirely happy with it. What do you think? Tricky to get the balance right between saying something correct and saying something comprehensible for a beginner.

thesamovar:
OK, I'm done with docs. Over to you @mstimberg for installation notes and anything else.

thesamovar:
Should we add a support page to the docs rather than just having that on the website? Maybe in the introduction?

mstimberg:
> I just added some explanation of xi and SDEs but I'm not entirely happy with it. What do you think? Tricky to get the balance right between saying something correct and saying something comprehensible for a beginner.

I think the start is great, than in becomes a bit frustrating after the example code. I don't have a good idea either, how to make it clearer, though :-/ Maybe add something about why one shouldn't use `randn` directly? It would be great if the text would answer the kind of question we had on the mailing list about 2 weeks ago (which we did not answer yet...). It would be great if we simply had some place to link to for a good explanation. http://www.scholarpedia.org/article/Stochastic_dynamical_systems#Ornstein-Uhlenbeck_process is not too bad, but also does not go into much depth.
Maybe the text should not only mention xi, but also that noise can be modeled at many places (noisy threshold, noisy synaptic propagation, ...).
I agree it is not easy!
Two minor remarks:
* I'd call it "Noise" (i.e., in quotation marks)
* we should link to it from the special symbols table

mstimberg:
> Should we add a support page to the docs rather than just having that on the website? Maybe in the introduction?

Ah yes, in fact I'd put it right on the start page, together with the other text of the overview.

thesamovar:
OK, support page added (as a page so we can link from the website).

mstimberg:
Great, support page looks good to me. I added `brian2tools` to the installation instructions and updated a few instances of `DimensionMismatchError` to match the actual error messages (which are really quite nice now!). 
I'll have a final look over the docs, but I think this PR is ready to get closed.

thesamovar:
Just working on the text for SDEs now!

mstimberg:
> Just working on the text for SDEs now!

Neat :)

thesamovar:
OK, done - what do you think?

Also, should we update the setup script etc. here or will you make another branch to finalise the release?

mstimberg:
Great, I think the SDE explanation is good enough, certainly much better than everything we had so far.

> Also, should we update the setup script etc. here or will you make another branch to finalise the release?

Let's stick to the docs here and do the setup change (and the new version number!)  separately (as the last thing).

mstimberg:
I did a last change to fix #713. You currently have to call `device.reinit()` *and* `device.activate()` to re-use a standalone device in the same script. This could certainly be more straightforward, but I don't want to add yet another method (we currently have `Device.reinit`, `Device.activate`, `reset_device`, `reinit_devices`...) and making e.g. just a single call to `device.reinit` work is not trivial. The situation is quite fragile with the clocks currently, as we have a bit of a chicken-egg problem with a lot of places needing a clock, but clocks are now normal state variables and have therefore the same limitations as other state variables in standalone mode.
Long story short: this is a really advanced and rare use case, and it works with the somewhat inconvenient reinit+activate -- I'd prefer to not touch this any further so late before a release...

Oh, the commit also now raises an error if you call `device.build` multiple times for standalone -- this previously worked but included all previous objects in the newly generated code. In short, a big mess, nothing that a user'd actually want.

With that, we are ready to merge here I think, still have to wait for the tests to run through...

thesamovar:
Oh, one last docs request. Can we update the docs so that they have the Brian logo like it does for the Brian 1 readthedocs page? https://brian.readthedocs.io/en/stable/

mstimberg:
> Oh, one last docs request. Can we update the docs so that they have the Brian logo like it does for the Brian 1 readthedocs page? https://brian.readthedocs.io/en/stable/

Of course, done!

mstimberg:
> Should I start running tests on my various machines or wait until we merge this one?

Actually, how about merging it now? The reported test status here on the github page is not correct (it does not work if the last commit is one with `[ci skip]`): travis ran through both branch and on the merged version, appveyor ran the branch but will still take > 1h30 to run the tests on the merged version. Given that it will run exactly the same test again after the merge, I think it is ok to take a short-cut here...

thesamovar:
Yep!

PR #749, opened 2016-09-09
[MRG] Small refactoring to only ask for the platform information once

mstimberg:
It's a small refactoring change, but we were duplicating some code to enable 64bit->32bit cross-compilation in several places. It also turns out that getting the platform information is somewhat costly (at least on my machine) -- it's only a couple of milliseconds but in the previous code it was asking for this information again and again for each `CodeObject` creation. In a small test with a simulation that included many short `run` statements in a loop (far from optimal with current Brian 2...), this made actually a significant performance difference!

### DIFF
--- a/brian2/codegen/codeobject.py
+++ b/brian2/codegen/codeobject.py
@@ -2,6 +2,7 @@
 Module providing the base `CodeObject` and related functions.
 '''
 import copy
+import platform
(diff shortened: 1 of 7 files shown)


thesamovar:
Looks fine to me - happy for this to be merged.

PR #750, opened 2016-09-10
[MRG] Warn about unstable integration

### LINKED ISSUE #554: Warn about unstable integration

thesamovar:
This isn't quite ready yet, but the basic ideas are there. Would be interested if you took a look and gave me some feedback @mstimberg.

The first thing is that by changing the default compiler options for MSVC we avoid the problem reported in #626 (see discussion there).

The next thing is that after a run, or on loading a state variable from disk, we check for nans or large values and warn if we find any. Currently, this isn't passing tests because it keeps raising the following error:

```
ERROR: brian2.tests.test_stateupdaters.test_refractory_stochastic
    self.test(*self.arg)
  File "C:\programming\brian2\brian2\tests\test_stateupdaters.py", line 683, in test_refractory_stochastic
    net.run(10*ms)
  File "C:\programming\brian2\brian2\core\base.py", line 278, in device_override_decorated_function
    return func(*args, **kwds)
  File "C:\programming\brian2\brian2\units\fundamentalunits.py", line 2424, in new_f
    result = f(*args, **kwds)
  File "C:\programming\brian2\brian2\core\network.py", line 896, in run
    obj._check_for_invalid_states()
  File "C:\programming\brian2\brian2\groups\group.py", line 943, in _check_for_invalid_states
    #     return
  File "C:\programming\brian2\brian2\groups\group.py", line 522, in get_states
    data = ImportExport.methods[format].export_data(self, vars, units=units, level=level)
  File "C:\programming\brian2\brian2\importexport\dictlike.py", line 22, in export_data
    level=level+1),
  File "C:\programming\brian2\brian2\groups\neurongroup.py", line 516, in state
    return Group.state(self, name, use_units=use_units, level=level+1)
  File "C:\programming\brian2\brian2\groups\group.py", line 349, in state
    return var.get_addressable_value_with_unit(name=name, group=self)
  File "C:\programming\brian2\brian2\core\variables.py", line 256, in get_addressable_value_with_unit
    return self.get_value_with_unit()
  File "C:\programming\brian2\brian2\core\variables.py", line 208, in get_value_with_unit
    return Quantity(self.get_value(), self.unit.dimensions)
  File "C:\programming\brian2\brian2\core\variables.py", line 374, in get_value
    raise TypeError('Cannot get the value for an auxiliary variable.')
TypeError: Cannot get the value for an auxiliary variable.
```

It seems this error is just being raised by calling ``get_states()`` on certain types of object.

The check for "large" values is ``abs(v)>1e150`` - I chose this value because SpikeGeneratorGroup has a period of 1e100. However, we could reduce the size of the default period instead of having such a massive threshold for "too large" values. A value of 1e50 seems like it would be better, since the largest SI prefix is 1e24.

Might also be worth adding some tests, and checking performance implications of this test.

Fixes #554

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -133,6 +133,9 @@ install:
     conda create -n travis_conda --yes --quiet pip python=$PYTHON numpy scipy nose sphinx ipython sympy pyparsing jinja2 setuptools coverage;
     fi
   - source activate travis_conda
+  - if [ ${PYTHON:0:1} == "2" ]; then
+    conda install --yes --quiet bsddb;  # makes weave code generation much faster
+    fi
   - conda install --yes --quiet -c brian-team py-cpuinfo  # install cpuinfo from our own channel
   - if [[ $CYTHON == 'yes' ]]; then conda install --yes cython; SETUP_ARGS=--with-cython; else SETUP_ARGS=''; fi
   - if [[ $REPORT_COVERAGE == 'yes' ]]; then pip install -q coveralls; fi
--- a/.travis.yml
+++ b/.travis.yml
@@ -133,8 +133,9 @@ install:
     conda create -n travis_conda --yes --quiet pip python=$PYTHON numpy scipy nose sphinx ipython sympy pyparsing jinja2 setuptools coverage;
     fi
   - source activate travis_conda
+  # makes weave code generation much faster:
   - if [ ${PYTHON:0:1} == "2" ]; then
-    conda install --yes --quiet bsddb;  # makes weave code generation much faster
+    conda install --yes --quiet bsddb;
     fi
   - conda install --yes --quiet -c brian-team py-cpuinfo  # install cpuinfo from our own channel
   - if [[ $CYTHON == 'yes' ]]; then conda install --yes cython; SETUP_ARGS=--with-cython; else SETUP_ARGS=''; fi
--- a/brian2/codegen/cpp_prefs.py
+++ b/brian2/codegen/cpp_prefs.py
@@ -56,7 +56,7 @@ prefs.register_preferences(
         '''
         ),
     extra_compile_args_msvc=BrianPreference(
-        default=['/Ox', '/w', '/fp:fast', msvc_arch_flag],
+        default=['/Ox', '/w', msvc_arch_flag],
         docs='''
         Extra compile arguments to pass to MSVC compiler (the default
         ``/arch:`` flag is determined based on the processor architecture)
--- a/brian2/codegen/cpp_prefs.py
+++ b/brian2/codegen/cpp_prefs.py
@@ -56,7 +56,7 @@ prefs.register_preferences(
         '''
         ),
     extra_compile_args_msvc=BrianPreference(
-        default=['/Ox', '/EHsc', '/w', '/fp:fast', msvc_arch_flag],
+        default=['/Ox', '/w', '/fp:fast', msvc_arch_flag],
         docs='''
         Extra compile arguments to pass to MSVC compiler (the default
         ``/arch:`` flag is determined based on the processor architecture)
--- a/brian2/codegen/cpp_prefs.py
+++ b/brian2/codegen/cpp_prefs.py
@@ -50,7 +50,7 @@ prefs.register_preferences(
         '''
         ),
     extra_compile_args_gcc=BrianPreference(
-        default=['-w', '-O3', '-ffast-math', '-march=native'],
+        default=['-w', '-O3', '-ffast-math', '-fno-finite-math-only', '-march=native'],
         docs='''
         Extra compile arguments to pass to GCC compiler
         '''
--- a/brian2/core/network.py
+++ b/brian2/core/network.py
@@ -18,6 +18,7 @@ from brian2.core.names import Nameable
 from brian2.core.base import BrianObject, brian_object_exception
 from brian2.core.clocks import Clock, defaultclock
 from brian2.devices.device import device
+from brian2.groups.group import Group
 from brian2.units.fundamentalunits import check_units, Quantity
 from brian2.units.allunits import second, msecond
 from brian2.core.preferences import prefs, BrianPreference
@@ -888,6 +889,11 @@ class Network(Nameable):
             logger.debug('\n' + str(profiling_summary(self)))
         else:
             self._profiling_info = None
+
+        # check for nans
+        for obj in self.objects:
+            if isinstance(obj, Group):
+                obj._check_for_invalid_states()
 
     @device_override('network_stop')
     def stop(self):
--- a/brian2/core/network.py
+++ b/brian2/core/network.py
@@ -874,6 +874,11 @@ class Network(Nameable):
         else:
             device._last_run_completed_fraction = 1.0
 
+        # check for nans
+        for obj in self.objects:
+            if isinstance(obj, Group):
+                obj._check_for_invalid_states()
+
         if report is not None:
             report_callback((end_time-start_time)*second, 1.0, t_start, duration)
         self.after_run()
@@ -889,11 +894,6 @@ class Network(Nameable):
             logger.debug('\n' + str(profiling_summary(self)))
         else:
             self._profiling_info = None
-
-        # check for nans
-        for obj in self.objects:
-            if isinstance(obj, Group):
-                obj._check_for_invalid_states()
 
     @device_override('network_stop')
     def stop(self):
--- a/brian2/core/variables.py
+++ b/brian2/core/variables.py
@@ -371,7 +371,7 @@ class AuxiliaryVariable(Variable):
                                                 scalar=scalar)
 
     def get_value(self):
-        raise TypeError('Cannot get the value for an auxiliary variable.')
+        raise TypeError('Cannot get the value for an auxiliary variable (%s).' % self.name)
 
 
 class ArrayVariable(Variable):
--- a/brian2/devices/cpp_standalone/device.py
+++ b/brian2/devices/cpp_standalone/device.py
@@ -851,10 +851,23 @@ class CPPStandaloneDevice(Device):
                 self._last_run_time, self._last_run_completed_fraction = map(float, last_run_info.split())
 
         # Make sure that integration did not create NaN or very large values
-        owners = [var.owner for var in self.arrays
-                  if isinstance(var.owner, Group)]
+        owners = [var.owner for var in self.arrays]
+        # We don't want to check the same owner twice but var.owner is a
+        # weakproxy which we can't put into a set. We therefore store the name
+        # of all objects we already checked. Furthermore, under some specific
+        # instances a variable might have been created whose owner no longer
+        # exists (e.g. a `_sub_idx` variable for a subgroup) -- we ignore the
+        # resulting reference error.
+        already_checked = set()
         for owner in owners:
-            owner._check_for_invalid_states()
+            try:
+                if owner.name in already_checked:
+                    continue
+                if isinstance(owner, Group):
+                    owner._check_for_invalid_states()
+                    already_checked.add(owner.name)
+            except ReferenceError:
+                pass
 
 
     def build(self, directory='output',
--- a/brian2/devices/cpp_standalone/device.py
+++ b/brian2/devices/cpp_standalone/device.py
@@ -415,9 +415,9 @@ class CPPStandaloneDevice(Device):
                     if var.size[0] * var.size[1] == len(data):
                         size = var.size
                     elif var.size[0] == 0:
-                        size = (len(data)/var.size[1], var.size[1])
-                    elif var.size[0] == 0:
-                        size = (var.size[0], len(data)/var.size[0])
+                        size = (len(data)//var.size[1], var.size[1])
+                    elif var.size[1] == 0:
+                        size = (var.size[0], len(data)//var.size[0])
                     else:
                         raise IndexError(('Do not now how to deal with 2d '
                                           'array of size %s, the array on disk '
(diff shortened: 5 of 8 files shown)


thesamovar:
The auxiliary variables were xi, xi_1, etc. I just added a check to skip these in ``VariableOwner.get_states`` and now all tests pass. Will just add a test and maybe do a quick performance check.

thesamovar:
This seems like it might be a problem with multicompartmental?

```
ERROR: Rallpack 2
    self.test(*self.arg)
  File "C:\programming\brian2\brian2\tests\test_spatialneuron.py", line 343, in test_rallpack2
    run(250*ms + defaultclock.dt)
  File "C:\programming\brian2\brian2\units\fundamentalunits.py", line 2424, in new_f
    result = f(*args, **kwds)
  File "C:\programming\brian2\brian2\core\magic.py", line 371, in run
  File "C:\programming\brian2\brian2\core\magic.py", line 231, in run
  File "C:\programming\brian2\brian2\core\base.py", line 278, in device_override_decorated_function
    return func(*args, **kwds)
  File "C:\programming\brian2\brian2\units\fundamentalunits.py", line 2424, in new_f
    result = f(*args, **kwds)
  File "C:\programming\brian2\brian2\core\network.py", line 880, in run
    obj._check_for_invalid_states()
  File "C:\programming\brian2\brian2\groups\group.py", line 945, in _check_for_invalid_states
    self._check_for_invalid_values(k, v)
  File "C:\programming\brian2\brian2\groups\group.py", line 961, in _check_for_invalid_values
    raise ValueError("nan for {name}.{k} = {v}".format(name=self.name, k=k, v=v))
ValueError: nan for spatialneuron.y = [  8.00000000e-06              nan              nan ...,              nan
              nan              nan]
```

thesamovar:
This is another error I don't quite understand yet...

```
ERROR: Test that `SpikeGeneratorGroup` connects properly.
    self.test(*self.arg)
  File "C:\programming\brian2\brian2\tests\test_spikegenerator.py", line 29, in test_spikegenerator_connected
    run(7*ms)
  File "C:\programming\brian2\brian2\units\fundamentalunits.py", line 2424, in new_f
    result = f(*args, **kwds)
  File "C:\programming\brian2\brian2\core\magic.py", line 371, in run
  File "C:\programming\brian2\brian2\core\magic.py", line 231, in run
  File "C:\programming\brian2\brian2\core\base.py", line 278, in device_override_decorated_function
    return func(*args, **kwds)
  File "C:\programming\brian2\brian2\units\fundamentalunits.py", line 2424, in new_f
    result = f(*args, **kwds)
  File "C:\programming\brian2\brian2\core\network.py", line 880, in run
    obj._check_for_invalid_states()
  File "C:\programming\brian2\brian2\groups\group.py", line 943, in _check_for_invalid_states
    state = self.get_states()
  File "C:\programming\brian2\brian2\groups\group.py", line 526, in get_states
    data = ImportExport.methods[format].export_data(self, vars, units=units, level=level)
  File "C:\programming\brian2\brian2\importexport\dictlike.py", line 23, in export_data
    copy=True, subok=True)
  File "C:\programming\brian2\brian2\core\variables.py", line 1186, in __array__
    value = self[:]
  File "C:\programming\brian2\brian2\core\variables.py", line 820, in __getitem__
    return self.get_item(item, level=1)
  File "C:\programming\brian2\brian2\core\variables.py", line 812, in get_item
    values = self.get_with_index_array(item)
  File "C:\programming\brian2\brian2\core\base.py", line 278, in device_override_decorated_function
    return func(*args, **kwds)
  File "C:\programming\brian2\brian2\core\variables.py", line 1096, in get_with_index_array
    return variable.get_value()[indices]
IndexError: index 6 is out of bounds for axis 0 with size 6
```

thesamovar:
OK I fixed the problems above by only running the nan check for objects with a state updater (which is what it's designed to deal with). However, they are both still issues. Namely, the ``test_rallpack2`` test is producing nans which it presumably isn't supposed to, and ``get_states()`` is raising an error for ``SpikeGeneratorGroup`` (which again, feels like a problem but doesn't cause an error any more because we just don't do this). Maybe open new issues for those?

Otherwise, I think this is ready to go. I added two new tests and checked that standalone passes too. I didn't check performance implications, but it just runs through every variable once at the end of a run, which must be much less work than a single timestep, so this must be pretty insignificant.

I think this is ready to merge (and maybe open a couple of new issues for the problems it exposed).

thesamovar:
Also once we merge this, we can remove the high priority label and 2.0 milestone from #626.

mstimberg:
Nice work, I'm not 100% convinced yet, though. I think we have two kind of potential problems: 
1. the use of a linear integrator that lead to square roots of negative values or similar mistakes in the code 
2. a simulation that is unstable (e.g. an Euler integration with a large dt)

IMO, the first one should lead to a hard error, a warning is somewhat easy to overlook and I don't see that this simulation could lead to useful results. You removed the fast math flag for Windows so that this should now lead to NaNs which are caught by the added checks, but why not `if(errno) throw...`? On the other hand, then we should probably use `numpy.seterr` as well to raise an error from numpy under the same condition. Hmm, maybe it is not a good idea, after all. I could imagine something like a parallel parameter exploration (think of model fitting) where some randomly chosen parameter leads to a division by zero or something like that. In that case, stopping the whole simulation seems excessive. Not sure, maybe your removal of the fast-math flag *is* the best short-term solution... 

About the general "is NaN or very large" check. Wouldn't it make sense to put this in `StateUpdater.after_run`? This way you don't need a somewhat hacky `if hasattr('state_updater')` check and we could for example have a similar check in `SpatialStateUpdater`. Also, do we have to warn about all values or should we rather warn about values *introduced by the simulation*? E.g. similar to the spike generator period, I could imagine that some user-defined parameter is set to a very high value or to `inf` (or even `NaN`, say to denote something like "no value" or "N/A") -- it would be weird to warn about that. Maybe we should use `before_run` to take note of all "already big/inf/NaN" values and only warn about those that changed? But then, we don't want to have a boolean array of the same size of all variables around either :-/ BTW: The NaN values in `test_rallpack2` are an example of a case where there should be no warning, the `y` values are set to NaN in the beginning and are unchanged during the simulation. You get NaN values in a morphology if it uses a mix of sections that have spatial coordinates and others that don't (which happens e.g. if you load a morphology and add an artificial axon) -- in this test this is of course not at all necessary, we could simply change it to not use any coordinates. Actually, at least variables marked as `constant` (like the coordinates) should be excluded from the check. Um, thinking about it, we only need to check variables defined by differential equations, no?

Two more thoughts:
* My proposal above to put in in `after_run` does not quite work for standalone...
* I'm not too worried about the performance penalty either

I thought I had already fixed `SpikeGeneratorGroup.get_states` at some point, but apparently not... It has something to do with the fact that a `SpikeGeneratorGroup` has a size `N`, but this is not the size of the stored indices and times. This is not correctly specified in the indices somewhere.

thesamovar:
I think we don't want to raise an error for large values or nans for the reason you state: for parameter exploration (which I do regularly and I'm sure others do too), this would be incredibly problematic with no obvious good way to work around it. I think it's OK though. We downgraded a lot of warning messages to info level precisely so that warnings would stand out as serious, and I think this qualifies perfectly for that level.

The reason I put the check where it was is for two reasons: (1) to make it work with standalone (this was quite tricky to get right actually and explains the slightly odd code), and (2) to make it work with magic (since MagicNetwork deletes all objects in ``after_run`` itself). I think the way I did it is a fairly reasonable solution for minimising the amount of duplicated code but looks a bit odd and maybe has some restrictions. I'd be inclined to leave it if we can fix the problem below because it works, we've got a tight deadline, and it can easily be changed in the future if we don't like it (there's no user-facing component).

The problem that there should be nans in spatial neurons is a serious one. Any thoughts on how best to address it? Does StateUpdater give a list of variables that it modifies? If so, the easiest think to do that would address many of the problems you mention would be to just check the diffeq variables as you say. Otherwise, we'd need some way of extracting the diffeq variables that works for both NeuronGroup and SpatialNeuron. In any case, my vote is for doing something along these lines. What do you think?

Open a new issue (not for 2.0) for SpikeGeneratorGroup.get_states?

mstimberg:
> I think we don't want to raise an error for large values or nans for the reason you state: for parameter exploration (which I do regularly and I'm sure others do too), this would be incredibly problematic with no obvious good way to work around it. I think it's OK though. We downgraded a lot of warning messages to info level precisely so that warnings would stand out as serious, and I think this qualifies perfectly for that level.

Ok, agreed, let's forget about the error here. I also realized that checking `errno` does not help us that much in the general case, one of the consequences of `-ffast-math` is `-fno-math-errno`...

> The reason I put the check where it was is for two reasons: (1) to make it work with standalone (this was quite tricky to get right actually and explains the slightly odd code), and (2) to make it work with magic (since MagicNetwork deletes all objects in after_run itself). I think the way I did it is a fairly reasonable solution for minimising the amount of duplicated code but looks a bit odd and maybe has some restrictions. I'd be inclined to leave it if we can fix the problem below because it works, we've got a tight deadline, and it can easily be changed in the future if we don't like it (there's no user-facing component).

I have to say the whole `after_run` handling is a bit of a mess (e.g. `MagicNetwork.after_run` does not actually call `Network.after_run`, which would break all `after_run` methods if we had any that weren't empty...) and is not correctly implemented for standalone (it is called once per build instead of once per run). I still think it is the only reasonable place to put such a check, since the check has to be specific to the type of object. I am close to thinking that it would be better to not have this check at all rather than a half-baked one (mostly because from my experience, these kind of "workarounds" tend to stay around for a long time :-) ).



> The problem that there should be nans in spatial neurons is a serious one.

I think it's not only a problem for spatial neurons, I don't see why setting a value to a high/inf/nan value *before a run* should get a warning message *after a run*, this would be very confusing (and I think almost worse than not having any warning). 
 
> Does StateUpdater give a list of variables that it modifies? If so, the easiest think to do that would address many of the problems you mention would be to just check the diffeq variables as you say. Otherwise, we'd need some way of extracting the diffeq variables that works for both NeuronGroup and SpatialNeuron. In any case, my vote is for doing something along these lines. What do you think?

In the long run, it would be great to have a general mechanism that notes for each `CodeObject` which variables it reads and writes (this would allow for fancy stuff like parallelization of `CodeObjects` that don't affect each other). For now, if we want to do a quick and dirty solution we can just use the `equations` object (all objects that have a `state_updater` happen to have `equations` as well) which provides the differential equations. This will work for `Synapses`, `NeuronGroup` and `SpatialNeuron` (at least for the integration of the channels: there's also `SpatialStateUpdater` that is updating `v` that would be nice to check for NaN, but in this case the warning message should also be different...). 

If we are only ever checking it for objects that have a state updater, do we need the `only_if_has_state_updater` keyword, actually?

Hmpf, I'm really not happy with this solution, maybe I should try to do it "properly" as the first thing to improve after the 2.0 release :)

mstimberg:
> Open a new issue (not for 2.0) for SpikeGeneratorGroup.get_states?

I wanted to open an issue but I was not able to reproduce the issue, do you have a simple example that triggers it?

thesamovar:
OK so how about we change the current hack to check for the existence of an ``equations`` attribute and just check the diffeq values in that case? We should avoid any user-introduced nans like that, and nans for non-diffeq variables. Happy to add a new "fix this post-2.0" issue too! Gotta run and do some marking now, but talk to you later.

thesamovar:
Very quickly, I think adding a ``get_values()`` for each relevant object after the run calls in ``test_spikegenerator_connected()`` should raise it.

mstimberg:
> Very quickly, I think adding a get_values() for each relevant object after the run calls in test_spikegenerator_connected() should raise it.

Ah thanks. Actually `SpikeGeneratorGroup.get_states` works, but using `Synapses.get_states` when connected to a `SpikeGeneratorGroup` does not. This is because of the somewhat weird `spike_number` variable, I opened an issue about it (if I am right then it is actually a trivial fix): #752

mstimberg:
So I checked whether `-ffast-math` still has the big performance impact that we saw earlier (given that we now pull expressions out of the loop) and, well, yes, it is still a big difference. As an example, the COBAHH example takes 7s on my machine with `-ffast-math` and 37s without... I also noticed that I can make the test you write fail with `-ffast-math` switched on, if I (re-)include the threshold and the reset. Odd, but well.
I am not sure, the safest solution would be of course to switch off `-ffast-math`, in the same way as you did for Windows. But this will give all of a sudden much slower simulations, and that does not make users happy. We can of course document how to switch `-ffast-math` on but then it becomes something for the advanced users only.
I currently opted for an intermediate solution: I kept `-ffast-math` but added `-fno-finite-math-only` (which overrides the opposite setting of `-ffast-math`). This means that gcc is still free to re-arrange stuff, replace divisions by multiplications, not use `errno`, etc. *but* it no longer assumes that math functions do not receive or result in NaN/Inf. And if I am not mistaken, this is the thing that really matters to us (e.g. the test no longer fails and correctly returns NaN). Unfortunately, we don't have this kind of control for msvc. What do you think?

mstimberg:
Forgot to mention it: COBAHH with `-ffast-math -fno-finite-math-only` takes 8s on my machine.

thesamovar:
OK, COBAHH on MSVC takes 21s versus 19s with ``/EHsc /fp:fast`` off versus on, so I think we can just use your updated settings for gcc and leave the current version for MSVC and it's no big deal. We can add an issue to look into common subexpression optimisations for post-2.0.

thesamovar:
OK, this is ready to merge when tests pass then.

PR #753, opened 2016-09-13
[MRG] Define `spike_number` as its own index in `SpikeGeneratorGroup`

### LINKED ISSUE #752: SpikeGeneratorGroup's "spike_number" variable is not correctly indexed
`spike_number` is only used as an index for `neuron_index` and `spike_time`, but since it does not define an index itself it will be indexed with the group's default index which won't work: the group's size `N` refers to the number of "neurons" in the group and not the number of spikes stored. Setting the index to itself should actually fix it...

mstimberg:
This avoids that it is indexed with the default index and therefore with the size of the group (which is the number of "neurons", not spikes)
Fixes #752


thesamovar:
Good to merge.

mstimberg:
I'll take the liberty and merge this right away even though the test suites for the PR have not run through yet (they already ran and passed on the branch, though). It's a tiny change and we are busy commiting and merging changes which trigger the test suite anyway.

Issue #764, opened 2016-10-04
Using "(unless refractory)" without refractory keyword does not give helpful error message


mstimberg:
In current master, you'll get the rather unhelpful error message
```Python
KeyError: 'not_refractory'
```
if you use `(unless refractory)` but forget about setting the `refractory` keyword.

PR #765, opened 2016-10-10
[MRG] Re-enable the warning about a missing `refractory` argument

mstimberg:
It turned out #764 was a regression, the `KeyError` was raised before the warning we already had in place. This commit fixes this and adds a test for the presence of the warning.

A general remark: I activated the new "review required" feature that github introduced recently. Instead of just commenting and saying "good to merge", we can now do an official review and set the status -- let's see how this goes!
Not relevant for the current PR:= but github also has a new nice "rebase and merge" feature, we should try this when a branch got outdated by other merges, this should make our development history easier to follow.

### DIFF
--- a/brian2/groups/neurongroup.py
+++ b/brian2/groups/neurongroup.py
@@ -750,10 +750,13 @@ class NeuronGroup(Group, SpikeSource):
 
         # Add the conditional-write attribute for variables with the
         # "unless refractory" flag
-        for eq in self.equations.itervalues():
-            if eq.type == DIFFERENTIAL_EQUATION and 'unless refractory' in eq.flags:
-                not_refractory_var = self.variables['not_refractory']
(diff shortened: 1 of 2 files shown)


thesamovar:
I'm not against "rebase and merge" provisionally ;) but I've had bad experiences of rebasing in the past so I'm reserving judgement...

mstimberg:
> I'm not against "rebase and merge" provisionally ;) but I've had bad experiences of rebasing in the past so I'm reserving judgement...

Rebasing and force-pushing on a branch can be annoying for others, but in this case it will be merged anyway, so I don't see where it could become a problem. If there are non-trivial changes, then rebasing will lead to conflicts that need manual resolution anyway, you cannot use the mechanism in that case. But let's wait and see, we don't have a use for this very often, I think.

PR #770, opened 2016-11-07
Added example for using TimedArray in PoissonGroup

flinz:
In follow up to #660, I added a minimal example for inhomogeneous Poisson processes to the docs.
I also mention inhomogeneous Poisson on purpose, since this is something bound to be text searched for (at least I did).

### DIFF
--- a/docs_sphinx/user/input.rst
+++ b/docs_sphinx/user/input.rst
@@ -11,8 +11,8 @@ Input stimuli
 
 There are various ways of providing "external" input to a network.
 
-Poisson input
--------------
+Poisson inputs
+--------------
 For generating spikes according to a Poisson point process, `PoissonGroup` can
 be used, e.g.::
 
@@ -21,14 +21,11 @@ be used, e.g.::
     S = Synapses(P, G, on_pre='v+=0.1')
     S.connect(j='i')
 
-It takes a rate, an array of rates (one rate per neuron), or a string
-expression evaluating to a rate as an argument and can be connected to a
-`NeuronGroup` via the usual `Synapses` mechanism.
+See `More on Poisson inputs`_ below for further information.
 
-Note that `PoissonGroup` won't work for very high rates where there is a
-non-negligible probability of two spikes occurring in the same time step,
-as only one spike is allowed to occur. See the advanced topics below for
-ways to get around this.
+For simulations where the individually generated spikes are just used as a 
+source of input to a neuron, the `PoissonInput` class provides a more efficient
+alternative: see :ref:`poisson_input_class` below for details.
 
 Spike generation
 ----------------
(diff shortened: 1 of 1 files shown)


docs_sphinx/user/input.rst
```diff
@@ -155,10 +155,21 @@ Poisson inputs
 If the given value for ``rates`` is a constant, then using
 ``PoissonGroup(N, rates)`` is equivalent to
 ``NeuronGroup(N, 'rates : Hz', threshold='rand()<rates*dt')`` and setting the
```
mstimberg: I know that this is not actually your change, but I think it would render better if the `NeuronGroup(...)` line would be on its own (i.e. with `::` and then indented), instead of just quoted.

docs_sphinx/user/input.rst
```diff
@@ -155,10 +155,21 @@ Poisson inputs
...
-group's ``rates`` attribute. If ``rates`` is a string, then it is equivalent
+group's ``rates`` attribute.
+
+If ``rates`` is a string, then it is equivalent
 to ``NeuronGroup(N, 'rates = ... Hz', threshold='rand()<rates*dt')`` with the
```
mstimberg: Same as above (i.e. please put it on its own line), also I think including the colon before the unit would improve readibility a bit (i.e. `'rates = ... : Hz'`).

docs_sphinx/user/input.rst
```diff
@@ -155,10 +155,21 @@ Poisson inputs
...
 respective expression for the rates (which will be evaluated at every time step
-and therefore allows time-dependent rates). Note that, as can be seen in its
+and therefore allows time-dependent rates, i.e. inhomogeneous Poisson 
+processes). For example, the following code (see also `Timed arrays`_) uses a 
+`TimedArray` to define the rates of `PoissonGroup` as a function of time, 
+resulting in five 100ms blocks of 100 Hz stimulation, followed by 100ms
+silence::
+
+    stimulus = TimedArray([100., 0.] * 5, dt=100.*ms)
```
mstimberg: I'd avoid using the "list multiplication" (it certainly has a better name?): in Brian 2 you can write things like `[0, 1]*mV` so using multiplication in a way that does not work for numpy arrays might be confusing. I'd either write it out completely or use `np.tile([100., 0.], 5)`.

flinz:
> And even though of course you did not introduce this problem, if you could find more distinguishable headings than "Poisson input" and "Poisson inputs", I'd be very grateful..

Mind if I just merge the two blocks (`Poisson Input` and `Poisson Inputs`)? The first block already mentions half of what is repeated in the second, and the usage of `PoissonInput` would actually be helpful even for non-advanced users.

mstimberg:
> Mind if I just merge the two blocks (Poisson Input and Poisson Inputs)? The first block already mentions half of what is repeated in the second, and the usage of PoissonInput would actually be helpful even for non-advanced users.

Well, that might indeed be the easiest solution... Yes, I'm fine with this.

thesamovar:
I'd rather keep them separate: I was trying to keep the bit before the "advanced" cutoff line really simple for first time readers. Why not just "More on Poisson inputs"?

mstimberg:
I think it was not only a problem of the title but also that the first half of the "More on Poisson inputs" section was a repetition of the "Poisson input" section in different words. But I agree that some part of the second section is implementation details that are not necessary for first-time readers. Hmm, I remember not being a big fan of the "advanced section" approach, since it is really arbitrary (e.g. I'm not sure that more users use `run_regulary` than `PoissonInput`)...

flinz:
I incorporated the line comments, keeping the sections split. I moved everything advanced to below, and split the topics a little.

Sorry for many pushes, I somehow can't build the sphinx docs locally to check the rst syntax on commits (this fails with `ImportError: No module named optimisation`). If you prefer a squashed commit, I can reopen a new PR.

docs_sphinx/user/input.rst
```diff
@@ -21,14 +21,11 @@ be used, e.g.::
     S = Synapses(P, G, on_pre='v+=0.1')
     S.connect(j='i')
 
-It takes a rate, an array of rates (one rate per neuron), or a string
-expression evaluating to a rate as an argument and can be connected to a
-`NeuronGroup` via the usual `Synapses` mechanism.
+See `More on Poisson inputs`_ below for other advanced topics.
```
mstimberg: "other advanced topics" sounds a bit odd, how about "for further information"?

mstimberg:
> Sorry for many pushes, I somehow can't build the sphinx docs locally to check the rst syntax on commits (this fails with ImportError: No module named optimisation).

Hmm, that's odd. Do you use the `build_html_brian2.py` script from the `dev/tools/docs` folder?

> If you prefer a squashed commit, I can reopen a new PR.

You can squash your commits with `git rebase -i` and then force push into your repository, this will update this PR, no need to open a new one. And yes, indeed, I'd prefer a smaller number of commits :smile: (it is not essential, though).

docs_sphinx/user/input.rst
flinz: ok, I agree, this was a little ad hoc. :)

flinz:
Addressed your line comment and squashed the commits.

> Hmm, that's odd. Do you use the build_html_brian2.py script from the dev/tools/docs folder?
I did, but somehow messed up my pythonpath. Works now.

thesamovar:
Looks great to me too!

mstimberg:
Thanks again!

flinz:
Sure, was about time to help out a little :)

PR #775, opened 2016-11-14
[MRG] Do not upload development versions to anaconda anymore

mstimberg:
I think uploading every single commit to master as a "development conda package" to anaconda.org wasn't a great idea. Since we don't give them unique version numbers  (everything after 2.0 is 2.1+git), you cannot use the standard system for updating packages, you need to force reinstall etc. Also, you cannot combine the "dev" channel with the standard channel, making everything quite messy.
Given that the users that are interested in an unreleased version of Brian 2 are probably also those that are fine using it directly from github, I don't think we have much of a target audience for these packages (download statistics seem to indicate that basically no one uses them).

So with this PR, non-release packages will no longer be uploaded to anaconda.org (the installation docs are also updated accordingly). With the next release, I'd then also remove all the development packages from anaconda.org to reduce confusion.

### DIFF
--- a/dev/continuous-integration/conda-server-push.py
+++ b/dev/continuous-integration/conda-server-push.py
@@ -12,7 +12,7 @@ options = ['-t', token, 'upload',
 filenames = glob.glob('*.tar.bz2')
 release = '+git' not in filenames[0]
 if not release:
-    options.extend(['--channel', 'dev', '--force'])
+    sys.exit(0)  # Let's not upload development versions to conda
 
 # Uploading sometimes fails due to server or network errors -- we try it five
 # times before giving up
--- a/docs_sphinx/introduction/install.rst
+++ b/docs_sphinx/introduction/install.rst
@@ -181,35 +181,38 @@ case, the Python version does not matter.
 Try running the test suite (see :ref:`testing_brian` below) after the
 installation to make sure everything is working as expected.
 
+.. _development_install:
+
 Development version
 -------------------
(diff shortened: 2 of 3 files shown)


thesamovar:
I'd say it's quite handy to have even if it's not used very often? How problematic is keeping it the way it is? Is it slowing down our other builds for example?

mstimberg:
I don't have a super strong opinion on this, the idea behind removing it was to not add unnecessary complexity to our already quite complex packaging/distribution story. The problem with this packages is that switching back to the standard release, or updating the development version becomes a somewhat complicated task, you have to use `install --force`, make sure to remove the standard `brian-team` channel, etc.

But it does not really affect the build times (we still build the packages to catch errors), except if there's some network problem with anaconda.org, and as you can see from this PR, it does not complicate scripts etc. that much.

thesamovar:
OK I don't really have strong feelings either, and you're the one that does all the work on packaging so I'll leave it to you. Will do an approve review now.

mstimberg:
Hmm, ok, let's keep it as it is for now -- we want to make the bugfix release as soon as possible, so we should probably avoid a not very important change like this that has the potential to break things in our packaging toolchain.

mstimberg:
This PR needs quite a bit of an update and we are not 100% convinced it is a good idea anyway. I'm closing it for now to clear up our list of pull requests. I leave the branch, so we can always come back to this.

PR #779, opened 2016-11-17
Documentation updates for 2.0.1

mstimberg:
A few last documentation updates. After this I'd like to do the 2.0.1 release -- or do you have anything to add?


mstimberg:
A wait, I just added one more thing: travis and appveyor now use `travis_retry` respectively `appveyor-retry` for commands that rely on working network/servers (`conda install`, `pip install`) -- this should hopefully reduce test failures due to anaconda server unavailability.
Since there is no way to really test this locally, let's see how this works out...

PR #786, opened 2016-11-25
[MRG] Parallel compile

mstimberg:
I was a bit annoyed with the standalone compilation taking so much time and it turned out that just using `make` with parallel jobs reduced the time drastically (at least on Linux). I added two new preferences for arguments to the `make`/`nmake` commands, and added `-j` for Linux by default. I also enabled parallel compilation for Windows, but this does not seem to have much of an effect (at least on our test server).

I think we could also drastically improve Cython runtime compilation, but this requires refactoring the code a bit (the `compile` step would trigger compilation in a separate process, a result would only be required when we reach `run`. This is a longer term project (and then, we might reconsider our codegen targets anyway), so let's not worry about that for now.

### DIFF
--- a/brian2/codegen/cpp_prefs.py
+++ b/brian2/codegen/cpp_prefs.py
@@ -57,7 +57,7 @@ prefs.register_preferences(
         '''
         ),
     extra_compile_args_msvc=BrianPreference(
-        default=['/Ox', '/w', msvc_arch_flag],
+        default=['/Ox', '/w', msvc_arch_flag, '/MP'],
         docs='''
         Extra compile arguments to pass to MSVC compiler (the default
         ``/arch:`` flag is determined based on the processor architecture)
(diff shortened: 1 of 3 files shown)


mstimberg:
Hmm, the tests on travis now sometimes seem to fail. The error message suggests problem a with the multiprocessing plugin for the nose test suite, even though I am not quite sure what the problem is. I don't see why it should care about make compiling in parallel or not. I'll relaunch the tests but without parallelization on the test level, let's see whether this changes anything.

thesamovar:
On my Windows machine the previous version runs all tests fine.

Also, code looks good so I'll approve.

mstimberg:
Ok, I still don't quite understand while it failed previously, but I think the speed difference on Linux is significant enough to warrant this change, so I'll merge it now. Either way, the parallel flag is just a preference that can be easily changed by the user if they have any problem with it. If it turns out to be problematic in practice, we can reconsider the default setting later.

PR #799, opened 2017-01-13
[MRG] Unit/dimensions refactoring

### LINKED ISSUE #797: Clear up use of dimensions vs. units
I was looking into the time that is spent during run preparation (which becomes relevant when using many short runs) and noticed that a significant time is spent in the `get_unit` function, e.g. to get the unit of external constants. Now, that function can certainly be optimized (e.g. with caching), but I realized that in most cases it is doing something unnecessary in the first place: we are interested in the *dimensions* (to check consistency) and do not need a `Unit` object.
It actually seems to me that most of the places where we store units should actually store dimensions instead, most importantly state variables. The only additional information a `Unit` stores in addition to the dimensions is a scale and a string name. However, state variables are always stored in the base scale and we cannot do much with the string name either, since our `Quantity` objects only store dimensions, not a unit.
So for example whether you define a variable as having units of `Hz` or `second**-1` *will* have an effect (it is stored differently), but as soon as you access the its values, you will get a `Quantity` that no longer has this information. When you print it, we will again search for a "nice" unit, but this does not depend on what unit was used for the definition in the first place.

So I wonder: should we get rid of most instances where units are currently stored (and going with that, most calls of `get_unit`) and instead store dimensions? For some use cases, this might have a noticeable effect on performance and I don't see us losing much -- whenever we display stuff to the user, i.e. when we transform things to strings, we use something like `in_best_unit` anyway.

Anything that I missed here?

PS: For the transition, we should probably keep a `unit` attribute around in places where we had it before to avoid breaking user code relying on it.

mstimberg:
This implements what I proposed in #797: `SingleEquation`, `Variable`, and `VariableView` now internally store dimensions instead of units. If I run a HH model repeatedly in a loop (with `run(0*ms)`) to coarsely matter preparation overhead, then this actually cuts the time in half! This mostly shows how inefficient `get_unit` was (which again and again went through the full registry to search for a suitable unit, checking and discarding scaled units along the way), but still. I also profiled a bit where the rest of the time is spend, and I think we can quite easily improve more with some simple caching. I might tackle this soon.
Note that you can still access a `unit` property of the above mentioned classes (since it was kind of part of our "public API" and I have already seen user code making use of this, e.g. to create equation strings programmatically), but this will raise a deprecation warning.

Fixes #797

### DIFF
--- a/brian2/core/variables.py
+++ b/brian2/core/variables.py
@@ -102,8 +102,8 @@ class Variable(object):
         name in the owning group. The same variable may be known under other
         names in other groups (e.g. the variable ``v`` of a `NeuronGroup` is
         known as ``v_post`` in a `Synapse` connecting to the group).
-    dimensions : `Dimensions`, optional
-        The dimensions of the variable.
+    dimensions : `Dimension`, optional
+        The physical dimensions of the variable.
     owner : `Nameable`, optional
         The object that "owns" this variable, e.g. the `NeuronGroup` or
         `Synapses` object that declares the variable in its model equations.
@@ -292,8 +292,8 @@ class Constant(Variable):
     name : str
         The name of the variable
     dimensions : `Dimension`, optional
-        The dimensions of the variable. Note that the variable itself (as
-        referenced by value) should never have units attached.
+        The physical dimensions of the variable. Note that the variable itself
+        (as referenced by value) should never have units attached.
     value: reference to the variable value
         The value of the constant.
     owner : `Nameable`, optional
@@ -352,7 +352,7 @@ class AuxiliaryVariable(Variable):
     name : str
         The name of the variable
     dimensions : `Dimension`, optional
-        The dimensions of the variable.
+        The physical dimensions of the variable.
     dtype : `dtype`, optional
         The dtype used for storing the variable. If none is given, defaults
         to `core.default_float_dtype`.
@@ -383,7 +383,7 @@ class ArrayVariable(Variable):
         names in other groups (e.g. the variable ``v`` of a `NeuronGroup` is
         known as ``v_post`` in a `Synapse` connecting to the group).
     dimensions : `Dimension`, optional
-        The dimensions of the variable
+        The physical dimensions of the variable
     owner : `Nameable`
         The object that "owns" this variable, e.g. the `NeuronGroup` or
         `Synapses` object that declares the variable in its model equations.
@@ -480,7 +480,7 @@ class DynamicArrayVariable(ArrayVariable):
         names in other groups (e.g. the variable ``v`` of a `NeuronGroup` is
         known as ``v_post`` in a `Synapse` connecting to the group).
     dimensions : `Dimension`, optional
-        The dimensinos of the variable
+        The physical dimensions of the variable.
     owner : `Nameable`
         The object that "owns" this variable, e.g. the `NeuronGroup` or
         `Synapses` object that declares the variable in its model equations.
@@ -579,7 +579,7 @@ class Subexpression(Variable):
     name : str
         The name of the subexpression.
     dimensions : `Dimension`, optional
-        The dimensions of the subexpression.
+        The physical dimensions of the subexpression.
     owner : `Group`
         The group to which the expression refers.
     expr : str
@@ -731,8 +731,9 @@ class VariableView(object):
         The group through which the variable is accessed (not necessarily the
         same as `variable.owner`).
     dimensions : `Dimension`, optional
-        The dimensinos to be used for the variable, should be `None` when a
-        variable is accessed without units (e.g. when accessing ``G.var_``).
+        The physical dimensions to be used for the variable, should be `None`
+        when a variable is accessed without units (e.g. when accessing
+        ``G.var_``).
     '''
     def __init__(self, name, variable, group, dimensions=None):
         self.name = name
@@ -1424,7 +1425,7 @@ class Variables(collections.Mapping):
         name : str
             The name of the variable.
         dimensions : `Dimension`, optional
-            The dimensions of the variable.
+            The physical dimensions of the variable.
         size : int
             The size of the array.
         values : `ndarray`, optional
@@ -1487,7 +1488,7 @@ class Variables(collections.Mapping):
         names : list of str
             The names of the variable.
         dimensions : `Dimension`, optional
-            The dimensions of the variable.
+            The physical dimensions of the variable.
         size : int
             The sizes of the arrays.
         dtype : `dtype`, optional
@@ -1527,7 +1528,7 @@ class Variables(collections.Mapping):
         name : str
             The name of the variable.
         dimensions : `Dimension`, optional
-            The dimensions of the variable.
+            The physical dimensions of the variable.
         size : int or tuple of int
             The (initital) size of the array.
         values : `ndarray`, optional
@@ -1610,7 +1611,6 @@ class Variables(collections.Mapping):
                        index=index)
         self.device.init_with_arange(self._variables[name], start, dtype=dtype)
 
-
     def add_constant(self, name, value, dimensions=DIMENSIONLESS):
         '''
         Add a scalar constant (e.g. the number of neurons `N`).
@@ -1622,10 +1622,11 @@ class Variables(collections.Mapping):
         value: reference to the variable value
             The value of the constant.
         dimensions : `Dimension`, optional
-            The dimensions of the variable. Note that the variable itself (as
-            referenced by value) should never have units attached.
+            The physical dimensions of the variable. Note that the variable
+            itself (as referenced by value) should never have units attached.
         '''
-        var = Constant(name=name, dimensions=dimensions, owner=self.owner, value=value)
+        var = Constant(name=name, dimensions=dimensions, owner=self.owner,
+                       value=value)
         self._add_variable(name, var)
 
     def add_subexpression(self, name, expr, dimensions=DIMENSIONLESS,
@@ -1638,7 +1639,7 @@ class Variables(collections.Mapping):
         name : str
             The name of the subexpression.
         dimensions : `Dimension`
-            The dimensions of the subexpression.
+            The physical dimensions of the subexpression.
         expr : str
             The subexpression itself.
         dtype : `dtype`, optional
@@ -1667,7 +1668,7 @@ class Variables(collections.Mapping):
         name : str
             The name of the variable
         dimensions : `Dimension`
-            The dimensions of the variable.
+            The physical dimensions of the variable.
         dtype : `dtype`, optional
             The dtype used for storing the variable. If none is given, defaults
             to `core.default_float_dtype`.
--- a/brian2/core/variables.py
+++ b/brian2/core/variables.py
@@ -179,11 +179,8 @@ class Variable(object):
     @property
     def unit(self):
         '''
-        Returns the `Unit` of this variable (deprecated, get the dimensions with
-        `Variable.dim` instead.)
+        The `Unit` of this variable
         '''
-        logger.warn("The 'unit' property is deprecated, use 'dim' instead.",
-                    'deprecated_variable_unit', once=True)
         return get_unit(self.dim)
 
     def get_value(self):
@@ -765,11 +762,8 @@ class VariableView(object):
     @property
     def unit(self):
         '''
-        Returns the `Unit` of this equation (deprecated, get the dimensions with
-        `SingleEquation.dim` instead.)
+        The `Unit` of this variable
         '''
-        logger.warn("The 'unit' property is deprecated, use 'dim' instead.",
-                    'deprecated_variableview_unit', once=True)
         return get_unit(self.dim)
 
     def get_item(self, item, level=0, namespace=None):
--- a/brian2/core/variables.py
+++ b/brian2/core/variables.py
@@ -133,7 +133,7 @@ class Variable(object):
         assert isinstance(dimensions, Dimension)
 
         #: The variable's dimensions.
-        self.dimensions = dimensions
+        self.dim = dimensions
 
         #: The variable's name.
         self.name = name
@@ -170,13 +170,6 @@ class Variable(object):
         return np.issubdtype(np.bool, self.dtype)
 
     @property
-    def dim(self):
-        '''
-        The dimensions of this variable.
-        '''
-        return self.dimensions
-    
-    @property
     def dtype_str(self):
         '''
         String representation of the numpy dtype
@@ -200,7 +193,7 @@ class Variable(object):
         '''
         Return the value associated with the variable (with units).
         '''
-        return Quantity(self.get_value(), self.dimensions)
+        return Quantity(self.get_value(), self.dim)
 
     def get_addressable_value(self, name, group):
         '''
@@ -268,7 +261,7 @@ class Variable(object):
                        ' dtype={dtype}, scalar={scalar}, constant={constant},'
                        ' read_only={read_only})>')
         return description.format(classname=self.__class__.__name__,
-                                  dimensions=repr(self.dimensions),
+                                  dimensions=repr(self.dim),
                                   dtype=repr(self.dtype),
                                   scalar=repr(self.scalar),
                                   constant=repr(self.constant),
@@ -641,7 +634,7 @@ class Subexpression(Variable):
                        'expr={expr}, owner=<{owner}>)>')
         return description.format(classname=self.__class__.__name__,
                                   name=repr(self.name),
-                                  dimensions=repr(self.dimensions),
+                                  dimensions=repr(self.dim),
                                   dtype=repr(self.dtype),
                                   expr=repr(self.expr),
                                   owner=self.owner.name)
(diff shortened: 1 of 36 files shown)


thesamovar:
Haven't looked at the code yet, but two possibilites:

1. Maybe we ought to improve ``get_unit`` even if we're not using it here if it's that inefficient? We could probably, given a quantity ``x`` construct the best unit explicitly via a lookup table of ``x.dimension`` followed by getting the best scale by inspecting the value of ``int(log10(float(x)))``? I guess this might not work with user constructed units quite as nicely?
2. I'd be happy with leaving a computed ``unit`` attribute and not having the ``DeprecationWarning`` - unless there's a problem with that?

mstimberg:
> Maybe we ought to improve get_unit even if we're not using it here if it's that inefficient? We could probably, given a quantity x construct the best unit explicitly via a lookup table of x.dimension followed by getting the best scale by inspecting the value of int(log10(float(x)))? I guess this might not work with user constructed units quite as nicely?

Actually, `get_unit` is only used to get the unscaled unit for dimensions, e.g. when we want to display `volt` instead of `m^2 kg s^-3 A^-1` in an error message. To get the "best unit" for a quantity, we use `Quantity.get_best_unit` (which I just made "public", it was called `_get_best_unit` before). The problem with `get_unit` was that the algorithm was the following:
```Python
for u in all_registered_units(*regs):
    if np.array(u, copy=False) == 1 and have_same_dimensions(u, x):
        return u
```
So it went through the whole list of units every time, making a lot of checks before finding the appropriate unit... A simple caching solution was not feasible given the flexibility to pass in arbitrary lists of `UnitRegistry` objects (the `*regs` argument). Well, we never actually use that (I can't really think of a use-case: we have a `UnitRegistry` for user-defined units, not sure what else we could need), so I got rid of this flexibility and also made use of the already existing dictionary that maps dimensions to units. Et voilà, `get_unit` is now about 500-4000 times faster :-) So I think we can close this chapter now. I made a few other minor related changes (e.g. removing the `all_registered_units` generator which was not used anywhere anymore) and also changed the order of lookup (now the user-defined units are first instead of last).

> I'd be happy with leaving a computed unit attribute and not having the DeprecationWarning - unless there's a problem with that?

That does make sense, I removed it. Given that `get_unit` is now fast it is not bad to use it anyway -- in our internal code we should simply take care to not use units where dimensions suffice.

That reminds me of something: I did actually not change units into dimensions in *all* places where it makes sense: the `check_units` decorator still uses units even though it strictly checks for correct dimensions. I think users don't generally care about this distinction and talking about "units" is more natural then talking about "dimensions" (also writing e.g. `volt` instead of `volt.dim`...).

thesamovar:
Agreed on continuing to use units for user-facing code rather than dimensions! Will take a look at the code asap, but sounds good to go.

brian2/core/variables.py
```diff
@@ -296,17 +291,17 @@ class Constant(Variable):
     ----------
     name : str
         The name of the variable
-    unit : `Unit`
-        The unit of the variable. Note that the variable itself (as referenced
-        by value) should never have units attached.
+    dimensions : `Dimension`, optional
+        The dimensions of the variable. Note that the variable itself (as
```
thesamovar: Again. I know that it just said dimensions before, but seems like a good opportunity to clean up potential source of confusion.

brian2/core/variables.py
```diff
@@ -356,17 +351,17 @@ class AuxiliaryVariable(Variable):
     ----------
     name : str
         The name of the variable
-    unit : `Unit`
-        The unit of the variable.
+    dimensions : `Dimension`, optional
+        The dimensions of the variable.
```
thesamovar: And here.

brian2/core/variables.py
```diff
@@ -387,8 +382,8 @@ class ArrayVariable(Variable):
         name in the owning group. The same variable may be known under other
         names in other groups (e.g. the variable ``v`` of a `NeuronGroup` is
         known as ``v_post`` in a `Synapse` connecting to the group).
-    unit : `Unit`
-        The unit of the variable
+    dimensions : `Dimension`, optional
+        The dimensions of the variable
```
thesamovar: Apparently we say this a lot.

brian2/core/variables.py
```diff
@@ -483,8 +479,8 @@ class DynamicArrayVariable(ArrayVariable):
         name in the owning group. The same variable may be known under other
         names in other groups (e.g. the variable ``v`` of a `NeuronGroup` is
         known as ``v_post`` in a `Synapse` connecting to the group).
-    unit : `Unit`
-        The unit of the variable
+    dimensions : `Dimension`, optional
+        The dimensinos of the variable
```
thesamovar: And again, but this time there is even a typo to correct! ;)

brian2/core/variables.py
```diff
@@ -519,18 +515,18 @@ class DynamicArrayVariable(ArrayVariable):
...
         if isinstance(size, int):
-            dimensions = 1
+            ndim = 1
         else:
-            dimensions = len(size)
+            ndim = len(size)
 
         #: The number of dimensions
-        self.dimensions = dimensions
+        self.ndim = ndim
```
thesamovar: Yeah I like using ndim instead of dimensions, that's much better!

brian2/core/variables.py
```diff
@@ -583,8 +578,8 @@ class Subexpression(Variable):
     ----------
     name : str
         The name of the subexpression.
-    unit : `Unit`
-        The unit of the subexpression.
+    dimensions : `Dimension`, optional
+        The dimensions of the subexpression.
```
thesamovar: !

brian2/core/variables.py
```diff
@@ -734,11 +730,11 @@ class VariableView(object):
     group : `Group`
         The group through which the variable is accessed (not necessarily the
         same as `variable.owner`).
-    unit : `Unit`, optional
-        The unit to be used for the variable, should be `None` when a variable
-         is accessed without units (e.g. when accessing ``G.var_``).
+    dimensions : `Dimension`, optional
+        The dimensinos to be used for the variable, should be `None` when a
```
thesamovar: And again with a typo.

brian2/core/variables.py
```diff
@@ -761,14 +757,14 @@ def __init__(self, name, variable, group, unit=None):
         # We keep a strong reference to the `Indexing` object so that basic
         # indexing is still possible, even if the group no longer exists
         self.indexing = self.group._indices
-        self.unit = unit
+        self.dim = dimensions
```
thesamovar: I'm actually slightly concerned that we have dim and ndim now that I think about it, but I think it's probably OK.

brian2/core/variables.py
```diff
@@ -1418,18 +1413,18 @@ def _add_variable(self, name, var, index=None):
...
         Add an array (initialized with zeros).
 
         Parameters
         ----------
         name : str
             The name of the variable.
-        unit : `Unit`
-            The unit of the variable
+        dimensions : `Dimension`, optional
+            The dimensions of the variable.
```
thesamovar: !

brian2/core/variables.py
```diff
@@ -1491,8 +1486,8 @@ def add_arrays(self, names, unit, size, values=None, dtype=None,
         ----------
         names : list of str
             The names of the variable.
-        unit : `Unit`
-            The unit of the variables
+        dimensions : `Dimension`, optional
+            The dimensions of the variable.
```
thesamovar: !

brian2/core/variables.py
```diff
@@ -1530,8 +1526,8 @@ def add_dynamic_array(self, name, unit, size, values=None, dtype=None,
         ----------
         name : str
             The name of the variable.
-        unit : `Unit`
-            The unit of the variable
+        dimensions : `Dimension`, optional
+            The dimensions of the variable.
```
thesamovar: !

brian2/core/variables.py
```diff
@@ -1609,40 +1605,40 @@ def add_arange(self, name, size, start=0, dtype=np.int32, constant=True,
...
         ----------
         name : str
             The name of the variable
-        unit : `Unit`
-            The unit of the variable. Note that the variable itself (as referenced
-            by value) should never have units attached.
         value: reference to the variable value
             The value of the constant.
+        dimensions : `Dimension`, optional
+            The dimensions of the variable. Note that the variable itself (as
```
thesamovar: !

brian2/equations/equations.py
```diff
@@ -203,7 +201,7 @@ def check_identifier_constants(identifier):
                          'variable name.' % identifier)
 
 
-def unit_and_type_from_string(unit_string):
+def dimensions_and_type_from_string(unit_string):
     '''
     Returns the unit that results from evaluating a string like
```
thesamovar: I guess this should now read physical dimensions instead of unit.

brian2/equations/equations.py
```diff
@@ -217,8 +215,8 @@ def unit_and_type_from_string(unit_string):
 
     Returns
     -------
-    u, type : (Unit, {FLOAT, INTEGER or BOOL})
-        The resulting unit and the type of the variable.
+    d, type : (`Dimension`, {FLOAT, INTEGER or BOOL})
+        The resulting dimensions and the type of the variable.
```
thesamovar: !

brian2/equations/equations.py
```diff
@@ -343,8 +341,8 @@ class SingleEquation(object):
         The type of the equation.
     varname : str
         The variable that is defined by this equation.
-    unit : Unit
-        The unit of the variable
+    dimensions : `Dimension`
+        The dimensions of the variable
```
thesamovar: !

brian2/equations/equations.py
```diff
@@ -767,9 +767,9 @@ def _get_stochastic_type(self):
                                              if eq.type == PARAMETER]),
                                doc='All parameter names.')
 
-    units = property(lambda self:dict([(var, eq.unit) for var, eq in
+    dimensions = property(lambda self: dict([(var, eq.dim) for var, eq in
                                        self._equations.iteritems()]),
-                     doc='Dictionary of all internal variables and their corresponding units.')
+                     doc='Dictionary of all internal variables and their corresponding dimensions.')
```
thesamovar: !

brian2/equations/unitcheck.py
```diff
@@ -23,8 +24,8 @@ def check_unit(expression, unit, variables):
     ----------
     expression : str
         The expression to evaluate.
-    unit : `Unit`
-        The expected unit for the `expression`.
+    dimensions : `Dimension`
+        The expected dimensions for the `expression`.
```
thesamovar: !

brian2/equations/unitcheck.py
```diff
@@ -36,10 +37,12 @@ def check_unit(expression, unit, variables):
     DimensionMismatchError
         If an unit mismatch occurs during the evaluation.
     '''
-    expr_unit = parse_expression_unit(expression, variables)
-    fail_for_dimension_mismatch(expr_unit, unit, ('Expression %s does not '
-                                                  'have the expected unit %r') %
-                                                  (expression.strip(), unit))
+    expr_dims = parse_expression_dimensions(expression, variables)
+    # FIXME: The error message is not great
```
thesamovar: Will fix before merging?

brian2/groups/neurongroup.py
```diff
@@ -111,10 +111,10 @@ def _get_refractory_code(self, run_namespace):
             variables = self.group.resolve_all(identifiers,
                                                run_namespace,
                                                user_identifiers=identifiers)
-            unit = parse_expression_unit(str(ref), variables)
-            if have_same_dimensions(unit, second):
+            dims = parse_expression_dimensions(str(ref), variables)
+            if dims is second.dim:
```
thesamovar: Same as before. For example, might we not have something that evaluates as e.g. ``sqrt(second*second)`` which would fail ``is second.dim``?

brian2/groups/neurongroup.py
```diff
@@ -111,10 +111,10 @@ def _get_refractory_code(self, run_namespace):
...
                 abstract_code = 'not_refractory = (t - lastspike) > %s\n' % ref
-            elif have_same_dimensions(unit, Unit(1)):
+            elif dims is DIMENSIONLESS:
```
thesamovar: Same issue?

brian2/groups/neurongroup.py
```diff
@@ -128,7 +128,7 @@ def _get_refractory_code(self, run_namespace):
             else:
                 raise TypeError(('Refractory expression has to evaluate to a '
                                  'timespan or a boolean value, expression'
-                                 '"%s" has units %s instead') % (ref, unit))
+                                 '"%s" has units %s instead') % (ref, dims))
```
thesamovar: Same issue as the fixme above?

brian2/groups/neurongroup.py
```diff
@@ -603,7 +602,7 @@ def __setattr__(self, key, value):
                                            'size.') % linked_var.name)
 
             eq = self.equations[key]
-            if eq.unit != linked_var.unit:
+            if eq.dim != linked_var.dim:
```
thesamovar: Here we're using ``!=`` instead of ``is not`` so I guess this can work then?

brian2/parsing/expressions.py
```diff
@@ -181,7 +182,7 @@ def _get_value_from_expression(expr, variables):
         raise SyntaxError('Unsupported operation ' + str(expr.__class__))
 
     
-def parse_expression_unit(expr, variables):
+def parse_expression_dimensions(expr, variables):
     '''
     Returns the unit value of an expression, and checks its validity
```
thesamovar: Again, change docstring to say physical dimensions?

brian2/tests/test_equations.py
```diff
@@ -124,7 +124,7 @@ def test_parse_equations():
     # A simple equation
     eqs = parse_string_equations('dv/dt = -v / tau : 1')
     assert len(eqs.keys()) == 1 and 'v' in eqs and eqs['v'].type == DIFFERENTIAL_EQUATION
-    assert get_dimensions(eqs['v'].unit) == DIMENSIONLESS
+    assert eqs['v'].dim is DIMENSIONLESS
```
thesamovar: is check again

brian2/units/fundamentalunits.py
```diff
@@ -191,7 +190,7 @@ def wrap_function_dimensionless(func):
     '''
     Returns a new function that wraps the given function `func` so that it
     raises a DimensionMismatchError if the function is called on a quantity
-    with dimensions (excluding dimensionless quantitities). Quantities are
+    with dimensions (excluding dimensionless quantities). Quantities are
```
thesamovar: Maybe physical dimensions here too?

brian2/units/fundamentalunits.py
```diff
@@ -460,7 +458,11 @@ def __ipow__(self, value):
 
     #### COMPARISON ####
     def __eq__(self, value):
-        return np.allclose(self._dims, value._dims)
+        try:
```
thesamovar: If we change the is checks to use an ``==`` check instead, we could first check if ``self._dims is value._dims`` and potentially get most of the speed still.

brian2/units/fundamentalunits.py
```diff
@@ -2237,69 +2240,50 @@ def unregister_unit(u):
...
-
-def get_unit_for_display(x):
+    s : str
+        A string representation of the respective unit or the string ``'1'``.
     '''
-    Return a string representation of the most appropriate unit or ``'1'`` for
-    a dimensionless quantity
-    '''
-    if x is 1 or x is DIMENSIONLESS or have_same_dimensions(x, 1):
+    if d is 1 or d is DIMENSIONLESS:
```
thesamovar: Same issue?

mstimberg:
I adapted the docstrings and changed "dimensions" to "physical dimensions" everywhere (I hope!). I agree that should make things clearer (well, "physical dimensions" could also mean something like width x height I guess, but that should be clear enough from the context...). I am not that worried about the `ndim` attribute, we only have in one place (`DynamicArrayVariable`) and either way it should not be directly used by user code.

I am not sure about the `Dimension.time` thing. It certainly would look cleaner than `second.dim`, but then we'd also need `Dimension.conductance`, etc. (given that we don't want anyone to write``Dimension.length**-2 * Dimension.mass**-1 * Dimension.time**3 * Dimension.electrical_current**2``...). For now, I tried to make the functions that expect dimensions to be "forgiving", i.e. you could actually write `dimensions=second` instead of `dimensions=second.dim`. I guess it is better to not add new layer of names; as we said earlier, users don't care that much about the distinction between units and dimensions.

About the check with `is` vs. `==` -- actually this is designed to be ok. No code should ever create new `Dimension` objects explicitly, but always pass through `get_or_create_dimensions` which will return a `Dimension` object from a cache if it already exists. So stuff like `second.dim * second.dim is second.dim**2` *will* evaluate to ``True``.

I improved one error message (the previous FIXME), so if the tests pass I am going to merge this.

thesamovar:
Great!

Issue #782, opened 2016-11-18
clip function returns double value for integer arguments


mstimberg:
This only applies to C++/Cython -- probably we should use fused types/templates here to return the type of the first argument?

thesamovar:
Yeah, should be similar to things like ``mod``, ``abs`` etc.

dsprenkels:
Would you like PRs for random easy bugs like this one? If so, then I would like to work on this.

mstimberg:
> Would you like PRs for random easy bugs like this one? If so, then I would like to work on this.

A PR would be most certainly welcome! In general, if it is a bug with a relatively clear fix, please don't hesitate to contribute. If it is something that needs discussion (most importantly everything that changes user-facing syntax), it is better to lay out the proposed solution first before spending much time on an implementation that might not be what we had in mind.

dsprenkels:
I opened a pull request in #810.

PR #810, opened 2017-02-06
Implement generic typing for `_clip`

dsprenkels:
Issue: #782

This pull request makes a more abstract version of the `_clip` function
to accommodate also for arguments of other types than `double`.

It would probably be a good idea to add test cases for each of the different
types, but I have not really looked into the internals of your test suite yet.

### DIFF
--- a/brian2/codegen/generators/cpp_generator.py
+++ b/brian2/codegen/generators/cpp_generator.py
@@ -438,7 +438,8 @@ DEFAULT_FUNCTIONS['abs'].implementations.add_implementation(CPPCodeGenerator,
                                                             name='_brian_abs')
 
 clip_code = '''
-        inline double _clip(const double value, const double a_min, const double a_max)
+        template <typename T>
+        inline T _clip(const T value, const double a_min, const double a_max)
         {
 	        if (value < a_min)
 	            return a_min;
--- a/brian2/codegen/generators/cython_generator.py
+++ b/brian2/codegen/generators/cython_generator.py
@@ -368,11 +368,18 @@ DEFAULT_FUNCTIONS['sign'].implementations.add_implementation(CythonCodeGenerator
                                                              name='_sign')
 
 clip_code = '''
-cdef double clip(double x, double low, double high):
-    if x<low:
-        return low
-    if x>high:
-        return high
+ctypedef fused _to_clip:
+    char
+    short
+    int
+    float
+    double
+
+cdef _to_clip clip(_to_clip x, double low, double high):
+    if x < low:
+        return <_to_clip?>low
+    if x > high:
+        return <_to_clip?>high
     return x
 '''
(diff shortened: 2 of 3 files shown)


mstimberg:
Hi, thanks the changes look very good to me! A test would be good -- there's currently no way to define variables as anything else than double or int in model descriptions, so I think testing those two would be good enough for now.
To add a test, you can add a function called `test_clip` to `brian2.tests.test_functions`. Try to follow the example of something like `test_bool_to_int`, i.e. it should ideally:
* have a single run call + only check results after the run so that it runs with standalone mode (note that the run could potentially be `run(0*ms)` ).
* be annotated with the `standalone-compatible` attribute and use the `reinit_devices` teardown function.

More info on all this [here](http://brian2.readthedocs.io/en/latest/developer/guidelines/testing.html#unit-tests), but no need to delve deeply into it, adapting `test_bool_to_int` should be all you need.

To run the test suite on your machine, use one of the scripts in `dev/tools/tests`.

Many thanks in advance!

dsprenkels:
During testing I found out that the Cython implementation was incorrect, because all the arguments and the return value used the same fused type. The argument types should of course be independent.

Now the `high`, `low` values need to be cast explicitly when they are returned to the caller.

mstimberg:
> During testing I found out that the Cython implementation was incorrect, because all the arguments and the return value used the same fused type. The argument types should of course be independent.

Good catch! Do we really need independent fused types/templates for the `low`/`high` arguments, though? We care about preserving the type of the value that has been passed in, but `low` and `high` could simply be typed as double and C would automatically cast whatever type is passed in? So I think it would be enough to have the first argument as a fused type/template and make sure that the return value has the same type (by explicitly casting it to the type, if necessary).

dsprenkels:
You're right. The other two fused types don't seem to be necessary. I changed the code accordingly.

mstimberg:
Thanks! Sorry for nitpicking once more: I think for consistency it'd be the best if the C++ code also only used a template for the first argument.

dsprenkels:
Updated the code.

One detail. Consider the following equation:

```
clipexpr1 = clip(integer_var1, 0, 1) : integer
```

This code uses integers for all types. But when the code is run, the runtime will perform an unnecessary cast four times (twice for `0`, twice for `1`; from and to `double`). This will probably happen very rarely, and the performance is probably negligible anyway, but it feels a bit weird to implement it this way.

mstimberg:
Thanks again, this looks good to me, I'll just wait for the tests to run through before merging. I get your point about the unnecessary casts, but I think in this case I prefer simplicity over the (as you say, probably negligible) performance benefit. The most common use case is a double variable for the first value (then the other variables have to be cast to double for the comparison, anyway) and this should be fast, but integer variables are relatively rare. In your example, the profiler probably optimizes the problem away, anyway.

I'd be open to change this if profiling shows that it makes a measurable diference, though :-)

PR #813, opened 2017-02-15
[MRG] Workaround glibc bug

### LINKED ISSUE #803: Performance issue in C++ standalone (probably platform/compiler-specific)
While running a standard HH model (see code in the end), I noted that the example slowed down dramatically when I switched to standalone mode. I could reproduce this issue on two Linux machines, but not on a Linux cluster where I had SSH access. The issue basically goes away when re-introducing `-ffinite-math-only` (which we removed as a standard option because it can hide NaN values).
I tried to dig into it a bit and find out what the differences to the code running on the cluster is, and it pointed to AVX instructions -- the code on the cluster seemed to use a SSE exp function, while the code on my machine used an AVX variant. Using `-mno-avx`, or removing `-march=native` does indeed make things a bit faster, but still not as fast as runtime. This is all very weird, given that runtime mode seems to be using AVX instructions without any speed penalty... Looking at the generated code, I really don't see any significant difference between the runtime and the standalone code, except that standalone code replaces constants by their values (which should make things *faster*, if anything...).
This is all really confusing. I could understand if AVX instructions made things slower instead of faster (apparently switching between AVX and SEE can slow things down, and proper AVX support for mathematical functions seem to depend on special libraries, i.e. we'd probably have to use the intel compiler and link against the MKL or something like that), but why on earth does this not affect runtime weave/cython ...‽‽‽

I guess this is Linux/gcc/whatever-specific, or can you confirm this on Windows?

Here quick benchmark values for various compiler options (1s of the below model, basically all spent in the state updater)

|   | runtime (weave) | runtime (cython) | standalone |
| --- | --- | --- | --- |
| default | 9s | 10s | 48s **(!)** |
| `-noavx` | 9s | 10s |  30s **(!)** |
| `-ffinite-math-only` | 8s | 10s | 7s |
 

```Python
defaultclock.dt = 0.01*ms

El = 10.613*mV
ENa = 115*mV
EK = -12*mV
gl = 0.3*msiemens/cm**2
gNa0 = 120*msiemens/cm**2
gK = 36*msiemens/cm**2
C = 1*uF/cm**2

eqs = '''
dv/dt = (gl * (El-v) + gNa * m**3 * h * (ENa-v) + gK * n**4 * (EK-v) + I) / C : volt
I : amp/meter**2
gNa : siemens/meter**2
dm/dt = alpham * (1-m) - betam * m : 1
dn/dt = alphan * (1-n) - betan * n : 1
dh/dt = alphah * (1-h) - betah * h : 1
alpham = (0.1/mV) * (-v+25*mV) / (exp((-v+25*mV) / (10*mV)) - 1)/ms : Hz
# alpham = alpha_fun(v) : Hz
betam = 4 * exp(-v/(18*mV))/ms : Hz
alphah = 0.07 * exp(-v/(20*mV))/ms : Hz
betah = 1/(exp((-v+30*mV) / (10*mV)) + 1)/ms : Hz
alphan = (0.01/mV) * (-v+10*mV) / (exp((-v+10*mV) / (10*mV)) - 1)/ms : Hz
betan = 0.125*exp(-v/(80*mV))/ms : Hz
'''

axon = NeuronGroup(500, eqs, method='exponential_euler', threshold='v>50*mV', refractory='v>50*mV')
```

mstimberg:
This fixes #803 by introducing a new preference for environment variables that will be set during the execution of standalone simulations which defaults to `LD_BIND_NOW=1`. I did not bother introducing different preferences for Windows and *nix, I don't think having a non-relevant environment variable (such as `LD_BIND_NOW` on Windows) should hurt.

### DIFF
--- a/brian2/devices/cpp_standalone/device.py
+++ b/brian2/devices/cpp_standalone/device.py
@@ -82,6 +82,13 @@ prefs.register_preferences(
         Additional flags to pass to the nmake command on Windows. By default, no
         additional flags are passed.
         '''
+        ),
+    run_environment_variables=BrianPreference(
+        default={'LD_BIND_NOW': '1'},
+        docs='''
+        Dictionary of environment variables and their values that will be set
+        during the execution of the standalone code.
+        '''
         )
     )
 
@@ -847,6 +854,13 @@ class CPPStandaloneDevice(Device):
 
     def run(self, directory, with_output, run_args):
         with in_directory(directory):
+            # Set environment variables
+            for key, value in prefs.devices.cpp_standalone.run_environment_variables.iteritems():
+                if key in os.environ and os.environ[key] != value:
+                    logger.info('Overwriting environment variable '
+                                '"{key}"'.format(key=key),
+                                name_suffix='overwritten_env_var', once=True)
+                os.environ[key] = value
             if not with_output:
                 stdout = open('results/stdout.txt', 'w')
             else:
--- a/docs_sphinx/introduction/known_issues.rst
+++ b/docs_sphinx/introduction/known_issues.rst
@@ -56,11 +56,11 @@ are not affected by this problem.
 Slow standalone simulations
 ---------------------------
 
-While we are not yet sure about the exact reason, we have observed that some
-simulations run very slow on modern hardware when compiled with the gcc
-compiler in C++ standalone mode (see #803). Workarounds are to switch to a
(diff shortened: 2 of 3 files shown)


mstimberg:
> Maybe also credit Oleg's github account as well as name, in case he some day wants to claim some credit for contributions to open source software linked to his github profile?

Good idea, I added this info to the release notes. Merging it now.

PR #814, opened 2017-02-15
[MRG] Reduce the number of supported data types for `brian_mod`

mstimberg:
In current master, we generate a whopping 169 variants of `brian_mod`, even though you cannot generate variables (or number literals) with data types such as `unsigned int` or `short` at the moment. This clutters the generated code quite a bit (#594 would of course deal with this as well) without a clear gain. This PR is still quite generous by allowing for six data types (i.e. a total of 36 variants), but we probably rather want to err on the side of caution here.

### DIFF
--- a/brian2/codegen/generators/cpp_generator.py
+++ b/brian2/codegen/generators/cpp_generator.py
@@ -89,8 +89,7 @@ prefs.register_preferences(
 
 
 mod_support_code = ''
-typestrs = ['unsigned char', 'char', 'unsigned short', 'short', 'unsigned int', 'int', 'unsigned long', 'long',
-            'unsigned long long', 'long long', 'float', 'double', 'long double']
+typestrs = ['int', 'long', 'long long', 'float', 'double', 'long double']
 floattypestrs = ['float', 'double', 'long double']
 for ix, xtype in enumerate(typestrs):
     for iy, ytype in enumerate(typestrs):
--- a/brian2/core/clocks.py
+++ b/brian2/core/clocks.py
@@ -49,8 +49,8 @@ def check_dt(new_dt, old_dt, target_t):
     ...
     ValueError: Cannot set dt from 100. us to 200. us, the time 10.1 ms is not a multiple of 200. us
     '''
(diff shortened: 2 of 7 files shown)


thesamovar:
Yeah I was also just erring on the side of caution originally, but I agree that fewer would be better. Maybe unsigned integer types worth keeping? If we create any index variables that are unsigned integers for example, that could come up?

mstimberg:
> Maybe unsigned integer types worth keeping? If we create any index variables that are unsigned integers for example, that could come up?

I don't think it should come up -- we don't use unsigned integers for any user-visible variables (except `timestep`, which we never documented, I think...) and users cannot declare unsigned integers (without going into the internals and manually adding a variable with dtype `np.uint32` or something like that. Since this only affects the use of `%` in abstract code (which will be translated to `brian_mod`), I don't see how unsigned integers could come up.

Actually, I wonder whether we should not get rid of all unsigned integers in our code in the first place (the main place where we use them is in the spike queue). I previously shared the widespread belief that variables that it is a good idea to declare non-negative variables as unsigned, but this seems to be more harmful than helpful in C. I think the current consensus is to use `unsigned` only for variables that you interpret as bit patterns, in all other cases one should rather use signed variables. See for example the following excerpt from Google's style guide (https://google.github.io/styleguide/cppguide.html#Integer_Types):
> Some people, including some textbook authors, recommend using unsigned types to represent numbers that are never negative. This is intended as a form of self-documentation. However, in C, the advantages of such documentation are outweighed by the real bugs it can introduce. Consider:
> `for (unsigned int i = foo.Length()-1; i >= 0; --i) ...`
> This code will never terminate! Sometimes gcc will notice this bug and warn you, but often it will not. Equally bad bugs can occur when comparing signed and unsigned variables. Basically, C's type-promotion scheme causes unsigned types to behave differently than one might expect.
> So, document that a variable is non-negative using assertions. Don't use an unsigned type.

Some more discussion here: http://www.soundsoftware.ac.uk/c-pitfall-unsigned

thesamovar:
OK great! I'm happy to avoid unsigned.

Issue #761, opened 2016-09-24
PoissonInput doesn't work with refractoriness and numpy target


thesamovar:
```
from brian2 import *
prefs.codegen.target = 'numpy'

tau = 10*ms
eqs = '''
dv/dt=-v/tau : 1 (unless refractory)
'''
G = NeuronGroup(100, eqs, reset='v=0', threshold='v>1', refractory=1*ms)
P = PoissonInput(G, 'v', 10, 100*Hz, weight=0.1)
run(100*ms)
```

gives the error:

```
BrianObjectException: Original error and traceback:
Traceback (most recent call last):
  File "E:\Dan\programming\brian2\brian2\codegen\runtime\numpy_rt\numpy_rt.py", line 122, in run
    exec self.compiled_code in self.namespace
  File "(string)", line 13, in <module>
ValueError: operands could not be broadcast together with shapes (99,) (100,) (99,) 

Error encountered with object named "neurongroup_1".
Object was created here (most recent call only, full details in debug log):
  File "<ipython-input-2-ff5506ac93b4>", line 9, in <module>
    G = NeuronGroup(100, eqs, reset='v=0', threshold='v>1', refractory=1*ms)

An exception occured during the execution of code object poissoninput_1_codeobject.
The error was raised in the following line:
v[not_refractory] += 0.1 * poissoninput_binomial_1(_vectorisation_idx)
 ValueError: operands could not be broadcast together with shapes (99,) (100,) (99,) 
(See above for original error message and traceback.)
```

The line with the error should probably read:

```
v[not_refractory] += 0.1 * poissoninput_binomial_1(_vectorisation_idx)[not_refractory]
```

So I think this is an issue with the numpy codegen path rather than `PoissonInput` itself.

mstimberg:
I had a quick look at it and as you say it is a general numpy codegen problem not something specific to `PoissonInput`. I think you can trigger it whenever you use an automatically vectorized function (e.g. `rand()`) in combination with an `unless refractory` variable. E.g. the following code triggers it as well:
```Python
from brian2 import *
prefs.codegen.target = 'numpy'

G = NeuronGroup(10, 'dv/dt = -v/(10*ms) : 1 (unless refractory)',
                reset='v=0', threshold='v>1', refractory=1*ms)
G.v[:5] = 1.1
S = Synapses(G, G, on_pre='v += rand()')
S.connect(j='i')
run(defaultclock.dt)
```

[...]

PR #816, opened 2017-02-27
[MRG] Fix `_vectorisation_idx` for numpy target

mstimberg:
Previously, the `_vectorisation_idx` variable in numpy templates was allowed to be a scalar variable which would *directly* be interpreted as the number of values that e.g. `rand()` should produce, or a vector whose *length* would determine the number of values. In addition, `rand()` would return a scalar value if the `_vectorisation_idx` had the value 1 or was of length 1, and an array otherwise.
This was a bit of mess, I now made this (hopefully) a bit clearer:
* `rand()` will return a scalar value if it receives a scalar value (remember, in the long run, we want to use `_vectorisation_idx` to get reproducible random numbers, so we might need the freedom to have different values for `_vectorisation_idx`, even though we do not need to determine how many numbers to produce.)
* If `rand()` receives an array (or rather: anything with a length), it will use the length of this array to determine how many values to produce.

In many cases (e.g. `group.Vm = 'Vr + randn()*5*mV'`), `_vectorisation_idx` needs to be "the size of the group". For neurons, we actually already have an array storing the values from 0 to `N` in the variable `i`, but we don't have the equivalent for synapses -- in the general case, we do not want to waste memory like this. This is why previously in this situation we used `_vectorisation_idx = N`. We cannot do this anymore with the approach outlines above, but we don't want to do an `np.arange(N)` for synapses, either. Therefore, this PR introduces a new `LazyArange` class that we can use as a replacement for `arange`. You can query its length and it also allows some slicing/indexing, but it is not meant to be a super-general all-purpose class. It does everything that we need for the numpy templates, all tests and examples run fine with it.
With this change, the fix for #761 became trivial, we only have to replace ``_vectorisation_idx`` by ``vectorisation_idx[not_refractory]`` (i.e., treat it like all other variables) in the "conditional write" code used for refractoriness.

### DIFF
--- a/brian2/codegen/generators/numpy_generator.py
+++ b/brian2/codegen/generators/numpy_generator.py
@@ -305,27 +305,19 @@ for func_name, func in [('sin', np.sin), ('cos', np.cos), ('tan', np.tan),
 def randn_func(vectorisation_idx):
     try:
         N = len(vectorisation_idx)
+        return np.random.randn(N)
     except TypeError:
-        N = int(vectorisation_idx)
-
-    numbers = np.random.randn(N)
-    if N == 1:
-        return numbers[0]
-    else:
-        return numbers
+        # scalar value
+        return np.random.randn()
 
 
 def rand_func(vectorisation_idx):
     try:
         N = len(vectorisation_idx)
+        return np.random.rand(N)
     except TypeError:
-        N = int(vectorisation_idx)
-
-    numbers = np.random.rand(N)
-    if N == 1:
-        return numbers[0]
-    else:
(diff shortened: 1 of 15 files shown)


mstimberg:
> These sorts of tricks make me a little nervous, but it's probably better than what was there before, and if it's passing all the tests then I think it's good to go.

Yes, I agree... The advantage of the new approach is that errors should be fairly clear (a `LazyArange` somewhere where we expect a numpy array, for example), whereas previously errors could be more subtle (e.g. the confusion between a single element and a list of length one).

Issue #574, opened 2015-10-15
Additional units


romainbrette:
For concentrations, the standard unit is molar (M). It's unfortunately not an ISI unit, as it's mol/liter. However in physiology one generally uses mM, which is `mol/m**3`. So maybe we could add mM as a unit? If it doesn't make sense, then we could just have it as shorthand for `mol/m**3`.
Speaking of which: mol could also be added, as it's the standard abbreviation of mole.
Also it seems that we don't have kelvin (temperature).
Finally, how about liter/litre?

Could we also have a few useful constants?
e.g.:
R (gaz constant)
F (Faraday constant)
There's of course the issue of how to name them (no so sure).

thesamovar:
When we do ``from brian2 import *`` I think we should only have unit names that are safe, i.e. no one letter names. What about something like ``from brian2.constants import R, F`` so that you have to explicitly say which constants you want? For molar I didn't quite follow it, but I don't think we should add non-SI units: so writing ``mM`` as a shorthand for ``mol/m3`` seems like the better solution to me.

romainbrette:
Yes absolutely. Your idea for constants is good.
For molar and others: yes I agree; we could have molar, mM, and uM as shorthands.
The only thing I'm wondering is for specifying units in equations: would mM work? (if not I think it's ok to have mol/m3)
How about kelvin?

mstimberg:
All this was implemented in #821

PR #821, opened 2017-03-07
[MRG] Additional units/constants, remove celsius

mstimberg:
This PR fixes unit-related issues we discussed in #817 and #574, and a few related issues I encountered along the way. Functionality-wise this PR is ready, but I did not update the documentation and tests yet. I first wanted to make sure that we'll go ahead with these changes as they are. My commits were quite atomic, so we can easily revert single commits if any of the changes detailed below is not consensual.

### Temperatures

As discussed in #817, this now removes `celsius` as a unit. If you use it outside of strings you get a (hopefully) quite clear error message:

```pycon
>>> from brian2.units.allunits import celsius
>>> temp = 27*celsius
    [...]
TypeError: The unit "celsius" is no longer supported to avoid ambiguities when mixed with absolute
temperatures defined in Kelvin. Directly use "kelvin" when you are only interested in temperature
differences, and add the "zero_celsius" constant from the brian2.units.constants module if you want to
convert a temperature from °C to °K.
```

I did not do anything to catch it in equations or other code strings, so these kind of things will unfortunately just give a general error message:

```pycon
>>> group = NeuronGroup(1, 'tadj : 1')
>>> group.tadj = '3**((24*celsius - 36*celsius)/(10*celsius))'
[...]
KeyError: 'The identifier "celsius" could not be resolved.'
```

But then, I am not sure it is worth adding special handling for this somewhere in the code generation process -- I guess most simulations that use `celsius` in  the equations will have used it to define a temperature constant before as well.

### Additional units

There are two new units: `molar` and `liter` (and `litre`), and of course `mmolar`, `mliter`, etc. I also added `mM`, `uM`, `nM` to `brian2.units.stdunits`. A few comments/open questions:

- Currently, `molar` (and its scaled variants), as well as `mM`, `uM`,`nM` are imported with `from brian2 import *`, whereas `liter` is not (i.e. it needs an `from brian2.units.allunits import liter`)

- Our unit system only stores the dimensionality of quantities, not the unit that was used to create it. Therefore, it cannot make a difference between something defined in `uliter`, and something in `mmeter**3` for example. This means that depending on the way we order things, either of the two will be displayed. So we can chose between always displaying things in liters, e.g. you'd get:

  ```pycon
  >>> 5*mmeter**3
  5. * uliter
  ```

  or always displaying things in `meter**3`, i.e. you'd get:

  ```pycon
  >>> 5*uliter
  5. * mmeter3
  ```

  The first version would be a bit odd, if we'd not import `liter` by default, since you could not directly copy&paste the output as a valid quantity.

- Since we store state variables in the base scale of the SI unit system, you are not allowed to use either `liter`, nor `molar` as the unit definition of a variable, i.e. it has to be `meter ** 3`, or `mole/meter**3` respectively. As Romain mentioned in #574, we *could* allow `mmolar` or `mM` (or `kliter`...), but I am not sure this is not more confusing than helpful ("why can I use `mM` but not `mV`?")

### Constants

The PR adds a new module `brian2.units.constants` with a few physical constants that I considered to be of potential relevance to biological modeling (e.g. there's no "speed of light" or "gravitational constant") -- I guess there are still too many. Apart from maybe getting rid of some of them (or adding more), there's one big question: currently, the module defines several alternative names for the constants. This means that depending on your preferred style, you can use either of the two following statements to get the Boltzmann, gas, and Faraday constants:

```pycon
>>> from brian2.units.constants import k, R, F
>>> from brian2.units.constants import Boltzmann, gas_constant, Faraday
```

I'm not too sure that's a good idea... Alternatively, we can either only define short names or only define long names. If a user prefers the other convention, they can always do something along the lines of:

```pycon
>>> from brian2.units.constants import Boltzmann as k, gas_constant as R, Faraday as F
```

I think I slightly prefer "long names only", maybe with a common suffix `_constant` or `_const` (i.e. `bolzmann_constant`, `faraday_constant`, `gas_constant`, ...), this would make it nicely explicit what a constant like `F` refers to (`from brian2.units.constants import faraday_constant as F`).

### Error message for wrong units

Quite often, users did not really understand the error message for incorrect unit specifications. IMO, one reason was that is was written from the developer point of view ("... does not evaluate to a unit"), and another reason that it presented the same error for incorrect unit names (e.g. using `mV` instead of `volt`) and completely wrong unit strings (e.g. with a syntax error). We can do much better than this (and maybe avoid a few questions on the mailing list this way :smirk:).

#### Current master

```pycon
>>> Equations('v : volt***2')
  ...
ValueError: "volt***2" does not evaluate to a unit when only using base units (e.g. volt but not mV):
invalid syntax (<string>, line 1)

>>> Equations('v : voltage')
  ...
ValueError: "voltage" does not evaluate to a unit when only using base units (e.g. volt but not mV):
name 'voltage' is not defined

>>> Equations('v : mV')
  ...
ValueError: "mV" does not evaluate to a unit when only using base units (e.g. volt but not mV): name
'mV' is not defined


```

#### This PR

```pycon
>>> from brian2 import *
>>> Equations('v : volt***2')
  ...
EquationError: Error parsing the unit specification for variable "v": Could not interpret "volt***2" as a unit
specification: invalid syntax (<string>, line 1)

>>> Equations('v : voltage')
  ...
EquationError: Error parsing the unit specification for variable "v": Unit specification refers to "voltage",
but this is not a base unit. The following base units are allowed: amp, coulomb, farad, henry, hertz,
joule, katal, kelvin, kilogram, lumen, lux, meter, mole, newton, ohm, pascal, radian, second, siemens,
sievert, tesla, volt, watt, weber (plus some variants of these, e.g. "Hz" instead of "hertz", or "metre"
instead of "meter").

>>> Equations('v : mV')
  ...
EquationError: Error parsing the unit specification for variable "v": Unit specification refers to "mV", but
this is not a base unit. Use "volt" instead.

>>> Equations('v : molar')
  ...
EquationError: Error parsing the unit specification for variable "v": Unit specification refers to "molar",
but this is not a base unit. Use "metre ** -3 * mole" instead.

```


mstimberg:
One more thing I forgot: I also added `mol` as an alternative name for `mole`, and `ampere` for `amp` -- both are not imported by default, though, which makes them less useful...

thesamovar:
Temperature: good!

Additional units:

* Happy with molar and mM, etc.
* Use m^3 not liter as volume is good I think.
* What about only insisting that ``float(u)==1`` as the condition for using ``u`` as the unit specifier in equations?

Constants:

Good idea to have the module. I would say we shouldn't have any constants that are a single letter. We should encourage people, as you suggest, to do ``from ... import ... as k`` if they want that. I think given that you have to explicitly import from ``.constants`` we don't need a uniform ``_constant`` suffix, just go with the name people usually use to refer to them. (e.g. ``h_bar_constant`` would seem weird, right?)

Error messages: great!

mstimberg:
> What about only insisting that float(u)==1 as the condition for using u as the unit specifier in equations?

We could do this, this certainly has the advantage of being straightforward. On the other hand, from the user's point of view it might look odd to see that `mM`, `kliter` are "legal", while `mV` or `umeter` are not. I think I could be convinced either way, having a clear and unambiguous rule (that automatically applies to new units that we might add at some point) certainly has its benefits.

> . I think given that you have to explicitly import from .constants we don't need a uniform _constant suffix, just go with the name people usually use to refer to them. (e.g. h_bar_constant would seem weird, right?)

It's a bit of mess currently, I think the most consistent might be to use `_constant` whenever the name does not already have a noun suffix (e.g. we would not convert `electron_mass` to `electron_mass_constant`, but `Faraday` would be `faraday_constant` -- BTW: I think not having capital letters is better). And `h_bar` would be called `dirac_constant` (if that's the constant you meant -- it's not included BTW ;) ). So here's my proposed list of names (but again, I'm also open to kick out some of these or add new ones):
```
avogadro_constant
boltzmann_constant
electric_constant
electron_mass
elementary_charge
faraday_constant
gas_constant
magnetic_constant
molar_mass
zero_celsius
```
Going by Wikipedia redirections, `electrical_constant` and `magnetic_constant` should instead be `vacuum_permittivity` and `vacuum_permeability` -- if I'd ever have to use them, I would not be able to chose the correct one, though :smirk:

@romainbrette Any opinion on all this?

thesamovar:
I don't have a strong opinion on the constant names, but your selection seems reasonable to me.

Agree that ``mM`` and ``kliter`` being legal while others are not would be weird, didn't think of that. Would it be coherent to say that only un-prefixed forms are allowed? Maybe that still has the problem that the scale isn't what you'd expect it to be?

romainbrette:
I think the initial reason why we didn't want people to use mV is they 
would expect v = 25 would mean 25 mV and not 25 volt. So in this sense 
we should indeed have mM as the base unit for concentration and not M. 
It turns out it is also the standard unit that is used in 
electrophysiology so I think it's ok. For kliter, well I don't think the 
problem will occur, no one uses this (and if they do, it's not a problem).

The problem is more one of presentation, ie explaining that mM is 
allowed but not mV. One option is to allow only mole*meter**-3. But 
that's not very convenient. I would rather be in favor of putting a 
cautionary note in the documentation, explaining that mM is a base unit 
because it equals mole/meter^3.

romainbrette:
All is very nice. It's a good choice and good names for constants, except maybe molar_mass (mole of what?). I think electric_constant and magnetic_constant are better names than vacuum_x.
I'm just slightly hesitant about whether we should also have short names (or R_constant, F_constant, epsilon0_constant).
But ok, I think the long names are still a better option, because the short names would have to have a suffix anyway, otherwise that would be a bit dangerous (eg F is often used for firing rate). We should have good examples for using them (import ... as R).

mstimberg:
About the units in the equation definitions: indeed, the original reason to only allow units with scale 1 here was to avoid ambiguities about their scale. This was quite important in Brian 1 because arrays did not have units, so when you typed `group.v` you got back an array of values in volt, but without the unit information. In Brian 2, this array will have units attached so confusion should be unlikely. Nevertheless, you could still use `group.v_`, or `array(group.v)` or some other operation that discards the units, so it is still safer to keep the convention. I think if we were to design this system from scratch right now, we might rather make the distinction between dimensions and units clearer, i.e. in equations you'd write something like `voltage`, `length`, `current`, `current/area`, etc. instead of units, but well.
Long story short, I got the impression that there is a consensus to allow all "scale 1" units, which would mean to allow `mmolar`/`mM` (and `kliter`, but no one cares) -- we'll add a note in the docs for this specific case.

About the constants, I also got an impression that we got a consensus to use the long names (mentioning in the docs/examples how to use them with short names). I'll replace `molar_mass` by `molar_mass_constant` ("a physical constant which relates relative atomic mass and molar mass" -- [Wikipedia](https://en.wikipedia.org/wiki/Molar_mass_constant)).

I'll clean up the PR and add docs and tests before proposing it for merging.

mstimberg:
I updated the PR to implement all the changes we discussed and also added documentation and tests. However, during testing I realized that our current `UnitRegistry` system is somewhat broken, and I'd like to take the opportunity to fix this once and for all. I'll try to explain in some detail what the problem is:
Brian (1 and 2) have a system to automatically register new units and it will use these units to when displaying values. E.g. if you'd do (a bit contrived of course):

```pycon
>>> 3e-7*meter**-4*kilogram**-1*second**4*amp**2
3.00000000e-07 * metre ** -4 * kilogram ** -1 * second ** 4 * amp ** 2
```

It doesn't have any "nice" unit to display it in. However, if you'd use something like:

```pycon
>>> pF/cm**2
```

It will register this unit (you could call `register_unit` explicitly, but `Unit.automatically_register_units` is `True` by default). From then on, it will be used for display if appropriate:

```pycon
>>> 3e-7*meter**-4*kilogram**-1*second**4*amp**2
30. * pfarad / cmetre ** 2
```

This (and that registration is automatic) is actually quite important because otherwise even if you'd type `30*(pfarad/cmeter**2)` directly, it wouldn't use the unit you used for specifying it for display and would fall back to the SI dimensions instead. This is because quantities only store the dimensions, so the information is no longer available when doing the display. Side note: you actually need the parentheses above if you want to have nice display with the "automatic registration" mechanism. If you omit them, `30*pfarad` will be calculated first, resulting in a `Quantity`, so no `Unit/Unit` division takes place.

Now, there was a bug in this mechanism in Brian 1 and early versions of B: all user units were added to a list, so *each use* of `pfarad/cmeter**2` would store it again. Together with the way "unit lookup" was organized, this also slowed down things. The straightforward solution was to store units in a set instead of in a list, which also meant implementing a proper hash function -- the previous hash function was inherited  from `np.float64` (Brian 1), respectively from `np.ndarray` (Brian 2), which meant that only the numerical value matter (e.g. `hash(msiemens) == hash(mvolt))`. In the current state of this branch, this is all working out, the hash function depends on all the fields of the `Unit` object (`name`, etc.), and everything works.

What's the problem? The docs advertise that you are allowed to change the displayed name of a `Unit` after its creation:

> A new unit defined by multiplication, division or taking powers
> generates a name for the unit automatically, so that for
> example the name for `pfarad/mmetre**2` is `"pF/mm^2"`, etc. If you
> don't like the automatically generated name, use the
> `Unit.set_display_name` method.



With the current implementation, this will break things... If you change the name of a unit, then its hash value changes and that's not allowed for objects stored in a set. I see two solutions:

1. We allow certain changes to a unit after its creation (`dispname` and `latexname` I guess), and the methods used for doing so will take care of unregistering and re-registering the unit.
2. We do not allow any changes to a unit after its creation, if you want to use a unit with a non-default display name, then you'll have to explicitly create the unit with `Unit.create`

I think I prefer the second solution, it is cleaner and it is compatible with the standard use of hash functions and sets (objects are supposed to be immutable). IMO, this is also a very rarely used feature (I don't think I ever used it myself) so it is probably not important to make this very convenient to use.

romainbrette:
I actually didn't know you could change the names of units, so I'd say 
#2 is fine.

thesamovar:
I also doubt anyone ever did this, so solution 2 is fine. On the other hand, should we not just do a lookup based on the dimensions alone? I think I didn't understand the issue well.

mstimberg:
> On the other hand, should we not just do a lookup based on the dimensions alone? I think I didn't understand the issue well.

It is complicated :nerd_face: We do a lookup based on the dimensions, but then for each dimension there's a list of units (e.g. for the dimensions `second` there's `msecond`, `usecond`, etc.). We also have an independent set of all units, but actually I don't think that's needed anymore now that we have the dimensions-based lookup.

Either way, dropping the option to rename units seems to be a consensus, I'll go ahead implementing this solution.

mstimberg:
Ok, I think I fixed everything we discussed and it is a bit more sensible now. I realized there were a few requirements that might be conflicting:
1. We want units to compare as equal when they are equal in the quantity sense (otherwise we'd have to explain to users why `meter == metre` is `False`, but `1*meter == metre` is `True`).
2. We need to keep track of all available units so that we can allow their use in strings
3. We need to keep track of units for presentation purposes, where we need one canonical representation for each dimensionality and scale
4. We do not want to store the same unit multiple times (e.g. when you use `pF/cm**2` repeatedly, it should not be stored again and again)
5. (somewhat optional, see below) We might want the user to overwrite the stored representation for a given unit.

A few solutions that we had over the course of the development of Brian 1/Brian 2, and some stuff I did here in this PR, do not work with these requirements for various reasons: For example, to remember all available units (requirement 2) without duplicating them (requirement 4), we cannot simply use a set, since such a set would not store both `meter` and `metre` (since they compare equal). Using a list (Brian 1's solution) would work for 2 but not for 4, etc.

So, here is my current approach: A `UnitRegistry` stores the available units in two data structures. The first is a dictionary with `repr(unit)` as the key and the `unit` as the value. This is what we use to construct namespaces for string evaluation, etc. The second data structure is a tiny bit more complicated, but still reasonable I hope :smirk: : it is a dictionary of dictionaries. The main key is the dimensionality, and the keys of the second dictionary is the scale as a floating point value. This is what we use to get the canonical representation for a quantity, based on its dimensions and scale. The advantage of this data structure is that if the user generates new units repeatedly, they will be only stored once, but you can still store multiple units for the same dimensionality (which we of course need for scaled units).

Now there's one more thing we might want to fix. The automatic unit registration works, i.e. as soon as you use a unit, it will be remembered for presentation purposes:
```pycon
>>> 3e-7*meter**-4*kilogram**-1*second**4*amp**2
3e-7 * metre ** -4 * kilogram ** -1 * second ** 4 * amp ** 2
>>> pF/cm**2  # This will automatically register the unit
pfarad / cmetre ** 2
>>> 3e-7*meter**-4*kilogram**-1*second**4*amp**2
30. * pfarad / cmetre ** 2
```

Now, that only works if no standard unit for the same dimensionality exists. The documentation talks about creating a new unit "Nm", but this is equivalent to Joule -- you can therefore not force the use of this unit for display purposes, even if you explicitly try to register it:
```pycon
>>> from brian2.units.allunits import newton
>>> Nm = newton*meter
>>> 3*Nm
3. * joule
>>> register_new_unit(Nm)
>>> 3*Nm
3. * joule
```

Now, we could simply turn around the order of the unit registries, i.e. user-defined units have preference over the standard units (I did this actually at some point). It turns out that this is a terrible idea if you also use automatic unit registration -- if (for some reason), you used `metre ** 2 * kilogram * second ** -3 * amp ** -1` at some point, all values in `volt` would now be displayed using the long form... So the only proper solution I see for that is to have *two* user registries, once for the explicitly defined units with `register_new_unit`, which has preference over standard units, and one for the automatically defined units, which does not overwrite standard units.

Having said all this, I'd also be fine if it were simply not possible to overwrite standard units for display purposes. There's always stuff like this you can do:
```pycon
>>> (3*Nm).in_unit(Nm)
'3. N m'
```

thesamovar:
> the scale as a floating point value

Shouldn't it be an int? i.e. it's something like ``int(log10(x))`` I guess? (Presumably slightly more complicated, but something like that.)

I suspect that nobody or virtually nobody has ever used ``register_new_unit`` or any of that stuff, and so we should optimise the solution for the typical use case of people doing things like ``pF/cm``. I'm quite happy to make all that API something we consider to be more or less internal to Brian and not something users should normally use. In particular then, your last suggestion seems like a good idea to me. I think that, together with your new UnitRegistry data structure covers everything?

mstimberg:
> Shouldn't it be an int? i.e. it's something like int(log10(x)) I guess? (Presumably slightly more complicated, but something like that.)

I was referring to the values stored in `_siprefixes`, they were previously:
```pycon
>>> _siprefixes
{'': 1,
 'E': 1e+18,
 'G': 1000000000.0,
 'M': 1000000.0,
 'P': 1000000000000000.0,
 'T': 1000000000000.0,
...}
```
And this fails with the most recent numpy version if you do `_siprefix[''] ** -1`. The way these values are currently used they have to be the actual floating point values, since they determine the `value` of the `Unit`. But you are right that it would probably be better to store integer numbers somewhere, and use them when e.g. multiplying two units, instead of using the floating point values.
I looked into this stuff a bit and I realized that our handling of scales internally is a bit odd, anyway: we have a `scales` attribute of strings which is supposed to hold the scale per dimension, but we are not actually using it:
```pycon
>>> (pF/cm**2).scale
['', '', '', '', '', '', '']
```
(I just checked, it's the same in Brian 1). Similar for the `scalefactor`, we only use it during construction. So how about: not storing the `scalefactor` at all (it's convenient for construction, but not needed afterwards), and storing the scale as an integer numbers (as you said, the exponent of 10) -- I don't think it is useful to store the scale independently for each dimension, or ist it? We could then use the scale for exact calculations of the value, which might avoid some floating point issues, e.g. currently using variants of liter and meter³ does not give the exact same values:
```pycon
>>> float(8*pliter), float(8000*umeter3)
(8e-15, 7.999999999999999e-15)
```

A final minor note on this: this "exponent of 10"-based system will prevent us from introducing units like `feet` or `inches` in the future, but I personally could not care less...

About `register_new_unit` -- I'm happy to take this option away from the user. As you said, I doubt anyone used it and it was not easy to use correctly in the first place.

thesamovar:
I'm not bothered about excluding feet and inches - we can always add them as ``Quantity`` constants rather than units (and I would think it would be better not to have them in any case).

That all sounds good to me.

mstimberg:
Ok, I *think* this should finally be it, this took quite a bit longer than expected (but of course it also does quite a bit more than initially proposed). The only thing that is still missing from my side are the release notes, but it would be good to manually test this branch a bit, our test suite is not covering display issues (i.e., does the unit used for display look "good") very well. I also realized that an earlier version in this branch completely botched up the values of scaled units (e.g. `float(mmeter2)` equaled 1.0), and no test was catching that. I added a new test for that.

brian2/equations/equations.py
```diff
@@ -223,17 +226,27 @@ def dimensions_and_type_from_string(unit_string):
     ValueError
         If the string cannot be evaluated to a unit.
     '''
+    # Lazy import to avoid circular dependency
```
thesamovar: Could we think about refactoring this stuff to the units code? Even though it's kind of specific to equations, it feels odd to have it here.

brian2/equations/equations.py
```diff
@@ -251,13 +264,38 @@ def dimensions_and_type_from_string(unit_string):
     if unit_string == 'integer':
         return DIMENSIONLESS, INTEGER
 
-    # Check first whether the expression evaluates at all, using only base units
+    # Check first whether the expression only refers to base units
```
thesamovar: Actually, this certainly belongs here so maybe the previous bit should stay here too.

dev/tools/static_codegen/units_template.py
```diff
@@ -41,7 +45,43 @@
 {additional_units}
 {all_units}
 
+class _Celsius(object):
+    '''
+    A dummy object to raise errors when ``celsius`` is used. The use of
+    `celsius` can lead to ambiguities when mixed with temperaturs in `kelvin`,
```
thesamovar: typo

dev/tools/static_codegen/units_template.py
```diff
@@ -41,7 +45,43 @@
...
+    so its use is no longer supported. See github issue #817 for details.
+    '''
+    error_text = ('The unit "celsius" is no longer supported to avoid '
+                  'ambiguities when mixed with absolute temperatures defined '
+                  'in Kelvin. Directly use "kelvin" when you are only '
+                  'interested in temperature differences, and add the '
+                  '"zero_celsius" constant from the brian2.units.constants '
+                  'module if you want to convert a temperature from °C to °K.')
```
thesamovar: avoid use of degree symbol?

docs_sphinx/user/equations.rst
```diff
@@ -29,6 +29,10 @@ the values out of a `NeuronGroup` and discarding the units. Compound units are o
 There are also three special "units" that can be used: ``1`` denotes a dimensionless floating point variable,
 ``boolean`` and ``integer`` denote dimensionless variables of the respective kind.
 
+.. note:: For molar concentration, the base unit that has to be used in the equations is ``mmolar`` (or ``mM``), *not*
+          ``molar``. This is because 1 molar is 10³ mol/m³ in SI units (i.e., it has a "scale" of 10³), whereas
```
thesamovar: same for cubed symbol?

mstimberg:
Thanks for the comments, because of the `: integer`, etc. stuff I'd indeed prefer that function to stay in the functions module. About the °, ³, etc. -- I removed it where it was used in an error message (because it might be shown messed up in a badly configured console) but I left it in the docs, it shouldn't be a problem in the HTML version. It's 2017 after all :)
  
You might have noticed that this PR also have a bunch of testing/package version related changes. These are mostly quick fixes to make the tests pass, I'll open an issue to discuss this. I'll merge the branch as soon as the tests finally pass.

PR #826, opened 2017-03-22
[MRG] Support for non-linear membrane currents in SpatialNeuron

mstimberg:
This branch adds support for non-linear membrane equations in `SpatialNeuron` following an approach that @romainbrette suggested a while ago. Previously, we required the membrane current `Im` to be linear in `v`, i.e. that it can be expressed in the form `-g_tot * v + I_0`. With the changes in the current branch, we now use sympy to linearize our equations, i.e. we set `g_tot = dIm/dv` and `I_0 = Im - g_tot*v`. For equations that are already linear in `v` (most importantly the standard HH-type equations) this does not change anything (apart from using sympy's derivation mechanism instead of sympy's pattern matching). However, it allows us to simulate more detailed models of channel dynamics, in particular GHK-type equations.
I added one example based on a paper that uses GHK equations for a low-threshold Calcium current (`examples/frompapers/Destexhe_et_al_1998.py`). The example is also a useful example (I hope) for a slightly more complex simulation setup, it uses store/restore to re-run the same model with different channel densities and injected currents. It is comparatively slow (full simulation takes ~5min with weave on my machine) since it cannot make use of standalone mode and has many short runs, but this also means that it can serve as a useful benchmark for future improvements of runtime mode/multiple runs (caching, etc.).

Apart from documentation and tests, I consider this branch finished, but of course happy to get some feedback.


mstimberg:
I changed a small thing (if the membrane current used functions such as `clip` that sympy cannot differentiate, the result contained unevaluated derivatives and would make compilation fail with an unclear error message), from my side this could be merged. @romainbrette: could you maybe have a look at it?

Issue #781, opened 2016-11-18
Cython compilation error for floating point arguments to range (in synapse generation)


mstimberg:
Code like the following will lead to a compilation error in Cython:
```Python
S.connect(j='k for k in range(i-3., i+4.) if i!=k', skip_if_invalid=True)
```
```
Error compiling Cython file:
------------------------------------------------------------
...
    for _i in range(N_pre):
        _raw_pre_idx = _i + _source_offset

                
        i = _array_neurongroup_i[_raw_pre_idx]
        _iter_high = 4.0 + i
                        ^
------------------------------------------------------------

/home/marcel/.cython/brian_extensions/_cython_magic_fa12725a6193fc9a6e17c2d0896ab48f.pyx:170:25: Cannot assign type 'double' to 'int32_t'

Error compiling Cython file:
------------------------------------------------------------
...

                
        i = _array_neurongroup_i[_raw_pre_idx]
        _iter_high = 4.0 + i
        _iter_step = 1
        _iter_low = (-3.0) + i
                          ^
------------------------------------------------------------

/home/marcel/.cython/brian_extensions/_cython_magic_fa12725a6193fc9a6e17c2d0896ab48f.pyx:172:27: Cannot assign type 'double' to 'int32_t'
```

We could either use our AST parsing system to raise an argument if the arguments are not integers, or simply force a conversion to integer (which is what e.g. the C++ code is implicitly doing now).

thesamovar:
I would tend to say raise an error: it's easy to wrap with an int if you want to.

dsprenkels:
I would like to try and fix this bug. I think it is important that all runtimes behave in the same way, so there are two obvious fixes:

  1. Implicitly cast floating point types to integers in the cython backend.
  2. Throw an error and also change the other backends to throw an error.

Option 2 will change the API and could (in a rare case) break existing production code. So my preference would be to implement 1. (In this case, printing a warning is an additional option.)

mstimberg:
Re-thinking about this, I tend to agree with @thesamovar and rather raise an error. The reason is that:
1. in our "abstract code" we mostly try to follow Python semantics, and Python's `range` function will raise an error for float arguments
2. casting to float might hide an user error (e.g. if the user wrote `range(0.5, 1.5)` then they probably did not understand what `range` is supposed to do here)

> Option 2 will change the API and could maybe break existing production code.

This is certainly an important point to consider. However, in the past we have taken the stance that breaking user code is ok if:
[...]

PR #827, opened 2017-03-25
synapse: range(...) arguments type check

dsprenkels:
Issue #781.

This PR adds a check and tests to ensure that during synapse generation, arguments to `range` are always `int`s. If not, it will raise with a **`TypeError`**.

### DIFF
--- a/brian2/synapses/synapses.py
+++ b/brian2/synapses/synapses.py
@@ -31,10 +31,11 @@ from brian2.units.fundamentalunits import (Quantity, DIMENSIONLESS,
                                            fail_for_dimension_mismatch)
 from brian2.units.allunits import second
 from brian2.utils.logger import get_logger
-from brian2.utils.stringtools import word_substitute
+from brian2.utils.stringtools import get_identifiers, word_substitute
 from brian2.utils.arrays import calc_repeats
 from brian2.core.spikesource import SpikeSource
 from brian2.synapses.parse_synaptic_generator_syntax import parse_synapse_generator
+from brian2.parsing.bast import brian_ast
 from brian2.parsing.rendering import NodeRenderer
 
 MAX_SYNAPSES = 2147483647
@@ -1463,8 +1464,14 @@ class Synapses(Group):
         return deps
 
     def _add_synapses_generator(self, j, n, skip_if_invalid=False, namespace=None, level=0):
-        template_kwds, needed_variables = self._get_multisynaptic_indices()
+        # Get the local namespace
+        if namespace is None:
+            namespace = get_local_namespace(level=level+1)
+
         parsed = parse_synapse_generator(j)
+        self._check_parsed_synapses_generator(parsed, namespace)
+
+        template_kwds, needed_variables = self._get_multisynaptic_indices()
         template_kwds.update(parsed)
         template_kwds['skip_if_invalid'] = skip_if_invalid
 
@@ -1557,9 +1564,7 @@ class Synapses(Group):
                       "using generator '%s'") % (self.source.name,
                                                  self.target.name,
                                                  parsed['original_expression']))
-        # Get the local namespace
-        if namespace is None:
-            namespace = get_local_namespace(level=level+1)
+
         codeobj = create_runner_codeobj(self,
                                         abstract_code,
                                         'synapses_create_generator',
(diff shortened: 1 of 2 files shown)


brian2/synapses/synapses.py
```diff
@@ -1570,3 +1575,17 @@ def _add_synapses_generator(self, j, n, skip_if_invalid=False, namespace=None, l
                                         check_units=False,
                                         run_namespace=namespace)
         codeobj()
+
+    def _check_synapse_parse(self, parsed, namespace):
```
mstimberg: Maybe we can find a clearer name for this function? `_check_generator_arguments` or something along these lines?

brian2/tests/test_synapses.py
```diff
@@ -2190,6 +2191,22 @@ def test_synapses_refractory_rand():
...
     assert_equal(target.v[5:], 0)
 
+
+@attr('codegen-independent')
+def test_range_noint_error():
+    # arguments to `range` should only be integers (issue #781)
+    G = NeuronGroup(42, 'v: 1', threshold='False')
+    S = Synapses(G, G, 'w:1', 'v+=w')
+    msg = 'The "{}" argument of the range function was .+, but it needs to be an integer.'
+    assert_raises_regex(TypeError, msg.format('high'), lambda: S.connect(j='k for k in range(42.0)'))
```
mstimberg: Good test, I wonder if it is not a bit too specific by matching the exact text of the error message. On the other hand, it should be pretty obvious what the issue is, if we ever to decide to change the text. So, ok, let's leave it that way.

brian2/synapses/synapses.py
```diff
@@ -1570,3 +1575,17 @@ def _add_synapses_generator(self, j, n, skip_if_invalid=False, namespace=None, l
...
+        if parsed['iterator_func'] == 'range':
+            # We expect all arguments of the range function to be integers
+            for argname, arg in parsed['iterator_kwds'].items():
+                identifiers = get_identifiers(arg)
+                variables = self.resolve_all(identifiers, run_namespace=namespace,
+                                             user_identifiers=identifiers)
+                annotated = brian_ast(arg, variables)
+                print(annotated, annotated.dtype)
```
mstimberg: Please remove the print statement.

brian2/tests/test_synapses.py
```diff
@@ -2264,4 +2281,6 @@ def test_synapses_refractory_rand():
     test_synapse_generator_random_with_condition()
     test_synapses_refractory()
     test_synapses_refractory_rand()
+    test_synapses_refractory()
```
mstimberg: This should call your new test, instead of calling the refractory tests twice.

dsprenkels: This must have gone wrong during merge conflict resolving. :+1:

brian2/synapses/synapses.py
dsprenkels: I'll change it to `_check_parsed_synapse_generator`. But I will add a suitable doc string to clarify the function's purpose.

dsprenkels: The idea would be to extend this function if any other type-checking is needed in the future. Maybe this is a bit too elaborate?

mstimberg: I wouldn't worry about this now. It's an internal function (made very clear by starting the name with an underscore), so we can always rename it or change the docstring if the function becomes more general.

brian2/synapses/synapses.py
dsprenkels: :+1:

brian2/tests/test_synapses.py
```diff
@@ -2264,4 +2280,5 @@ def test_synapses_refractory_rand():
     test_synapse_generator_random_with_condition()
     test_synapses_refractory()
     test_synapses_refractory_rand()
+    test_synapse_generator_range_noint()
```
mstimberg: Almost there: the function is called ` test_range_noint_error` and not `test_synapse_generator_range_noint()` (but I'd actually prefer the latter name) :smile:

dsprenkels: Oops. Should be fixed now. :)

brian2/tests/test_synapses.py
```diff
@@ -2190,6 +2190,22 @@ def test_synapses_refractory_rand():
     assert all(target.v[:5] > 0)
     assert_equal(target.v[5:], 0)
 
+
+@attr('codegen-independent')
+def test_synapse_generator_range_noint():
+    # arguments to `range` should only be integers (issue #781)
+    G = NeuronGroup(42, 'v: 1', threshold='False')
+    S = Synapses(G, G, 'w:1', 'v+=w')
+    msg = 'The "{}" argument of the range function was .+, but it needs to be an integer\.'
```
dsprenkels: I just changed this to escape the last '.'.

mstimberg: Good catch!

mstimberg:
Ok, great, merging this now. Many thanks again!

PR #828, opened 2017-03-25
Expose capacitive current in spatial neuron

### LINKED ISSUE #677: Expose capacitive current in SpatialNeuron
To calculate extracellular fields (LFP), we need to know the capacitive current `C*dv/dt`. There's an example (`lfp.py`) that does that using `run_regularly` to store `v` at the previous step etc, but it's not elegant. So it would be nicer to have a state variable `Ic` in the group, standing for the capacitance current (in `amp/meter**2`), which is updated in the template, when the equation is integrated.

phreeza:
PR to fix #677 

Some caveats apply, as discussed in the issue. I will add the equivalent code to the other backends.

### DIFF
--- a/brian2/codegen/runtime/cython_rt/templates/spatialstateupdate.pyx
+++ b/brian2/codegen/runtime/cython_rt/templates/spatialstateupdate.pyx
@@ -198,4 +198,5 @@
     for _i in range(0, N):
         {{Ic}}[_i] = {{Cm}}[_i]*({{v}}[_i] - _v_previous[_i])/{{dt}}
 
+    free(_v_previous)
 {% endblock %}
--- a/brian2/codegen/runtime/cython_rt/templates/spatialstateupdate.pyx
+++ b/brian2/codegen/runtime/cython_rt/templates/spatialstateupdate.pyx
@@ -2,6 +2,7 @@
                   _ab_star0, _ab_star1, _ab_star2, _b_plus,
                   _a_plus0, _a_plus1, _a_plus2, _b_minus,
                   _a_minus0, _a_minus1, _a_minus2, _v_star, _u_plus, _u_minus,
+                  _v_previous,
                   _gtot_all, _I0_all,
                   _c1, _c2, _c3,
                   _P_diag, _P_parent, _P_children,
@@ -27,7 +28,6 @@
     cdef int _num_children
     cdef double _subfac
     cdef int _children_rowlength
-    cdef double *_v_previous = <double *>malloc(N * sizeof(double))
 
     # MAIN CODE
     _vectorisation_idx = 1
@@ -42,7 +42,7 @@
         {{vector_code|autoindent}}
         {{_gtot_all}}[_idx] = _gtot
         {{_I0_all}}[_idx] = _I0
-        _v_previous[_idx] = {{v}}[_idx]
+        {{_v_previous}}[_idx] = {{v}}[_idx]
 
     # STEP 2: for each section: solve three tridiagonal systems
 
@@ -196,7 +196,6 @@
                                             + {{_B}}[_i+1] * {{_u_plus}}[_j])
 
     for _i in range(0, N):
-        {{Ic}}[_i] = {{Cm}}[_i]*({{v}}[_i] - _v_previous[_i])/{{dt}}
+        {{Ic}}[_i] = {{Cm}}[_i]*({{v}}[_i] - {{_v_previous}}[_i])/{{dt}}
 
-    free(_v_previous)
 {% endblock %}
--- a/brian2/codegen/runtime/cython_rt/templates/spatialstateupdate.pyx
+++ b/brian2/codegen/runtime/cython_rt/templates/spatialstateupdate.pyx
@@ -1,4 +1,4 @@
-{# USES_VARIABLES { Cm, dt, v, N,
+{# USES_VARIABLES { Cm, dt, v, N, Ic,
                   _ab_star0, _ab_star1, _ab_star2, _b_plus,
                   _a_plus0, _a_plus1, _a_plus2, _b_minus,
                   _a_minus0, _a_minus1, _a_minus2, _v_star, _u_plus, _u_minus,
@@ -12,7 +12,6 @@
 {% extends 'common.pyx' %}
 
 {% block maincode %}
-
     cdef int _i
     cdef int _j
     cdef int _k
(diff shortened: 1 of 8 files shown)


phreeza:
Note sure why this failed a test, looks like a glitch on the CI server. I will add some comments to the code and a test and then it should be done.

mstimberg:
> Note sure why this failed a test, looks like a glitch on the CI server.

Yes, unfortunately that happens every once in a while. I restarted the job, that should be enough.

Either way, thanks for the Pull Request, I'll try to review it soon.

mstimberg:
Oh, just one more thing: while finishing up the PR, would you mind updating the `lfp.py` example to use `Ic` instead of the `run_regularly` hack? It would be odd to keep this around when there is a more convenient way. Thanks!

mstimberg:
I thought a bit about the order issue, and I think it would indeed be the best solution to change the order in general, i.e. add `self.diffusion_state_updater.order = 1` to the `__init__` method of `SpatialStateUpdater`. It makes more sense for comparing `Ic` and `Im`, and it does not change results in any meaningful way (except when initial conditions are not at equilibrium values, but simulations normally have an initialization phase anyway). We'll make a clear note in the documentation/release notes for the users that rely on precisely reproducible results on the sub-timestep scale.

phreeza:
OK sounds good. Want me to do that change in this PR as well?

mstimberg:
> OK sounds good. Want me to do that change in this PR as well?

Yes, please, I think it makes sense to do it here.

phreeza:
There still seems to be a dependency on when in the timestep the current is recorded. You can check this by removing the `when='end'` in [test_spatialneuron.py line 779 ](https://github.com/phreeza/brian2/blob/88e5372fc10878a16d5ac04bf520e2ed534f33cd/brian2/tests/test_spatialneuron.py#L779), this increases the discrepancy between the currents. Not sure how to fix it, any ideas @mstimberg ?

mstimberg:
> There still seems to be a dependency on when in the timestep the current is recorded. You can check this by removing the `when='end'` in `test_spatialneuron.py` line 779 , this increases the discrepancy between the currents. Not sure how to fix it, any ideas @mstimberg ?

That is weird -- nothing should be happening that affects `Ic` or `Im` between `end` and `start`, so recording at `start` should simply give you the values from the preceding time step. I'll look into this tomorrow.

Ah wait, there's a current, that's why: at the first time step where the current changes, `Im` will already take into account the changed current (since it is evaluated with the new current in place), whereas `Ic` does not yet. I don't see any way around this, there is some fuzziness about what happens within one timestep when you have things that are explicitly time-dependent like a changing current. I think that's fine, I can't think of any solution that makes this problem completely go away.

phreeza:
OK, in that case, I think that would be all from my side, do you want to
review?

brian2/codegen/runtime/cython_rt/templates/spatialstateupdate.pyx
```diff
@@ -28,11 +27,13 @@
     cdef int _num_children
     cdef double _subfac
     cdef int _children_rowlength
+    cdef double *_v_previous = <double *>malloc(N * sizeof(double))
```
mstimberg: I just realized that we should probably take the same approach as for the other temporary variables (e.g. `_ab_star0`, etc.), i.e. not allocate/free `_v_previous` every time step, but instead declare it as a "proper" variable in the model. This would mean adding it to the arrays [here](https://github.com/phreeza/brian2/blob/88e5372fc10878a16d5ac04bf520e2ed534f33cd/brian2/spatialneuron/spatialneuron.py#L375), and then referring to it via `{{_v_previous}}` in the template.

brian2/tests/test_spatialneuron.py
```diff
@@ -745,6 +745,44 @@ def test_spatialneuron_morphology_assignment():
     assert_allclose(neuron.sec1.sec11.v[:], np.ones(2)*volt)
     assert_allclose(neuron.sec1.sec12.v[:], np.zeros(2)*volt)
 
+@attr('codegen-independent')
```
mstimberg: The test is good, but currently it is actually only tested for the `numpy` target because it is marked as `codegen-independent` (which is meant for tests that check things at the description level and don't do a run). It should be  marked `@attr('standalone-compatible', 'multiple-runs')` and include an explicit `device.build` call (as described [here](http://brian2.readthedocs.io/en/latest/developer/guidelines/testing.html#unit-tests))

examples/compartmental/lfp.py
```diff
@@ -57,7 +54,7 @@
 lfp.y = [1*mm, 2*mm, 4*mm, 8*mm, 16*mm]
 # Synapses are normally executed after state update, so v-previous_v = dv
 S = Synapses(neuron,lfp,model='''w : ohm*meter**2 (constant) # Weight in the LFP calculation
-                                 v_post = w*(Cm_pre*(v_pre-previous_v_pre)/dt-Im_pre) : volt (summed)''')
+                                 v_post = w*(Ic_pre-Im_pre) : volt (summed)''')
 S.summed_updaters['v_post'].when = 'after_groups'  # otherwise v and previous_v would be identical
```
mstimberg: The comment should be changed, since now this is only necessary to avoid the mismatch between `Ic` and `Im` due to a current. Maybe simply: "otherwise Ic has not yet been updated for the current time step."

phreeza:
Thanks for your feedback, I think I have addressed everything. Haven't tested the standalone code locally, I hope the tests pass.

brian2/codegen/runtime/numpy_rt/templates/spatialstateupdate.py_
```diff
@@ -26,7 +27,7 @@ _vectorisation_idx = 1
 _vectorisation_idx = LazyArange(N)
 {{vector_code|autoindent}}
 
-_v_previous = {{v}}.copy()
+{{_v_previous}} = {{v}}.copy()
```
mstimberg: It's doesn't matter much here, but this will not store the copy in the reserved memory but instead overwrite the variable name. This means that e.g. during debugging, a user looking at the value of `neuron._v_previous` would not see the values set here. I'd therefore replace this with `{{_v_previous}}[:] = {{v}}`

brian2/devices/cpp_standalone/templates/spatialstateupdate.cpp
```diff
@@ -5,6 +5,7 @@
                   _ab_star0, _ab_star1, _ab_star2, _b_plus,
```
mstimberg: `Ic` also has to be added to the `USES_VARIABLES` block.

brian2/codegen/runtime/numpy_rt/templates/spatialstateupdate.py_
phreeza: Ah, yes that makes total sense. Didn't think properly.

mstimberg:
Great, thanks. Looking good to merge, all tests pass on my machine. I'll wait for the tests on the test servers to run through, and then merge.

PR #829, opened 2017-04-03
small change in grammar of documentation

CharleeSF:
just to show I can do a pull request


mstimberg:
Thanks, looks good to me :) Pro-tip for future "documentation only" commits/PRs: if you add `[ci skip]` somewhere in your commit message, it will not trigger the test suite on our servers. This does not really matter at the moment, but in periods of high development activity it can reduce waiting time for the commits that changed stuff that needs to be tested.

Issue #680, opened 2016-04-16
Don't use group names in runtime code


thesamovar:
If we can avoid using names of groups in generated code then we can avoid some recompilations that are currently happening. We could, for example, if there were two groups with names ``neurongroup_3`` and ``neurongroup_4`` map ``_array_neurongroup_3`` to ``_array_NeuronGroup1`` and ``_array_neurongroup_4`` to ``_array_NeuronGroup2`` (i.e. we would look up the class name of the object pointed to, and give them a numerical index that increases from 1). Now if we ran again but it would ``neurongroup_5`` and ``neurongroup_6`` then it would still map to the same ``_array_NeuronGroup1`` etc. and a recompilation wouldn't be necessary.

I think we're already not freezing constants right? #480 suggests this, but on the other hand when I run some code with an increasing number of neurons ``N`` I seem to be getting recompilations sometimes. I think it would be good to avoid this for people running parameter space searches.

mstimberg:
I'd rather not make the link between the names in the simulation and the names in the code less clear if we can avoid it. Do you really think this is a problem in practice? If you run a simulation in a for loop, then you should only end up with two different sets of names, regardless of the total number of iterations (the second set of names because at construction time, the old objects still exist). Will this be noticeable in any real setting? The only situation where I can image that it somewhat matters would be in a jupyter notebook or a similar interactive setting. 

> I think we're already not freezing constants right? #480 suggests this, but on the other hand when I run some code with an increasing number of neurons N I seem to be getting recompilations sometimes.

We don't freeze constants in runtime (we do in standalone, though) -- are you sure you are not getting the recompilations due to the changed names?

thesamovar:
Do we reuse old but now deactivated ``neurongroup_i`` names or do we just add more at the end always? I feel like I am getting more than one recompilation in a loop. I see it very clearly on the speed test that the first time I run it, for all N weave and cython are much slower for the first run than the second. If it were just two recompilations then I think that would be fine! Not ideal, but good enough.

Another solution: we can add a warning to the user if they are getting potentially unnecessary recompilations by taking the code string X and removing all names ``neurongroup_i`` to get Y say, and then counting how many different X map to each Y. If it's more than 2 then we could warn about potentially unnecessary recompilations and that the user should explicitly name the objects.

mstimberg:
[...]

PR #832, opened 2017-04-05
[MRG] Caching and small run preparation improvements

mstimberg:
I got around to work a bit on the caching stuff we have been discussing for a long time (in particular, see #124), to avoid the overhead of repeated preparations when doing multiple runs.

For somewhat extreme cases (running a short simulation for a HH model in a loop), this PR has a great impact on performance (e.g. running a 20ms simulation of 100 HH neurons took ~2s instead of ~9s). Simulations with a single run are only affected in a minor way, and of course all this is only relevant for the preparation period, so it only matters when the run time for your actual simulation is short.

Contrary to the title of #124, I did not use joblib to do the caching, because I did not really see what this would buy us. The caching is only done in memory with simple dictionaries, I did not try to implement caching on disk (for which of course joblib would be helpful). The main reason is that I am seeing a lot of potential problems (e.g., we fix a bug in a state updater or in the code generation, and users still access an old cached version) and not much benefit.

Caching is done on multiple levels. The lowest levels are helpful even for single runs:
* `sympy_to_str` and `str_to_sympy` cache their results -- the mapping from one to the other should be fixed, I only see one hypothetical problem: if the user changes a function in `DEFAULT_FUNCTIONS` so that it maps to another sympy name, this would not be taken into account. But I don't see why a user would do this in the first place, and even less so between two runs. I also made some minor tweaks to the two functions, making them slightly faster for non-cached calls.
* `Equations.get_substituted_expressions` is cached. This is called several times during the preparation period and since `Equations` are immutable by design, caching is safe here.

Then there is caching on the abstract code level:
* The state updater application (same equations, same variables, same method) is cached in `apply_stateupdater`, which also means that it remembers which method was chosen if none was specified explicitly. For complex models and/or integration methods, this part is where most of the time was spent during run preparation, so this has a big influence on the run time for repeated runs. To make this work, I had to make `CodeString`, `SingleEquation`, and `Equations` properly comparable/hashable, so that they compare equal when their content is equal (it probably would have worked with a simpler identity-based comparison as well, since we don't recreate these objects normally, but well).
* The transformation of code into a series of statements is cached as well (i.e. `make_statements`). Since the time when we introduced optimizations this is a very costly operation as well.

Finally, there is caching on the target-specific code generation level: if a `CodeObject` is created for the same code, same variables, same template, same target, etc., then it is re-used instead of re-created. Re-creating these objects was also taking a significant time. However, this caching was a bit trickier to get right, since previously the transformation from the variable dictionary to the namespace that was used for the `CodeObject` was done during object initialization. We have to do this for every run, though, because things like e.g. the `num_...` variables for synapses have to be updated in case they changed between runs. This resulted in the changes you see in the last commit.

I tried to err on the side of caution, and much of this relies on they way how `Variable` objects are compared/hashed. Standard variables (arrays) etc. are compared by identity -- if the object is the same, it is considered the same from the code generation point of view (of course the values it refers to may have changed, but our generated code only references them). The same is true for user-defined functions, however there it comes with a very minor disadvantage: user-defined Python-only functions (i.e., just defined with `check_units`) will create a new `Function` object every time, so code referring to them will not benefit from caching. I don't think that is a problem that needs fixing, though. Finally, `Constant` objects are compared by their actual value -- we re-create them at every run and codegen might have hard-coded them into the code, so we want to make sure to not hide changes in constant values by caching, while not making our caching completely useless at the same time :)

That's it, it seems to be working fine as far as I can tell, so good to merge from my side.


mstimberg:
> That's it, it seems to be working fine as far as I can tell, so good to merge from my side.

Hah, the latest mail on the mailing list made me realize there's actually a problem with my current implementation that we need to fix first: all the caches hold strong references to the `Variable` objects, so this means that memory will not be freed if you get rid of objects...

thesamovar:
This is the sort of change that needs careful thought and testing, so let's not rush but... great work! It will be fantastic to have this in Brian.

I would argue that there is a benefit to cacheing to disk (e.g. people who run a script multiple times for parameter search rather than doing multiple run calls from within a single run of a script), and we can potentially make this safe. In the key for the (on-disk) dictionary, we can include Brian's version number (or better a hash of the source code if this is reasonably quick to compute once at the beginning). Another alternative is to include in the install script something that deletes the cache. There'll be some users who could shoot themselves in the foot, i.e. those who run from git and have their python path point directly to the git directory rather than installing. But that's probably only developers!

The other really huge benefit is to standalone mode. There are a few things we could do to improve this. For example, if we detect that the code we are about to write to a file is the same as the code already in that file, we can just choose not to write to it. Like that the compiler knows that the file is already compiled and doesn't have to recompile it. Hmm, not sure how much benefit we get from that though given that the ``objects.h`` file is the most likely to change, and almost all files depend on this. Maybe there's a way to reorganise to avoid this problem though? Presumably the current cacheing of the code generation pathway will already help for standalone. Since we don't yet cache to disk this is of relatively minimal use right now, but with cacheing to disk we'd already get some benefit there. Combining these different bits of cacheing could make a huge difference to standalone.

Last (techical) point. For ``DEFAULT_FUNCTIONS`` if there is no conceivable need to change it, and that can only cause problems, why not make it a frozen dict after it's created?

thesamovar:
Oh, and it might take me a bit of time to get around to testing and reviewing this because it's a biggie, but it's also important so will try to get to it as soon as possible.

mstimberg:
> This is the sort of change that needs careful thought and testing, so let's not rush

Yes, agreed! I will probably also add a few more tests with potential problems in mind (changing stuff between runs). Apart from that, there's the memory problem and apparently the test server failed for Python 3 and standalone, so there's still some work to do ;-)

> I would argue that there is a benefit to cacheing to disk (e.g. people who run a script multiple times for parameter search rather than doing multiple run calls from within a single run of a script), and we can potentially make this safe.

I have to say I am quite hesitant to add on-disk-caching for now. This is really a non-trivial problem to get right:
1. it has to be safe for multiprocessing (joblib could take care of this, though)
2. Brian's version/source code is potentially not enough, we'd also need to take into account dependencies like sympy (an install script would not help here)
3. Packages like Brian2GeNN would also need to be taken into account, or we'd need a whole system where packages can register their relevance for caching, or we'd need to version `CodeObject`/`Generator` classes or something like that

Standalone is also a whole different beast, for stuff like parameter explorations I think there it would make sense to have a more explicit system that does not rely on automatic caching. Something like the parameter exploration stuff we discussed previously (run an already compiled simulation but inject different parameters).

How about first getting the memory-based caching right and giving it some exposure? We can then tackle the disk-based cache as a new issue.

> For example, if we detect that the code we are about to write to a file is the same as the code already in that file, we can just choose not to write to it.

Dan-of-the-past already implemented this ~3 years ago :smile: https://github.com/brian-team/brian2/blob/master/brian2/devices/cpp_standalone/device.py#L112

> For `DEFAULT_FUNCTIONS` if there is no conceivable need to change it, and that can only cause problems, why not make it a frozen dict after it's created?

Possibly. I wonder whether *adding* new functions would not be something reasonable to allow. If it is done at the beginning of a script, that should not lead to any problem (assuming only in-memory caching, of course). Another easy fix would be to simply use the `DEFAULT_FUNCTIONS` dictionary as part the cache key, so changes would invalidate the cache automatically.

thesamovar:
> Brian's version/source code is potentially not enough, we'd also need to take into account dependencies like sympy (an install script would not help here)

This probably shouldn't matter though, right? In this case, certainly a version number for sympy would do the trick.

> Packages like Brian2GeNN would also need to be taken into account, or we'd need a whole system where packages can register their relevance for caching, or we'd need to version CodeObject/Generator classes or something like that

Agreed that we don't want to get into that! Is it definitely necessary though since those libraries won't use the cacheing mechanism by default? Or maybe they will because they'll be called by Brian which will do cacheing.

Another idea. We make the cacheing to disk off by default and you need to explicitly enable it, either with a preference or maybe (even safer) with something in the code (e.g. ``set_brian_disk_cache('directory_name')``. Like that you're reminded about it. We could then transition to default-on cacheing after some testing and resolving some of these issues. The benefit is that this gives a way for people who want to do multiple runs to get the speed improvements, even if it's off by default.

> Standalone is also a whole different beast, for stuff like parameter explorations I think there it would make sense to have a more explicit system that does not rely on automatic caching. Something like the parameter exploration stuff we discussed previously (run an already compiled simulation but inject different parameters).

Yes, but if we could - at relatively low effort - get an intermediate solution that improves it a bit that would still be good because the compilation times for standalone can be pretty long and negates a lot of the benefit.

> Dan-of-the-past already implemented this ~3 years ago

I'm getting to be like one of those old men who keep telling the same stories as if you'd never heard them before.

In any case, apparently that alone isn't enough to stop recompilation. I might think about that idea for optimising the header files to save on recompilations.

> Another easy fix would be to simply use the DEFAULT_FUNCTIONS dictionary as part the cache key, so changes would invalidate the cache automatically.

Yeah that's probably good, although you don't necessarily want to have to hash that dictionary every time you access the cache?

mstimberg:
> Brian's version/source code is potentially not enough, we'd also need to take into account dependencies like sympy (an install script would not help here)
> 
> This probably shouldn't matter though, right? In this case, certainly a version number for sympy would do the trick.

Sure, that was just to point out that there are many potential pitfalls that are not obvious at first sight.

> Agreed that we don't want to get into that! Is it definitely necessary though since those libraries won't use the cacheing mechanism by default? Or maybe they will because they'll be called by Brian which will do cacheing.

Yes, the way it is currently implemented, a `GeNNCodeObject` for example would be cached by Brian.

> Another idea. We make the cacheing to disk off by default and you need to explicitly enable it, either with a preference or maybe (even safer) with something in the code (e.g. set_brian_disk_cache('directory_name'). Like that you're reminded about it.

Something like this might be a good solution (or a preference or something like that). However, I'd still prefer to finish/merge the memory-based caching before going the next step. I also wonder whether a proper explicit `Network.write_to_disk` and `Network.load_from_disk` (which would store all the caches on disk as well) would not be worth it. For runtime mode, the caching does not help you much for repeated runs if your code spends a lot of time generating synapses, for example. Ok, you can do that part with `store/restore` already actually. It still feels like a waste to do all the object creation etc. again, but maybe this is not really an issue if all the computationally demanding stuff is cached.

> In any case, apparently that alone isn't enough to stop recompilation. I might think about that idea for optimising the header files to save on recompilations.

Right, that's actually a bug. It should not recompile if you run exactly the same script twice (assuming that you used `set_device(...)` or `build(...)` with `clean=False` which is *not* the default (maybe it should?)). I think the only reason why it does this right now is because in some places, we did not fix the sorting for elements. So the contents of a dictionary are printed in an order that can change between runs, triggering the re-compile. As soon as one single file changed, we'll have to link all the object files again which seems to take most of the time of the compilation process.

> Yeah that's probably good, although you don't necessarily want to have to hash that dictionary every time you access the cache?

I was worried about this for the `Variables` dictionary as well, but hashing is really fast. I just tested hashing `DEFAULT_FUNCTIONS` and it takes 1.3us, so I think this negligible.

thesamovar:
> Something like this might be a good solution (or a preference or something like that). However, I'd still prefer to finish/merge the memory-based caching before going the next step.

That seems reasonable. Maybe let's not close #124 when this is merged then?

>  I also wonder whether a proper explicit Network.write_to_disk and Network.load_from_disk (which would store all the caches on disk as well) would not be worth it. For runtime mode, the caching does not help you much for repeated runs if your code spends a lot of time generating synapses, for example. Ok, you can do that part with store/restore already actually. It still feels like a waste to do all the object creation etc. again, but maybe this is not really an issue if all the computationally demanding stuff is cached.

Indeed! That would be even better.

> Right, that's actually a bug.

OK, will open a separate issue for that. Yeah, maybe ``clean=False`` by default is a good idea!

Let me know when you've got this to a state where it's mergeable and I'll do a more detailed review?

mstimberg:
> That seems reasonable. Maybe let's not close #124 when this is merged then?

Sure. It will be closed automatically due to my issue comment, but we can of course reopen it.

> Let me know when you've got this to a state where it's mergeable and I'll do a more detailed review?

Will do!

mstimberg:
Well, well, this was considerably more complicated than expected... I fixed all test failures and the memory leak in the changes up to commit 402de76. It was quite a bit of work, since the `CodeObject` that we were caching had references to the variables dictionary, the namespace, etc, and we could not use weak references easily in a few places. I think the end result was ok, but it broke `brian2genn`, and the changes to make it work again weren't obvious. I felt that caching `CodeObject` was maybe a bit too ambitious, given that it is not a well-separated object, but is tightly linked into the code generation/execution process in several places. So I decided to revert basically all of the changes I did (I left the commits in place so that we can go back, though) and instead cache the `CodeGenerator.translate` function which (together with `make_statements` which is already cached) takes up most of the time of the `CodeObject` translation but it is better separated from the rest, it has a fairly clear input/output flow (abstract code in, target language code out). This worked fine (commit 3b34b53) but then I became a bit more ambitious, also thinking about future extensions to on-disk caching.

The limitation of the previous conservative "object identity for `Variable` objects" approach was that it was good enough for running the same objects several times (e.g. using `store`/`restore`, or simply multiple run statements where a variable changes between runs), but it did not help at all when objects where re-created (which of course also creates new `Variable` objects). Now this is something that users should generally avoid, but of course often they don't. There are also legitimate use cases of re-creating objects, see for example the memory-leak related question on the list recently where the user wanted to re-create `StateMonitor` objects to free their memory. I therefore implemented a new way of comparing `Variable` objects, where only all the information that describes the variable in an abstract way counts (name, size, type, etc.), but not its actual object identity or the identity of the "owner". This means that loops that re-create the same objects are now also sped up, limited only by the changing names between runs (#680). However, I soon realized that the simple "abstract code + owner name + language (+ a few other details)" --> "target language code" mapping does not quite work in several corner cases. For example, the `CodeObject` representing a `pre` statement might have the same name and be connected to a `Synapses` object of the same name, but still produce different code because the source/target `NeuronGroup` has a different name. We were also changing the `variables` object in-place in `CodeGenerator.translate` to add the variables for function namespaces (e.g. the array that a `TimedArray` refers to), which of course screwed everything up since the `variables` dictionary was used as part of the key of our cache and supposed to never change. Similarly, we actually changed the content of `Equations` objects in some places, even though the `Equations` object is supposed to be immutable. Finally, even something as global as Brian's preferences matter for `CodeGenerator.translate` (e.g. it decides whether code for OpenMP will be created or not).

I was close to giving up and saying that it is too dangerous to cache the code generation process for now (i.e., before we refactor it to have clearer inputs/outputs), and only keeping the caching of the state updater, some parsing functions, etc. I *think* I now did fix all the issues, though, at least the test suites run through. Still, this certainly needs some rigorous testing. Apart from adding a few tests that are tailored to potential caching problems, I don't think I will add anything else to this branch. So this is up for review now, whenever you find some time.

thesamovar:
This sounds like it does indeed need some rather careful review. Will get on that as soon as I can. And indeed, it also sounds like refactoring codegen might make this a whole lot safer.

Did you do any performance comparisons with it yet?

mstimberg:
> Did you do any performance comparisons with it yet?

Yes, I did -- don't take the numbers below too literally, though, it can easily vary by several 1/100s between runs (for the first run in particular). The benchmark is a HH model, either with no method specified (which will chose Euler, not really appropriate here but well...), or with `method='rk4'`. The simulation is repeated in a loop (the simulation only runs for 0ms, so all this is only preparation time), either without or with object re-creation:

```Python
# Same object
start = time.time()
times = []
for i in range(10):
    run(0*ms)
    times.append(time.time() - start)
    start = time.time()
    
# New object
start = time.time()
times = []
for i in range(10):
    axon = NeuronGroup(100, eqs, method='rk4', threshold='v>50*mV')
    run(0*ms)
    times.append(time.time() - start)
    start = time.time()
```



The measured times on my system (note that all this uses pre-compiled weave code already on disk):

### Same objects / RK4

|                   | First run | Subsequent runs |
| ----------------- | --------: | --------------: |
| **Release 2.0.1** |     1.58s |           1.33s |
| **master**        |     1.46s |           1.17s |
| **this PR**       |     1.27s |           0.05s |



### New objects / RK4

|                   | First run | Subsequent runs |
| ----------------- | --------: | --------------: |
| **Release 2.0.1** |     1.61s |           1.33s |
| **master**        |     1.45s |           1.16s |
| **this PR**       |     1.38s |           0.06s |

### Same objects / automatic method choice

|                   | First run | Subsequent runs |
| ----------------- | --------: | --------------: |
| **Release 2.0.1** |     0.62s |           0.37s |
| **master**        |     0.31s |           0.18s |
| **this PR**       |     0.26s |           0.05s |

### New objects / automatic method choice

|                   | First run | Subsequent runs |
| ----------------- | --------: | --------------: |
| **Release 2.0.1** |     0.58s |           0.40s |
| **master**        |     0.40s |           0.21s |
| **this PR**       |     0.30s |           0.05s |

Especially if you use an expensive method such as `rk4`, this does really make a clearly noticeable difference, having one additional second of preparation time at each iteration will matter if your simulation run is short (and of course the preparation time will be longer for more complex networks).

thesamovar:
Well I think that could be a big win for certain parameter searching or behaviour mapping type scenarios, so well worth it.

mstimberg:
Yes, I agree. Oh, and there is one more thing I realized during benchmarking: running the benchmark with the "magic" network is slower than using an explicitly constructed `Network` object. This in itself is not too surprising, but the speed difference comes mainly from the `gc.collect()` we do after each run of the magic network. We put this in to make sure that all unused objects are cleared from memory, but maybe this is unnecessary now. By itself, it does not take care of the issue we discussed in #680, anyway.
Just for reference: without garbage collection, the last example (new objects / automatic method choice) takes 0.26s for the first and 0.02s for the second run, i.e. the garbage collection takes about 0.03s for each run. Of course this is not much in absolute terms, but it is more than all the other preparation time combined if all elements are already in the cache.

brian2/codegen/generators/base.py
```diff
@@ -215,10 +216,27 @@ def has_repeated_indices(self, statements):
...
+                               self.allows_scalar_write,
+                               self.iterate_all,
+                               self.template_name,
+                               self.name,
+                               self.owner.name,
+                               {var.owner.name
+                                for var in self.variables.itervalues()
+                                if getattr(var, 'owner', None) is not None},
+                               prefs,
+                               self.__class__])
```
thesamovar: This feels unsafe because it's not obvious from this function alone that the outcome of code generation depends solely on these variables, and if you change another function somewhere else to add a new dependency, you might not remember to update this here. As an alternative, what about the following. Make the ``CodeGenerator`` object hashable and give it a hash function. This hash function will then call the hash functions of any objects that are part of the code generator, and like that the code that generates the key is distributed to the place where the items in that key are actually used. In addition, modify the code so that any external variables must be extracted in a single function call which returns a value that is considered part of any key. We can't enforce that in Python, but the big ones would be prefs and ``current_device``, so we could add a switch that makes those temporarily inaccessible, and wrap any code generation functions with something that turns this switch on and then off again at the end. Like that, if we forget and try to access prefs or device in a place it doesn't know about, it will cause an error.

thesamovar: Alternatively, we consider this sort of change as part of a future larger refactor of code generation, but it makes me feel very wary.

thesamovar: Another alternative. If there is some (potentially expensive) way of writing a program that keeps track of which variables a function call actually accesses, then we could use that as part of the test script to assert that it really does only access variables that are part of the key. Not sure how feasible this is.

thesamovar:
OK well the tests pass on my machine too, including the long and standalone ones. Now that I've looked at the code, I'm still concerned about safety. It's really nicely done and indeed I think you've caught everything, but it just still feels like it would be so easy to slip up here without some way of forcing ourselves not to. Looks to me like there are three options (mentioned in the inline comments above, but maybe better to have part of the main thread so I'm repeating slightly):

* Refactor to make everything that is cached more clearly a pure function. There is a Python library that can check if a function call is pure or not: https://dan.bravender.net/2009/8/23/Purely_Functional_Programming_for_Python.html. It's a bit strict because any external function call violates the purity (which is good). This would be quite a lot of refactoring though.
* Plan to do this refactoring fairly soon, but not quite yet. This feels risky to me. We could add a few more safety items to the key, like today's date, but that won't catch everything, only changes to Brian's source code.
* Write some code that traces execution of the code generation pipeline and checks whether or not the variables that we think are the only ones being accessed (and are part of the key) are indeed the only ones being accessed. This would then be added to the test script as it would presumably be expensive. Not sure if this is feasible, but it would be a nice option if it were.

Thoughts on these options?

mstimberg:
I agree with the concerns about the safety, and as I mentioned earlier I was close to giving up on caching the `CodeGenerator.translate` part. I'd certainly like to refactor things to make them (as much as possible) purely functional, this would also allow much easier testing (currently, we mostly test the full code generation pipeline instead of the individual units individually). But this is something that will take quite some time and will also most likely break projects like Brian2GeNN, so I don't think this is a short-time solution. For now, I think we can chose between:
1. remove caching for translate (the other caching should be safe and already safes a lot of time)
2. trust our test suite (and maybe add more tests like the ones in my last commit)

Some specific comments on some of your points/ideas:

> As an alternative, what about the following. Make the CodeGenerator object hashable and give it a hash function. This hash function will then call the hash functions of any objects that are part of the code generator, and like that the code that generates the key is distributed to the place where the items in that key are actually used.

I tried something like this before, but the problem is that for different caching situations we want different hash functions: from the point of view of numerical integration, `'dv/dt = -v/(10*ms)'` is the same regardless of the `NeuronGroup` that defines it, or more specifically independent of the `owner` of the `ArrayVariable` that represents `v`; for code generation, however if everything is the same but `v` now belongs to an `owner` named `neurongroup_1` instead of `neurongroup`, this changes the code that is generated and therefore has to make a difference.

> We could add a few more safety items to the key, like today's date, but that won't catch everything, only changes to Brian's source code.

You are talking about on-disk caching here, right? This would open a whole different can of worms, let's stick with in-memory caching for now, ok :smile: 

About all the ideas about tracing: I think in principle we can do anything, but setting up a framework for that would be quite complex and I am not sure it is worth the effort -- I'd rather prefer the inputs/outputs of the functions to be clear enough that we don't need any tracing to convince ourselves that it doesn't change anything that it should not change.

So, my conclusion/preferred solution:
* revert the caching of `CodeGenerator.translate` -- the main time in this function is spent in `make_statements` (where all the optimization happens), and this we already cache. 
* The remaining functions that are cached (`make_statements`, `apply_state_updater`, `parse_string_equations`, etc.) have quite clear inputs and (with the fixes in this branch) should not modify their arguments
* in the long-term: refactor the code generation process and look into on-disk caching

What do you think?

thesamovar:
That sounds like the solution to me!

mstimberg:
I reverted the caching of `CodeGenerator.translate`, so please have another look!

thesamovar:
I haven't forgotten! I promise.

thesamovar:
I'm going to start reviewing this now. Might take a few days for me to get it done. Also, I'm planning to ask a few more questions about particular lines of code than I normally would so as to be sure that everything is safe.

brian2/core/variables.py
```diff
@@ -274,6 +275,9 @@ def __repr__(self):
                                   constant=repr(self.constant),
                                   read_only=repr(self.read_only))
 
+    _state_tuple = property(lambda self: (self.dim, self.dtype, self.scalar,
+                                          self.constant, self.read_only,
+                                          self.dynamic, self.name))
```
thesamovar: Similar comment to above, is there any way we can make this more safe against our own future changes by automating the definition of ``_state_tuple``?

brian2/devices/cpp_standalone/device.py
```diff
@@ -766,18 +767,23 @@ def copy_source_files(self, writer, directory):
...
-        # # this works for the namespaces for functions with C++ (e.g. TimedArray
-        # # treats it as a C array) but does not work in places that are
-        # # implicitly vectorized (state updaters, resets, etc.). But arrays
-        # # shouldn't be used there anyway.
+        # Write Function namespaces as static arrays
         for code_object in self.code_objects.itervalues():
-            for name, value in code_object.variables.iteritems():
-                if isinstance(value, np.ndarray):
-                    self.static_arrays[name] = value
-
```
thesamovar: Is the idea here that previously we looked for static arrays, but since the only way they can be accessed is by a function that we can just insert function namespaces instead?

brian2/equations/codestrings.py
```diff
@@ -29,18 +31,31 @@ class CodeString(object):
...
 
-        # : The code string
-        self.code = code
+        self._code = code
 
         # : Set of identifiers in the code string
         self.identifiers = get_identifiers(code)
 
+    code = property(lambda self: self._code,
+                    doc='The code string')
```
thesamovar: Is there a reason for changing code from an attribute to a property?

brian2/equations/equations.py
```diff
@@ -433,6 +437,25 @@ def __init__(self, type, varname, dimensions, var_type=FLOAT, expr=None,
                                                       if variable =='xi' or variable.startswith('xi_')]),
                                     doc='Stochastic variables in the RHS of this equation')
 
+
+    _state_tuple = property(lambda self: (self.type, self.varname,
+                                          self.dim, self.var_type,
+                                          self.expr, tuple(self.flags)),
+                            doc='A tuple representing the full state of this '
+                                'object, used for hashing and equality '
+                                'testing.')
```
thesamovar: Same comment again. Here we could automate the creation of ``_state_tuple`` from the arguments passed to ``__init__``, so either put a decorator there or have it derive from some class that automates the creation of the ``_state_tuple``. This solution could probably also be used for ``Variable`` actually.

brian2/parsing/rendering.py
```diff
@@ -181,52 +181,102 @@ def render_UnaryOp(self, node):
...
+        op_name = node.op.__class__.__name__
+        if op_name == 'UAdd':
+            # Nothing to do
+            return self.render_node(node.operand)
+        elif op_name == 'USub':
+            return -self.render_node(node.operand)
+        elif op_name == 'Not':
+            return sympy.Not(self.render_node(node.operand))
+        else:
+            raise ValueError('Unknown unary operator: ' + op_name)
```
thesamovar: This is all rewritten to avoid going to a string and then turning them into a sympy object, right? That seems sensible, but just wondering why it was necessary.

brian2/parsing/sympytools.py
```diff
@@ -148,7 +141,11 @@ def sympy_to_str(sympy_expr):
     str_expr : str
         A string representing the sympy expression.
     '''
-    
+    if sympy_expr in _sympy_to_str_cache:
+        return _sympy_to_str_cache[sympy_expr]
```
thesamovar: How safe is this in cases of changes in version of sympy?

brian2/stateupdaters/base.py
```diff
@@ -100,6 +103,16 @@ def apply_stateupdater(equations, variables, method, group_name=None):
...
+        if isinstance(method, basestring):
+            method_key = method
+        elif hasattr(method, '__call__'):
+            method_key = method.__class__.__name__
+        else:
+            method_key = tuple(method)
+        cache_key = (equations, _hashable(variables), method_key)
+        if cache_key in StateUpdateMethod._state_update_cache:
+            return StateUpdateMethod._state_update_cache[cache_key]
+
```
thesamovar: This looks like another case where wrapping ``_init__`` or the class could work to automate this.

brian2/codegen/translation.py
```diff
@@ -211,6 +210,10 @@ def make_statements(code, variables, dtype, optimise=True, blockname=''):
     describing how the statement can be reformulated as a sequence of if/then
     statements. Calls `~brian2.codegen.optimisation.optimise_statements`.
     '''
+    cache_key = (code, _hashable(variables), dtype, optimise, blockname)
+    if cache_key in _make_statements_cache:
+        return _make_statements_cache[cache_key]
+
```
mstimberg: Ok, I did this (I'll push the commit in a minute) -- we still have to be careful whenever we add arguments and think about how they relate to the caching, but I think the most common issue that we could have is that the cache no longer works.

brian2/core/variables.py
mstimberg: I don't think we can automatize the `_state_tuple` here: it's very point is to make the distinction between attributes that matter for cacheing and those that don't. E.g., for arrays all describing attributes (`dtype`, `dim`, etc.) matter, but `value` does not. For a `Constant`, on the other hand, the `value` matters (because it might be used with its value in the generated code).

brian2/devices/cpp_standalone/device.py
mstimberg: More or less: it is part of a fundamental change how function namespaces (most important use case is `TimedArray`) where added. Previously, we added them to the original `variables` dictionary at some point. This screwed up caching, because the `variables` were used as part of the cache key earlier and therefore should not change. For this specific bit of code we therefore no longer search for those arrays as part of the `variables` dictionary, but explicitly add them from the `Function`.

brian2/equations/codestrings.py
mstimberg: It's just to make the immutability of the `CodeString` very clear: `code` is a read-only attribute and should never be changed. We use this idiom in a few places and it seemed to make sense to use here as well.

brian2/equations/equations.py
mstimberg: As mentioned above, I don't think we can automate this in a straightforward way (we could have a function where we define which attributes should be considered for caching and which not, but then I don't think this would make things clearer).

brian2/parsing/rendering.py
mstimberg: Yes, before we took a string, parsed it's AST, created a string from it and then eval'ed it which again parsed its AST and created the sympy objects. Since this stuff is called a lot of times (less with the caching now, though), it made a measurable time difference. It also feels more logical to do it this way instead of making the string roundtrip. Oh, and I think it might even prevent some loss of precision due to converting floats to strings and back, but not sure that was actually an issue.

brian2/parsing/sympytools.py
mstimberg: Not sure what you mean -- this is still all about caching in memory, so I don't see how version changes can come into play.

brian2/stateupdaters/base.py
mstimberg: yep, see my upcoming commit.

brian2/codegen/translation.py
thesamovar: Yeah that's ideal if we make a change and it causes an error rather than silently making a mistake.

brian2/core/variables.py
thesamovar: You mention this a bit later, but what about a system where all attributes are assumed to matter unless we explicitly exclude them? Like that, if we change the class in the future we are forced to make a decision about the new attributes. I feel like this isn't a lot more work and minimises the chance of silent errors in the future.

brian2/devices/cpp_standalone/device.py
thesamovar: OK!

brian2/equations/equations.py
thesamovar: As above - it's not about making it clearer it's about protecting ourselves from our future selves. I think this is particularly important for this cacheing stuff because it would be so easy to make a mistake here that we have to go to some efforts to avoid doing so.

brian2/parsing/rendering.py
thesamovar: Great!

brian2/parsing/sympytools.py
thesamovar: Ah good point!

brian2/utils/caching.py
```diff
@@ -0,0 +1,92 @@
...
+Module to support caching of function results to memory (used to cache results
+of parsing, generation of state update code, etc.). Provides the `cached`
+decorator.
+'''
+
+import functools
+import collections
+
+
+class _CacheStatistics(object):
```
thesamovar: Do we use this or just there for potential future performance/debugging?

thesamovar:
That's all looking great - I would also have the system we mentioned for marking some variables/attributes as not part of the cache for ``Variable`` and derivatives (can maybe add arguments to the ``@cached`` decorator to list ignored function arguments?), and then probably good to go. I already like the new version with your latest commit much better - I find it much cleaner.

brian2/equations/equations.py
mstimberg: I see your point, but any change to those core classes has to be done very carefully -- there are a lot of other possibilities to break this apart from the `_state_tuple`. I don't see a straightforward way to make this specific thing safer with regard to future changes, IMO having extensive tests is the most important...

brian2/utils/caching.py
mstimberg: I used this just for seeing that the caching actually worked. We could potentially use this in tests I guess?

brian2/core/variables.py
mstimberg: Ah, sorry, overlooked this comment... Hmm, I guess we could use `__dict__` to get all attributes and do something along these lines. I'll give it a go!

brian2/utils/caching.py
thesamovar: Yep - fine to leave it there anyway. We might find a use for it one way or the other.

brian2/equations/codestrings.py
thesamovar: OK!

brian2/core/variables.py
```diff
@@ -165,6 +164,16 @@ def __init__(self, name, dimensions=DIMENSIONLESS, owner=None, dtype=None,
         #: Whether the variable is an array
         self.array = array
 
+        #: Mark the list of attributes that should not be considered for caching
+        #: of state update code, etc.
+        self._cache_irrelevant_attributes = ['owner']
```
thesamovar: This could be a class attribute maybe? Not sure if it's worth doing but maybe slightly more logical.

brian2/core/variables.py
```diff
@@ -165,6 +164,16 @@ def __init__(self, name, dimensions=DIMENSIONLESS, owner=None, dtype=None,
...
+
+    @property
+    def _state_tuple(self):
+        return tuple(value for key, value in self.__dict__.iteritems()
+                     if key not in (self._cache_irrelevant_attributes +
```
thesamovar: Do we need to sort the items in a dict or is the order deterministic?

brian2/equations/equations.py
```diff
@@ -426,6 +426,18 @@ def __init__(self, type, varname, dimensions, var_type=FLOAT, expr=None,
...
+        #: of state update code, etc.
+        self._cache_irrelevant_attributes = ['update_order']
+
+    @property
+    def _state_tuple(self):
+        '''A tuple representing the full state of this object, used for
+        hashing and equality testing.'''
+        return tuple(value for key, value in self.__dict__.iteritems()
+                     if key not in (self._cache_irrelevant_attributes +
+                                    ['_cache_irrelevant_attributes']))
```
thesamovar: This could potentially be refactored since it's the same code as before?

brian2/core/variables.py
mstimberg: Indeed! It would even simplify the code a bit, since it wouldn't appear in `__dict__` itself!

thesamovar:
I like the new changes, this is feeling much safer and cleaner to me. Hope you didn't find that too tedious to do. I made a couple of non-essential suggestions. Let me know when you're ready and I can do a final review.

brian2/core/variables.py
mstimberg: It shouldn't matter here since the order is arbitrary but deterministic, but I'm not 100% sure there are no corner cases. So, using `sorted` sounds reasonable.

brian2/equations/equations.py
mstimberg: I considered it, but given that the function is almost trivial, I'm not sure it's worth introducing a common super-class for `SingleEquation` and `Variable` given that the classes don't have much in common... Hmm, maybe something like `CacheKey` could make sense, I'll have a look.

thesamovar: Yeah I was thinking a sort of mixin class.

mstimberg:
Ok, moved stuff into a `CacheKey` mixin class -- I agree that it's cleaner this way! Ready for final review I guess :)

mstimberg:
No worries, it was a quite substantial change and I think the code improved quite a bit during the discussion.
I might delay the merge a bit until @CharleeSF's PR, I could imagine that it leads to quite a few merge conflicts...

thesamovar:
Sure!

PR #834, opened 2017-04-11
Minor changes to the documentation

MihirVaidya94:
Just to show I am comfortable using Git. 
Some of the methods mentioned in the "List of methods" and "Math operations" section could not be read as they were enclosed in single quotations(``...``) and hence occupied a single line. So I broke them up into groups of 90-100 characters and enclosed each group in different quotations.

### DIFF
--- a/docs_sphinx/developer/units.rst
+++ b/docs_sphinx/developer/units.rst
@@ -65,11 +65,12 @@ appropriate. Most of the methods are overwritten using thin function wrappers:
 
 **List of methods**
 
-``all, any, argmax, argmax, argsort, clip, compress, conj, conjugate, copy,
-cumsum, diagonal, dot, dump, dumps,`` ``fill, flatten, getfield, item, itemset, max,
-mean, min, newbyteorder, nonzero, prod, ptp, put, ravel,`` ``repeat, reshape, round,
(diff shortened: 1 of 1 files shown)


mstimberg:
Hi Mihir,
many thanks for your changes, these extremely long lines definitely need to be taken care of! Your solution is certainly better than the current state, but not optimal for two reasons:
1. The source code is a bit hard to read, with the line wraps not matching the quotation marks
2. The length of the line is still fixed, i.e. depending on window/font size we might still have words that are not visible

I'd say the best solution (although unfortunately quite laborious to do, I guess that's why I did not do it myself in the first place...) is to put each individual word in the quotation marks, i.e. something like:
```
``all``, ``any``, ``argmax``, ...
```
instead of 
```
``all, any, argmax, ...``
```

(Also I just realized that `argmax` is duplicated in that list...)

It'd be great if you could do these changes, but otherwise I'd merge it as it is -- as I said, it's definitely an improvement! Thanks again!

MihirVaidya94:
Hi Marcel,
I was planning to enclose each word in quotation marks. But as you said, it is quite laborious. I did this using a few lines of code now :). To maintain uniformity, I have made the same changes to the remaining list of methods too. Hope this resolves the issue

mstimberg:
Thanks, looking great now!

PR #837, opened 2017-04-27
[MRG] Avoid standalone recompilations

mstimberg:
On Linux, standalone's `main` file was always re-compilated, since it was not mentioned by name (instead there was an `executable` target). Also, the ordering of arrays in `object.cpp` was not completely deterministic: it was sorted by "name", but this was just the name of the variable (e.g. many variables have the name "i"). I changed all this, and also changed the default for `clean` to `False`. I am not 100% about this one, because it can actually lead to compilation errors if you re-run the same simulation but change compilation parameters (or activate OpenMP). I mention this in the error message now. Either way, I'm not seeing any recompilation if I run the same script twice.

So, should stay with `clean=True` by default for safety or switch to `clean=False`? I rarely have compilation times that are longer than 1 or 2 seconds, and adding `clean=False` is not very difficult. But with the new error message, I am happy either way.

Maybe an alternative solution: we keep `clean=True` by default, but when it is activated, we display an `INFO` message at the start of the run, telling the user that they can turn it off.

### DIFF
--- a/brian2/codegen/runtime/cython_rt/templates/spikemonitor.pyx
+++ b/brian2/codegen/runtime/cython_rt/templates/spikemonitor.pyx
@@ -9,7 +9,7 @@
 
     cdef int _num_events = {{_eventspace}}[_num{{_eventspace}}-1]
     cdef int _start_idx, _end_idx, _curlen, _newlen, _j
-    {% for varname, var in record_variables.items() %}
+    {% for varname, var in record_variables | dictsort %}
     cdef {{cpp_dtype(var.dtype)}}[:] _{{varname}}_view
     {% endfor %}
     if _num_events > 0:
@@ -37,7 +37,7 @@
             # Resize the arrays
             _owner.resize(_newlen)
             {{N}} = _newlen
-            {% for varname, var in record_variables.items() %}
+            {% for varname, var in record_variables | dictsort %}
             _{{varname}}_view = {{get_array_name(var, access_data=False)}}.data
             {% endfor %}
             # Copy the values across
@@ -45,7 +45,7 @@
                 _idx = {{_eventspace}}[_j]
                 _vectorisation_idx = _idx
                 {{ vector_code|autoindent }}
-                {% for varname in record_variables %}
+                {% for varname in record_variables | sort %}
                 _{{varname}}_view [_curlen + _j - _start_idx] = _to_record_{{varname}}
                 {% endfor %}
                 {{count}}[_idx - _source_start] += 1
--- a/brian2/codegen/runtime/cython_rt/templates/statemonitor.pyx
+++ b/brian2/codegen/runtime/cython_rt/templates/statemonitor.pyx
@@ -16,7 +16,7 @@
 
     cdef int _i
 
-    {% for varname, var in _recorded_variables.items() %}
+    {% for varname, var in _recorded_variables | dictsort %}
     {% set c_type = cpp_dtype(variables[varname].dtype) %}
     {% set np_type = numpy_dtype(variables[varname].dtype) %}
     # Resize the recorded variable "{{varname}}" and get the (potentially
--- a/brian2/codegen/runtime/weave_rt/templates/spikemonitor.cpp
+++ b/brian2/codegen/runtime/weave_rt/templates/spikemonitor.cpp
@@ -42,7 +42,7 @@
             _owner.mcall("resize", _newlen_tuple);
             // Set N explicitly (it is referenced with a restricted pointer)
             {{N}} = _newlen;
-            {% for varname, var in record_variables.items() %}
+            {% for varname, var in record_variables | dictsort %}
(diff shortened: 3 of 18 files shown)


thesamovar:
How can it lead to compilation errors? Shouldn't gcc or MSVC handle that? If not, can we include (in comments) the relevant compilation parameters in the source so that it triggers a recompile if they change? It would be weird if we had to do that, but I guess we can if it's the only way. I would prefer to put ``clean=False`` by default if we can do so safely.

Another question: if we only change the values of some parameters, does this now run without a recompilation? I guess the answer is yes for changing a parameter that is stored in one of the data files, and no if it's a parameter that gets inserted into the source code.

mstimberg:
> How can it lead to compilation errors? Shouldn't gcc or MSVC handle that?

No, the make utility does a very simple check for every file that it is supposed to build: is the file newer than its dependencies? If yes, don't rebuild it, if no, rebuild it. The compilation flags are not dependencies, so they don't trigger a rebuild. But for this specific problem, there is actually an easy solution: we can add the makefile itself as a dependency to all object files. In a quick test, this seemed to work well.

But I actually realized that there were places which still have non-deterministic ordering, so I'm changing the status of this PR back to "WIP". I also realized that we should probably do the same thing for Cython/weave!

mstimberg:
> Another question: if we only change the values of some parameters, does this now run without a recompilation? I guess the answer is yes for changing a parameter that is stored in one of the data files, and no if it's a parameter that gets inserted into the source code.

Ah, forgot about this. Nothing changed in this regard: it only works without recompilation if the parameter is set with a numpy array and therefore loaded from disk at the start. Each parameter change that is set with a string expression or with a scalar value will lead to changes in `main.cpp` and therefore a re-compilation. Changing an externally defined constant will lead to changes in all files that use this constant. This doesn't actually make much of a difference, since most of time is spent in linking everything together into the single `main` file.
In principle, we could use dynamic instead of static linking, then only the files that are updated (e.g. `main.cpp` for parameter changes) need to be recompiled, and no linking takes place. However, I don't see how we could easily detect the difference between a change that leaves the ABI intact (parameter changes) and one that breaks it (e.g. adding/removing an object). Maybe conditioning everything on `objects.cpp` would work, but I am worried that we could introduce subtle bugs that are hard to debug with this approach...

mstimberg:
Ok, this is now again ready-to-merge from my side, hoping that I did not introduce any new errors. I made the compiled files depend on the makefile, and updated all the `{%for ... %}` loops in the templates (weave and Cython included) to make sure they always generate sorted (and therefore deterministic) code.

thesamovar:
Looks like there might be a problem on WIndows since AppVeyor builds not working.

I think the current way you've done it is great for now, and we might make a few changes that will improve it in terms of recompilations later. I'm wondering about giving externally defined constants names (and internally defined ones like ``N`` the number of neurons) and putting their values in files rather than freezing them as constants inside the source code. We avoided doing that originally because it made it harder for the compiler to optimise the code, but actually with all the loop invariant optimisation stuff we have in there now, I'm not sure that this will make much of a difference. The only time where it might is if either (a) we generate an ``if(cond)`` where ``const bool cond = true;`` using freezing, but I'm not sure this comes up? or (b) we currently know the number of neurons at compile time and the compiler can use this to set up SSE instructions nicely, but actually it can still do it with a runtime loop size and I doubt the efficiency difference would be great.

Anyway, that's all something to do later. Can you look into Windows build and when it's working let me know and I'll do a review (including running it locally) at that time?

mstimberg:
> I'm wondering about giving externally defined constants names (and internally defined ones like N the number of neurons) and putting their values in files rather than freezing them as constants inside the source code.

What I like about our current approach is that the source code is quite readable, and `main.cpp` uses lines like this to set initial values:
```C++
                for(int i=0; i<_num__array_poissongroup_rates; i++)
                {
                    _array_poissongroup_rates[i] = 15.0;
                }
```
It would be a shame if this would instead load a binary file from disk. However, I think a nice solution could be to have all the parameters/constants in a human-readable JSON file. This could also be useful thing for the whole parameter-exploration/fitting idea. The only disadvantage is that big arrays (e.g. predefined connectivity or spike patterns) are very inefficiently represented in an ASCII file. A good solution might be to only have scalar values (or small arrays, say < 100 values) in the ASCII file, and for everything else state a file name with the binary values. All this is a somewhat bigger project, but not *too* difficult either.

> Can you look into Windows build and when it's working let me know and I'll do a review (including running it locally) at that time?

Windows builds should be working again (I inadvertently added an empty line to the make file...), so please go ahead and test/review.

thesamovar:
JSON file idea is nice (and yes to doing that for scalars but not for arrays), but maybe not straightforward to parse without a library in C++?

Will review asap.

mstimberg:
> JSON file idea is nice (and yes to doing that for scalars but not for arrays), but maybe not straightforward to parse without a library in C++?

I think we could use this library: https://github.com/DaveGamble/cJSON  It's a single file under a MIT license, so we could include it in the same way as we do for randomkit.

PR #844, opened 2017-05-22
Complex eigenvalues bug

CharleeSF:
The equations 
```Python
eqs = '''dv/dt      = (ge+gi-(v+49*mV))/(20*ms) : volt
         dge/dt     = -ge/(5*ms) : volt
         dgi/dt     = Dgi/(5*ms) : volt
         dDgi/dt    = ((-2./5) * Dgi - (1./5**2)*gi)/(10*ms) : volt''' 
```
now fail with an error message (if the method is specified).
I guess I also have to add this somewhere as a test case? in the test_stateupdaters.py file?

### DIFF
--- a/brian2/stateupdaters/exact.py
+++ b/brian2/stateupdaters/exact.py
@@ -4,7 +4,7 @@ Exact integration for linear equations.
 import itertools
 
 import sympy as sp
-from sympy import Wild, Symbol
+from sympy import Wild, Symbol, I
 
 from brian2.equations.codestrings import is_constant_over_dt
 from brian2.parsing.sympytools import sympy_to_str, str_to_sympy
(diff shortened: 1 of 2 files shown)


mstimberg:
Thanks for the PR. Indeed, before merging we usually require a test that covers the issue the PR is supposed to fix. `test_stateupdaters.py` would be the right place to put a test, see the [developer docs](http://brian2.readthedocs.io/en/latest/developer/guidelines/testing.html) for more details.

Two other minor hints/suggestions:
* If after commiting, but before pushing to github, you realize that you did some formatting mistakes, etc., you can "amend" the previous commit instead of doing a new one. If you already did several commits and realize you did a mistake in an earlier commit (including for example things like typos in commit messages), you can also do an "interactive rebase" where you can "squash" several commits into one, change commit messages, etc. I'd refrain from doing this after pushing to github, though, since it can make life difficult for others who already cloned your changes.
* For code in issue descriptions/comments, you should use triple backticks to mark the block as a code block, or even better as a Python code block to get syntax highlighting:
````
```Python
for x in range(3):
   do_something()
```
````

CharleeSF:
I added the test case to test_priority in a similar way previous equations are tested. I am not sure if this is how you want it. I can also only test whether linear fails rather than testing all integrators, and/or making a separate test function. I realize deciding these things takes some getting used to.

Thanks a lot for the git-tips!!

mstimberg:
I think is ok to put it there, there are various ways of testing it, and this is certainly a reasonable one. This is all that is necessary to fix the issue, I so I'll merge after the test suite ran through. Thanks again :)

Issue #856, opened 2017-07-11
Exact integration fails with sympy 1.1


mstimberg:
The latest sympy version seems to break the exact integration. The symbolic solution involves a call of `as_explicit` on the sympy matrix. This fails with an error message that seems to be a bug in sympy:
```
...
File ".../sympy/matrices/expressions/matmul.py", line 67, in _entry
    if not X.cols.is_number:
AttributeError: 'int' object has no attribute 'is_number'
```

I don't see any straightforward way to fix this right now, I consider just doing a quick release that requires sympy < 1.1.

thesamovar:
That seems like a reasonable short term solution!

PR #857, opened 2017-07-12
[MRG] Avoid sympy 1.1 for now

mstimberg:
Require sympy < 1.1 as a workaround for #856

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -124,7 +124,7 @@ install:
   - if [[ $MINIMAL_VERSIONS == 'yes' ]]; then
     travis_retry conda create -n travis_conda --yes --quiet pip python=$PYTHON numpy==1.9 scipy==0.14 libgfortran==1 nose sphinx ipython sympy==0.7.6 jinja2==2.7 pyparsing setuptools coverage;
     else
-    travis_retry conda create -n travis_conda --yes --quiet pip python=$PYTHON numpy nose sphinx ipython sympy pyparsing jinja2 setuptools coverage;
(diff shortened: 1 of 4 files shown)


mstimberg:
Sympy got fixed in the meantime (the bug is only in 1.1.0, 1.1.1 has the fix), so I excluded the buggy version explicitly instead of requiring an older version. I also fixed another packaging issue: we were previously requiring an old version of scipy that still ships with weave, but this was no longer compatible with newer versions of numpy (or at least we did not have compatible packages). Instead, I'm now installing the standalone weave package (which is not part of anaconda itself but packages for conda-forge, I copied the packages to our brian-team channel).
I'll wait for the tests to pass and then merge, ok?

BTW: In the longer term we should probably get rid of our brian-specific channel and add Brian to conda-forge, this would also increase its visibility (not that important for Brian itself, but more so if we decide to finally split off `brian2.units` into its own package)

thesamovar:
Yes to all that!

What do we need to do to get Brian into conda-forge? Does it imply any limitations?

mstimberg:
> What do we need to do to get Brian into conda-forge? Does it imply any limitations?

It's quite straightforward, I did the same for a few packages already. The main limitation is that we can only depend on other packages in conda-forge, i.e. we lose the option to depend on specific packages that exist only in our own channel. I don't really think that's an issue anymore, since `py-cpuinfo` was the only package that we shipped ourselves, and I added it to conda-forge a while ago. Having it in conda-forge actually will also simplify our release procedure a bit, since we no longer upload the packages ourselves from travis/appveyor. Only downside I see at the moment: conda-forge does not build linux-32 packages. But then, for our latest Brian2 package, only 4 out of 1125 downloads were for linux-32, so we might as well drop those...

thesamovar:
OK, yeah - let's do that then. Would make the installation even easier.

Issue #481, opened 2015-06-10
Options for numerical integration methods


mstimberg:
Especially for more complex integration methods, we might want to give the user the option to set some parameters for the chosen numerical integration method. For example, if at some point we add support for implicit methods that use fix-point iteration, the user might want to chose the error tolerance for a single step. Also, the option to replace constants by their values (#480) could be configured in this way.

I think the most straight-forward way would be if every class that takes a `method` argument (at the moment: `NeuronGroup` and `Synapses`) would also take a `method_options` keyword that can be passed a dictionary of options. Something like:

``` Python
G = NeuronGroup(N, model='...', method='rk2', method_options={'replace_constants': True})
```

PR #858, opened 2017-07-19
idea of how to get method_options to stateupdatemethod and generator

CharleeSF:
I think getting method_options to the stateupdatemethod and generator could happen in many ways and this is one of them.. not sure if I know when to send it as argument and when as an attribute of the objects involved. Commentary is very welcome! This is in response to issue #481


mstimberg:
I think this PR is now obsolete as this is dealt with as part of PR #870.

PR #864, opened 2017-08-12
[MRG] Documentation on a workaround for a bug that occurs with some version…

thesamovar:
…s of numpy on windows with weave

[ci skip]

### DIFF
--- a/docs_sphinx/introduction/known_issues.rst
+++ b/docs_sphinx/introduction/known_issues.rst
@@ -16,6 +16,13 @@ and modify the line ``msvcr_dbg_success = build_msvcr_library(debug=True)`` to r
 ``msvcr_dbg_success = False`` (you can comment out the existing line and add the new line
 immediately after).
 
+"AttributeError: MSVCCompiler instance has no attribute 'compiler_cxx'"
+-----------------------------------------------------------------------
+
+This may be a bug in some versions of numpy on Windows. The following change may fix the problem.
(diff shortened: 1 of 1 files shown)


mstimberg:
Didn't we have a workaround for this that we removed with b52ec680 because the issue had been fixed in numpy? I'm not sure we should tell users to manually edit library code, how about just telling them to update numpy?

thesamovar:
The commit message says it should have been fixed with numpy 1.10.2 (actually it says 0.10.2 but I assume that means 1.10.2) but I'm getting the problem on numpy 1.11.1.

mstimberg:
> The commit message says it should have been fixed with numpy 1.10.2 (actually it says 0.10.2 but I assume that means 1.10.2) but I'm getting the problem on numpy 1.11.1.

Are you sure you are using 1.11.1? That version does not seem to have the `elif not self.compiler_cxx:` line anymore that you refer to in the docs: https://github.com/numpy/numpy/blob/v1.11.1/numpy/distutils/ccompiler.py

thesamovar:
Well that's really weird because my local copy does but it reports itself as being 1.11.1. Anyhow, upgrading it seems to have solved the problem, so indeed let's go with your suggestion. Will make the change in a moment.

thesamovar:
OK changed the text but forgot to put ``[ci skip]`` in commit message. What do you think of the change?

PR #865, opened 2017-08-15
changing cpp_standalone device's code_object_class method

CharleeSF:
Only thing I had to change was removing referrals to codegen.target, since these will give wrong codeobjects when using the standalone device. After changing them to None the test suite still seems to pass..


brian2/devices/cpp_standalone/device.py
```diff
@@ -481,6 +481,21 @@ def variableview_get_with_expression(self, variableview, code,
                                   'standalone scripts.')
 
     def code_object_class(self, codeobj_class=None, pref=None):
```
mstimberg: This needs to be `fallback_pref` as well.

brian2/devices/cpp_standalone/device.py
```diff
@@ -481,6 +481,21 @@ def variableview_get_with_expression(self, variableview, code,
...
+        ----------
+        codeobj_class : None or a CodeObject class.
+            If this is keyword is set to None or no arguments are given, this method will return
+            the default (`CPPStandaloneCodeObject` class).
+        fallback_pref : str
+            For the cpp_standalone device this option is ignored.
+
+        Returns
+        -------
+        `CodeObject` class
```
mstimberg: The syntax here is a bit wrong, it works like for `Parameters`, i.e. something along the lines of:
```
Returns
-------
codeobj_class : class
    The `CodeObject` class that should be used.
```

brian2/devices/device.py
```diff
@@ -224,14 +224,31 @@ def seed(self, seed=None):
...
+            If this is keyword is set to None or no arguments are given, this method will return
+            the default.
+        fallback_pref : str
+            String describing which attribute of prefs to access to retrieve the 'default' target.
+            Usually this is codegen.target, but in some cases we want to use object-specific targets
+            such as codegen.string_expression_target.
+
+        Returns
+        -------
+        `CodeObject` class
```
mstimberg: See comment above.

brian2/devices/cpp_standalone/device.py
```diff
@@ -481,6 +481,21 @@ def variableview_get_with_expression(self, variableview, code,
                                   'standalone scripts.')
 
     def code_object_class(self, codeobj_class=None, pref=None):
+        '''
+        Return `CodeObject` class (either `CPPStandaloneCodeObject` class or input)
+
+        Parameters
+        ----------
+        codeobj_class : None or a CodeObject class.
```
mstimberg: The standard way to describe optional parameters in numpy doc is to not mention `None`, but instead write:
```
codeobj_class : a `CodeObject` class, optional
```

brian2/devices/cpp_standalone/device.py
```diff
@@ -481,6 +481,21 @@ def variableview_get_with_expression(self, variableview, code,
...
+            If this is keyword is set to None or no arguments are given, this method will return
+            the default (`CPPStandaloneCodeObject` class).
+        fallback_pref : str
```
mstimberg: See remark above:
```
fall_back_pre : str, optional
```

brian2/devices/device.py
```diff
@@ -224,14 +224,31 @@ def seed(self, seed=None):
         '''
         raise NotImplementedError()
 
-    def code_object_class(self, codeobj_class=None, pref='codegen.target'):
+    def code_object_class(self, codeobj_class=None, fallback_pref='codegen.target'):
+        '''
+        Return `CodeObject` class according to input/default settings
+
+        Parameters
```
mstimberg: See comments on the parameters above.

mstimberg:
Everything looks good to me now, I'll do a final more thorough check after lunch.

One more thing: Unfortunately the very first test that travis runs (which is only about building the documentation) currently fails on master due to incompatibilities of the latest version of Sphinx (the package for building the docs) with one of the extensions we are using. This fails for our current master and will therefore fail in your branch as well. We can ignore this test failure for now (if it is the only test failing, of course), I'll fix it in master later.

brian2/devices/cpp_standalone/device.py
```diff
@@ -480,9 +480,28 @@ def variableview_get_with_expression(self, variableview, code,
                                   'variables with string expressions in '
                                   'standalone scripts.')
 
-    def code_object_class(self, codeobj_class=None):
-        # Ignore the requested codeobj_class
-        return CPPStandaloneCodeObject
+    def code_object_class(self, codeobj_class=None, pref=None):
```
mstimberg: This still says `pref` instead of `fallback_pref` -- standalone simulations therefore fail.

PR #875, opened 2017-09-01
[MRG] Minor runtime improvements (mostly for tests)

mstimberg:
Tests are running quite slow, so I looked a bit into the long-running tests and made some improvements to run them faster. The changes I did don't change much for the total test runtime, but I think they are nevertheless worth having. I made a specific change about integration method/time for the Rallpack2 test (i.e., multicompartmental models), but the other changes are more general:
* I switch off the profiling by default. For tests that don't do much each time step (like many of our tests), it added a considerable overhead. Most of our documentation actually suggested that it was off by default, and if you use `profiling_summary` without switching it on, you get an error message telling you what to do.
* I did some micro-optimisations to the run loop, special-casing simulations with a single clock. I'm not sure I hit the best trade off between performance and readability (the fastest solution would be to have one complete run loop for each of the 6 combinations of "single clock?", "profiling?", "progress report?" to avoid any `if` statements within the loop, but I don't feel like this is worth the horrible redundancy in code -- maybe we should actually generate code for the run loop...?).

Well, that's it. As I said, I don't think this will have a noticeable impact on the total test runtime, but for some tests it makes a big difference: the most extreme case is `test_long_run_dt_change()` which runs an empty simulation for 100s -- its runtime went down from ~12s to ~3s on my machine.

### DIFF
--- a/brian2/core/magic.py
+++ b/brian2/core/magic.py
@@ -225,7 +225,7 @@ class MagicNetwork(Network):
         gc.collect()  # Make sure that all unused objects are cleared
 
     def run(self, duration, report=None, report_period=10*second,
-            namespace=None, profile=True, level=0):
+            namespace=None, profile=False, level=0):
         self._update_magic_objects(level=level+1)
         Network.run(self, duration, report=report, report_period=report_period,
                     namespace=namespace, profile=profile, level=level+1)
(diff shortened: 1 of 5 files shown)


thesamovar:
I think we actually discussed code generating the run loop at some point, and decided against it on the basis that it was below the tradeoff threshold.

mstimberg:
> I think we actually discussed code generating the run loop at some point, and decided against it on the basis that it was below the tradeoff threshold.

I guess that's still the case. But just to make sure there's no misunderstanding: I did not consider generating code for cython/weave, the run loop would be a Python-only thing (not using our code generation pipeline etc.).

thesamovar:
Yep, that's how I understood it.

PR #883, opened 2017-09-18
[MRG] Sphinx doc fixes

mstimberg:
This branch fixes a few minor issues in the documentation, most importantly:
* mybinder links are now only shown in the HTML output (i.e. not for the epub or PDF target), which together with some other small changes allows us to build the[ PDF version](http://readthedocs.org/projects/brian2/downloads/pdf/sphinx_doc_fixes/) of the docs again (642 pages... I don't think it is *super*-useful, but some users seemed to make use of it as an offline copy of the documentation).
* It replaces the `sphinxcontrib-issuetracker` extension by a simpler use of the builtin "external links" mechanism. This has the disadvantage that we have to write something like `` :issue:`33` `` instead of simply `#33` when we want issue numbers to link to github, but it removes one dependency and should be more future proof (`sphinxcontrib-issuetracker` does not seem to be maintained anymore, it does not work with recent sphinx releases). 

For easier reviewing, I've enabled builds of the documentation in this branch here: http://brian2.readthedocs.io/en/sphinx_doc_fixes/


tutorials/2-intro-to-brian-synapses.ipynb
```diff
@@ -583,7 +607,9 @@
   {
    "cell_type": "code",
    "execution_count": null,
-   "metadata": {},
+   "metadata": {
+    "collapsed": true
```
thesamovar: Does this have any significance?

mstimberg: It's a new field that recent jupyter notebook versions added, but I'm not sure about what it actually does... Maybe we should use a filter (see e.g. http://timstaley.co.uk/posts/making-git-and-jupyter-notebooks-play-nice/) to make this cleaner in the future?

Issue #878, opened 2017-09-07
Remove "independent" state updater?


mstimberg:
The "independent" state updater can be used for systems of differential equations where all equations are independent of each other, and individual equations are analytically solvable but *not necessarily linear* (e.g. it could be something like `dx/dt = sin(2*pi*t*freq)/ms : 1`).

This state updater has sometimes led to problems in the past due to changes in sympy. But most importantly, I think I have never seen it in actual use except for the carefully crafted examples we use in the test suite. In addition, it has the same naming issue described in #877.

My vote would be to simply remove it (with an error that explains that it has removed instead of just failing with "unknown state updater"), or maybe just mark it as deprecated for now. But then, I honestly don't think anyone uses it.

thesamovar:
Sure, mark it as deprecated and see if anyone complains. If so, we can reconsider.

Issue #877, opened 2017-09-07
Rename 'linear' state updater?


mstimberg:
During the review of the recent paper comparing neural simulators, the authors mentioned that they were somewhat confused that our state updater is called "linear". I never gave this much thought, but there is a certain inconsistency: most of our state updater use as a name the method they use ("Euler", "RK2", etc.), but the "linear" state updater is named after the type of equations it can solve. The authors mentioned above were actually thinking about "linear" as a method, i.e. they thought it would be a linear approximation (or in other words, the forward Euler method).

Following the logic for the names of all other updaters (except the "independent" updater, but I'll open another issue about that one right away), I have to agree that a better name might be "exact" (my preference), or maybe "analytic"?

Of course we'd have to be careful changing this, given that it is in examples all over the place. I would just have it raise a warning first, and change it everywhere in the documentation/examples/tutorials. When/if we finally remove it, it should still give an error message that states what to use instead.

@thesamovar , @romainbrette: thoughts?

thesamovar:
It's logical but I don't think it's important enough to justify a backwards incompatibility. I'd be happy to introduce a new name, but not eliminate the old name. If we do that, I'd be happy with exact but might we ever want to have an exact solver that can handle more complicated equations than linear? Would that be an issue?

mstimberg:
> It's logical but I don't think it's important enough to justify a backwards incompatibility.

Fully agree, we don't want to break scripts using it. For me it would be just a question of adding a new name "silently", or raising a warning if the old name is used.

> If we do that, I'd be happy with exact but might we ever want to have an exact solver that can handle more complicated equations than linear? Would that be an issue?

What kind of equations would that be? If there is a relevant class of such equations, I'd be happy to add a solver for them rather sooner than later (not impossible that we can do this via sympy already). Either way, we'd probably include such a solver in the `exact` solver instead of creating a new one?

thesamovar:
OK, I'd be happy with this then. Indeed, if we ever did extend our exact solver to cover more equations then we could include that with the same name.

PR #890, opened 2017-10-18
[MRG] Rename/deprecate state updaters

mstimberg:
This branch:
* renames the `'linear'` state updater to `'exact'` (#877), but still accepts `'linear'`, without raising a warning (I think we should keep the annoyance to a minimum for such a "cosmetic" change)
* deprecates the `'independent'` state updater (#878), its use now raises a warning (note that the state updater gets never chosen automatically, so that should only affect users that chose it manually -- if there are any...)

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -147,7 +147,7 @@ install:
 script:
 - if [[ $CONDA_BUILD == 'yes' ]]; then
     source deactivate;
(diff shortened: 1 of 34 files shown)


mstimberg:
> Looks great but I think we need to update the tutorials too.

Indeed! Should be done now.

PR #894, opened 2017-10-25
[WIP] Cython types

### LINKED ISSUE #893: Cython: potential compilation failures with fused types
In the generated Cython code, we do not define the types for the temporary variables (e.g. the intermediate result `_v` for the update of the membrane potential `v` before it is written back). Normally, Cython figures out the type by itself just fine (i.e. this is not about a performance issue where Cython generates Python instead of C code). However, the use of fused types (e.g. for the `clip` function) can confuse Cython sufficiently so it raises an error during compilation. The error will complain about "Invalid use of fused types, type cannot be specialized" or the compiler will crash. The latter is actually a bug in Cython, but I think fixing it would just mean to raise the "Invalid use of fused types" error.
We actually get this with the `frompapers/Platkiewicz_Brette_2011.py` example:
```pyconsole
Error compiling Cython file:
------------------------------------------------------------
...
        t = _array_defaultclock_t[0]
        v = _array_neurongroup_v[_idx]
        not_refractory = (t - lastspike) > 0.005
        xi = _lio_1 * _randn(_vectorisation_idx)
        _I = ((_lio_2 * (- I)) + I) + (_lio_3 * xi)
        _theta = (_lio_4 * ((vT + (a * clip(_lio_5 + v, 0.0, _lio_6))) - theta)) + theta
                               ^
------------------------------------------------------------

/home/marcel/.cython/brian_extensions/_cython_magic_ca70be6786b10bc4b580e0b5c5af56e0.pyx:183:32: Compiler crash in AnalyseExpressionsTransform

ModuleNode.body = StatListNode(_cython_magic_ca70be6786b10bc4b580e0b5c5af56e0.pyx:6:0)
StatListNode.stats[15] = StatListNode(_cython_magic_ca70be6786b10bc4b580e0b5c5af56e0.pyx:79:0)
StatListNode.stats[0] = DefNode(_cython_magic_ca70be6786b10bc4b580e0b5c5af56e0.pyx:79:0,
    modifiers = [...]/0,
    name = u'main',
    np_args_idx = [...]/0,
    num_required_args = 1,
    py_wrapper_required = True,
    reqd_kw_flags_cname = '0',
    used = True)
File 'ExprNodes.py', line 10680, in infer_type: AddNode(_cython_magic_ca70be6786b10bc4b580e0b5c5af56e0.pyx:183:81,
    infix = True,
    operator = u'+',
    result_is_used = True,
    use_managed_ref = True)
File 'ExprNodes.py', line 10681, in infer_type: MulNode(_cython_magic_ca70be6786b10bc4b580e0b5c5af56e0.pyx:183:25,
    infix = True,
    operator = u'*',
    result_is_used = True,
    use_managed_ref = True)
File 'ExprNodes.py', line 10680, in infer_type: SubNode(_cython_magic_ca70be6786b10bc4b580e0b5c5af56e0.pyx:183:71,
    infix = True,
    operator = u'-',
    result_is_used = True,
    use_managed_ref = True)
File 'ExprNodes.py', line 10681, in infer_type: AddNode(_cython_magic_ca70be6786b10bc4b580e0b5c5af56e0.pyx:183:32,
    infix = True,
    operator = u'+',
    result_is_used = True,
    use_managed_ref = True)
File 'ExprNodes.py', line 10751, in result_type: AddNode(_cython_magic_ca70be6786b10bc4b580e0b5c5af56e0.pyx:183:32,
    infix = True,
    operator = u'+',
    result_is_used = True,
    use_managed_ref = True)
File 'ExprNodes.py', line 11027, in is_py_operation_types: AddNode(_cython_magic_ca70be6786b10bc4b580e0b5c5af56e0.pyx:183:32,
    infix = True,
    operator = u'+',
    result_is_used = True,
    use_managed_ref = True)
File 'ExprNodes.py', line 10978, in is_py_operation_types: AddNode(_cython_magic_ca70be6786b10bc4b580e0b5c5af56e0.pyx:183:32,
    infix = True,
    operator = u'+',
    result_is_used = True,
    use_managed_ref = True)

Compiler crash traceback from this point on:
  File "/home/marcel/anaconda2/envs/brian2/lib/python2.7/site-packages/Cython/Compiler/ExprNodes.py", line 10978, in is_py_operation_types
    type2.is_unicode_char or
AttributeError: 'NoneType' object has no attribute 'is_unicode_char'
```

A fix seems to be to define the types of all temporary variables in the Cython code, I did this in a branch already, will open a PR shortly.

mstimberg:
Should fix #893 by properly cdef'ing all temporary variables in Cython. See that issue report for more details.

### DIFF
--- a/brian2/codegen/generators/cython_generator.py
+++ b/brian2/codegen/generators/cython_generator.py
@@ -59,6 +59,10 @@ class CythonCodeGenerator(CodeGenerator):
 
     class_name = 'cython'
 
+    def __init__(self, *args, **kwds):
+        self.temporary_vars = set()
+        super(CythonCodeGenerator, self).__init__(*args, **kwds)
+
     def translate_expression(self, expr):
         expr = word_substitute(expr, self.func_name_replacements)
         return CythonNodeRenderer().render_expr(expr, self.variables).strip()
@@ -120,10 +124,9 @@ class CythonCodeGenerator(CodeGenerator):
                                                              index=index)
             lines.append(line)
         # the actual code
-        created_vars = set([])
         for stmt in statements:
-            if stmt.op==':=':
-                created_vars.add(stmt.var)
+            if stmt.op == ':=' and not stmt.var in variables:
+                self.temporary_vars.add((stmt.var, stmt.dtype))
             line = self.translate_statement(stmt)
             if stmt.var in conditional_write_vars:
                 subs = {}
@@ -222,7 +225,7 @@ class CythonCodeGenerator(CodeGenerator):
         support_code = []
         handled_pointers = set()
         user_functions = []
-        for varname, var in self.variables.items():
+        for varname, var in sorted(self.variables.items()):
             if isinstance(var, Variable) and not isinstance(var, (Subexpression, AuxiliaryVariable)):
                 load_namespace.append('_var_{0} = _namespace["_var_{1}"]'.format(varname, varname))
             if isinstance(var, AuxiliaryVariable):
@@ -284,6 +287,12 @@ class CythonCodeGenerator(CodeGenerator):
             else:
                 # fallback to Python object
                 load_namespace.append('{0} = _namespace["{1}"]'.format(varname, varname))
+
+        for varname, dtype in sorted(self.temporary_vars):
+            cpp_dtype = get_cpp_dtype(dtype)
+            line = "cdef {cpp_dtype} {varname}".format(cpp_dtype=cpp_dtype,
+                                                       varname=varname)
+            load_namespace.append(line)
 
         return {'load_namespace': '\n'.join(load_namespace),
                 'support_code_lines': support_code}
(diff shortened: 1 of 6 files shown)


mstimberg:
> This all looks great to me. I was surprised there wasn't already machinery in place so that we didn't need a new is_integer_expression function though?

Yes, we are actually doing similar things in several places... I did a quick-and-dirty implementation that changed the code as little as possible, copying the `is_boolean_expression` approach. But looking at it again we could actually also re-use the `BrianASTRenderer` which annotates the expression with information including the `dtype` -- that would actually also replace `is_boolean_expression`. I'll have a look whether that's feasible or whether there's a problem I overlooked.

Either way, in the long run we should do some serious refactoring of all this, we do parse expressions way too often, extracting the same kind of information several times.

thesamovar:
Sounds good to me.

mstimberg:
So indeed we can simply replace `is_integer_expression` by using `BrianASTRenderer` -- I did this and removed the code for `is_integer_expression`. We could also replace all uses of `is_boolean_expression` in the same way, but I'll leave this for the bigger refactoring.

thesamovar:
OK nice!

PR #895, opened 2017-11-02
[MRG] Re-add stateupdate caching

mstimberg:
Gnarg, I just realized that a small but important line got lost when resolving the merge conflicts between the GSL branch and master: the results of the state update code generation were actually no longer cached ;-( This is particularly annoying as it had the most noticeable effect on the runtime improvements for repeated runs. I think it would therefore be good to do a "hotfix" release as soon as possible, hopefully not too many users already tested the performance and were disappointed :-/


thesamovar:
Agreed!

PR #899, opened 2017-11-23
Added a python script that uses `doc2dash` to build a `Dash.app`-compatible docset for Brian2

DylanMuir:
This script builds a nice docset from the Brian2 docs, which can be used in `Dash.app` or similar clones.

### DIFF
--- a/dev/tools/docs/build_dash_docset.py
+++ b/dev/tools/docs/build_dash_docset.py
@@ -1,0 +1,13 @@
+import brian2
+import sphinx
+import doc2dash
+import sys
+import os
+
+# - Build documentation with Sphinx
+os.chdir('../../../docs_sphinx')
+sphinx.main(['sphinx-build', '-b', 'html', '-d', '_build/doctrees', '.', '_build/html'])
+
+# - Run doc2dash on the built documentation
+os.system('doc2dash _build/html/ -n Brian2 -I index.html -d ..')
+


thesamovar:
Thanks for the PR!

This isn't something I'll ever use, but I'm not against it being included. On the other hand, maybe it makes more sense to be included in another repository? What are your thoughts @mstimberg?

mstimberg:
I think in general `dev/tools/doc` is fine for this kind of script. Does it do more than run `doc2dash` on the standard HTML docs? Maybe the command could instead be stated in the developer docs somewhere (somewhere around here: http://brian2.readthedocs.io/en/stable/developer/guidelines/documentation.html)?

DylanMuir:
The script ensures that brian, sphinx and doc2dash are installed properly (by `import`ing them in python), then runs the commands from the appropriate location. It took me a little bit of playing around to work out how to do it properly, which is why I thought it was worth 'immortalising'. :)

zifeo:
It would be great to have Brian2 as user contributed documentation on Dash!

mstimberg:
Oops, I forgot about this PR... I'll include the script, it does not do any harm and might be useful for some. 

> It would be great to have Brian2 as user contributed documentation on Dash!

I'm not 100% sure what that means, but if you want to provide the Brian2 docs to some contributed-docs repository somewhere, please feel free to do so :)

Issue #900, opened 2017-12-07
Incorrectly connected synapses with numpy target


mstimberg:
The following code does not correctly connect synapses on Brian 2.1.2 when using the numpy target:
```Python
LayerNum = 100
NeuronNum = 3*LayerNum

Neuron = NeuronGroup(NeuronNum, model='')
Link = Synapses(Neuron, Neuron)

Link.connect('i>=LayerNum and i<2*LayerNum and j>=2*LayerNum and j<3*LayerNum')
print('{} new links (start with: {}...)'.format(len(Link),
                                                str([(pre, post)
                                                     for pre, post in zip(Link.i[:10],
                                                                          Link.j[:10])])))
```
With Cython/weave/C++ standalone this does what it is supposed to do:
```
10000 new links (start with: [(100, 200), (100, 201), (100, 202), (100, 203), (100, 204), (100, 205), (100, 206), (100, 207), (100, 208), (100, 209)]...)
```
However, with numpy this results in:
```
90000 new links (start with: [(0, 0), (0, 0), (0, 0), (0, 0), (0, 0), (0, 0), (0, 0), (0, 0), (0, 0), (0, 0)]...)
```
This is a regression, it still worked correctly in 2.0.2.1.

PR #901, opened 2017-12-08
[MRG] Use `&` and `|` instead of `*` and `+` for logical operations in numpy

mstimberg:
In numpy, we were using `*` and `+` to implement logical and/or. This works in general, but can become problematic when the result is used as an index because something like `[0, 0, 1]` is not equivalent to `[False, False, True]`. This could lead to incorrect creation of synapses (see #900).

Why this only manifested recently is unclear, but using arithmetic operators for boolean operations is quite fragile and simply changing the order can change the type:
```pycon
>>> True * False
0

>>> True * False * np.array([True, False])
array([0, 0])

>>> np.array([True, False]) * True * False
array([False, False], dtype=bool)
```

### DIFF
--- a/brian2/parsing/rendering.py
+++ b/brian2/parsing/rendering.py
@@ -145,17 +145,17 @@ class NodeRenderer(object):
         if len(node.comparators)>1:
             raise SyntaxError("Can only handle single comparisons like a<b not a<b<c")
         return self.render_BinOp_parentheses(node.left, node.comparators[0], node.ops[0])
-        
+
     def render_UnaryOp(self, node):
         return '%s %s' % (self.expression_ops[node.op.__class__.__name__],
                           self.render_element_parentheses(node.operand))
-                
+
     def render_Assign(self, node):
         if len(node.targets)>1:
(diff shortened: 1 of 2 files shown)


thesamovar:
I think & and | didn't work in an older version of numpy which is why we didn't use them - I guess the newer version is responsible for it breaking now? Anyway, this looks good to merge to me.

mstimberg:
> I think & and | didn't work in an older version of numpy which is why we didn't use them - I guess the newer version is responsible for it breaking now?

I don't think it's a numpy version issue (the test runs fine with old Brian versions and the most recent numpy), it's a more subtle thing where the order of evaluations can change based on the type of individual variables (`True` vs. `np.bool(True)` or something like that) which is enough to trigger the problem in some cases. But I think you could also come up with examples which fail in earlier Brian versions.

I looked a bit at the discussions back when you wrote this render module, but I did not find a justification for using `*` and `+` (we discussed using `&` vs. `and` in the user code and stuff like that, but not the generated numpy code). I think our handling of boolean variables changed quite a bit since then anyway.

PR #903, opened 2017-12-12
`TimedArray`s that support a temporal offset

DylanMuir:
Added an argument to `TimedArray`, such that an array of values can be shifted in time to begin at a fixed temporal offset. When continuing a simulation from t > 0, one would like to provide a new input signal as a `TimedArray`. Without the temporal offset, the `TimedArray` needs to be padded out to the current simulation time.

### DIFF
--- a/brian2/input/timedarray.py
+++ b/brian2/input/timedarray.py
@@ -33,7 +33,7 @@ def _find_K(group_dt, dt):
 
 class TimedArray(Function, Nameable):
     '''
-    TimedArray(values, dt, name=None)
+    TimedArray(values, dt, name=None, tOffset=0*second)
 
     A function of time built from an array of values. The returned object can
     be used as a function, including in model equations etc. The resulting
@@ -51,6 +51,9 @@ class TimedArray(Function, Nameable):
     name : str, optional
         A unique name for this object, see `Nameable` for details. Defaults
         to ``'_timedarray*'``.
+    tOffset : float, optional
+        A temporal offset at which the data in the TimedArray should begin. Prior to
+        t = tOffset, the first element of the array will be used
 
     Notes
     -----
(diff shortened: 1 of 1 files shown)


mstimberg:
Thanks for the pull request! I did not look at it in detail yet, but let's first make sure that I understand what problem it is solving given that in Brian 2 a `TimedArray` is used as a function of time. With the standard `TimedArray`, you can shift things around in time like this:
```Python
ta = TimedArray(...)
offset = 100*ms
neurons = NeuronGroup(..., '''... ta(t - offset)...''')
```
What kind of model do you have in mind where this would not be enough?

DylanMuir:
Hi @mstimberg, that's true you could modify the equations in that way.

When generating inputs to the network, it would be nice to provide an input displaced in time without having to modify the network namespace. Or if two inputs were provided with different temporal offsets, etc... It's nice to decouple the input signals from the network equations.

mstimberg:
I'm very sorry for not having been responsive on this pull request... There are some issues in the code that needs fixing (as indicated by the failed tests), but before getting to this I want to be sure we really want this feature. When we implemented the `TimedArray` as an input of time, we had situations like offsets or periodic stimulation in mind. We thought that it would be better to represent such details explicitly in the equations instead of adding options to the `TimedArray` initializer because it is 1) more explicit in the model equations which are the core part of the model description and 2) it gives you a lot of flexibility (e.g. the offset could be a global constant, a constant per neuron, a function of something, etc.). So I'm still not really convinced that we need this special case as an additional feature. @thesamovar , @romainbrette : what do you think?

On the other hand, the use case in your initial comment is certainly a valid one and it might be a good idea to make these kind of situations easier to handle. A simple thing I could think of would be to give access to a variable `start_t` that stores the start time of the current run, so that you could use that variable as an offset.

thesamovar:
I'm also not sure we need a special case. Your ``ta(t-offset)`` and ``offset`` is a variable seems to work perfectly well as far as I'm concerned. The ``offset`` could also be a variable of the group, for example.

DylanMuir:
From a model design cleanliness point of view, I think it would be beneficial to decouple the model equations from the input signals.

mstimberg:
Thanks again for your contribution, but I'm closing this pull request for now as I think the core developers agree that this is not a feature we want right now. I'll open an issue about the `start_t` thing, though, it could be interesting to have access from within the equations to the simulation start/end time.

PR #910, opened 2018-01-02
[MRG] Simplify SpatialNeuron

mstimberg:
Big thanks to @moritzaugustin for pointing out that our three arrays `_ab_star`, `_a_minus` and `_a_plus` (the LHS of the three systems of equations that we solve at every time step) were actually identical! So instead of solving three systems (with a lot of redundant code), we now solve only one (for three RHSs). This simplifies the code quite a bit and also makes things a bit faster.

We previously optionally parallelized over these 3 systems with OpenMP. There was a preference and a heuristic to chose between this parallelization and a parallelization over the branches of the morphology (which should be more efficient with more than 3 branches and more then 3 threads). Since the three systems are now basically solved as one, this choice is obsolete and I documented the preference accordingly (+ raise a warning when someone is actually changing its value from `"auto"` -- I doubt that many users did that...).

### DIFF
--- a/brian2/devices/cpp_standalone/device.py
+++ b/brian2/devices/cpp_standalone/device.py
@@ -58,17 +58,9 @@ prefs.register_preferences(
         default=None,
         validator=lambda val: val in [None, 'branches', 'systems'],
         docs='''
-        Which strategy to chose for solving the three tridiagonal systems with
-        OpenMP: `'branches'` means to solve the three systems sequentially, but
-        for all the branches in parallel, `'systems'` means to solve the three
-        systems in parallel, but all the branches within each system
-        sequentially. The `'branches'` approach is usually better for
-        morphologies with many branches and a large number of threads, while the
-        `'systems'` strategy should be better for morphologies with few
-        branches (e.g. cables) and/or simulations with no more than three
-        threads. If not specified (the default), the `'systems'` strategy will
-        be used when using no more than three threads or when the morphology
-        has less than three branches in total.
+        DEPRECATED. Previously used to chose the strategy to parallelize the
+        solution of the three tridiagonal systems for multicompartmental
+        neurons. Now, its value is ignored.
         '''
         ),
     extra_make_args_unix=BrianPreference(
@@ -572,7 +564,12 @@ class CPPStandaloneDevice(Device):
         if nb_threads > 0:
             logger.warn("OpenMP code is not yet well tested, and may be inaccurate.", "openmp", once=True)
             logger.diagnostic("Using OpenMP with %d threads " % nb_threads)
-    
+            if prefs.devices.cpp_standalone.openmp_spatialneuron_strategy != None:
+                logger.warn("The devices.cpp_standalone.openmp_spatialneuron_strategy "
+                            "preference is no longer used and will be removed in "
+                            "future versions of Brian.", "openmp_spatialneuron_strategy",
+                            once=True)
+
     def generate_objects_source(self, writer, arange_arrays, synapses, static_array_specs, networks):
         arr_tmp = CPPStandaloneCodeObject.templater.objects(
                         None, None,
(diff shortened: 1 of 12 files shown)


thesamovar:
This blows my mind! I'm happy to do the review, but maybe it would be better to have @romainbrette do it since you two wrote the multicompartmental code together and I'm not so familiar with it?

mstimberg:
I know that you did not work much on the multicompartmental stuff, but I'm not sure whether Romain will have time to review the code anytime soon. It is actually rather straightforward to see that it does the same thing as before (which makes it even more embarrassing that I never realized it...): if you look at the `spatialneuron_prepare` templates, you see that `_a_plus0`, `_a_minus0`, etc. simply copy over the values from `_ab_star0`, etc. and apply the same "corrections ". These values are only calculated once at the beginning of the simulation, they never change. The code in the `spatialstateupdate` has three loops that are three identical copies apart from the variable changes -- if you replace `_a_plus0` by `_ab_star0`, etc. you see that the only difference that remains are the lines that set `_v_star`/`_u_plus`/`_u_minus`.

After all, all the tests (including the long ones) pass and the examples I checked calculate the exact same result (with `-ffast-math` there can be small differences, though).

moritzaugustin:
once these improvements (in particular those of the cpp_standalone templates) are confirmed I could start updating the cuda_standalone version of the spatialstateupdate template, so let me know (no need to hurry from my side though) 

[anyway, I guess I will be informed by github automatically when the pull request will be accepted ;)]

brian2/devices/cpp_standalone/device.py
```diff
@@ -572,7 +564,12 @@ def check_openmp_compatible(self, nb_threads):
         if nb_threads > 0:
             logger.warn("OpenMP code is not yet well tested, and may be inaccurate.", "openmp", once=True)
             logger.diagnostic("Using OpenMP with %d threads " % nb_threads)
-    
+            if prefs.devices.cpp_standalone.openmp_spatialneuron_strategy != None:
```
thesamovar: This should probably be ``is not`` rather than ``!=``?

brian2/devices/cpp_standalone/templates/spatialneuron_prepare.cpp
```diff
@@ -1,7 +1,5 @@
-{# USES_VARIABLES { N, _invr, Ri, Cm, dt, area, r_length_1, r_length_2
+{# USES_VARIABLES { N, _invr, Ri, Cm, dt, area, r_length_1, r_length_2,
```
thesamovar: Unrelated bug fix?

mstimberg:
> Code looks fine, I still feel like I can't really judge the maths without spending more time than I have... However if two of you have already looked it over carefully maybe that's OK?

I am not too worried -- @moritzaugustin and I have looked at it in quite some detail and I talked about it at a group meeting where @romainbrette did not see anything wrong with it either. But what convinces me most is the empirical evidence: before the change `_ab_star`, `_a_minus` and `_a_plus` were bit-by-bit identical, and all the tests and examples give the same results (with some small floating point discrepancies). I'll do a final check with `-O0`, this should give us bitwise identical results for the examples/tests.

brian2/devices/cpp_standalone/templates/spatialneuron_prepare.cpp
mstimberg: It's actually only a cosmetic fix, our parsing does not require a comma at the end of the line.

brian2/devices/cpp_standalone/device.py
mstimberg: Agreed.

mstimberg:
> I'll do a final check with -O0, this should give us bitwise identical results for the examples/tests.

Just for the record: I ran a comparison for the `bipolar_with_inputs` example and the `rall2` test (with a highly branched tree), and all values before and after the change are identical for all the code generation targets. There's a small (up to 1e-8mV) numerical difference between numpy and the other code generation targets (in master as well as in this branch), but this is not surprising because the algorithms used for solving the banded linear system are not exactly the same.

thesamovar:
That all seems fine to me!

PR #913, opened 2018-01-25
[MRG] Fix numpy 1.14 compatibility issues

mstimberg:
This PR fixes three issues, the first two directly about the [recent numpy 1.14 release](https://github.com/numpy/numpy/releases/tag/v1.14.0) (~3 weeks ago) and the last a bit more general:
* Our use of `np.issubdtype` raised warnings because of a potential future incompatibility (if I understood correctly that incompatibility does not concern us, but the warning was annoying either way): 
* numpy's print output improved (less unnecessary spaces, etc.), but this of course makes almost all of our doctests fail. For the doctests, I therefore enabled the "legacy mode". Note that the tests only work correctly when run with `brian2.test()` or with one of our scripts, running something like `nosetests --with-doctests brian2` manually will fail (with numpy 1.14).
* Switching between numpy versions (as I did for testing these changes) can lead to runtime or link errors when weave/Cython reuse code compiled against a different numpy version. I now embed the ABI version directly in the code comment where we also put compiler options etc. (for weave), respectively consider the ABI version part of the key that we use for hashing (Cython). Unfortunately, I did not find an elegant way to ask for this ABI version from Python, so I directly read it out from numpy's header file... If this is too hacky, an alternative would be to embed the full numpy version, even though it would mean to recompile everything (e.g. for tests) after each minor numpy update.

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -101,7 +101,9 @@ install:
   - 'python -c "import struct; print(struct.calcsize(''P'') * 8)"'
 
   # Install the build dependencies of the project via conda
-  - 'appveyor-retry conda install --yes --quiet numpy nose "sphinx<1.6.3" "sympy!=1.1.0" pyparsing jinja2 ipython setuptools cython'
+  - 'appveyor-retry conda install --yes --quiet numpy nose "sphinx<1.6.3" "sympy!=1.1.0" pyparsing jinja2 ipython setuptools cython "conda==4.3.21"'
+  # Pin conda version
+  - 'echo conda ==4.3.21 >> %PYTHON%\conda-meta\pinned'
   # Install an older version of scipy (which still has weave) on Python 2
   - 'if "%PYTHON_VERSION:~0,1%" == "2" appveyor-retry conda install --yes --quiet -c brian-team weave scipy'
   - 'if "%PYTHON_VERSION:~0,1%" == "3" appveyor-retry conda install --yes --quiet scipy'
--- a/brian2/codegen/generators/base.py
+++ b/brian2/codegen/generators/base.py
@@ -20,24 +20,6 @@ __all__ = ['CodeGenerator']
 logger = get_logger(__name__)
 
 
-def get_numpy_ABI_version():
-    # I don't see a way to get numpy's ABI version from Python,
-    # so we look it up in the numpy header file that we are using
-    # later...
-    header_name = os.path.join(numpy.get_include(), 'numpy', '_numpyconfig.h')
-    if os.path.exists(header_name):
-        with open(header_name, 'r') as f:
-            for line in f.readlines():
-                if line.startswith('#define NPY_ABI_VERSION'):
-                    elements = line.split(' ')
-                    if len(elements) == 3 and elements[2].startswith('0x'):
-                        return elements[2]
-
-    # Something went wrong, we don't want to raise an error because
-    # it might be that we'll never actually use this information
-    return 'Unknown'
-
-
 class CodeGenerator(object):
     '''
     Base class for all languages.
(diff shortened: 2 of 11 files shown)


thesamovar:
The print output thing is kind of annoying, but as long as everything works on our own test suite I think it's fine.

I would have no objection to including full numpy version rather than ABI if that seems safer. To be honest, I think people update their numpy fairly infrequently so this is only a minor inconvenience. Does ABI only ever bump on point updates (i.e. x.y not x.y.z)? If so, we could just use the first two numbers from the version string?

mstimberg:
> The print output thing is kind of annoying, but as long as everything works on our own test suite I think it's fine.

I think they handled it as well as they possibly could have, giving us a single global option. We could also globally activate this in Brian itself, but it would be a bit weird if users would get "old-style" output from numpy whenever they use Brian... I think we can potentially do something to switch it on automatically regardless of the way the tests are called, but I'd leave this at least until we switched to pytests.

> I would have no objection to including full numpy version rather than ABI if that seems safer. To be honest, I think people update their numpy fairly infrequently so this is only a minor inconvenience. Does ABI only ever bump on point updates (i.e. x.y not x.y.z)? If so, we could just use the first two numbers from the version string?

Actually now that I looked into it in a bit more detail, I realized that my understanding of this was too basic. It turns out that the ABI version did not change since version 1.5 or so, so embedding it into the code does not actually help that much against the errors I was seeing! I think the problem is that it is fine do use an old extension (or weave code, etc.) that has been compiled with an *older* version of numpy (except if the ABI version changed, but it did not do so for many years), but this does not necessarily means that it is fine to use an extension that has been compiled against a *newer* version of numpy with an older version. So, using the ABI version is not actually good enough for us (updates are of course much more common than downgrades, but if you switch between environments, you may well switch to an older numpy version). So there's the C API version, but this mostly correponds to the x.y release scheme, as you said (1.11 and 1.12 did not bump the API version). Given all that, I think just including numpy's version is actually the best, I don't mind that much whether it is x.y or the full x.y.z version -- after all it is a cache, nothing bad happens if we compile unnecessarily (except for filling up the cache directory, I think at some point we need to establish a mechanism to clear old files...). Maybe let's go for the x.y version as a good compromise?

But all that also made another thing clearer to me (no need to deal with it in this PR): we are currently building one conda brian2 package for every numpy version, establishing an unnecessarily strict link between brian2 and numpy. We could instead build our packages for the oldest supported numpy version, it should work fine with later versions (as long as they stay ABI compatible). But then the fact that our conda package does not work with numpy 1.14 also means that users did not run into the problems fixed with this PR...

thesamovar:
Agree on printing and doctests.

Also happy with using x.y for the moment. I'd say investigate whether a simpler build strategy would work for the future, but let's make that a separate issue and PR?

I also agree that a way to limit cache size would be great, but as far as I can tell there's no way to do that with Weave. For Cython, we've written our own cache so we could have control over this. Maybe open another issues for these points?

mstimberg:
> Also happy with using x.y for the moment.

Ok, I updated the PR.

> I'd say investigate whether a simpler build strategy would work for the future, but let's make that a separate issue and PR?

Agreed.

> I also agree that a way to limit cache size would be great, but as far as I can tell there's no way to do that with Weave. For Cython, we've written our own cache so we could have control over this. Maybe open another issues for these points?

I will -- I think we don't need to care that much about weave, looking at our anaconda download stats it seems that finally more users are using Python 3 instead of Python 2 (especially under Linux). Which reminds me that we should make our code Python3-compatible without the 2to3 crutch...

thesamovar:
Indeed I wonder if we should go Python 3 only when numpy Python 2 support ends?

Will review this early next week. Currently writing a lecture on compilers. :(

mstimberg:
> Indeed I wonder if we should go Python 3 only when numpy Python 2 support ends?

Possibly. I think the most important thing for the near future is that we switch our development model to "Python 3"-first. Then, we I don't think there's any harm in continuing to support Python 2 until we actually have a reasonable use case that does not work in Python 2 anymore  (i.e. not just convenience, but something like the type annotations) .

Issue #914, opened 2018-01-25
Reduce the size of the Cython cache


mstimberg:
After running a lot of code with various versions of Brian, in particular the test suite, the Cython cache directory fills up with a lot of files. While disk space is not a major issue in general, a couple of GBs of cache can be annoying, especially on relatively small SSDs.

There's one relatively easy thing that we could to improve things: Currently, we write a `.pyx` file to disk, transform it into a `.cpp` and then compile it to a `.o` and `.so` file (named differently on Windows, I guess). All we need later is the `.so` file -- this is the module we import. For debugging purposes it can sometimes be useful to have the actual source files around, but it could be made optional via a preference. Deleting everything we don't need would make quite a difference. For example, when I run the `CUBA.py` example with Cython, it creates 10 code objects. With all the huge files that Cython tends to create, this already takes ~18MB of space. When I delete everything except for the `.so` files that are needed to avoid later recompilation, then we are down to ~4M, i.e. quite a significant decrease.

A more complex problem is to keep this cache from continuously growing. We can't run any kind of regular "cleanup code", but we could delete old code whenever we create new code, depending on some criterion about the size of the cache. Checking the actual disk size would be too costly, but just limiting the number of cached code objects might be enough. Unfortunately it would mean to also keep track of the use of the code (we don't want to delete a code just because it was created a long time ago, it should have been *used* a long time ago). A single text file or similar would be good enough for this, but it needs some serious thought to avoid the errors that weave made (e.g. stuff got slower with the number of code objects that were cached, updating the database was not threadsafe). An easier solution to doing this "online" might be to have a warning after a certain number of cache entries (e.g. checked at import time) and to provide an explicit "offline" command to clear old elements from the cache.

thesamovar:
I like that last idea - I think that would be the best solution.

mstimberg:
> I like that last idea - I think that would be the best solution.

I agree, a manual offline cleanup would be more in the spirit of Brian and is less difficult to implement (no need to worry about multithreading, etc.). We could still keep track of the use of cache elements to allow the user to do a fancy "delete everything but the last 1000 entries", but this would still be tricky do do get right (again, multhreading, etc.) I tend towards a simple solution, something where we don't track usage at all, and just check the number of elements in the cache directory at import time (configurable via a preference) + a warning if the number exceeds some threshold. For convenience, we can provide a `clear_cache()` command, but it would just delete everything from the respective directory.

PR #920, opened 2018-02-27
[MRG] Reduce codegen cache size

mstimberg:
As discussed in #914, this PR improves Cython's and (to a lesser extent) weave's code generation cache:
1. After successful compilation of a library file for Cython we now delete the corresponding `.cpp` and `.pyx` files, plus any additional temporary library files (I still need to check whether this is working correctly on Windows as well). The library file (e.g. `.so` on Linux) is the only file that we'll ever need later. When I run the Cython test suite with an empty cache, I now get 166MB of cached files instead of 774MB before the change. Quite an improvement! I did not see any noticeable slowdown, and the deletion can be switched off with a preference for debugging (`codegen.runtime.cython.delete_source_files`).
2. When importing Brian 2, if the Cython or weave cache directory take up more than 1GB (`codegen.max_cache_dir_size` preference), the user gets an INFO message like this:
```
INFO       Cache size for target "cython": 1584 MB.
You can call "clear_cache('cython')" to delete all files from the cache or manually
delete files in the "/home/marcel/.cython/brian_extensions" directory. [brian2]
```
3. As the message suggests, `clear_cache('cython')` or `clear_cache('weave')` can be used to delete the respective cache directory.

[Unrelated to the changes here, I had to bump up the numpy dependency to version 1.10, we were actually already using a numpy function in one place that did not exist in numpy 1.9. **EDIT**: I pushed this change directly to master, as tests on all branches were failing because of a related issue]

### DIFF
--- a/brian2/__init__.py
+++ b/brian2/__init__.py
@@ -97,3 +97,14 @@ for name, version in [('numpy',  '1.10'),
 
 # Initialize the logging system
 BrianLogger.initialize()
+
+
+# Check the caches
+if prefs.codegen.max_cache_dir_size > 0:
+    from brian2.codegen import check_cache
+    from brian2.codegen.runtime.weave_rt.weave_rt import get_weave_cache_dir as _get_weave_cache_dir
+    from brian2.codegen.runtime.cython_rt.extension_manager import get_cython_cache_dir as _get_cython_cache_dir
+    for target, dir in [('weave', _get_weave_cache_dir()),
+                        ('cython', _get_cython_cache_dir())]:
+        if dir is not None:
+            check_cache(target, dir)
--- a/brian2/__init__.py
+++ b/brian2/__init__.py
@@ -120,7 +120,12 @@ def check_cache(target):
     size = _get_size_recursively(cache_dir)
     size_in_mb = int(round(size/1024./1024.))
     if size_in_mb > prefs.codegen.max_cache_dir_size:
-        logger.warn('Cache size for target "%s": %s MB' % (target, size_in_mb))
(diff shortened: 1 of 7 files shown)


brian2/__init__.py
```diff
@@ -96,4 +99,51 @@ def _check_dependency_version(name, version):
...
+        logger.debug('Cache size for target "%s": %s MB' % (target, size_in_mb))
+
+
+def clear_cache(target):
+    cache_dir = _cache_dirs.get(target, None)
+    if cache_dir is None:
+        raise ValueError('No cache directory registered for target "%s".' % target)
+    logger.debug('Clearing cache for target "%s" (directory "%s").' %
+                 (target, cache_dir))
+    shutil.rmtree(cache_dir)
```
thesamovar: Are we absolutely 100% sure that ``cache_dir`` could never be something it shouldn't be?! :)

mstimberg: Good point. Maybe we should ask the user for confirmation after displaying the name of the directory? Checking for file types sounds like a good idea as well.

mstimberg:
I added a check in `clear_cache` that all the files in the directory have an expected extension as you suggested. It bails out with an error otherwise (`IOError`, not sure that's a great choice but I did not immediately find a better fit). I also added a test that checks that it refuses to delete the directory when a file is present that should not be there. There's no test that clearing the cache actually works because this is not so easy to test: we run tests in parallel and if one test deletes the cache directory then the other tests might not be happy about it...

I think it should be safe enough like this, asking the user for explicit confirmation is maybe too much -- after all, they need to explicitly call `clear_cache`.

thesamovar:
Yep - I'm happy with that!

Issue #922, opened 2018-03-06
run_regularly() not working for subgroups


AI-pha:
Hello,

In #766 (May 23, 2017) run_regulary on subgroups was called a "legitimate use case", so I think this is expected to work. However, in the following minimal example, the state variable is not updated by run_regularly. (It does not raise an error and I did not find anything in the documentation concerning run_regulary for subgroups).

Here is the code that reproduces the issue (v just stays at 1, instead of rising by 1 every 10 ms):
```python
from brian2 import NeuronGroup, StateMonitor, run, ms, figure, plot, show

neurons = NeuronGroup(10, model="v:1")
neurons[0:5].v = 1
neurons[0:5].run_regularly("v += 1", dt=10 * ms)

statemon = StateMonitor(neurons, variables=["v"], record=True)

run(100 * ms)

print(neurons[0:5].v)

fig = figure()
plot(statemon.t / ms, statemon.v.T)
show()
```

Thanks a lot for having a look at this issue.

mstimberg:
Hi, many thanks for the report, this in indeed a bug, your code should work as you intended. I had a quick look at it and I know what's going wrong, it will be quite easy to fix.
I'll give some more details on what is happening later, but fortunately there's a quite simple workaround: store the `StateUpdater` object that `run_regularly` returns in a variable to make the `run` function aware of it (if you explicitly construct a `Network` object, add the returned object via `Network.add`). I.e., instead of:
```Python
neurons[0:5].run_regularly("v += 1", dt=10 * ms)
```
use:
```Python
v_updater = neurons[0:5].run_regularly("v += 1", dt=10 * ms)
```

AI-pha:
Thanks for the very quick reply.

When I use:
```python
v_updater = neurons[0:5].run_regularly("v += 1", dt=10 * ms) 
```
I get
`An error occurred when preparing an object. ReferenceError: weakly-referenced object no longer exists
`
This also happens, when I add the 'v_updater' to a Network.


It does however work fine (both for magic and Network), when a subgroup is explicitly created like that:
```python
subgr = neurons[0:5]
v_updater = subgr.run_regularly("v += 1", dt=10 * ms) 
```

It also works without the 'v_updater', if the subgroup is explicitly created and manually added to a Network (which is anyway my preferred way to do it).

mstimberg:
You are right, sorry for the half-working workaround.

[...]

PR #924, opened 2018-03-07
[MRG] Properly support `run_regularly` on subgroups

mstimberg:
See #922

### DIFF
--- a/brian2/groups/group.py
+++ b/brian2/groups/group.py
@@ -13,7 +13,7 @@ import inspect
 
 import numpy as np
 
-from brian2.core.base import BrianObject
+from brian2.core.base import BrianObject, weakproxy_with_fallback
 from brian2.core.names import Nameable
 from brian2.core.preferences import prefs
 from brian2.core.variables import (Variables, Constant, Variable,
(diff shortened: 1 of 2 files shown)


brian2/groups/group.py
```diff
@@ -940,6 +940,20 @@ def run_regularly(self, code, dt=None, clock=None, when='start',
...
+        # include this operation automatically, i.e. with the parent group
+        # (adding just the run_regularly operation to the parent group's
+        # contained objects would no be enough, since the CodeObject needs a
+        # reference to the group providing the context for the operation, i.e.
+        # the subgroup instead of the parent group. See github issue #922
+        source_group = getattr(self, 'source', None)
+        if source_group is not None:
+            if not self in source_group.contained_objects:
+                source_group.contained_objects.append(self)
+
```
thesamovar: This feels a bit hacky and prone to reintroduce the same error in future. What about we just always add subgroups to ``contained_objects`` at the point of creating them, but add a new attribute/mixin class ``object_container_only`` or ``ObjectContainerOnly`` that tells the run scheduler to only include the contained objects of this object, but not the object itself?

thesamovar: If we add a weakref to ``contained_objects`` that could work? We could also cull weakrefs that no longer point to anything occasionally?

mstimberg: The problem with using weakrefs is that it does not work for code like the following (see #922):
```Python
neurons[0:5].run_regularly("v += 1", dt=10 * ms)
```

thesamovar: Another idea: when we create a subgroup, we set the ``contained_objects`` list to be the source ``NeuronGroup.contained_object`` list?

mstimberg: This is something I tried but it does not quite work: code objects like the `CodeRunner` that is created by `run_regularly` keep a reference to their "owner". However, to avoid circular dependencies, we store these only as weak references. If the `run_regularly` only adds itself to the `contained_object` list of the parent `NeuronGroup`, then the `Subgroup` will be garbage collected and we'll get a "weakly-referenced object no longer exists" during code generation.

thesamovar: OK so how about: ``run_regularly`` has a full ref to ``Subgroup``, but ``run_regularly`` is inserted into the parent ``NeuronGroup.contained_objects`` (possibly by having ``Subgroup.contained_objects`` be a computed attribute that picks out ``parent.contained_objects``). Would that work? Feels like that keeps the needed objects alive, but avoids circular references?

mstimberg: In principle this should work, I think. But it would mean that we have to make a difference between a subgroup (strong reference) and a standard group (weak reference) -- not sure that's cleaner than the solution in this PR? It would also mean we'd have to change things in `CodeRunner` which is widely used (there's no "run_regularly class").

Issue #906, opened 2017-12-20
Synaptic variables are incorrectly initialized for connects after store/restore


mstimberg:
If synapses are connected after a `restore()`, their values are not initialized as they should, instead they take the values they had before the restore (initially reported on the [mailing list](https://groups.google.com/d/msg/briansupport/U82m2Zln7A8/NN3xhT6eAgAJ)).
A simple example:
```Python
from brian2 import *

G = NeuronGroup(1, '')
S = Synapses(G, G, 'w: 1')
store()

S.connect(True)
S.w = 1

restore()

S.connect(True)
print(S.w[:])  # Should be 0
```
This will print `[ 1. ]`, instead of `[ 0. ]`, i.e. the value of `w` is still remembered from before `restore()`. This is also the case for internally used variables such as `lastupdate` which will lead to incorrect integration for event-driven synaptic variables.

mstimberg:
It's actually a more general `DynamicArray` issue: we guarantee that such an array is initialized to 0, but this will not be the case if we shrink (not with `shrink` but with `resize`) and regrow it -- the temporarily "invisible" memory will keep its previous content.

I prepared a PR that will initialize the "to be forgotten" memory during shrinking -- I think this is a rare event (except for the `store`/`restore` mechanism, I don't think we are ever doing it) and therefore should not affect performance notably (potentially less than initializing memory when resizing over previously used memory).

PR #928, opened 2018-03-15
Avoid that the caching system keeps references to the numerical values of TimedArrays alive

### LINKED ISSUE #923: Deleting `TimedArray` objects does not free memory
Code like this will leak memory, i.e. memory usage continually increases:
```Python
from brian2 import *

while True:
    ta = TimedArray(np.random.randn(10000, 1000), dt=defaultclock.dt)
    g = NeuronGroup(1000, 'dv/dt = (-v + ta(t, i))/(10*ms) : 1', method='exact')
    mon = StateMonitor(g, 'v', record=True)
    run(10000*defaultclock.dt)
    print(np.mean(mon.v))
```
Originally reported by Wilhelm Braun on the mailing list: https://groups.google.com/d/msg/briansupport/VyOTZR0HaMs/ujKWsiDLAwAJ

mstimberg:
It turned out that the reason for the memory leak seen in #923 was the caching system. `TimedArray` objects are `Function` objects, and their information (e.g. dimensionality) is used during the code generation process. This kept references to the data they used alive, even if the `TimedArray` object was deleted. Fortunately the solution is quite simple, we can exclude the attributes that refer to the data from the "state tuple" that is used for the cache with the same mechanism we use for `ArrayVariable` objects.

Fixes #923


brian2/input/timedarray.py
```diff
@@ -82,6 +83,8 @@ class TimedArray(Function, Nameable):
      [ 1.  3.]
      [ 2.  4.]] mV
     '''
+    _cache_irrelevant_attributes = {'_id', 'values', 'pyfunc',
+                                    'implementations'}
```
thesamovar: This line could be a little risky if we decide to modify cacheing to be persistent across sessions? I mean specifically ``implementations`` (although maybe others). If this is specific to this case, we should consider changing, but if we're universally doing this then adding this one is no problem.

mstimberg:
Update: the problem did not fix itself with the merge of #933. However, when I ran my test script with all the recent changes, it turned out that it now runs massively *slower* than before... Bisection shows that the problem was introduced with commit 9c99e8 (fix for #906) -- I'll have a closer look.

mstimberg:
Ok, this was one of the most annoying multi-day debugging sessions I did in a while :disappointed:...

The error I was seeing did not really have to do with the caching, but the previous caching (that kept `TimedArray` instances alive and led to the memory leak) actually prevented it from happening in the tests... The problem appeared only on OS X (and could disappear when I added debug output in certain places, making the debugging super annoying...) so I had to debug on travis. What I saw was that `TimedArray`s were sometimes giving back wrong values. After a lot of trial-and-error I found that the issue manifested itself like the following. We had two tests that looked like this (one had units in the `TimedArray`, the other not, but that doesn't matter here):
```Python
    ta = TimedArray(np.arange(10), dt=0.1*ms)
    G = NeuronGroup(1, 'value = ta(t) + 1: 1', dt=0.1*ms)
    mon = StateMonitor(G, 'value', record=True, dt=0.1*ms)
    run(1.1*ms)
```
In this case, only the `StateMonitor` is running any code, and that code for weave looks like this (highly simplified of course):
```C++
  static double* _namespace_timedarray_values;
  inline double _timedarray(const double t)
  {
      ...
      return _namespace_timedarray_values[i];
  }

static PyObject* compiled_func(PyObject*self, PyObject* args) {
  ...
  _namespace_timedarray_values = _timedarray_values;
  ...
  const double _lio_1 = _timedarray(__source_value_neurongroup_t) + 1;
  ... // record stuff in StateMonitor
}
```
The way the namespace is handled is a bit un-intuitive, but weave only deals with the namespace within the `compiled_func` so we have to get it somehow out to the `_timedarray` function. Now, with plenty of debugging output it turned out that when things went wrong, `_namespace_timedarray_values` updated in `compiled_func` and `_namespace_timedarray_values` in `_timedarray`did not point to the same address in memory! Instead, the pointer in `_timedarray` pointed to the global variable that was used in a *previous* test... (see this test run, for example: https://travis-ci.org/brian-team/brian2/builds/369196230)
This is really confusing, because no actual linking between files takes place, all the code is in a single file...

To make a long story short, I think the problem is that we are using `inline` for the `_timedarray` function -- this is supposedly only a hint to the compiler, but it turns out that clang uses by default C99 semantics, where having just an `inline` function is actually incorrect (http://clang.llvm.org/compatibility.html#inline):
> In C99, inline means that a function's definition is provided only for inlining, and that there is another definition (without inline) somewhere else in the program.

According to the documentation, this should rather lead to some "undefined symbol" issues, but well, it seems it screwed up things in our tests. So now, I define the function rather as "static inline", which is supposedly the right thing to do (we could simply leave away "inline" as well -- the compiler will decide about that itself, anyway).

I sincerely hope that this fixes it once and for all (in my debugging branch it did, but I did not yet run the full test suite on OS X)!

thesamovar:
That's weird, that page says that it only applies to C not C++, and I thought Weave generated C++. Certainly we use C++ in some of our templates. In any case, if static inline works, great!

mstimberg:
I don't claim that I really understand this bug, I'm just happy that it is working now :smile: I think it is reasonable that the `inline` had something to do with it, but it might not be the full story. Anyway, merging this thing.

Issue #927, opened 2018-03-13
User defined `delay` variable in `Synapses` object creates ambigiuity


denisalevi:
The `delay` variable (as shortcut for the pre pathways delay) is accessible in the `Synapses` objects equations (`model`, `on_event` etc). But `delay` can also be defined as a synaptic variable, which create ambiguity about which variable is used when.

Quick example I was just running. When using `delay` in a differential equation like this
```python
Synapses(group, group, model='''                          
                dA/dt = delay/ms : second                        
                ''',
                delay=2*ms)
```
`delay=2*ms` is used for the state update of `A`. But when defining another `delay` variable, e.g. like this
```python
Synapses(group, group, model='''                          
                dA/dt = delay/ms : second
                delay = 4*ms : second                  
                ''',
                delay=2*ms)
```
`delay = 4*ms` is used in the state update of `A`.

I guess banning `delay` as synaptic variables as done for `lastupdate` would solve this. Or removing the accessibility to the synaptic delay in synapse code since other pathways delay variables (e.g. `post.delay`) are not accessible in synapse code either.

mstimberg:
> I guess banning delay as synaptic variables as done for lastupdate would solve this. 

Uh, indeed -- I'm actually surprised that this was not already the case... :unamused: 

> Or removing the accessibility to the synaptic delay in synapse code since other pathways delay variables (e.g. post.delay) are not accessible in synapse code either.

This is not that trivial, unfortunately. We add a reference to the pre pathway's delay to the `Synapses.variables` dictionary. The `variables` dictionary has two uses: it determines what you can access as attributes of the `Synapses` object, and it determines the variables you can access in the code. For `delay`, we are only interested in the former (to allow the convenience of `syn.delay = ...` instead of `syn.pre.delay = ...`) but it comes necessarily with the latter.

That said, I think we should fix both things, i.e.: no longer allow to declare variables called "delay", and change the mechanism so that the `delay` of the pre-synaptic pathway is not exposed via `variables` but in a more direct way (`Synapses.delay` is a property, delegating things to the pre-synaptic pathway).

Issue #355, opened 2014-10-31
Allow time varying delays in Synapses as an option


thesamovar:

phreeza:
Are there any plans on integrating this feature? I think brian 1 had it, right? If so, would it be hard porting it from there? I would be interested in doing it.

mstimberg:
Yes, we are certainly interested in this (just ignore that this issue was already opened 2 1/2 years ago ;) ). The feature was in Brian 1 as an experimental feature, but I don't think we necessarily have to port it from there. The basic change is relatively obvious, the difficulty comes from getting the details right.

The core reason why delays are not allowed to change during the run is that at the start of a run, we transform all delays into "offsets", i.e. simply integers representing the delay in timesteps (which is used to insert arriving spikes into the `SpikeQueue`): https://github.com/brian-team/brian2/blob/master/brian2/synapses/cspikequeue.cpp#L90

The easiest way to implement support for changing delays would be to make this preparation optional and only do it when the user uses constant delays, otherwise we transform the delay into an integer value for each spike. It would probably a good idea to profile this a bit and see whether it makes a meaningful speed difference (by comparing two simulations where delays are constant, but in one case, they are marked as variable). If it does not, we could also *never* pre-calculate integer delays. On the other hand, if it makes a big difference, we might e.g. keep a look-up table to replace a relatively  costly division/rounding operation by a look-up.

Then, there are a few syntax questions: should it be an explicit choice by the user to allow the use of variable delays, or do we switch automatically if there is any code that modifies the delays? I'd probably vote for the former, but then again if we decide that pre-calculation does not save us anything, we could simply always allow variable delays. Finally, how much do we check the delay values that the user dynamically sets in their code? Setting negative values would certainly give weird results, but our code-generation framework has no general mechanism to control for this [1]. Also, our `SpikeQueue` data structure is based on a maximum delay -- what happens if the dynamic change makes a delay bigger than the previously maximal delay? Do we have a dynamic data structure with a bit of buffer for this situation (e.g. like a dynamic array, whenever we exceed the current maximal delay, we double it)? Or does the user need to manually set a maximal delay in the beginning? The former can be potentially costly in memory, and the latter can be possibly annoying if it raises an error for delays exceeding the set maximum value (imagine this error at some point during a long-running simulation just because a single delay became a tiny bit bigger than the maximal delay), or lead to hard-to-find bugs if it silently clips delays to the maximum.

[...]

PR #931, opened 2018-03-20
[MRG] Handle delay variable differently

mstimberg:
As discussed in #927 , this PR changes two things:
* `delay` can no longer be used as a variable name in `Synapses`
* The pre-synaptic delay can still be set/accessed using the shorthand `synapses.delay` (instead of `synapses.pre.delay`), but you can no longer reference `delay` in code.

I'm not 100% sure about the latter change: I don't see much use of accessing it in code, but in the long-term we want to have access to it for models of "delay plastiticity" (#355). However, we'd still have to be unambiguous about the syntax to distinguish `pre` and `post` delay. We could simply use `delay_pre` and `delay_post` but then this would be confusing with respect to other `_pre` and `_post` variables which refer neuronal and not synaptic variables. But I think this is something that we should discuss/fix independent of this PR.

### DIFF
--- a/brian2/equations/equations.py
+++ b/brian2/equations/equations.py
@@ -177,7 +177,7 @@ def check_identifier_reserved(identifier):
     if (identifier in ('t', 'dt', 'xi', 'i', 'N') or
             identifier.startswith('xi_')):
         raise SyntaxError(('"%s" has a special meaning in equations and cannot '
-                           ' be used as a variable name.') % identifier)
+                           'be used as a variable name.') % identifier)
 
 
 def check_identifier_units(identifier):
@@ -675,7 +675,8 @@ class Equations(collections.Hashable, collections.Mapping):
     #: `Equations.register_identifier_check` and will be automatically
     #: used when checking identifiers
     identifier_checks = {check_identifier_basic, check_identifier_reserved,
-                         check_identifier_functions, check_identifier_units}
+                         check_identifier_functions, check_identifier_constants,
+                         check_identifier_units}
 
     @staticmethod
     def register_identifier_check(func):
--- a/brian2/equations/equations.py
+++ b/brian2/equations/equations.py
@@ -133,7 +133,7 @@ def check_identifier_basic(identifier):
     
     Raises
     ------
-    ValueError    
+    SyntaxError
         If the identifier does not conform to the above rules.
     '''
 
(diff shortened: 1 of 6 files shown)


thesamovar:
I agree we want to have it in the long term but need to decide a good syntax for it. Do you think that's covered in existing issues or should we add a new issue for that?

mstimberg:
> Do we want to change a whole bunch of ValueError to SyntaxError since strictly speaking this is a backwards-incompatible change? I think it's probably OK in this case but just wanted to raise the point.

Yes, I was wondering about this as well -- I certainly agree that backwards-compatiblity is important. I think in this case it is fine because I don't see a reason for a user or e.g. a developer of a device to catch this error. It is raised if the user uses an illegal name for a variable, and I don't see a reason to not propagate it to the user.

The situation before was really quite ugly:
```pycon
>>> g = NeuronGroup(1, 'i : volt')
...
KeyError: 'The name "i" is already present in the variables dictionary.'
>>> g = NeuronGroup(1, 'exp : volt')
...
ValueError: "exp" is the name of a function, cannot be used as a variable name.
>>> g = NeuronGroup(1, '')
>>> s = Synapses(g, g, 'lastupdate : second')
...
SyntaxError: lastupdate is a reserved name.
```

In contrast, with this PR:
```pycon
>>> g = NeuronGroup(1, 'i : volt')
...
SyntaxError: "i" has a special meaning in equations and cannot  be used as a variable name.
>>> g = NeuronGroup(1, 'exp : volt')
...
SyntaxError: "exp" is the name of a function, cannot be used as a variable name.
>>> g = NeuronGroup(1, '')
s = Synapses(g, g, 'lastupdate : second')
...
SyntaxError: "lastupdate" is a reserved name that cannot be used as a variable name.
```

However, I just realized that our check for the name of constants (to avoid variables named `pi`, `inf`, or `e`), was actually not applied... I'll fix this here as well.

thesamovar:
Agreed!

mstimberg:
Ha, I found another issue in this PR... The PR broke all non-trivial indexing (e.g. with strings or with pre-/post-synaptic neuron indices) for the `delay` variable. Unfortunately, this was not covered in the tests, I only discovered it because the Licklider example no longer worked. Looking into this more, I actually realized that this was already partly broken before. The following works in master (but was broken in this PR):
```Python
>>> S.delay[1, :] = 5*ms
```
But the following does not:
```Python
>>> S.pre.delay[1, :] = 5*ms
```

With the fix I just submitted, both work now.

thesamovar:
Good spot!

mstimberg:
All tests pass, merging it now.

PR #932, opened 2018-03-23
argument goes to diameter param if argument name is not specified int…

kapilkd13:
In this implementation of `Cylinder` function, first argument is unnamed and it tries to get attached with param `diameter`, which is already defined.
It throws
```
DimensionMismatchError                    Traceback (most recent call last)
<ipython-input-34-4d594a6997e6> in <module>()
      1 # biophsical properties and their potential representation is in attached file
      2 
----> 3 Cylinder(10, x=[0, 100]*um, diameter=1*um)

~/anaconda3/envs/brian2/lib/python3.6/site-packages/brian2/units/fundamentalunits.py in new_f(*args, **kwds)
   2356                                                            value=newkeyset[k])
   2357                         raise DimensionMismatchError(error_message,
-> 2358                                                      get_dimensions(newkeyset[k]))
   2359 
   2360             result = f(*args, **kwds)

DimensionMismatchError: Function "__init__" expected a quantitity with unit meter for argument "diameter" but got 10. (unit is 1).

```

fixing it by naming that argument correctly

### DIFF
--- a/docs_sphinx/user/multicompartmental.rst
+++ b/docs_sphinx/user/multicompartmental.rst
@@ -206,7 +206,7 @@ Morphologies with coordinates can also be created section by section, following 
 morphologies::
 
     soma = Soma(diameter=30*um, x=50*um, y=20*um)
-    cylinder = Cylinder(10, x=[0, 100]*um, diameter=1*um)
+    cylinder = Cylinder(n=10, x=[0, 100]*um, diameter=1*um)
     section = Section(5,
                       x=[0, 10, 20, 30, 40, 50]*um,
                       y=[0, 10, 20, 30, 40, 50]*um,
--- a/docs_sphinx/user/multicompartmental.rst
+++ b/docs_sphinx/user/multicompartmental.rst
@@ -56,7 +56,7 @@ start of the section represents the parent compartment with diameter 15 μm, no
 |**Cylinder** |  ::                                                                               |
 |             |                                                                                   |
 |             |     # Each compartment has fixed length and diameter                              |
-|             |     Cylinder(5, diameter=10*um, length=50*um)                                     |
+|             |     Cylinder(n=5, diameter=10*um, length=50*um)                                   |
 |             |                                                                                   |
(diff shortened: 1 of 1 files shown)


mstimberg:
Thanks, good catch! I noticed that the line following the one you fixed repeats the same mistake for `Section` -- would you mind fixing this as well? A final small thing: for documentation fixes it is best if you include `[ci skip]` somewhere in the commit message (ideally in a separate line after the main message). This will skip the running of our test suite which takes quite a long time and if you only changed documentation, there's nothing that needs to be tested. Thanks!

docs_sphinx/user/multicompartmental.rst
```diff
@@ -56,7 +56,7 @@ start of the section represents the parent compartment with diameter 15 μm, no
 |**Cylinder** |  ::                                                                               |
 |             |                                                                                   |
 |             |     # Each compartment has fixed length and diameter                              |
-|             |     Cylinder(5, diameter=10*um, length=50*um)                                     |
+|             |     Cylinder(n=5, diameter=10*um, length=50*um)                                     |
```
mstimberg: The alignment of the `|` on the right is off, the table will therefore not be displayed.

docs_sphinx/user/multicompartmental.rst
```diff
@@ -65,7 +65,7 @@ start of the section represents the parent compartment with diameter 15 μm, no
 |             |                                                                                   |
 |             |     # Length and diameter individually defined for each compartment (at start     |
 |             |     # and end)                                                                    |
-|             |     Section(5, diameter=[15, 5, 10, 5, 10, 5]*um,                                 |
+|             |     Section(n=5, diameter=[15, 5, 10, 5, 10, 5]*um,                                 |
```
mstimberg: Same as above.

kapilkd13:
Done :smiley:

mstimberg:
Great, I don't have anything else to complain about ;)

PR #934, opened 2018-03-29
Fix cache memory use

### LINKED ISSUE #933: Caching leads to important memory leak
Simulations that recreate objects from scratch repeatedly show a significant memory leak with the new caching system, with the cache preventing objects from being garbage-collected. This type of simulation is often used for parameter explorations or optimization problems, as it can be easier to write than the `store/restore` mechanism.
Here's a simple script that repeatedly runs the CUBA example:
```Python
from brian2 import *

def run_sim():
    taum = 20 * ms; taue = 5 * ms; taui = 10 * ms
    Vt = -50 * mV; Vr = -60 * mV; El = -49 * mV
    eqs = '''
    dv/dt  = (ge+gi-(v-El))/taum : volt (unless refractory)
    dge/dt = -ge/taue : volt
    dgi/dt = -gi/taui : volt
    '''
    P = NeuronGroup(4000, eqs, threshold='v>Vt', reset='v = Vr',
                    refractory=5 * ms, method='linear')
    P.v = 'Vr + rand() * (Vt - Vr)'
    Ce = Synapses(P, P, on_pre='ge += (60 * 0.27 / 10) * mV')
    Ci = Synapses(P, P, on_pre='gi += (-20 * 4.5 / 10) * mV')
    Ce.connect('i<3200', p=0.02)
    Ci.connect('i>=3200', p=0.02)
    s_mon = SpikeMonitor(P)
    run(1 * second)
    return s_mon.num_spikes


for _ in range(100):
    spikes = run_sim()
    print('Total number of spikes: %d' % spikes)
```

Plotting the memory usage (via [memory_profiler](https://pypi.org/project/memory_profiler/)) shows that there is a quite obvious problem (but the cached version at least runs faster ;) ):
mem_comparison

I've looked into things a bit and I think I fixed the main part of the issue, but I'll test things a bit more before opening the PR.

mstimberg:
This fixes the memory leak discussed in #933. It turned out that the biggest issue was the way the `SpikeQueue` was handled. Whenever code referenced to the `SpikeQueue` (all `on_pre`/`on_post` code), then the caching mechanism used an explicit reference to the `SpikeQueue` as part of the "cache key". This led to two problems:
1. Recreating exactly the same network with the same code would not use the cache at all for all synaptic statements, because the `SpikeQueue` was no longer the same.
2. The reference to the `SpikeQueue` prevented it from getting garbage collected (since the `SpikeQueue` stores synaptic delays for each synapse, it can take quite a bit of memory).

This was fixed by using our existing mechanism (the `_state_tuple` attribute), to replace the `SpikeQueue` reference by a simple tuple with all information relevant to code generation. I also made the caching mechanism use weak references if possible to avoid the problem in a more general way. There were some other smaller issues (e.g. our `_hasheable` function did not recurse down into tuples) that I fixed as well, together with some minor performance optimisation (which became necessary because the `_hasheable` function is now called more often than before).

Performance/memory usage looks pretty good now.

Recreated objects (see #933 for code):
full_mem_comparison

Store/Restore mechanism:
full_mem_comparison_restore

Fixes #933

### DIFF
--- a/brian2/core/variables.py
+++ b/brian2/core/variables.py
@@ -525,8 +525,8 @@ class DynamicArrayVariable(ArrayVariable):
     '''
     # The size of a dynamic variable can of course change and changes in
     # size should not invalidate the cache
-    cache_irrelevant_attributes = (ArrayVariable._cache_irrelevant_attributes |
-                                   {'size'})
+    _cache_irrelevant_attributes = (ArrayVariable._cache_irrelevant_attributes |
+                                    {'size'})
 
     def __init__(self, name, owner, size, device, dimensions=DIMENSIONLESS,
                  dtype=None, constant=False, needs_reference_update=False,
--- a/brian2/devices/device.py
+++ b/brian2/devices/device.py
@@ -89,6 +89,8 @@ class Device(object):
 
         self._maximum_run_time = None
 
+        self._state_tuple = (self.__module__, self.__class__.__name__)
+
     def _set_maximum_run_time(self, maximum_run_time):
         '''
         Sets a maximum time for a run before it will break. Used primarily for testing purposes. Not guaranteed to be
--- a/brian2/synapses/cythonspikequeue.pyx
+++ b/brian2/synapses/cythonspikequeue.pyx
@@ -35,9 +35,13 @@ cdef extern from "cspikequeue.cpp":
 cdef class SpikeQueue:
     # TODO: Currently, the data type for dt and delays is fixed
     cdef CSpikeQueue[double] *thisptr
+    cdef readonly tuple _state_tuple
 
     def __cinit__(self, int source_start, int source_end):
         self.thisptr = new CSpikeQueue[double](source_start, source_end)
+
+    def __init__(self, source_start, source_end):
+        self._state_tuple = (source_start, source_end, np.int32)
 
     def __dealloc__(self):
         del self.thisptr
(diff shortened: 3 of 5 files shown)


mstimberg:
> Why is SpikeQueue part of a key? And if it should be, why is it enough to only start source start and end as the state of that class? This doesn't seem entirely obvious to me.

Fully agree that it isn't obvious... We cache functions like `make_statements` that basically get a code string and a dictionary of `Variable` objects. So these are transformed into the key of the caching dictionary. `SpikeQueue` is part of the variables for the synaptic propagation code, because this code calls `peek()` and `advance()` on the spike queue. Now, if we did a code refactoring this might change: the `SpikeQueue` is only used in the fixed code in the template, so `make_statements` doesn't actually have to know about it. Why can we describe the state of the `SpikeQueue` as only source start/end? The term "state" is actually not very appropriate here: what we call state for the caching mechanism is "everything that could change the generated code" -- in this specific case, we could actually even have an empty state, as the code involving the spike queue is entirely independent of the actual queue.

> Also, is there any way we can add a test suite for this sort of error in the future?

That would be great, but I think we shouldn't do more complex stuff like memory profiling in the general test suite -- we are already running into time outs every now and then. It could be part of #33 ?

thesamovar:
Wow, issue #33 was written almost 5 years ago! :)

OK that all makes sense. Finishing review now.

PR #935, opened 2018-04-03
[MRG] Make profiling optional in standalone

mstimberg:
This is somewhat embarrassing, but in contrast to what I thought, measuring the profiling information in C++ standalone *can* take a significant amount of time. Profiling has been always optional in runtime and recently we switched the default to off for performance reasons. However, in C++ standalone switching off profiling did not change anything. With this PR, profiling becomes optional (it also fixes a minor corner-case issue, where the profiling info would not be reloaded if you reinitialize the device and re-run a s imulation).

The performance difference can be quite striking:
`cuba_openmp`: 0.26-0.27s (this PR); 0.36-0.37s (master)
`STDP_standalone`: 13.4s (this PR); 19.1s (master)
For other examples where the computation within a single code object takes a very long time (e.g. `Kremer_et_al_2011_barrel_cortex`, `COBAHH`) there's not much of a difference, though.

### DIFF
--- a/brian2/core/network.py
+++ b/brian2/core/network.py
@@ -362,7 +362,7 @@ class Network(Nameable):
         is to allow devices (e.g. `CPPStandaloneDevice`) to overwrite this.
         '''
         if self._profiling_info is None:
-            raise ValueError('(No profiling info collected (did you run with '
+            raise ValueError('No profiling info collected (did you run with '
                              'profile=True?)')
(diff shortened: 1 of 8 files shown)


mstimberg:
> Surprising that it makes so much difference though.

I think the rule of thumb is that it makes a difference when runtime vs. standalone makes a big difference. The overhead is of the same complexity -- one timing operation per time step and per code object.

Issue #941, opened 2018-05-04
Equation Error in brian2/examples/frompapers/Rothman_Manis_2003.py


matrec4:
Hi brian-team, 

In line 82 of the brian2/examples/frompapers/Rothman_Manis_2003.py, I noticed that an equation didn't match the paper. 

I believe 
dz/dt=q10*(zinf-z)/wtau : 1 

should be 

dz/dt=q10*(zinf-z)/ztau : 1

I can create a branch with the modification if required.

Thanks!

mstimberg:
Hi,
yes, I think you are right. Happy to receive a pull request about this (minor pro tip: if you include `[ci skip]` somewhere in the commit message -- e.g. in a separate line after the main text -- it will not unnecessarily start the full test suite on the test servers. Examples are not checked as part of the test suite anyway).

matrec4:
Hi @mstimberg,

I branched locally and committed the fix but can't push the branch to the repo as I don't have permissions for this repo. 
How should I proceed to create the pull request without the branch being pushed (sorry this is my first contrib so not really aware of the process)?

Happy to let you fix otherwise.

matrec4:
Hi @mstimberg,

Thanks for the clarification!
All done now.

I took a look at the github repo README before asking but didn't find the info on there. And I didn't think of going in the readthedocs documentation to find it. 
Maybe it could help others to add the link to the readthedocs contrib workflow on the repo README? 
My guess is that contribs usually come from github... 

Thanks for sharing the github documentation, didn't think to look for that either. 
I'll know for next time.

Cheers.

mstimberg:
> I took a look at the github repo README before asking but didn't find the info on there. And I didn't think of going in the readthedocs documentation to find it.
> Maybe it could help others to add the link to the readthedocs contrib workflow on the repo README?
> My guess is that contribs usually come from github...

Indeed, the thing to do nowadays seems to be to add a "CONTRIBUTING" file that will then be prominently displayed in the github interface (https://help.github.com/articles/setting-guidelines-for-repository-contributors/)

mstimberg:
Fixed via #944.

mstimberg:
Hi @matrec4, a bit of an odd question: in our upcoming paper on Brian 2, we want to acknowledge everyone who contributed in one way or another – big or small – to the Brian 2 code and documentation. I think you are the only one where I only have a github handle, would you mind giving your real name (here or per mail at marcel.stimberg@inserm.fr)? Thanks!

PR #944, opened 2018-05-15
Fixing equation mistake in example file frompapers/Rothman_Manis_2003…

matrec4:
….py at line 82 from wtau to ztau according to equation in the reference publication [ci skip] #941


mstimberg:
Thanks for the fix, merging it now.

Issue #949, opened 2018-05-17
Times that are multiples of dt are sensitive to floating point issues for refractoriness and SpikeGeneratorGroup


mstimberg:
Our check whether a neuron is not refractory anymore is not taking care to deal with values at dt boundaries, e.g. for a refractory time of 1ms, it will simply check
```Python
not_refractory = (t - lastspike) > 0.0001
```

This means that for a simulation dt of 0.1ms and a refractory time of 1ms, the actual refractory period will not necessarily be 10 time steps. For example, in the below code it is 11 time steps in 10% of the cases
```Python
from brian2 import *
import collections
G = NeuronGroup(1, '', threshold='True', refractory=1*ms)
spike_mon = SpikeMonitor(G)
run(1*second)
print(collections.Counter(np.diff(np.int_(np.round(spike_mon.t/defaultclock.dt)))))
```
```console
Counter({10: 899, 11: 91})
```

A workaround is to use a refractory time slightly below the requested time, e.g. `refractory=0.99*ms`.

I think the best fix is to subtract a small amount from the refractory period that we are testing against, say 1e-6*dt -- this should fix the problem for the "at the boundary" values without changing anything for values that weren't a multiple of dt before. The downside of the fix is that some older simulations will no longer be exactly reproducible with newer versions of Brian, but I think our policy so far has been to accept this as long as the previous behaviour was clearly buggy (as I think is the case here).

This was first reported by @zhouyanasd in issue #943 who observed that network activity changed depending on the absolute time of the network (the above check's floating point issues are sensitive to the absolute time).

thesamovar:
Modified to add that the same problem occurs with ``SpikeGeneratorGroup``. We're doing the problematic hack at the moment:

```
            # We shift all the spikes by a tiny amount to make sure that spikes
            # at exact multiples of dt do not end up in the previous time bin
            # This shift has to be quite significant relative to machine
            # epsilon, we use 1e-3 of the dt here
            shift = 1e-3*dt
            timebins = np.asarray(np.asarray(self._spike_time + shift)/dt,
                                  dtype=np.int32)
```

thesamovar:
[...]

PR #951, opened 2018-05-17
Fix #948

### LINKED ISSUE #948: SpikeGeneratorGroup sorting is very slow
The problem is in ``SpikeGeneratorGroup.__init__`` and ``before_run`` (and also duplicated code in ``set_spikes``) like this:

```
        if not sorted:
            # sort times and indices first by time, then by indices
            rec = np.rec.fromarrays([times, indices], names=['t', 'i'])
            rec.sort()
            times = np.ascontiguousarray(rec.t)
            indices = np.ascontiguousarray(rec.i)
```

and

```
            shift = 1e-3*dt
            timebins = np.asarray(np.asarray(self._spike_time + shift)/dt,
                                  dtype=np.int32)
            index_timebins = np.rec.fromarrays([self._neuron_index,
                                                timebins], names=['i', 't'])
            if not len(np.unique(index_timebins)) == len(timebins):
                raise ValueError('Using a dt of %s, some neurons of '
                                 'SpikeGeneratorGroup "%s" spike more than '
                                 'once during a time step.' % (str(self.dt),
                                                               self.name))
            self._previous_dt = dt
            self._spikes_changed = False
```

Huge amounts of time are spent in ``ndarray.sort()`` and ``unique``. The problem appears to be the use of recarrays, which are apparently very slow. For the first code block, it can be replaced with this code and is much faster:

```
I = lexsort((indices, times))
indices = indices[I]
times = times[I]
```

In some code I was working on, the time drops from 90s to about 17s.

I will submit a PR soon since it shouldn't take long to implement this.

thesamovar:
Fixes #948.

I also added a new test because we weren't testing the ``sorted=True`` keyword before and I think it would fail in some cases before (but should work now).

Incidentally, at the moment I can't run the full test suite on my Windows machine because of a whole bunch of errors (existing in master), so I'm going to look into why that is too.

### DIFF
--- a/brian2/input/spikegeneratorgroup.py
+++ b/brian2/input/spikegeneratorgroup.py
@@ -205,11 +205,6 @@ class SpikeGeneratorGroup(Group, CodeRunner, SpikeSource):
         self.variables['spike_time'].set_value(times)
         self.variables['_lastindex'].set_value(0)
 
-        # Update the internal variables used in `SpikeGeneratorGroup.before_run`
-        self._neuron_index = indices
-        self._spike_time = times
-        self._spikes_changed = True
-
     def _check_args(self, indices, times, period, N, sorted):
         times = Quantity(times)
         if len(indices) != len(times):
--- a/brian2/input/spikegeneratorgroup.py
+++ b/brian2/input/spikegeneratorgroup.py
@@ -96,12 +96,13 @@ class SpikeGeneratorGroup(Group, CodeRunner, SpikeSource):
         self.start = 0
         self.stop = N
 
+        times = np.asarray(times)
+        indices = np.asarray(indices)
         if not sorted:
             # sort times and indices first by time, then by indices
-            rec = np.rec.fromarrays([times, indices], names=['t', 'i'])
-            rec.sort()
-            times = np.ascontiguousarray(rec.t)
-            indices = np.ascontiguousarray(rec.i)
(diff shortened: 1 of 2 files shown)


thesamovar:
Good point! How about we merge this as it is (assuming tests pass) and modify #949 to be a bit more general and cover ``SpikeGeneratorGroup`` as well as refractory (since the discussion about what to do will be partly the same)?

mstimberg:
> I didn't realise that it would be problematic. Shall I shift those back into __init__ in that case? What would be the preferred solution?

It's not super problematic, but it's not considered "best practice". The standard solution is to have something like `self._spike_time = None  # will be set in _check_args` in `__init__` -- this should also include the special `#:` comments used for the reference.

> By the way test errors entirely fixed by installing gsl. Any way we can skip those tests by default or if gsl not installed? Some users will install Brian and then run the test suite and it will scare them!

Yeah, probably a good idea. Users installing via conda get `gsl` as a dependency, though. I'm curious how many install via pip vs. via conda nowadays.

mstimberg:
Great, all looking good to me, will merge this now. One of the Travis tests did not complete, but I've seen that regularly (but not always) for that test, I think it has to do with the Cython locking implementation. I've already pushed a fix that seems to work to another branch, but unfortunately in that branch I have a weird issue with standalone tests running in an infinite loop on appveyor :-/

Issue #945, opened 2018-05-16
Problem installing Brian 2 with Python 2.7


jarishna:
Using latest (16/5/18) anaconda distribution on a mac with python 2.7.14 and trying to install via conda gives an error:

```
conda install -c brian-team brian2
Solving environment: failed

UnsatisfiableError: The following specifications were found to be in conflict:
  - blaze
  - brian2
Use "conda info <package>" to see the dependencies for each package.
```

any help appreciated!

mstimberg:
Hi. This is similar to #930 I think, your blaze package seems to depend on numpy 1.14 but brian2 requires numpy < 1.14 (for now, this will change with the next release which should happen next week or so). The best workaround at the moment is to install brian2 into its own environment which avoids such problems. Something along the lines of:
```
conda create -c brian-team -n brian2 python=2 brian2 matplotlib jupyter ipython
```
Should give you a `brian2` environment with all the necessary tools which you can then activate with `conda activate brian2` (or `source activate brian2` -- depends on your configuration). See https://conda.io/docs/user-guide/tasks/manage-environments.html for more details on environments.

mstimberg:
This should be fix with the latest release (currently: 2.1.3.1), please re-open this issue if you still have problems.

Issue #930, opened 2018-03-20
A problem when I install Brian2


hustyanghui:
Hello everyone,
       Recently I want to use anaconda3 and brian2 ,for reference:
            http://brian2.readthedocs.io/en/stable/introduction/install.html
       And I have installed Anaconda successfully!
      But when I want to install brian2 as the link suppose : conda install -c brian-team brian2
      There comes the  question.The result after is showed as follow:
       "
     Solving environment: failed

UnsatisfiableError: The following specifications were found to be in conflict:
  - brian2
  - numba
Use "conda info <package>" to see the dependencies for each package.

        "

So ,is there anyone else have meet this problem,would you like to give me some useful suggestions?Thanks very much!

hustyanghui:
For supplement,I installed Anaconda in Linux

mstimberg:
Hi. From the error message it is difficult to see what exactly is going wrong. Most likely the problem is that a package depends on numpy 1.14, while the current Brian2 package only supports numpy 1.13.
Your error message seems to state that there's a problem with the numba package, but that's odd given that it could downgrade numba to the numpy 1.13 version.
I'd try to manually downgrade numpy to version 1.13:
```
conda install numpy==1.13
```
If that doesn't work because some package requires the most recent numpy version, you can always create a conda environment (see https://conda.io/docs/user-guide/tasks/manage-environments.html) just for Brian. Try:
```
conda create -n brian2 -c brian-team python=3 brian2 matplotlib
```
(you can add other packages that you want to use at the end, e.g. `notebook` for the jupyter notebook)

Then, you can switch to this environment with
```
source activate brian2
```

mstimberg:
This should be fix with the latest release (currently: 2.1.3.1), please re-open this issue if you still have problems.

PR #954, opened 2018-05-22
[MRG] Do not tie conda packages to numpy version

### LINKED ISSUE #915: Reconsider packaging with respect to numpy versions
Currently, we build one Brian2 version for each numpy version, but building one for the oldest supported version might be enough. See the discussion here: https://github.com/brian-team/brian2/pull/913#issuecomment-360494739.

mstimberg:
As discussed earlier (https://github.com/brian-team/brian2/pull/913#issuecomment-360494739), this PR changes our conda packaging to a single package for each architecture, independent of the numpy version. The advantage is that numpy can be updated independently of Brian, and we won't run into problems when anaconda updates to a numpy version that did not exist at the time of our release (currently, this seems to prevent users from installing Brian on top of a recent full Anaconda distribution, see #930, #945). The disadvantage is that nothing prevents a user to run Brian with a new and untested numpy version -- on the other hand, that's already the case when they don't use conda for installation.

Note that in this branch (as in master) the Cython tests still hang sometimes for a certain combination of packages -- I thought I fixed it by explicitly unlocking the Cython lock files, but this does not seem to be enough. I'm currently testing using the [filelock](https://pypi.org/project/filelock/) package in a separate branch and will open another PR if that is successful.

Closes #915

### DIFF
--- a/brian2/codegen/runtime/cython_rt/extension_manager.py
+++ b/brian2/codegen/runtime/cython_rt/extension_manager.py
@@ -113,12 +113,13 @@ class CythonExtensionManager(object):
         if prefs['codegen.runtime.cython.multiprocess_safe']:
             lock_file = os.path.join(lib_dir, module_name + '.lock')
             with open(lock_file, 'w') as f:
+                # Lock
                 if msvcrt:
                     msvcrt.locking(f.fileno(), msvcrt.LK_RLCK,
                                    os.stat(lock_file).st_size)
                 else:
                     fcntl.flock(f, fcntl.LOCK_EX)
-                return self._load_module(module_path,
+                module = self._load_module(module_path,
                                          define_macros=define_macros,
                                          include_dirs=include_dirs,
                                          library_dirs=library_dirs,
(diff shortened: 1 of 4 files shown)


brian2/codegen/runtime/cython_rt/extension_manager.py
```diff
@@ -131,6 +132,13 @@ def create_extension(self, code, force=False, name=None,
                                          runtime_library_dirs=runtime_library_dirs,
                                          compiler=compiler,
                                          key=key)
+                # Unlock
+                if msvcrt:
+                    msvcrt.locking(f.fileno(), msvcrt.LK_UNLCK,
+                                   os.stat(lock_file).st_size)
+                else:
+                    fcntl.flock(f, fcntl.LOCK_UN)
+            return module
```
thesamovar: Did you mean to include these changes in this PR or is this something that was supposed to be in the other one?

mstimberg: This change was supposed to fix the build hang that I saw, but it turned out not to work. I'll open a new PR in a sec to use a dedicated module for file locking which seems to correctly deal with the issue.

PR #955, opened 2018-05-22
[MRG] Only use `profiled_codeobjects` attribute in `objects.cpp` template if defined

mstimberg:
Brian2GeNN does not support profiling but re-uses the `objects.cpp` template.


mstimberg:
[Note: I'm merging this despite several test failures -- those failures are due to the use of `CONDA_BUILD` as a variable name, an issue that has been fixed with commit cadaa79d3b38b2ffa1744cd62435afe282058fe7 in the `conda_numpy_build` branch.]

PR #957, opened 2018-05-23
[WIP] Use filelocking module for locking in Cython

mstimberg:
We regularly had one specific travis test hang during the Cython tests (e.g. [this one](https://travis-ci.org/brian-team/brian2/jobs/382208904)) and I think it has to due with our file locking implementation for Cython that we need to make running things with parallel processes work. In this PR, I use the [filelock](https://pypi.org/project/filelock/) module to do the locking as a replacement for our own implementation. I decided to declare it as a dependency (it is available both via pypi and conda, and installing it should always work given that it is a Python-only package), but if you run directly from source, it should only fail when you actually use Cython. An alternative would be to include the module directly in our code (it is a single file and public domain), but I think it could be useful to benefit from future enhancements and bug fixes. I don't feel strongly about this, though.

Note that I used a timeout of 9min for aquiring the lock which is quite long, but during testing I saw that e.g. a timeout of 1min is not enough in all cases (we might want to investigate why this is the case at some point, though...). Either way, we cannot do much useful with the timeout information anyway, I simply added it so that we can distinguish hangs due to lock acquisition from other problems on travis/appveyor (where tests will fail after 10min without output).

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -124,9 +124,9 @@ install:
        conda install --yes --quiet anaconda-client conda-build jinja2 pip setuptools;
     fi
   - if [[ $MINIMAL_VERSIONS == 'yes' ]]; then
-    travis_retry conda create -n travis_conda --yes --quiet python=$PYTHON numpy==1.10 scipy==0.16 libgfortran==1 nose "sphinx<1.6.3" ipython sympy==0.7.6 jinja2==2.7 pyparsing setuptools coverage;
+    travis_retry conda create -n travis_conda --yes --quiet python=$PYTHON numpy==1.10 scipy==0.16 libgfortran==1 nose "sphinx<1.6.3" ipython sympy==0.7.6 jinja2==2.7 pyparsing setuptools coverage filelock==2.0.6;
     else
-    travis_retry conda create -n travis_conda --yes --quiet python=$PYTHON numpy nose "sphinx<1.6.3" ipython "sympy!=1.1.0" pyparsing jinja2 setuptools coverage;
+    travis_retry conda create -n travis_conda --yes --quiet python=$PYTHON numpy nose "sphinx<1.6.3" ipython "sympy!=1.1.0" pyparsing jinja2 setuptools coverage filelock;
     fi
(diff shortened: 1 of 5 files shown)


mstimberg:
Um, something is odd, the test that this PR was supposed to fix failed again... No locking timeout, maybe the test is just very slow for the combination of packages and Cython (the test is using the "minimal versions")...? Will look into this tomorrow.

https://travis-ci.org/brian-team/brian2/jobs/382732493

mstimberg:
This is getting really annoying... So the `filelock` module did indeed not fix the problem. I checked which tests were not yet executed when the test hangs, and there were only two left. These tests are very simple and fast (<0.5s on my machine), but take a long time to compile when they are not in Cython's cache (>40s on my machine, several minutes on travis) because they use many simple code strings for `Synapses.connect` etc. This together with the interaction between multiple processes could well have been the reason for the observed timeout. I therefore split up the test into shorter ones, but the tests now simply hang elsewhere (even though before everything until then executed fine...). I still think the basic issue has to do with the fact that our Cython compilation takes a looong time -- no wonder given that even our simplest templates will lead to > 7000 lines of C++ code... If two processes run code that overlap in the code they need to compile I can easily imagine getting into a situation where a single test can take a considerable amount of time (even though 10min seems a bit extreme...).

I have to say I don't feel motivated to investigate this further in any detail. I think in the long run we should implement our own "simple version of weave" that replaces both weave and Cython for Python 2 and 3 -- this would make our life easier in many ways and should also lead to a better "perceived performance" as our generated code would be more compact than the one generated by Cython (and, to a lesser extent, weave). In the short run, I created a new branch were I switched off parallel testing for Cython on the test servers and added a few optimisations that slightly improve the runtime of the test infrastructure (less overhead from the conda installation process). I considered adding caching on the test servers as well (i.e. store the content of the weave/cython cache directories on the server and reload them at the start of a test run), but this seems to be problematic as our caches can get big quickly, and at least on appveyor everything will fail if we exceed 1GB.

thesamovar:
Agreed!

PR #962, opened 2018-05-28
[MRG] Introduce timestep function to avoid rounding issues in refractoriness

mstimberg:
This introduces a new `timestep` function and uses it for the implementation of refractoriness, as discussed in #949. It's basically done except for a bit of documentation.

A few things to discuss before merging:
* I used a shift of 1e-3dt to be consistent what we are doing in `SpikeGeneratorGroup`, not sure whether a smaller shift would be better...
* the implementation of the `timestep` function is slightly complicated by the fact that the time of the last spike is `-inf` before the first timestep. My handling of all this in the numpy implementation is probably not optimal, I'm seeing something like a 2% slowdown for the COBAHH example.
* I added a preference to switch back to the old behaviour -- not sure it's well explained and named, though (`core.legacy_refractory_timing`). Maybe we should have a preference category for this kind of thing?
* GSL integration was a bit tricky to get to work again, the way we use `timestep(t, dt)` in the refractoriness condition means that `dt` is used in the scalar and in the vector code, we never had this before.

### DIFF
--- a/brian2/core/core_preferences.py
+++ b/brian2/core/core_preferences.py
@@ -37,5 +37,17 @@ prefs.register_preferences('core', 'Core Brian preferences',
         Whether to raise an error for outdated dependencies (``True``) or just
         a warning (``False``).
         '''
-        )
+        ),
+    legacy_refractory_timing=BrianPreference(
+        default=False,
+        docs='''
+        Whether to use the semantics for checking the refractoriness condition
+        that were in place up until (including) version 2.1.2. In that
+        implementation, refractory periods that were multiples of dt could lead
+        to a varying number of refractory timesteps due to the nature of
+        floating point comparisons). This preference is only provided for exact
+        reproducibility of previously obtained results, new simulations should
+        use the improved mechanism which uses a more robust mechanism to
+        convert refractoriness into timesteps. Defaults to ``False``.
+        ''')
     )
--- a/brian2/core/core_preferences.py
+++ b/brian2/core/core_preferences.py
@@ -37,8 +37,11 @@ prefs.register_preferences('core', 'Core Brian preferences',
         Whether to raise an error for outdated dependencies (``True``) or just
         a warning (``False``).
         '''
-        ),
-    legacy_refractory_timing=BrianPreference(
+        )
+    )
+
+prefs.register_preferences('legacy', 'Preferences to enable legacy behaviour',
+    refractory_timing=BrianPreference(
         default=False,
         docs='''
         Whether to use the semantics for checking the refractoriness condition
--- a/brian2/core/functions.py
+++ b/brian2/core/functions.py
@@ -544,6 +544,34 @@ class log10(sympy_Function):
 _infinity_int = np.iinfo(int).max//2
 
 def timestep(t, dt):
+    '''
+    Converts a given time to an integer time step. This function slightly shifts
+    the time before dividing it by ``dt`` to make sure that multiples of ``dt``
+    do not end up in the preceding time step due to floating point issues. This
+    function is used in the refractoriness calculation.
+
+    .. versionadded:: 2.1.3
+
+    Parameters
+    ----------
+    t : np.ndarray or Quantity
+        The time to convert.
+    dt : float or Quantity
+        The length of a simulation time step.
+
+    Returns
+    -------
+    ts : int
+        The time step corresponding to the given time.
+
+    Notes
+    -----
+    This function can handle infinity values, it will return a value equal to
+    half of the maximal integer value. This assures that an expression such as
+    ``timestep(t) - timestep(lastspike)`` will result in a reasonable value even
+    if ``lastspike`` is ``-inf``. If ``timestep(lastspike)`` were equal to the
+    minimal representable integer value, this expression would overflow.
+    '''
     elapsed_steps = (t + 1e-3*dt)/dt
     if np.isscalar(elapsed_steps) or elapsed_steps.shape == ():
         if np.isinf(elapsed_steps):
--- a/brian2/core/functions.py
+++ b/brian2/core/functions.py
@@ -554,14 +554,14 @@ def timestep(t, dt):
 
     Parameters
     ----------
-    t : np.ndarray or Quantity
+    t : np.ndarray, float, Quantity
         The time to convert.
     dt : float or Quantity
         The length of a simulation time step.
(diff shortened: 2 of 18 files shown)


brian2/core/functions.py
```diff
@@ -540,6 +541,27 @@ def eval(cls, args):
...
+            else:
+                return _iinfo.min
+        else:
+            return np.int_(elapsed_steps)
+    else:
+        are_inf = np.isinf(elapsed_steps)
+        int_steps = np.asarray(elapsed_steps, dtype=int)
+        if any(are_inf):
+            int_steps[are_inf & (elapsed_steps < 0)] = _iinfo.min
+            int_steps[are_inf & (elapsed_steps > 0)] = _iinfo.max
```
thesamovar: Do we ever have t=+inf? Maybe we can leave this?

brian2/core/core_preferences.py
```diff
@@ -37,5 +37,17 @@ def default_float_dtype_validator(dtype):
...
+        docs='''
+        Whether to use the semantics for checking the refractoriness condition
+        that were in place up until (including) version 2.1.2. In that
+        implementation, refractory periods that were multiples of dt could lead
+        to a varying number of refractory timesteps due to the nature of
+        floating point comparisons). This preference is only provided for exact
+        reproducibility of previously obtained results, new simulations should
+        use the improved mechanism which uses a more robust mechanism to
+        convert refractoriness into timesteps. Defaults to ``False``.
+        ''')
```
thesamovar: I agree it would be great to have a ``legacy`` category to put this in.

brian2/tests/test_GSL.py
```diff
@@ -358,7 +358,7 @@ def test_GSL_non_autonomous():
 @with_setup(teardown=reinit_devices)
 @skip_if_not_implemented
 def test_GSL_refractory():
-    eqs = '''dv/dt = 100*Hz : 1 (unless refractory)'''
+    eqs = '''dv/dt = 99.99*Hz : 1 (unless refractory)'''
```
thesamovar: What's going on here?

brian2/tests/test_refractory.py
```diff
@@ -36,8 +37,8 @@ def test_missing_refractory_warning():
 @with_setup(teardown=reinit_devices)
 def test_refractoriness_basic():
     G = NeuronGroup(1, '''
-                       dv/dt = 100*Hz : 1 (unless refractory)
-                       dw/dt = 100*Hz : 1
+                       dv/dt = 99.999*Hz : 1 (unless refractory)
+                       dw/dt = 99.999*Hz : 1
```
thesamovar: These slightly concern me: is this something users might need to be doing?

brian2/core/functions.py
mstimberg: Actually with the refractoriness we only ever have `+inf` since we pass `t - lastspike` and `lastspike` is `-inf` if no spike has occurred yet. So that would mean we could ignore the `-inf` case. But since the plan is to provide the `timestep` function to the user as well, I find it dangerous to ignore the other corner case. E.g. instead of doing `timestep(t - lastspke, dt)`, what if the user uses `timestep(t, dt) - timestep(lastspike, dt)`? Oh, actually that would lead to an overflow even with the correct implementation :-/ Maybe we should convert `inf` into something like `1e9` instead of the maximum integer value? Seems a bit arbitrary, though...

brian2/tests/test_GSL.py
mstimberg: We were sloppy about this before, I always slightly tweaked the tests so that they passed. Now that we have made sure that refractoriness is exact up to a time step, this seemed to be an odd thing to do (even more than before, I mean). The problem is that with `dv/dt = 100*Hz` we are supposed to arrive at exactly 1 after 100 time steps, but due to floating point issues, we will end up slightly above or below. E.g. with 100Hz it will be at 1.0000000000000007 after 10ms, but with 1000Hz it will be at 0.9999999999999999 after 1ms... In principle, we could apply a similar kind of shift approach to threshold checks, but this would become messy since the user already provides a string condition. Also, I think that differential equations should rather be used for "realistic" dynamics where you'll never have *exactly* 100Hz as a firing rate. For regular spike trains, rather use `SpikeGeneratorGroup(1, [0], [0]*ms, period=10*ms).`

brian2/core/functions.py
thesamovar: I think in the C++ and Cython it's fine because it's just a branch that isn't usually taken, but it's a lot of work for numpy.

We could use +/- ``ourmaxint = min(intmax, -intmin)/2`` as the converted value so that arithmetic with it would work? We could also maybe set ``int_steps[are_inf] = sign(elapsed_steps[are_inf])*ourmaxint``? Might be even quicker to compute ``I,=are_inf.nonzero()`` first and then use that. Basically, at the moment we have to do N-element operations five times and we should only do that once since usually most ``are_inf`` will be false.

brian2/tests/test_GSL.py
thesamovar: Ah I see, OK that's fine then. In that case, perhaps just a slight performance tweak for numpy and we're done I think.

mstimberg:
Ok, I'll try out some performance improvements for the numpy implementation, move the preference to a new category, and document the new refractoriness implementation etc.

Unfortunately there are also still build issues both Windows and OS X have some issues with the `isinf` function/macro, let's see whether I can fix them "remotely".

mstimberg:
> Just running test locally but I think I have a fix for the MSVC build issue. Shall I commit it?

Sure, go for it.

thesamovar:
Just waiting for tests to pass locally, but looks like it will.

Also, we should document timestep I think.

thesamovar:
OK fix pushed, slightly different for weave and cython. Fix for Cython might help on Mac but fix for weave surely won't.

mstimberg:
Ok, great, weave was working fine on OS X, I think. But let's wait for the tests. And yes, I agree the  `timestep` function should be documented.

mstimberg:
> Your Cython change did not fix the issue

"did not fix the issue on OS X" I meant -- all seems to work fine on Windows.

thesamovar:
Looks good to me - did the numpy optimisation make any difference?

Also, do we have any other 'legacy' behaviours that we allow in an option?

PR #965, opened 2018-05-31
[MRG] The `timestep` function should not use `Quantity` objects during a simulation

mstimberg:
While I only saw a minor slow down for the numpy target with the COBAHH example due to the introduction of the `timestep`, I just realized that simpler examples are much more affected. One of them is part of our benchmarking: https://brian-team.github.io/brian2_benchmarks/#benchmarks.FullExamples.time_reliability_example

I found that due to the way that I added the `timestep` function, Brian thought that the function *requires* quantities instead of simple floats/arrays and was therefore helpfully converting the floats/arrays that we are using during simulation to quantities and back. Not only the conversion itself itself is somewhat costly, but arithmetic operation on quantities are also significantly slower due to the additional checks and unit conversions.

### DIFF
--- a/brian2/codegen/generators/numpy_generator.py
+++ b/brian2/codegen/generators/numpy_generator.py
@@ -4,7 +4,7 @@ import numpy as np
 
 from brian2.parsing.bast import brian_dtype_from_dtype
 from brian2.parsing.rendering import NumpyNodeRenderer
-from brian2.core.functions import DEFAULT_FUNCTIONS, Function
+from brian2.core.functions import DEFAULT_FUNCTIONS, timestep
 from brian2.core.variables import ArrayVariable
 from brian2.utils.stringtools import get_identifiers, word_substitute, indent
 from brian2.utils.logger import get_logger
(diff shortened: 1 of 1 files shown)


thesamovar:
Oops! Well noticed. Does that solve the issue with the slight slowdown too?

mstimberg:
> Does that solve the issue with the slight slowdown too?

I think it is still a bit slower than before, but it's hard to say since the run times always vary a bit. From a couple of trials I'd say that for the COBAHH example the slow down is now rather 1-2% compared to 2-3% I was seeing before.
Actually, I also observed the following: the bottleneck does not seem to be the handling of the inf values; instead, the division takes most of the time. This leads to the counter-intuitive result that calculating the `timestep` for an array is actually cheaper the *more* inf values are in it -- dividing inf by anything is faster than doing a division with actual numbers.

mstimberg:
Just to confirm, after the merge the performance impact for numpy looks much more reasonable:
https://brian-team.github.io/brian2_benchmarks/#benchmarks.FullExamples.time_reliability_example
https://brian-team.github.io/brian2_benchmarks/#benchmarks.FullExamples.time_COBAHH_example

Issue #966, opened 2018-06-05
Quantity.get_best_unit is very inefficient for large arrays


mstimberg:
When we convert `Quantity` arrays to strings, we try to find the "best unit" to represent them. This is done by checking which unit makes the sum of deviations from 1 the smallest (e.g. `[1, 2, 3]*ms` would be better represented by `[1, 2, 3]` (in `ms`) than by `[0.001, 0.002, 0.003]` (in `second`)). This is a reasonably well working heuristic, but it is computed by creating an intermediate array with the size of the original array × the number of units with the same dimensions. This can be very slow for big arrays and potentially even fill up the available memory. Note that e.g. for `second`, there are 17 units that are checked (second, msecond, usecond, nsecond, etc.).

I think we could potentially come up with a better heuristic, but there is one big inefficiency that we could address immediately: big arrays will never be converted to strings in their entirety, so we don't need to look at all values to find the best unit either. With standard setting, numpy will shorten arrays with more than 1000 elements to something like `[0 1 2 ... 9997 9998 9999]`, so we only need to check the magnitude of those values.

The problem is actually more important than it looks at the first sight, because it does not only affect situations where the user explicitly asks for a conversion to a string. In some places, we use code like this:
```Python
try:
   ....
except DimensionMismatchError:
   pass  # or do something else
```
The problem: the mere creation of a `DimensionMismatchError` will ask for the best units of the involved quantities to give nice error messages. We might want to fix this as well (e.g. `__str__` of a `DimensionMismatchError` could compute this only on-demand), but if we fix the more general issue, this should not make a noticeable difference.

A recent [post on the mailing list](https://groups.google.com/d/msg/briansupport/LpbHBqBLizQ/zUa14ak9AAAJ) made me aware of this issue.

Issue #841, opened 2017-05-05
Convert `brian2.units` into a standalone package


mstimberg:
We've discussed this several times, but we've never opened an issue for it: in the long term, I think it would be a good idea to separate `brian2.units` into a separate package, potentially making it useful outside of the Brian 2 community.
If we do this (which will take a bit of work to set up the infrastructure, but afterwards should not be very), I think we should adopt a development model that is very strict, i.e. keep it backwards-compatible as much as possible, don't reconsider the basic approach, etc.

thesamovar:
Yes to all.

mstimberg:
For reference, this is ongoing work in https://github.com/brian-team/brian2units (not necessarily the final package name…)

PR #967, opened 2018-06-05
[MRG] Fix inefficiencies in the Quantity/Unit implementation

mstimberg:
See problem description in #966. In addition to that issue, this PR also fixes an inefficiency when dividing large quantities by a unit (most commonly used to remove units, e.g. `mon.t/second`)

### DIFF
--- a/brian2/tests/test_units.py
+++ b/brian2/tests/test_units.py
@@ -1021,6 +1021,22 @@ def test_get_unit():
 
 
 @attr('codegen-independent')
+def test_get_best_unit():
+    # get_best_unit should not check all values for long arrays, since it is
+    # a function used for display purposes only. Instead, only the first and
+    # last few values should matter (see github issue #966)
+    long_ar = np.ones(10000)*siemens
+    long_ar[:10] = 1*nS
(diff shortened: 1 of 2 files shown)


mstimberg:
Ok, the benchmarks for the two issues fixed with this PR now look a bit ridiculous:
https://brian-team.github.io/brian2_benchmarks/#benchmarks.UnitBenchmarks.time_units_to_string
https://brian-team.github.io/brian2_benchmarks/#benchmarks.UnitBenchmarks.time_unit_removal_by_division

thesamovar:
The good sort of ridiculous!

Is it worth considering a more systematic performance analysis of units? It does pop up in profiling a bit, but in my cases never enough to worry about.

mstimberg:
> Is it worth considering a more systematic performance analysis of units? It does pop up in profiling a bit, but in my cases never enough to worry about.

Probably, how about doing this when we transition it to its own package (#841)? While it shouldn't matter for the simulation time, inefficiencies like the one fixed here can be really annoying in the analysis/plotting part.

thesamovar:
Agreed.

PR #972, opened 2018-06-14
[MRG] `timestep` function: implement the "overflow-safe" infinity value for C++/Cython

mstimberg:
I noticed two bugs in our recently introduced `timestep` function and fix them in this PR:

1. In the documentation we state: 
> This function can handle infinity values, it will return a value equal to half of the maximal integer value. This assures that an expression such as `timestep(t) - timestep(lastspike)` will result in a reasonable value even if `lastspike` is `-inf`.

This was actually only true for the numpy implementation, C++ and Cython still returned the minimum/maximum possible integer, leading to potential overflow issues.

2. If you declare a variables as `: integer` in equations, it will by default use a 32bit integer. Now, such a variable might be a natural way to store a timestep calculated with the `timestep` value. Unfortunately,  the "maximum integer" that we divide by 2 and use as the "infinity value" in numpy asks for the maximum value of a *Python* integer, which corresponds to a `long` in C and therefore to a 64bit integer on Linux/OS X... I know simply hardcoded the maximum 32bit integer / 2 everywhere...

Both issues would have been detected by the test I just added.

### DIFF
--- a/brian2/codegen/generators/cpp_generator.py
+++ b/brian2/codegen/generators/cpp_generator.py
@@ -463,17 +463,18 @@ namespace std {
 #endif
 #endif
 #endif
-        static inline int _timestep(double t, double dt)
-        {
-            if (std::isinf(t))
-            {
-                if (t < 0)
-                    return INT_MIN;
-                else
-                    return INT_MAX;
-            }
-            return (int)((t + 1e-3*dt)/dt); 
-        }
+static inline int _timestep(double t, double dt)
+{
+    const int _infinity_int = 1073741823;  // maximum 32bit integer divided by 2
+    if (std::isinf(t))
+    {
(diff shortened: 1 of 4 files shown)


mstimberg:
> Wondering if we should make it max int / 4, or max/2-1000 just to give a bit of wiggle room?

Hmm, not sure. Every choice will be arbitrary, but I think users will not do much arithmetics with these time steps. For me, the only reasonable use case is to add or subtract two values and this should be "safe" with the max/2 solution (even though `timestep(inf) - timestep(inf)` will give zero, etc.). Maybe we should make it a bit clearer in the documentation? But then, noone reads it anyway...

thesamovar:
OK fair enough! No need to document this particular detail I think.

Issue #976, opened 2018-06-27
Documentation fix: ProgressBar code


paulbrodersen:
The ProgressBar code snippet on http://brian2.readthedocs.io/en/stable/advanced/scheduling.html should be: 

```python
import sys

class ProgressBar(object):
    def __init__(self, toolbar_width):
        self.toolbar_width = toolbar_width
        self.ticks = 0

    def __call__(self, elapsed, complete, start, duration):
        if complete == 0.0:
            # setup toolbar
            sys.stdout.write("[%s]" % (" " * self.toolbar_width))
            sys.stdout.flush()
            sys.stdout.write("\b" * (self.toolbar_width + 1)) # return to start of line, after '['
        else:
            ticks_needed = int(round(complete * self.toolbar_width))
            if self.ticks < ticks_needed:
                sys.stdout.write("-" * (ticks_needed-self.ticks))
                sys.stdout.flush()
                self.ticks = ticks_needed
        if complete == 1.0:
            sys.stdout.write("\n")

net.run(duration, report=ProgressBar(40), report_period=1*second)
```

This fixes the following original lines:

```python
ticks_needed = int(round(complete * 40))
```

and

```python
net.run(duration, report=progress_bar, report_period=1*second)
```

Personally, I would provide a default value for `toolbar_width`, i.e. 

```python 
class ProgressBar(object):
    def __init__(self, toolbar_width=40):
        ...
net.run(duration, report=ProgressBar(), report_period=1*second)
```

mstimberg:
Hi @paulbrodersen : thanks for the note. Would you mind creating a pull request for this (default argument sounds very good to me as well)? The file to edit is `docs_sphinx/advanced/scheduling.rst` (https://github.com/brian-team/brian2/blob/master/docs_sphinx/advanced/scheduling.rst).

paulbrodersen:
Sorry, I didn't know the dev culture here so I didn't want to throw a PR nilly-willy at you before having cleared the changes. I will raise PRs more freely in the future. ;-)

mstimberg:
No worries, we definitely have to work on our documentation for potential contributors a bit to make clearer that such contributions are indeed very welcome.

mstimberg:
Closed via #977

PR #977, opened 2018-06-28
Fix issue #976 (Documentation fix: ProgressBar code)

paulbrodersen:
As discussed in issue #976, I made the following changes (in order of file differences):
1) Add a default `toolbar_width`.
2) Actually use the `toolbar_width` in `__call__` instead of a preset.
3) Pass the progress bar in the correct manner to `run`.


mstimberg:
Thank you very much again.

PR #978, opened 2018-07-06
List default functions and constants in namespace documentation

paulbrodersen:
This PR adds a list of the default namespace functions and constants in the documentation for namespaces (units are well documented in other places). It took me some time (not long but long enough to be annoying) to track down the default functions that are imported into any namespace and I thought I might save the next person some time. 

I don't have any experience with sphinx nor do I have it set up so I can't really check the formatting. It would be good if somebody else could. 
 
##### Additional comment:

Quite frankly, the standard doc-string for `namespace` needs some more love. It has taken me a long time and reading through pretty much all of brians documentation and then some experimentation to realize it's expressiveness. I think it would be good to have a discussion about what could and should be included in a new doc-string for `namespace`. I am not confident enough in my understanding of all the things that can be done with `namespace` to write a new doc-string by myself.

### DIFF
--- a/docs_sphinx/advanced/namespaces.rst
+++ b/docs_sphinx/advanced/namespaces.rst
@@ -5,7 +5,7 @@ Namespaces
 external parameters or functions. During the initialisation of a `NeuronGroup`
 or a `Synapses` object, this *namespace* can be provided as an argument. This
 is a group-specific namespace that will only be used for names in the context
-of the respective group. Note that units and a set of standard functions are
+of the respective group. Note that units and a set of standard functions (listed below) are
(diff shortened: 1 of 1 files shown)


paulbrodersen:
5 minutes after writing this I find the `functions` documentation. My bad -- I am blind today.

mstimberg:
No worries, I think your point is still valid: this should not be that difficult to find. We should probably mention the set of functions/constants that can be used more prominently when introducing equations and neural models (http://brian2.readthedocs.io/en/stable/user/models.html), at least with the link to the functions docs.

PR #979, opened 2018-07-11
Don't create `lastupdate` variables if they are not needed

denisalevi:
For `Synapses` without model equations, the `lastupdate` variable makes up around `~1/3` of the memory demand of a `Synapses` object if I'm not mistaken (`lastupdate`: 8 byte, `preID/postID`: 4 byte each, `delay`: 8 byte). But the `lastupdate` variable is only needed for `event-driven` dynamics or if it is used as identifier in the `Synapses` model equations of the `SynapticPathways` update code.

In this PR I tried to get rid of the `lastupdate`, when its not needed. I'm not sure if I caught all cases (guess the tests wills show). Because of GPU memory limits, this PR would be quite useful for us. But I guess generally also for brian2? Would appreciate some comments.

And in this version I'm checking naively for `lastupdate` being in any of `SynapticPathway` codes. This should be done in a safe way (right now it would also catch `wronglastupdate`). Before I start searching brian for the correct parser function, I was hoping you could just point me to it :)

### DIFF
--- a/brian2/synapses/synapses.py
+++ b/brian2/synapses/synapses.py
@@ -732,8 +732,11 @@ class Synapses(Group):
                 raise SyntaxError('"%s" is a reserved name that cannot be '
                                   'used as a variable name.' % name)
 
-        # Add the lastupdate variable, needed for event-driven updates
-        model = model + Equations('lastupdate : second')
+        # Check if a `lastupdate` variable is needed
+        need_lastupdate = self._check_lastupdate(model, on_pre, on_post)
+        if need_lastupdate:
+            model = model + Equations('lastupdate : second')
+
         # Add the "multisynaptic index", if desired
         self.multisynaptic_index = multisynaptic_index
         if multisynaptic_index is not None:
@@ -922,6 +925,28 @@ class Synapses(Group):
     def __getitem__(self, item):
         indices = self.indices[item]
         return SynapticSubgroup(self, indices)
+
+    def _check_lastupdate(self, model, on_pre, on_post):
+        # check if `lastupdate` is used as identifier in synapse model equations
+        if 'lastupdate' in model.identifiers:
+            return True
+        # check if we have `event-driven` dynamics
+        for eq in model.itervalues():
+            if 'event-driven' in eq.flags:
+                return True
+        # get all pathway event update codes
+        pathway_codes = []
+        for argument in [on_pre, on_post]:
+            if isinstance(argument, basestring):
+                pathway_codes.append(argument)
+            elif isinstance(argument, collections.Mapping):
+                for value in argument.intervalues():
+                    pathway_codes.append(value)
+        # check if lastupdate is needed for event update code
+        for code in pathway_codes:
+            # TODO: this also catches e.g. 'mylastupdate', use some brian parser!
+                return True
+        return False
 
     def _set_delay(self, delay, with_unit):
         if 'pre' not in self._synaptic_updaters:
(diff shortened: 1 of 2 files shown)


denisalevi:
Modified and added some test. Got a problem when `run_regularly` is using `lastupdate` (see last added test). One would have to add `lastupdate` manually when `run_regularly` uses it. I'm getting the feeling it would have maybe been easier to do the check for `lastupdate` on device side?

mstimberg:
I actually stumbled across the unnecessary `lastupdate` variable a few days ago as well, it is not only wasting memory but also the update takes a bit of time.

I wonder whether it makes sense to try to catch all the possible uses, given that they should be corner cases. I guess instead my preferred approach would be:
* The `lastupdate` variable is automatically added when you use event-based synaptic equations
* In all other cases (e.g. [this example](http://brian2.readthedocs.io/en/stable/user/synapses.html#explicit-event-driven-updates)), the user has to explicitly add the variable themselves and add `lastupdate = t` to the end of their `on_pre`/`on_post` statements.

While the latter change is not backwards-compatible, I think it would be ok if we special-case the "unknown identifier" error message for a `lastupdate` variable in synapses, telling the user what to do and/or linking to a page in the documentation.

denisalevi:
Your approach seems easier, otherwise there are just a lot of corner cases to catch.

But I just realized that there are also the `N_incoming` and `N_outgoing` variables which are not used most of the time. In standalone mode one could just write them to disk after synapse creation and free their memory. Currently we end up copying them into device memory just for the rare case of someone referencing them in synapses model / update code. And for `N_incoming`/`N_outgoing` one can't just add the variable manually. One could check for it the same way I implemented the `lastupdate` test here, but then we would again check for corner cases. Any idea how to go about that?

mstimberg:
When we introduced `N_incoming` and `N_outgoing`, we discussed the memory waste given that they are not used most of the time. We decided to not worry, given that they are only stored once per neuron, not once per synapse (they are accessible as synaptic variables, but that's just indexing magic). 

In principle it might be useful to have a general mechanism to just copy over variables to device code that are actually used on the device side? You wouldn't have to parse any code for that, you could look at the `variables` dictionaries of all code objects.

denisalevi:
> When we introduced N_incoming and N_outgoing, we discussed the memory waste given that they are not used most of the time. We decided to not worry, given that they are only stored once per neuron, not once per synapse (they are accessible as synaptic variables, but that's just indexing magic).

Yes. Forgot that when I wrote my last comment. I agree.

> In principle it might be useful to have a general mechanism to just copy over variables to device code that are actually used on the device side? You wouldn't have to parse any code for that, you could look at the variables dictionaries of all code objects.

Yes I would love to refactor that variable initialisation code and have a good mechanism for host to device copies. It just grew out of Konrads code and I just added more and more exceptions or started deleting variables that are initialised and not used etc... its quite messy. But I think we don't have time for such a change right now. And for `lastupdate` such a variable inspection mechanism wouldn't work as it is used in the update code.

So the implementation here (which I am currently using) breaks brian2genn performance (see brian-team/brian2genn#64). If there is some decision to remove `lastupdate`, that will need some change in `brian2genn` I guess.

How about implementing your suggestion of getting rid of `lastupdate` entirely for not event-driven synapses? The check is very easily implemented, I guess it would be more about documenting it (and well, the decision to do it). So what do you think? I would love to see `lastupdate` gone for our result benchmarks :)

mstimberg:
This has been made obsolete by PR #1003. Thanks for brining this up @denisalevi !

PR #981, opened 2018-07-12
[MRG] Float32 support

mstimberg:
Due to summer vacations this will still take a while to merge I think, but I'm putting it out here since it is basically ready from my side and it should be of high interest to the @brian-team/brian2cuda and @brian-team/brian2genn backends.

There are a few tests (mostly the multicompartmental stuff) that test numerical results vs. theoretical predictions that are just off too far with single precision, so I decided to simply skip them when single precision is used.  In the long run, we should also probably split these type of tests from the more basic unit/integration tests.

Travis and appveyor run tests with single precision (but only for Python 2.7, the test suits already take way too long...) and everything seems to work fine.

A remark about the clock: The variables `t` and `dt` always use double precision regardless of the preference setting, because everything gets very imprecise otherwise. To avoid problems, other variables or subexpressions that are direct functions of time (e.g. `lastspike`) are using the same dtype.


denisalevi:
Hi @mstimberg. Does brian2genn already work out of the box with this brian2 preference by any chance? :) or would it be possible to add it for our comparisons (I imagine it to be easy to translate the brian pref to genn pref?)

mstimberg:
No, it doesn't work with Brian2GeNN (it's not even merged in Brian 2 yet :) ). In principle, it should be easy to translate, but I see potential problems with variables like `lastupdate` that we force to be double precision in Brian.

brian2/core/core_preferences.py
```diff
@@ -14,16 +14,14 @@ def dtype_repr(dtype):
 
 
 def default_float_dtype_validator(dtype):
-    return dtype is float64
+    return dtype in [float32, float64]
```
thesamovar: I don't think we need it as a priority, but I wonder how much work it would be to allow for other types such as float16 or float128?

brian2/monitors/spikemonitor.py
```diff
@@ -237,16 +237,18 @@ def values(self, var):
         Examples
         --------
         >>> from brian2 import *
-        >>> G = NeuronGroup(2, """dv/dt = 100*Hz : 1
+        >>> G = NeuronGroup(2, """dv/dt = 100.0001*Hz : 1
```
thesamovar: This is really ugly because it's actually in the documentation. Is there no way around this?

brian2/monitors/spikemonitor.py
```diff
@@ -237,16 +237,18 @@ def values(self, var):
...
-        >>> G.v_th = [0.5, 1]
+        >>> G.v_th = [0.5, 1.0]
         >>> mon = EventMonitor(G, event='spike', variables='v')
         >>> run(20*ms)
         >>> v_values = mon.values('v')
-        >>> v_values[0]
-        array([ 0.5,  0.5,  0.5,  0.5])
-        >>> v_values[1]
-        array([ 1.,  1.])
+        >>> np.set_printoptions(precision=4)  # show fewer digits than default
```
thesamovar: Same here. I hate to have stuff like this actually in our docs.

brian2/tests/test_GSL.py
```diff
@@ -103,8 +103,8 @@ def test_GSL_default_function():
 
         assert max(trace_conventional.v[0]-trace_GSL.v[0]) < max_difference/mV, \
                 ('difference between conventional and GSL output is larger than max_difference')
-        assert not all(trace_conventional.v[0]==trace_GSL.v[0]), \
-                ('output of GSL stateupdater is exactly the same as Brians stateupdater (unlikely to be right)')
+        # assert not all(trace_conventional.v[0]==trace_GSL.v[0]), \
+        #         ('output of GSL stateupdater is exactly the same as Brians stateupdater (unlikely to be right)')
```
thesamovar: Why is this commented out?

brian2/tests/test_neurongroup.py
```diff
@@ -654,7 +655,7 @@ def test_threshold_reset():
                     threshold='v > 1', reset='v=0.5')
     G.v = np.array([0, 1, 2])
     run(defaultclock.dt)
-    assert_equal(G.v[:], np.array([0, 1, 0.5]))
+    assert_allclose(G.v[:], np.array([0, 1, 0.5]))
```
thesamovar: Comment not about this particular instance, but I wonder if we lose anything by weakening our tests like this? My feeling is probably not, but it makes me mildly anxious. I guess you've already pretty much covered this with the float-type dependent version of allclose.

thesamovar:
Oh, by the way do the CI tests now run for float32 and float64?

brian2/core/core_preferences.py
mstimberg: I don't think we'd need any additional changes for that, it's just more effort for testing. But we could certainly think about it again when a use case arises.

brian2/monitors/spikemonitor.py
mstimberg: I'll have a look, we can certainly come up with a different kind of example.

brian2/tests/test_neurongroup.py
mstimberg: In general I think we need to do a better job separating our tests for basic functionality from the tests for mathematical/numerical correctness. Currently, we only have a few tests (e.g. for STDP) that compare the exact results for non-trivial simulations.

brian2/tests/test_GSL.py
mstimberg: Ah, right. The test was a bit weird, checking that the results of two state updaters for the same equations were *not* exactly the same. It was more a "during-development" test, Charlee used it to make sure that the simulation was actually using her state updater. I think this test failed with `float32`, i.e. the results were exactly the same. We can simply delete it.

mstimberg:
> Oh, by the way do the CI tests now run for float32 and float64?

They do, or almost: we are testing float32 on all codegen targets and on all platforms, but only for Python 2 (our test suite is already huge as it is...)

thesamovar:
If I just ``run_nose_tests.py`` locally will I get float64? How can I run with ``float32`` locally?

brian2/monitors/spikemonitor.py
thesamovar: There are quite a few examples like this (not just this one).

brian2/tests/test_neurongroup.py
thesamovar: Worth creating an issue for? I'm not entirely sure it's necessary, I feel like ``allclose`` is probably as good as ``equal`` practically. Can we think of an example where it might be a problem that it passes allclose but not equal?

mstimberg:
> If I just run_nose_tests.py locally will I get float64? How can I run with float32 locally?

By default it uses float64, to get `float32` you'll have to pass `float_dtype=np.float32` to `brian2.test`

thesamovar:
Can we make that argument a list so that it runs all dtypes in the list? Then we can modify the standard ``run_nose_tests`` to be the default, but maybe ``run_nose_tests_long_and_standalone`` could run for both? It would be nice to have the ``run_examples`` have this option as well. I think it's good to do a more complete test run of this sort before releases. I usually do these last two.

thesamovar:
I'm getting a lot of this error locally:

```
ERROR: brian2.tests.test_subgroup.test_synapses_access_subgroups_problematic
    self.test(*self.arg)
  File "D:\programming\brian2\brian2\tests\test_subgroup.py", line 467, in test_synapses_access_subgroups_problematic
    S.w.__setitem__(item, value)
  File "D:\programming\brian2\brian2\core\variables.py", line 928, in __setitem__
    self.set_item(item, value, level=1)
  File "D:\programming\brian2\brian2\core\variables.py", line 922, in set_item
    run_namespace=namespace)
  File "D:\programming\brian2\brian2\core\base.py", line 276, in device_override_decorated_function
    return func(*args, **kwds)
  File "D:\programming\brian2\brian2\core\variables.py", line 993, in set_with_expression
    indices = np.atleast_1d(self.indexing(item))
  File "D:\programming\brian2\brian2\synapses\synapses.py", line 543, in __call__
    assume_unique=True)
  File "C:\Anaconda\lib\site-packages\numpy\lib\arraysetops.py", line 368, in intersect1d
    ar1 = ar1.ravel()
AttributeError: 'list' object has no attribute 'ravel'
```

mstimberg:
> I'm getting a lot of this error locally:

That's already fixed in master, I'll merge it into this branch in a bit.

mstimberg:
> Can we make that argument a list so that it runs all dtypes in the list? Then we can modify the standard run_nose_tests to be the default, but maybe run_nose_tests_long_and_standalone could run for both? It would be nice to have the run_examples have this option as well.

That would make the `brian2.test` function a bit cumbersome, with everything moving into an additional loop. How about just modifying the `run_nose_tests_long_and_standalone` script so that it runs `brian2.test` twice?

thesamovar:
Also this:

```
ERROR: brian2.tests.test_stateupdaters.test_check_for_invalid_values_linear_integrator
    self.test(*self.arg)
  File "D:\programming\brian2\brian2\tests\test_stateupdaters.py", line 727, in test_check_for_invalid_values_linear_integrator
    run(1*ms)
  File "D:\programming\brian2\brian2\units\fundamentalunits.py", line 2375, in new_f
    result = f(*args, **kwds)
  File "D:\programming\brian2\brian2\core\magic.py", line 371, in run
  File "D:\programming\brian2\brian2\core\magic.py", line 231, in run
  File "D:\programming\brian2\brian2\core\base.py", line 276, in device_override_decorated_function
    return func(*args, **kwds)
  File "D:\programming\brian2\brian2\units\fundamentalunits.py", line 2375, in new_f
    result = f(*args, **kwds)
  File "D:\programming\brian2\brian2\core\network.py", line 951, in run
    self.before_run(namespace)
  File "D:\programming\brian2\brian2\core\base.py", line 276, in device_override_decorated_function
    return func(*args, **kwds)
  File "D:\programming\brian2\brian2\core\network.py", line 843, in before_run
  File "D:\programming\brian2\brian2\core\network.py", line 841, in before_run
    obj.before_run(run_namespace)
  File "D:\programming\brian2\brian2\groups\group.py", line 1115, in before_run
    codeobj_class=self.codeobj_class
  File "D:\programming\brian2\brian2\codegen\codeobject.py", line 240, in create_runner_codeobj
    run_namespace=run_namespace)
  File "D:\programming\brian2\brian2\groups\group.py", line 739, in resolve_all
    run_namespace=run_namespace)
  File "D:\programming\brian2\brian2\groups\group.py", line 696, in _resolve
    return self._resolve_external(identifier, run_namespace=run_namespace)
  File "D:\programming\brian2\brian2\groups\group.py", line 807, in _resolve_external
    (identifier))
KeyError: 'The identifier "re" could not be resolved.'

Error encountered with object named "neurongroup_1_stateupdater".
    self.test(*self.arg)

An error occurred when preparing an object. KeyError: 'The identifier "re" could not be resolved.'
```

mstimberg:
> Also this:

Same, fixed in master (or run with sympy < 1.2)

thesamovar:
OK let me know when you've merged master into this and I'll rerun everything.

thesamovar:
Well it's nice to be able to look at the final result of ``brian2.test()`` and see that it says "All tests passed successfully" which wouldn't work so easily like that. Worth the effort?

mstimberg:
> Well it's nice to be able to look at the final result of brian2.test() and see that it says "All tests passed successfully" which wouldn't work so easily like that. Worth the effort?

How about we add addional summarizing output to `run_nose_tests_long_and_standalone`? `brian2.test()` returns the success or failure, so we can say something like:
```
Test for float32: FAILED, Test for float64: PASSED
Not all tests passed (see above)
```

thesamovar:
That works too.

mstimberg:
I pushed all changes and merged in recent master. How about we merge as soon as you confirm that tests pass on your machine? There are two more small PRs that I'd like to merge, and waiting >5h between each merge is a bit long. We should certainly do thorough testing (long tests, examples) before a release, of course.

thesamovar:
Running now, but will take a while.

mstimberg:
Ok, no worries. I'll actually have to go now, so we'll have to finish all this up on Monday :-/

thesamovar:
OK, the modification to the long test didn't work because of device stuff I think. Many errors of the form ``NOT_IMPLEMENTED: Cannot retrieve the values of state variables in standalone code before the simulation has been run.``

mstimberg:
Apologies, there were indeed a few issues (they were not really related to the changes in this PR, our test suite was simply not working when run repeatedly...). The long+standalone test suite runs through for me now, and tests on the CI servers seem to be working fine as well.

mstimberg:
Oh, and it seems to require only a trivial change in Brian2GeNN to make it work with the new preference setting. But I'm not sure about one thing, maybe @tnowotny can shed some light on this: what does it actually change to use `setPrecision(GENN_FLOAT)` or `setPrecision(GENN_DOUBLE)` on a model? For all individual variables, we set their type already with something like `n.varTypes.push_back("float");` or `n.varTypes.push_back("double");`...

thesamovar:
Still getting ``NotImplementedError`` on standalone when running long tests.

mstimberg:
> Still getting NotImplementedError on standalone when running long tests.

When exactly, on what kind of tests?

thesamovar:
Full output here (up to the point where I cancelled the test): https://gist.github.com/thesamovar/a2f18ae1f326c6638b0c0d0f87b92f30

thesamovar:
Git says it's up to date.

thesamovar:
That line is there, yep. Might need a bit of help debugging this one. Will be around tomorrow afternoon.

mstimberg:
For debugging this, can you see whether you see the same problem when you run the `run_nose_tests_long_and_standalone.py` but replace:
```python
success = [brian2.test(long_tests=True, test_standalone='cpp_standalone', float_dtype=np.float32),
           brian2.test(long_tests=True, test_standalone='cpp_standalone', float_dtype=np.float64)]
```
by
```python
success = [brian2.test([], test_standalone='cpp_standalone', float_dtype=np.float32),
           brian2.test([], test_standalone='cpp_standalone', float_dtype=np.float64)]
```
If yes, that should at least make the debugging much faster!

mstimberg:
Thanks for clearing this up @tnowotny . I don't see us using `scalar x` anywhere in the Brian2GeNN code, so I think it should work just fine even without us setting the precision. Still, it makes sense to set it to avoid confusion.

thesamovar:
Yep that works! Will open brian dev gitter if you want to debug together.

mstimberg:
> Will open brian dev gitter if you want to debug together.

I'm in the gitter chat (not sure whether it's the right one :) )

tnowotny:
However, I just remembered a couple of other uses of the ```setPrecision``` information:
- we replace literal constants with their appropriate single precision float versions in model code, e.g. ```3.1415``` becomes ```3.1415f```
- we replace math function names with explicit precision variants, e.g. ```sin()``` becomes ```sinf()``` 
- the global variable ```t``` that handles time (even though as derived from a long integer variable ```iT``` that does the actual counting) is defined according to the current precision. 

All of these make it a good idea to set the precision.

mstimberg:
> Tests now pass!

Same here, I'll merge it right away even though the appveyor tests did not pass yet (this will take > 5h still...)

mstimberg:
> the global variable t that handles time (even though as derived from a long integer variable iT that does the actual counting) is defined according to the current precision.

This is a problem, unfortunately, since in Brian `t` is always stored as a double variable... I don't think we will change this, since `t` is so ubiqutous in user expressions and with single precision, stuff like `mon.v[mon.t>=1*ms]` will often not do what the user thinks it should do (since `1*ms` is a double). Even just for display, things get ugly:
```pycon
>>> Quantity(1e-3, dtype=np.float32, dim=second.dim)
1.00000005 * msecond
```

I see two solutions:
1. always use `GENN_DOUBLE` as the precision for GeNN -- individual variables will still have the dtype that Brian defines
2. fix Brian2GeNN so that it can work with GeNN's single precision `t`

(1) is trivial to do, but of course we lose most of the performance benefits when every function call etc. still uses double precision.

I'll look into (2), maybe it is not that difficult to fix. We can still raise a warning to make users aware that Brian2GeNN is not calculating exactly the same thing as Brian.

tnowotny:
I can see the logic of using double precision for ```t```. We made the "clean" float version originally for older GPUs which *did not support doubles at all*. In that case we had to use float for everything. I suspect today that is pure history - opinions @neworderofjamie?
I think solution (1) would be quite counter-productive even though I am also not sure how the different compilers handle calls to ```sin(x)``` where ```x``` is a single precision float. Do they auto-cast ```x``` to ```double``` or resolve to a ```float``` precision ```sin``` function "C++"-style?

neworderofjamie:
Woah this is prescient! In the past (https://github.com/genn-team/genn/issues/183) we talked about abandoning floating point time altogether and I started implementing that. However it totally breaks backwards compatibility and ends up making the model code really ugly so **I** was thinking adding an option to make time double precision is probably a pragmatic solution - single precision t gets totally mangled after running my STDP model for 2000s :cry:

mstimberg:
> I think solution (1) would be quite counter-productive even though I am also not sure how the different compilers handle calls to sin(x) where x is a single precision float. Do they auto-cast x to double or resolve to a float precision sin function "C++-style"?

C++11 (which I think is required by GeNN?) has overloaded math functions for single precision.

neworderofjamie:
Sadly I think the CUDA maths functions are C-style - as in the float ones have an f at the end rather than relying on overloading

tnowotny:
hum ... @neworderofjamie, how were you thinking to make a double precision time version? We could introduce a new GeNN flag for it and then just make ```t``` double precision based on it - something like that?

tnowotny:
Brian2GeNN then could just set that GeNN flag and of we would go ... I am probably missing something.

mstimberg:
This sounds good to me. But a general question: is there a way to find out about the GeNN version "at runtime"? Even a simple thing like a `version.txt` file in the GeNN directory would be enough. This way we could check from Brian2GeNN what GeNN version is used and only set an option if it is supported, warn the user otherwise, etc.

neworderofjamie:
That would be exactly my suggestion

tnowotny:
@neworderofjamie, that is very simple to do. I guess we would rely on the compiler to auto-cast whenever ```t``` is used in some simulation code, but may not be the worst thing.
I guess we should do it.
@mstimberg about the version: I don't think we have anything in place but sounds like a really sensible thing to have.

denisalevi:
> Sadly I think the CUDA maths functions are C-style - as in the float ones have an f at the end rather than relying on overloading

As far as I know, CUDA math functions are overloaded for single and double precision. At least in CUDA 9.0 (no idea about pervious versions). See [here](https://docs.nvidia.com/cuda/cuda-math-api/group__CUDA__MATH.html). Even though there is only an example for `log` given. Correct me if I'm wrong.

tnowotny:
Looks like it in the documentation you referenced. In that case we could relax with our ```ensureFtype``` logic and not modify function names for single precision (at least from CUDA 9 onwards).
Independently, this would make @mstimberg's option 1) viable as well (at least from CUDA 9 onwards).

neworderofjamie:
Sadly, for reasons best known to NVIDIA, that seems to **only** apply to log - see https://docs.nvidia.com/cuda/cuda-math-api/group__CUDA__MATH__SINGLE.html#group__CUDA__MATH__SINGLE

neworderofjamie:
Actually, apologies, that is just terrible documentation! In ``maths_functions.hpp`` there's lots of the following:
```c++
__host__ __device__ __cudart_builtin__ float     sin(float in)           { return sinf(in); }
```
I will check older CUDA's and see if this is consistent - I would IMAGINE it came in when CUDA started supporting C++.

mstimberg:
I don't think that's the case, their documentation is just... not that good. The `log` function is just an example (on the page you linked, there's also `logf`, and the double precision docs also pretend that `log` is double-only).

It seems that this has been in CUDA for a long time, I've found "Introduction to CUDA" slides from 2011 that mention that math functions are overloaded.

mstimberg:
Let's continue discussions about float32 support in Brian2GeNN here: brian-team/brian2genn#69

PR #982, opened 2018-07-18
Make translation of abstract code codeobject-dependent in `CodeGenerator`

denisalevi:
The `CodeGenerator` class translates all abstract code blocks in the same way for all codeobjects. For brian2cuda I need this to be template dependent, since some templates execute host code and others device code, for which I sometimes want to use `atomicAdd` in the created vector code.

I have simply added an optional keyword argument dictionary to the `Device.code_object()`, which gets passed down through `CodeObject.translate()` until `CodeObject.translate_one_statement_sequence()`, which gets overwritten in the `CUDACodeObject(CodeObject)` class.

What do you think? Is there another way to achieve what I want or are there concerns about this?

### DIFF
--- a/brian2/codegen/generators/base.py
+++ b/brian2/codegen/generators/base.py
@@ -100,10 +100,12 @@ class CodeGenerator(object):
         '''
         raise NotImplementedError
 
-    def translate_one_statement_sequence(self, statements, scalar=False):
+    def translate_one_statement_sequence(self, statements, scalar=False,
+                                         **additional_kwds):
         raise NotImplementedError
 
-    def translate_statement_sequence(self, scalar_statements, vector_statements):
+    def translate_statement_sequence(self, scalar_statements, vector_statements,
+                                     additional_kwds):
         '''
         Translate a sequence of `Statement` into the target language, taking
         care to declare variables, etc. if necessary.
@@ -117,9 +119,13 @@ class CodeGenerator(object):
         scalar_code = {}
         vector_code = {}
         for name, block in scalar_statements.iteritems():
-            scalar_code[name] = self.translate_one_statement_sequence(block, scalar=True)
+            scalar_code[name] = self.translate_one_statement_sequence(block,
+                                                                      scalar=True,
+                                                                      **additional_kwds)
         for name, block in vector_statements.iteritems():
-            vector_code[name] = self.translate_one_statement_sequence(block, scalar=False)
+            vector_code[name] = self.translate_one_statement_sequence(block,
+                                                                      scalar=False,
+                                                                      **additional_kwds)
 
         kwds = self.determine_keywords()
 
@@ -215,9 +221,12 @@ class CodeGenerator(object):
                          if index not in ('_idx', '0'))
         return not all_unique
 
-    def translate(self, code, dtype):
+    def translate(self, code, dtype, additional_kwds):
         '''
         Translates an abstract code block into the target language.
+
+        In additional_kwds, codeobject dependent parameters can be passed (for
+        devices that inherit from the CodeObject class).
         '''
         scalar_statements = {}
         vector_statements = {}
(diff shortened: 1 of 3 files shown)


denisalevi:
I just realised that the `scalar` keyword arguments are coming from a similar situation you had with the numpy vectorisation, right? If you want to give the flexibility to add whatever arguments (as my implementation right now does), one should probably move the `scalar` keyword in the `additional_kwds` dictionary?

mstimberg:
I'm not completely opposed to handing down additional keywords, but it goes a bit against the idea of a `Generator` -- those should translate code snippets and not worry about templates. Do you only need two types of generators? If this is the case, I think the best solution would be to use separate `CodeObject`/`CodeGenerator` classes for both, and decide which of them to use in your `Device` based on the template. Brian2GeNN does this as well: https://github.com/brian-team/brian2genn/blob/feaa8bc90cf927718e5ed2c53ccdbad38447d629/brian2genn/device.py#L372

denisalevi:
All I currently need is to decide weather or not I try to use cuda atomic operations, which is basically the same as telling the `NumpyCodeGenerator` to try to vectorise the code only for some codeobjects (and for other directly use the Python loop fallback). Just that for CUDA we don't have a "bad" fallback option, but our fallback option is to use the standard version for vector code (as cpp_standalone does) and depending on the success, we call the kernel with different numbers of threads and blocks.

So yes, I could just inherit from `CUDACodeGenerator` and change a global class attribute and have two generators in that case. If that's the preferred way?

denisalevi:
But would I have to add all the `DEFAULT_FUNCTION` implementations to the second `CodeGenerator` class as well?

mstimberg:
If your generator class inherits from another generator class, then it should inherit the function definitions.

denisalevi:
Hi @mstimberg I have implemented two `CodeObject`/`CodeGenerator` classes and I change the default in `Device` when needed (as you suggested). I just see one problem: When a user sets the `codeobj_class` through `Synapses(..., codeobj_class=...)`, this is simply ignored without any warning. I guess thats the same for brian2genn.
I didn't find documentation for the `codeobj_class` keyword so I'm no sure if it is actually used. But it would in general allow us to decide on some preference on a synapses object basis. We will probably have two synaptic propagation modes for heterogeneous delays. The default is a bundle mode, but this might be inefficient in certain delay distribution cases. And I'm not sure if we will have a good heuristic for that. So it would be nice to be able to set the propagation mode to 'not bundles' for a specific `Synapses` object. And one could achieve this through different code object classes. But is this intended for that? Or is there another mechanism that would allow this?

mstimberg:
Hi, do you mean that the user would chose it for a specific `Synapses` object? That sounds feasible. 
The `codeobj_class` argument is used in general, but Brian2GeNN does indeed ignore it. But this just sloppiness, Brian2GeNN's `Device.codeobj_class` function should simply be defined as done for `CPPStandaloneDevice`:
```Python
    def code_object_class(self, codeobj_class=None, *args, **kwds):
        if codeobj_class is None:
            codeobj_class = GeNNUserCodeObject

        return codeobj_class
```
But then, I think `Brian2GeNN` might overwrite this choice later... Either way, feel free to *not* do this for Brian2CUDA and only set the `codeobj_class` explicitly in `Device.code_object` if it isn't `None` :)

denisalevi:
Our `code_object_class` method looks exaclty like that. My problem was another, but I just realized its no problem... :)
The thing is that the `codeobj` is set already in the `codeobject_runner` function (or similarly called). By the time it reaches our `CUDAStandaloneDevice.code_object` method, we can't differentiate weather a user passed the default `codeobject` class to e.g. a `Synapses` object or just `None` (and the default was chosen in `code_object_class`). But I guess the solution is just to have another `codeobject` class that behaves like the default but is called different, to separate those two cases.
Anyways, not so important right now. I guess this is done. I'm closing the PR. Thanks for your help :)

Issue #815, opened 2017-02-16
integer division has different semantics in Python 2 and 3 -- only for the numpy codegen target


mstimberg:
It should not come as a surprise, but I never actually thought about it in detail: we have some ugly inconsistency for integer division between Python 2 and 3. At the first sight, this could be ok because after all Python 2 and 3 *do* have different semantics in that regard, so this could be reflected in our interpretation of the code. The problem is that the semantics in C code are always "Python 2"-like and too complicate further, C-based runtime simulations by default use numpy for assigning to variables (to avoid the compilation overhead for trivial computation). 
For example, consider the following code:
```Python
from brian2 import *

G = NeuronGroup(10, '''x : 1
                       y : 1''')
G.x = 'i / 10'  # this uses numpy by default (prefs.codegen.string_expression_target)
G.run_regularly('y = i / 10')  # this uses the standard codegen target
run(defaultclock.dt)
print(G.x)
print(G.y)
```
In Python 2, regardless of the codegen target, `G.x` and `G.y` will always be 0 (floor division). In Python 3 however, you'll get this for `prefs.codegen.target = 'numpy'`:
```
<neurongroup.x: array([ 0. ,  0.1,  0.2,  0.3,  0.4,  0.5,  0.6,  0.7,  0.8,  0.9])>
<neurongroup.y: array([ 0. ,  0.1,  0.2,  0.3,  0.4,  0.5,  0.6,  0.7,  0.8,  0.9])>
```
(floating point division: Python3-semantics)
You'll get this for `prefs.codegen.target = 'cython'` *and* `prefs.codegen.string_expression_target = 'cython'`:
```
<neurongroup.x: array([ 0.,  0.,  0.,  0.,  0.,  0.,  0.,  0.,  0.,  0.])>
<neurongroup.y: array([ 0.,  0.,  0.,  0.,  0.,  0.,  0.,  0.,  0.,  0.])>
```
(floor division: C semantics)
And, probably worst of all, you'll get this if you don't set any preference and have a working Cython installation (i.e. `prefs.codegen.target` will be `'auto'` and use Cython, `prefs.codgen.string_expression_target` will be at its default `'numpy'`):
```
<neurongroup.x: array([ 0. ,  0.1,  0.2,  0.3,  0.4,  0.5,  0.6,  0.7,  0.8,  0.9])>
<neurongroup.y: array([ 0.,  0.,  0.,  0.,  0.,  0.,  0.,  0.,  0.,  0.])>
```
[...]

Issue #812, opened 2017-02-15
str_to_sympy converts integers into floats


mstimberg:
When converting a string to sympy (i.e., for use in equations), all integer literals are transformed to float literals and when variables are involved in certain operations, they are multiplied by 1.0. E.g.:
```Python
>>> from brian2 import *
>>> from brian2.parsing.sympytools import str_to_sympy
>>> from brian2.core.variables import ArrayVariable
>>> str_to_sympy('1 + i % 25', variables={'i': ArrayVariable('i', None, 100, get_device(),
...                                                          dtype=np.int32)})
Mod(1.0*i, 25.0) + 1.0
```

We do of course rarely deal with integers in differential equations and if we do, it normally doesn't make a difference whether they are represented as integers or as equivalent floating point variables. It *could* make a difference though e.g. when the user wants to use integer division in an equation:
```
dv/dt = (i/100) / second : 1
```

To do this properly, we also have to take care of functions that return integers. E.g. even the following workaround does not work at the moment:
```
dv/dt = (int(i)/int(100)) / second : 1
```
Since sympy does not know that these are integers and that there is anything special about integer division it will happily convert this to:
```
int(i)/(second*int(100.0))
```
which defeats the purpose.

You could of course use `int(i/100)` here, but this is still different from other places (e.g. synaptic `connect` calls) where it will use integer division without any explicit `int` call.

I think it is important that sympy does not screw with the type of literals/variables (we take some care to not do this in our code gen pipeline), but there's also the somewhat orthogonal question of how to deal with integer division. Currently, we use C/Python2 semantics and have example where we use this e.g. to arrange neurons in a grid (something like `x = i / gridsize`, `y = i % gridsize`).

Maybe it would be better to make this clear once and for all and use `/` for floating point and `//` for integer division as in Python 3? We have to be careful when introducing this, e.g. use multiple steps with a `FutureWarning` that the semantics will change and later a warning when the semantics did change. This certainly needs some careful thought and implementation (it is also related to #661 )

thesamovar:
Another possibility would be to shield all integer subexpressions from sympy, i.e. any subexpression that matches as integer would be replaced by a named variable and so there wouldn't be any simplifications involving it, and then at the end that named variable would be replaced by the original expression. Bit convoluted, but probably matches intended behaviour. Also easy to explain: sympy will simplify any float expression but not any integer expression. I'm hesitant to introduce the Python 3 semantics while we still support Python 2, because many people won't have encountered it.

[...]

Issue #597, opened 2015-11-19
Put the code for brian_mod into a separate h file?


thesamovar:
At the moment the debug logs are hard to read because of it.

mstimberg:
I agree that this makes things hard to read currently. How about creating a header file for every code object and including that instead of using weave's support code? We could use something like the hash of this file in the comments of the support code to make sure we get a recompile if something in the functions changed.

thesamovar:
Well I'd like to have most of the support code shown because sometimes that's what you're debugging, but the mod stuff is working fine now so could be hidden in a header.

thesamovar:
Actually this isn't really a high priority so I'm removing the 2.0 milestone.

mstimberg:
I agree. It is quite ugly whenever I look at generated code, but users tend to do this less often than we do ;)

PR #984, opened 2018-07-25
[MRG] Establish consistent division semantics

### LINKED ISSUE #661: Integer division is different for different languages
`-1/2==-1` in Python (and numpy), but `-1/2==0` in C++.

We handled this for `a%b` but not for integer division. We should either handle this or at least warn about it somewhere.

mstimberg:
This fixes #661 and #815 by enforcing Python3 division semantics everywhere for integers, i.e. `/` is floating point division, `//` is integer division. I think this is the best solution and we seem to have somewhat converged on it in the discussions in #815. As a side note, about 2/3 of our Anaconda package downloads are for Python 3 now.

Unfortunately, this PR changes the division semantics for *all* code generation targets except for the numpy target on Python 3. On the other hand, I think it was probably mostly used in assignments, e.g. to make a grid of neuron positions. As these assignments are using the numpy target by default, the change will primarily affect Python 2 users (and they are now a minority). 

Either way, this is an important change that can affect the results, I have therefore added warnings. E.g. in the unchanged Kremer et al. example, you'd get warnings like this:

> WARNING    The expression "i / N4" divides two integer values. In previous versions of Brian, this would  have used either an integer ("flooring") or a floating point division, depending on the Python version and the code generation target. In the current version, it always uses a floating point division. Explicitly ask for an  integer division ("//"), or turn one of the operands into a floating point value (e.g. replace "1/2" by "1.0/2") to no longer receive this warning. [brian2.parsing.bast.floating_point_division]

As the warning states, you only get the warning when there is room for ambiguity, i.e. the division of two integers via `/`. In the above example, the user should replace the expression by `i // N4`, but if they actually wanted floating point division, then something like `1.0*i/N4` would of course still work. I added a little extra check to avoid a warning for expressions like `int(i / N4)` and `floor(i / N4)`, because I've been using them myself and recommending them to other users as unambiguous integer division in current Brian without `//` support.

From my side, the PR is basically complete (and tests pass at least on my machine), but I still need to mention the changed semantics in the documentation, and check the examples.


thesamovar:
I agree with all this. I'm wondering if we should bump the version to 2.2 with this change? I'm also wondering if the warning should be permanent or removed at some future release? For example, when official support for Python 2 ends in 2020 perhaps?

Will do the review when everything passes and docs etc. are done.

mstimberg:
@thesamovar This is now done from my side.

thesamovar:
OK, will review soon. What do you think about my previous comment?

thesamovar:
Numpy floor returns a float not an int, so maybe that shouldn't be an exception?

mstimberg:
> Numpy floor returns a float not an int, so maybe that shouldn't be an exception?

You mean an exception to the warning?

brian2/codegen/optimisation.py
```diff
@@ -248,6 +248,16 @@ def render_BinOp(self, node):
...
+        elif op.__class__.__name__ == 'FloorDiv':
+            if left.__class__.__name__ == 'Num' and left.n == 0:  # 0//x
+                if node.stateless:
+                    # Do not remove stateful functions
+                    return _replace_with_zero(left, node)
+            # Only optimise floor division by 1 if the number is an integer, for
+            # floating point values, floor division by 1 changes the value
+            if (left.dtype == 'integer' and
+                    right.__class__.__name__ == 'Num' and right.n == 1):  # x//1
+                    return left
```
thesamovar: I think this needs an extra check, because ``3//1.0`` should be ``3.0`` not ``3`` and I think this last bit will replace it with ``3``?

brian2/parsing/bast.py
```diff
@@ -192,19 +204,53 @@ def render_Call(self, node):
...
         node.right = self.render_node(node.right)
         # TODO: we could capture some syntax errors here, e.g. bool+bool
         # captures, e.g. int+float->float
         newdtype = dtype_hierarchy[max(dtype_hierarchy[subnode.dtype] for subnode in [node.left, node.right])]
+        if node.op.__class__.__name__ == 'Div':
+            # Division turns integers into floating point values
+            newdtype = 'float'
+            # Give a warning if the code uses floating point division where it
+            # previously might have used floor division
+            if node.left.dtype == node.right.dtype == 'integer':
```
thesamovar: Is it worth also warning on things like ``True/2`` that would have returned 0 but now 0.5? In other words, dealing with bool as well as integer? Actually, we might like to think about that throughout?

brian2/parsing/expressions.py
```diff
@@ -336,6 +338,11 @@ def parse_expression_dimensions(expr, variables):
             u = left_dim*right_dim
         elif op == 'Div':
             u = left_dim/right_dim
+        elif op == 'FloorDiv':
+            if not (left_dim is DIMENSIONLESS and right_dim is DIMENSIONLESS):
+                raise SyntaxError('Floor division can only be used on '
+                                  'dimensionless integer values.')
```
thesamovar: Can also be used on non-integer values? In principle if the output is dimensionless it still makes sense?

brian2/parsing/rendering.py
```diff
@@ -254,6 +255,11 @@ def render_BinOp(self, node):
             op = self.expression_ops['Mult']
             return op(self.render_node(node.left),
                       1 / self.render_node(node.right))
+        elif op_name == 'FloorDiv':
+            op = self.expression_ops['Mult']
+            left = self.render_node(node.left)
+            right = self.render_node(node.right)
+            return sympy.floor(op(left, 1 / right))
```
thesamovar: This is a general comment but is it an issue that we convert everything into floats in the SympyRenderer? I had a feeling that we excluded all integer operations from being simplified or operated on by sympy but I can't find that now.

brian2/parsing/rendering.py
```diff
@@ -287,15 +293,22 @@ class CPPNodeRenderer(NodeRenderer):
...
                                     self.render_node(node.right))
-        elif node.op.__class__.__name__=='Mod':
+        elif node.op.__class__.__name__ == 'Mod':
             return '_brian_mod(%s, %s)' % (self.render_node(node.left),
                                            self.render_node(node.right))
+        elif node.op.__class__.__name__ == 'Div':
+            # C uses integer division, this is a quick and dirty way to assure
+            # it uses floating point division for integers
+            return '1.0f*%s/%s' % (self.render_element_parentheses(node.left),
+                                   self.render_element_parentheses(node.right))
```
thesamovar: I don't think we ever fixed https://github.com/brian-team/brian2/issues/661 did we? If so, perhaps this PR is the right time to do that? The code above for Div is correct, but we'd (I guess) need to do something a bit more for FloorDiv since ``-1//2==-1`` in Python but ``-1/2==0`` in C.

brian2/parsing/rendering.py
```diff
@@ -287,15 +293,22 @@ class CPPNodeRenderer(NodeRenderer):
           # Bool ops
           'And': '&&',
           'Or': '||',
+          # C does not have a floor division operator (but see render_BinOp)
+          'FloorDiv': '/',
```
thesamovar: We should also support FloorDiv on float values in C I think? At the moment it looks like ``1.0//2.0`` would get ``0.5`` in C whereas it should be ``0.0``.

docs_sphinx/user/equations.rst
```diff
@@ -51,10 +51,44 @@ you use more than one plain ``xi``. Note that noise is always independent across
...
+
+.. warning::
+
+    Brian versions up to 2.1.3.1 did not support ``//`` as the floor division
+    operator and potentially used different semantics for the ``/`` operator
+    depending on whether Python 2 or 3 was used. To write code that correctly
+    and unambiguously works with both newer and older Brian versions, you can
+    use expressions such as ``1.0*a/b`` to enforce floating point division (if
+    one of the operands is a floating point number, both Python 2 and 3 will use
+    floating point division), or ``floor(a/b)`` to enforce flooring division.
```
thesamovar: Maybe add a note that the result of ``floor(a/b)`` will always be a float not an integer?

brian2/codegen/optimisation.py
mstimberg: Good catch!

brian2/parsing/bast.py
mstimberg: I think doing arithmetic operations with integers directly should raise an error, that's why we introduced the `int` function in the first place (that was back in the days when Achilleas worked on BrianDroid, Java does not support this kind of arithmetic). Let's deal with this in a separate PR?

brian2/parsing/expressions.py
mstimberg: Yes, indeed.

brian2/parsing/rendering.py
mstimberg: The float/integer issue in sympy is ugly, but I think we should postpone this until we do our parsing/rendering refactoring. We have #812 open for that, even though the issue becomes less important with the changes here -- the main problem with having floats instead of integers was that it changed division semantics.

brian2/parsing/rendering.py
mstimberg: This PR actually has the side effect of fixing #661 in a few situations. But you are right, we still have to do something for C.

brian2/parsing/rendering.py
mstimberg: Agreed.

docs_sphinx/user/equations.rst
mstimberg: Not sure -- with the new semantics, having a 3.0 or a 3 should not make a difference any more, should it?

thesamovar:
When I run all tests I get errors on ``test_division_semantics()`` in ``test_neurongroup.py`` because the warnings are only shown once, so there are two warnings for numpy target, but then none for weave or cython raising an error. Not an actual problem but test needs to be fixed.

I'm also getting this one for standalone:

```
FAIL: brian2.tests.test_cpp_standalone.test_changing_profile_arg
    self.test(*self.arg)
  File "D:\programming\brian2\brian2\tests\test_cpp_standalone.py", line 381, in test_changing_profile_arg
    profiling_dict['op3_codeobject_1'] > 0*second)
AssertionError
```

My guess is that timing resolution is lower in Windows and something is being rounded down to 0?

brian2/parsing/bast.py
thesamovar: With bools you mean? OK.

brian2/parsing/rendering.py
thesamovar: OK indeed #812 covers this.

docs_sphinx/user/equations.rst
thesamovar: It might if you passed it to a function afterwards? Or if you tried to do ``i = floor(a/b) : integer`` in equations?

brian2/parsing/bast.py
mstimberg: Sorry, yes, with bools.

thesamovar:
Let me know when you want me to re-review.

mstimberg:
> Let me know when you want me to re-review.

I think I addressed all your points, so please have another look. Unfortunately, to fix the division semantics for negative numbers and avoid changing from integer to floating point types I had to 1) introduce a new `_brian_floordiv` function, and 2) switch off `cdivision` for Cython (which will slow down things). Regarding `_brian_floordiv` (and `_brian_mod`), it would be nice to do #597 soonish...

> When I run all tests I get errors on test_division_semantics() in test_neurongroup.py because the warnings are only shown once, so there are two warnings for numpy target, but then none for weave or cython raising an error. Not an actual problem but test needs to be fixed.

Ah, sorry, this should be fixed with the changes in #987 (it clears the logging cache between tests).

Re your standalone problem: let's postpone looking into this, I guess we can simply avoid it by running a computation that runs a bit longer.

brian2/codegen/generators/cpp_generator.py
```diff
@@ -107,7 +107,18 @@ def c_data_type(dtype):
...
 
-_universal_support_code = deindent(mod_support_code)+'''
+floordiv_support_code = ''
+for ix, xtype in enumerate(typestrs):
+    for iy, ytype in enumerate(typestrs):
+        hightype = typestrs[max(ix, iy)]
+        floordiv_support_code += '''
+        {hightype} _brian_floordiv({xtype} x, {ytype} y)
+        {{
+            return ({hightype})floor(1.0*x/y);
```
thesamovar: Can we not do better than this by taking into account whether or not x and y are integer types (similar to how we do it for mod just above)? Also, should be inlined I think.

brian2/codegen/runtime/cython_rt/templates/common.pyx
```diff
@@ -1,7 +1,8 @@
 #cython: boundscheck=False
 #cython: wraparound=False
-#cython: cdivision=True
+#cython: cdivision=False
```
thesamovar: Perhaps we could have a Brian preference that controls whether we do this? Actually, maybe the same preference could control how much care we take to get exactly the same semantics for mod and floordiv too? If the user promised to only use positive values things could be much simpler.

thesamovar:
By the way, I'm running the ``run_examples.py`` test and getting quite a few errors. It hasn't finished yet, so I don't know what's causing them. Will report when it's done. They may have been introduced in other releases since I haven't run this in a while.

brian2/codegen/generators/cpp_generator.py
mstimberg: Unfortunately at this point we can't, we don't have this information here. We only have it when we do the BAST parsing for optimization... With a refactoring, this would be possible, but I'm not sure we should get into this right now. Another possibility: forget about the negative floor division for now and add it as a known issue -- I think this will be a rather uncommon feature to use.

> (similar to how we do it for mod just above)

Not sure what you mean by that -- mod also introduces an extra function `_brian_mod`.

> Also, should be inlined I think.

As a lesson from that most annoying debugging session of my life, I'm now very hesitant to add `inline` for hypothetical performance benefits :)

brian2/codegen/runtime/cython_rt/templates/common.pyx
mstimberg: I'm not very fond of this, I think this is the kind of detail that users should normally not get into and I'd prefer if model semantics would not depend on a preference (imagine setting it in a configuration file and getting different results depending on where you run a simulation!).

thesamovar:
First problem found with the OpenCV example raises a division warning that we should probably fix.

Second problem was Stimberg et al. 2018 example 1 is missing an import sympy.

The rest I think are just an out of date matplotlib on my system. Updating and re-running now.

brian2/codegen/generators/cpp_generator.py
mstimberg: > The C++ compiler will though. Look at the definition of mod above: it creates a different expression depending if the dtypes are integer or float types.

Ok, I now get what you mean. But what could we improve here -- get rid of the 1.0 multiplication if one of the values is already a floating point variable? I think the compiler would easily optimize this away.

> And for inline, we have it on for mod and I don't think it's causing problems? I would guess that for very simple functions like this it should be OK to inline?

There is something odd about the linking and inline functions (remember my horror story from #928), even though I never quite understood why it would be an issue with C++. So I'd say rather `static inline` or no inline at all (same for `_brian_mod`, just to be sure) -- I have the strong feeling that the compiler will make a smarter decision than we do anyway.

brian2/codegen/runtime/cython_rt/templates/common.pyx
mstimberg: Right. I guess I'm fine with it as long as the default is "slow but correct". However, this will complicate the Cython code a bit, since with `cdivision=True` we cannot use `from __future__ import division` and have floating point division for integer values `a/b`.

mstimberg:
> First problem found with the OpenCV example raises a division warning that we should probably fix.
> 
> Second problem was Stimberg et al. 2018 example 1 is missing an import sympy

Ok, let's deal with this separately, we should make sure that all examples run correctly before the next release.

brian2/codegen/generators/cpp_generator.py
mstimberg: > I think we could get rid of the conversion to float and call to floor if both arguments are integer types which would be a big saving. Or am I missing something?

This would give the "truncate to 0" instead of "floor" semantics for negative values, unfortunately. We could add an if statement -- not sure whether this is faster, though?

Happy to use `static inline`, of course.

thesamovar:
Fine for examples. By the way, I just updated to last matplotlib and now the run examples script has a problem: the plots are showing and waiting for user input.

mstimberg:
> By the way, I just updated to last matplotlib and now the run examples script has a problem: the plots are showing and waiting for user input.

Huh, that's odd -- I definitely don't see this. Each example has the following added to its beginning to avoid exactly this:
```python
import matplotlib as _mpl
_mpl.use('Agg')
```

thesamovar:
It's weird - it doesn't happen on every example either. Will look into it another time.

brian2/codegen/generators/cpp_generator.py
thesamovar: It would, but we could use ``a//b==(a-_brian_mod(a, b))/b``, and ``_brian_mod`` doesn't need to convert to float and then back to int (the latter is apparently particularly slow but I didn't check).

mstimberg: Hmm, we could actually use this expression for floating point values as well, no? Maybe avoiding the `floor` function call makes sense there too? Only one way to find out: benchmarking :)

thesamovar: In principle yes, but I assume that ``floor`` is less expensive than ``fmod`` and ``_brian_mod`` on floats calls ``fmod`` twice.

mstimberg: The verdict is pretty clear...
Here's my test code:
```Python
set_device('cpp_standalone', clean=True)
seed(4321)
G = NeuronGroup(10000000, '''x : 1
                             y : 1
                             a : integer
                             b : integer
                             fvalue : 1
                             ivalue : 1''',
                dtype={'x': np.float, 'y': np.double,
                       'a': np.int32, 'b': np.int64})
G.fvalue = 'rand()*100000'
G.ivalue = 'int(rand()*100000)'
G.run_regularly('b = ivalue//10', name='b')
G.run_regularly('a = ivalue//10', name='a')
G.run_regularly('y = fvalue//10', name='y')
G.run_regularly('x = fvalue//10', name='x')
run(defaultclock.dt, profile=True)
print(profiling_summary())
```
If I run this with `floor(1.0*x/y)` as implementation for `_brian_floordiv`, the codeobjects each take about 10ms (a bit less for the 32bit integer, a bit more for the others). If I run it with `(x-_brian_mod(x, y))/y`, it takes about... 550ms! It seems the `floor` function is quite efficient after all.
As a side note, if we required C++11 (as Brian2GeNN already does), we wouldn't have to loop over the type combinations but could simply write:
```C++
template <typename T, typename U>
static inline typename std::common_type<T, U>::type _brian_floordiv(T x, U y)
{
    return floor(1.0*x/y);
}
```

mstimberg: Oh, and removing the 1.0 multiplication for floating point values and just using `x/y` for integer values does not change the runtime in any measurable way for me -- maybe a preference is not needed after all? But there is quite a bit of fluctuation, there might be a small effect.

thesamovar: My mind is blown!

The test above is not quite right because ``ivalue`` should be an integer, but the conclusion is the same. With ``ivalue`` an integer the ``a`` and ``b`` codeobjects are now taking twice as long instead of 10x as long.

And now that we know how fast ``floor`` is on floats, I revisited ``_brian_mod`` and it turns out replacing with the single function ``({hightype})(x-y*floor(1.0*x/y))`` is faster in all cases (including the integer case). So I propose we check that it gives the same values and replace.

mstimberg: Oh, sorry about `ivalue`. I didn't even think about replacing `_brian_mod` as well but that makes of course perfect sense! What do you think about replacing both functions with a single template function and requiring C++11? If I am not mistaken, then the feature we need is "type traits" (for `std::common_type`, and this has been supported by all compilers for a long time (e.g. for gcc since 4.3 and for MSVC since 14.0, i.e. Visual Studio 2005).

thesamovar: I think the only reason we didn't require C++11 already is that Python 2 on Windows is built with MSVC 2008. I have to say I assumed that since C++11 has 11 in the name, it wasn't supported in MSVC 2008 but if that's not the case I would have no objection to using that instead! Would generate much nicer code than the horrible situation we have at the moment.

mstimberg:
I reimplemented the `_brian_mod` function with `floor` and changed the way we handle the different types a bit. I think that's all that is needed for this PR? Or do you still think we should introduce a preference + division functions for Cython?

I'll fix the issue with the "Stimberg et al. 2018" example directly in master.

mstimberg:
How about merging this as it is (with the `cdivision=False` for Cython), and deciding whether it is worth adding the extra code for Cython after seeing what effect it has on benchmarks?

thesamovar:
I'm OK with that. Running the complete test suite and will OK if it passes.

By the way - neat trick with the templated typedef. Much cleaner.

thesamovar:
Currently seeing this error but it doesn't seem to be in a particular test. Some weird feature of the latest version of nose maybe?

```
ERROR: Failure: ImportError (cannot import name AttributeVariable)
  File "C:\Anaconda\lib\site-packages\nose\loader.py", line 418, in loadTestsFromName
    addr.filename, addr.module)
  File "C:\Anaconda\lib\site-packages\nose\importer.py", line 47, in importFromPath
    return self.importFromDir(dir_path, fqname)
  File "C:\Anaconda\lib\site-packages\nose\importer.py", line 94, in importFromDir
    mod = load_module(part_fqname, fh, filename, desc)
  File "D:\programming\brian2\brian2\codegen\languages\__init__.py", line 1, in <module>
  File "D:\programming\brian2\brian2\codegen\languages\base.py", line 5, in <module>
ImportError: cannot import name AttributeVariable
```

mstimberg:
Do you have some files from a (very) old version of Brian lying around? We don't have a `languages` package in `brian2.codegen` anymore.

thesamovar:
Ohhhh probably from when I was bisecting to find that weird bug with the script running twice.

PR #995, opened 2018-08-24
Create Izhikevich_2007.py

deep-introspection:
Script adapted from Fig. 1c of:
Eugene M. Izhikevich 
Solving the distal reward problem through linkage of STDP and dopamine signaling. 
Cerebral cortex 17, no. 10 (2007): 2443-2452.

### DIFF
--- a/examples/frompapers/Izhikevich_2007.py
+++ b/examples/frompapers/Izhikevich_2007.py
@@ -112,7 +112,7 @@ run(simulation_duration/2)
 # Visualisation
 dopamine_indices, dopamine_times = dopamine_monitor.it
 neurons_indices, neurons_times = neurons_monitor.it
-figure(figsize=(10,5))
+figure(figsize=(12,6))
 subplot(411)
 plot([0.05, 2.95], [2.7, 2.7], linewidth=5, color='k')
 text(1.5, 3, 'Classical STDP', horizontalalignment='center', fontsize=20)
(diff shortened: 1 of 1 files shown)


mstimberg:
Thanks @deep-introspection, looking very good! Two very minor comments:
* We recently changed the name of the `linear` method to `exact` -- for consistency with the other examples, could you change it here as well?
* Could you reduce the size of the figure a bit? The figure gets scaled down to fit into the readthedocs column (see e.g. [here](file:///home/marcel/programming/brian2/docs/examples/COBAHH.html)). For your figure that means that the x/y labels and ticks get very small.

examples/frompapers/Izhikevich_2007.py
```diff
@@ -0,0 +1,144 @@
...
+dApre *= gmax
+
+## Dopamine signaling
+tauc = 1000*ms
+taud = 200*ms
+taus = 1*ms
+epsilon_dopa = 5e-3
+
+# Setting the stage
+network = Network()
```
thesamovar: This line doesn't do anything and can be removed.

examples/frompapers/Izhikevich_2007.py
```diff
@@ -0,0 +1,144 @@
...
+
+## Dopamine signaling section
+dopamine_indices = array([0, 0, 0])
+dopamine_times = array([3520, 4020, 4520])*ms
+dopamine = SpikeGeneratorGroup(1, dopamine_indices, dopamine_times)
+dopamine_monitor = SpikeMonitor(dopamine)
+reward = Synapses(dopamine, synapse_stdp, model='''''',
+                            on_pre='''d_post += epsilon_dopa''',
+                            method='exact')
+reward.connect(p=1.)
```
thesamovar: This could be simply ``reward.connect()`` (does all-to-all).

deep-introspection:
Hi @mstimberg & @thesamovar !
Thanks for your comments. Is it better now?
Cheers :)

thesamovar:
Good to merge for me.

thesamovar:
And many thanks for the contribution!

mstimberg:
Merging it now, the figure will appear in the online docs with a little delay, but it will definitely be included with the next release. Thanks again!

Issue #991, opened 2018-08-23
Refractoriness with `timestep` is slow


mstimberg:
We've put some effort into making the newly introduced `timestep` function fairly fast, but I realized that it does still lead to a big slowdown for simple models that use refractoriness, e.g. CUBA. I'm still updating our benchmark suite with new tests that should show this, I'll post a link here when they are ready.

It seems that instead of having a `timestep` function that deals well with infinity it is faster to not have it deal with infinity at all, but instead formulate the refractoriness update as:
```Python
not_refractory = (t == -inf) or (timestep(t - lastspike, dt) >= timestep(refractory_time, dt))
```

**Pros:**
* seems to be faster (no comparison necessary for t values that are -infinity anyway)
* no arbitrary timestep value that we return for infinity

**Cons:**
* Not perfectly backwards compatible for users that started using `timestep` directly in their code since the 2.1.3 release (I doubt there are many)
* In principle this would break if someone gives `inf*second` as the refractory time (as an---albeit inefficient---way to switch it off). We could add another `or refractory_time == inf`, though (not sure how much that slows things down)

mstimberg:
It's not the comparison, I think `isinf` is as efficient as it can be. My hunch is that the difference is that by short-circuiting, we don't do an unnecessary subtraction/comparison/`timestep` call for neurons with `lastspike == -inf`. In the CUBA example, these are quite a few, e.g. after 1s ~30% of the cells never spiked, and after 10s it is still ~5%.

But it seems that the change only has a strong effect when I do another more general optimization at the same time. Our code currently generates stuff like:
```C++
const double dt = _ptr_array_defaultclock_dt[0];
```
which is a bit of a pity, since `dt` does not change during the run. I replaced this instead by the value, this probably helps the compiler optimizing stuff. Maybe also it inlines the simpler function whereas it doesn't when it contains the `if (std::isinf(...))`? I'll do some more detailed benchmarks later.

PS: Yes, it should have been `lastspike == -inf`

thesamovar:
I'm not sure inserting the value of dt is worth it, and it means adding recompilations if the user is changing dt.

thesamovar:
Will wait for detailed benchmarks before further comment.

mstimberg:
[...]

PR #996, opened 2018-08-28
[MRG] Optimization for refractoriness calculation

mstimberg:
See discussion in #991 . The changes in this PR are:
* `timestep` function is no longer safe to use with infinity values
* `timestep` function uses `int64` to avoid overflow
* Instead of setting `lastspike` to `-inf*second` for neurons that never spiked, we now set this value to `-10000*second`
* C++ and Cython code generation has been refactored so that scalar variables are read in the scalar code block (i.e. outside of the loop), regardless of whether they are used in the scalar or vector code. This is particularly relevant for `dt` and `t`.

The last change speeds up things considerably, even though the use of `int64` slows down things again. Nevertheless, if we are looking at the time the stateupdater takes in the context of a real simulation (CUBA example), this all now looks pretty reasonable (rough "best of 3" times, do not take them too seriously):

| *target/branch*        | master+legacy        | master | This PR  |
| ------------- | ------------------:| -----------------------:| ---------------:|
| **weave**       | 126ms | 203ms | 132ms | 
| **Cython**      | 196ms | 910ms | 209ms |
| **numpy**     | 970ms | 1610ms | 1250ms |
| **C++ standalone** | 104ms | 203ms | 109ms |

"master+legacy" means activating the preference that switches off the use of the `timestep` function.

### DIFF
--- a/brian2/codegen/generators/GSL_generator.py
+++ b/brian2/codegen/generators/GSL_generator.py
@@ -905,7 +905,9 @@ class GSLCodeGenerator(object):
         GSL_main_code = self.unpack_namespace(variables_in_vector, variables_in_scalar, ['t'])
 
         # rewrite actual calculations described by vector_code and put them in _GSL_func
-        GSL_support_code += self.make_function_code(self.translate_vector_code(vector_code[None],
+        func_code = self.translate_one_statement_sequence(vector_statements[None],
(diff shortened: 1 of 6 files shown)


thesamovar:
I think you've nailed it. Great work!

PR #997, opened 2018-08-28
[MRG] Raise an error for event-driven equations that are non-linear and not independent

mstimberg:
If you use non-linear synaptic equations like the ones in PR #995, currently Brian will allow you to mark them as `(event-driven)`, but since it cannot solve them exactly, it will solve all equations independently. This gives wrong results, because it does not take the development of the other variables during the time between updates into account.

This PR does nothing fancy to handle such equations, it just raises an error if the situation occurs.


mstimberg:
Oops, I added a change to the CUBA example here that shouldn't have been committed, let me close this and re-open a new PR.

PR #1001, opened 2018-08-29
[MRG] New float32/64 benchmark example

thesamovar:
I was curious to know how float32 and float64 compare, so I created a benchmark file for CUBA and COBA. Turns out it makes very little difference but float64 slightly faster.

figure_1

This PR adds the example.

### DIFF
--- a/examples/advanced/float_32_64_benchmark.py
+++ b/examples/advanced/float_32_64_benchmark.py
@@ -120,14 +120,6 @@ def generate_results(num_repeats):
 
     return results
 
-# Cache this if we have joblib installed, but don't worry if not
-try:
-    import joblib
-    mem = joblib.Memory(location='.', verbose=0)
-    generate_results = mem.cache(generate_results)
-except ImportError:
-    pass
-
 results = generate_results(3)
 
 bar_width = 0.9
--- a/examples/advanced/float_32_64_benchmark.py
+++ b/examples/advanced/float_32_64_benchmark.py
@@ -1,0 +1,168 @@
+'''
+Benchmark showing the performance of float32 versus float64.
+'''
+
+from brian2 import *
+from brian2.devices.device import reset_device, reinit_devices
+
+# CUBA benchmark
+def run_benchmark(name):
+    if name=='CUBA':
+
+        taum = 20*ms
+        taue = 5*ms
+        taui = 10*ms
+        Vt = -50*mV
+        Vr = -60*mV
+        El = -49*mV
+
+        eqs = '''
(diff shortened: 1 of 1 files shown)


examples/advanced/float_32_64_benchmark.py
```diff
@@ -0,0 +1,168 @@
...
+                prefs.core.default_float_dtype = dtype
+                times.append(run_benchmark(name))
+            results[name, 'cpp_standalone', dtype.__name__] = amin(times)
+
+    return results
+
+# Cache this if we have joblib installed, but don't worry if not
+try:
+    import joblib
+    mem = joblib.Memory(location='.', verbose=0)
```
mstimberg: In older joblib versions, the first argument is called `cachedir` not `location`, this can therefore fail with a `TypeError`. Simply leaving away the argument name should work for all versions.

thesamovar: Maybe I should just remove the joblib stuff actually? It's nice to have for developing the script but not that useful in the end.

mstimberg: Good point, I agree that it's probably better to keep this as simple as possible.

thesamovar: Done!

thesamovar: Got distracted while filling out the commit message so it's a bit cryptic

mstimberg:
Two post-merge remarks:
* Looking at the generated code, it is not too surprising that we are not seeing speed benefits from `float32`: most values used in the calculations, in particular the `lio` variables, are still double, so calculations are done with double values and only cast to float in the end. That said, I did a quick&dirty trial which tries to avoid this (in the `single_precision_everywhere` branch), and the results are not that different, but `float32` is a bit faster now.
* COBA with single precision is probably not a good benchmark, since it generates NaN values.

thesamovar:
Oh, why are the lio variables not float32? I guess all intermediate variables should also be float32 if the default dtype is that?

Ah, right. Well - that's good to know about COBA. What do you think might be good as a second benchmark?

mstimberg:
> Oh, why are the lio variables not float32? I guess all intermediate variables should also be float32 if the default dtype is that?

The problem is that the lio optimisation did not use the default dtype, but had a hardcoded double type. I think I fixed this in the `single_precision_everywhere` branch, I'll open a PR in a sec.

> Ah, right. Well - that's good to know about COBA. What do you think might be good as a second benchmark?

Not sure, maybe STPD? Or for something more complete, the Clopath et al. 2010 example?

PR #1002, opened 2018-08-29
[MRG] Measure/store preparation and "after run time" for C++ standalone

mstimberg:
This measures the time spent within a standalone simulation but before the first run (i.e., synapse creation, initialization), and after the last run (i.e., writing results to disk). The implementation is not that great and for consistency we should probably do the same thing for runtime, but this should be good enough for benchmarking right now (also, we don't advertise this as a mechanism that users should use).

### DIFF
--- a/brian2/devices/cpp_standalone/device.py
+++ b/brian2/devices/cpp_standalone/device.py
@@ -603,13 +603,38 @@ class CPPStandaloneDevice(Device):
         main_lines = []
         procedures = [('', main_lines)]
         runfuncs = {}
-        for func, args in self.main_queue:
+        # Note down the first and last runs in case of multiple runs -- this is
+        # necessary to break down the total simulation time into phases
+        first_run, last_run = None, None
+        for idx, (func, _) in enumerate(self.main_queue):
+            if func == 'run_network':
+                if first_run is None:
+                    first_run = idx
+                last_run = idx
+
+        for idx, (func, args) in enumerate(self.main_queue):
             if func=='run_code_object':
                 codeobj, = args
                 main_lines.append('_run_%s();' % codeobj.name)
             elif func=='run_network':
                 net, netcode = args
+                # Add additional code to stop the recording of
+                # "before-run-time", respectively start the recording of
+                # "after-run-time"
+                nb_threads = prefs.devices.cpp_standalone.openmp_threads
+                if idx == first_run:
+                    if nb_threads == 0:  # no OpenMP
+                        main_lines.append('Network::_before_run_time = ((double)(std::clock() - _clock_start) / CLOCKS_PER_SEC);')
+                    else:
+                        main_lines.append('Network::_before_run_time = omp_get_wtime() - _clock_start;')
+
                 main_lines.extend(netcode)
+
+                if idx == last_run:
+                    if nb_threads == 0:
+                        main_lines.append('_clock_start = std::clock();')
+                    else:
+                        main_lines.append('_clock_start = omp_get_wtime();')
             elif func=='set_by_constant':
                 arrayname, value, is_dynamic = args
                 size_str = arrayname+'.size()' if is_dynamic else '_num_'+arrayname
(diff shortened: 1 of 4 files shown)


thesamovar:
I'm wondering: if this is just so we can do benchmarking for brian2genn, could we do it using the code insertion mechanism in the benchmarking scripts rather than modifying brian2 itself? Then we could put off making changes like this until we can do it better?

mstimberg:
Good point! We could just do the same thing that brian2genn does currently. Actually, we could even use the same mechanism for brian2genn (it also supports `device.insert_code` in the main function), this way we could run exactly the same script for both benchmarks! It would also mean that we would not need to use a hacked-together branch of Brian2GeNN, instead we could put all the benchmarks in a separate repository. Sounds good to me, or too much change at this point? CC @tnowotny

thesamovar:
Sounds like a good idea to me. Although, I think it means we wouldn't be able to catch the times for loading and saving arrays to disk (because insert code always happens between these I think?). We could improve this though.

tnowotny:
I don't think it should take too much time to implement and sounds like a cleaner solution.
You can obviously see the insertions I made into the main.cpp template, all have the ```{% if benchmark %}``` tag. (Whether these are all necessary and/or optimal measurement points can also still be debated by the way).
Would you @mstimberg want to have a go at this and I join in as necessary?
PS: I see that there were a lot of other changes done in the benchmarking branch which we may want to merge into brian2genn master at this point?

tnowotny:
@thesamovar - that brings home that I have no idea how the insert_code method works ... how do you indicate where the code should be inserted?

thesamovar:
In principle the first argument is a ``slot`` where the code gets inserted, but at the moment there's only one slot ``'main'``. Code gets inserted into the main flow in the order it's called. So for example, if you call ``Synapses.connect`` and then ``insert_code`` the code will be inserted after the ``Synapses.connect`` call, and vice versa. But, all of this takes place within the space where, e.g. a ``Synapses.connect`` or ``run`` call could be inserted, which is after arrays have been loaded and before they are saved. We could very easily add new slots though!

mstimberg:
I had another look at the brian2genn code, and unfortunately we won't be able to use the same `device.insert_code` statements for brian2 and brian2genn. In brian2genn, a lot of copying/conversion happens between what brian2 calls the `main` slot and the run, whereas in brian2 the run is part of the `main` slot. We could decide that we don't need benchmark information that is as fine-grained, though. Either way, I think it would be good if we add the benchmarking code "externally", i.e. in a way that you can run it with a released brian2genn version. For that, we'll only have to add a bunch of slots for `device.insert_code` into the `main.cpp` template and then we can call benchmarking code for these slots in our test script.
The only disadvantage I'm seeing is that this means that we have to prepare every script that we want to use for (detailed) benchmarking individually, whereas with @tnowotny's current solution you only have to set a preference for whatever script you want to benchmark.
Another option: we make the benchmarking preference "official", and put it into the brian2genn release.

tnowotny:
Hum ... difficult choices this. I can see advantages in both approaches and have no strong opinion against either. If we make the benchmarking preference official it would not hurt performance or anything as we would switch it off by default and then it has no effect on running the simulation at all. So far so good. Only question then is how you replicate the same type of measurements for brian 2 cpp_standalone?
In terms of granularity of measurements, I think the separate measurements for converting brian 2 to GeNN format arrays and vice versa and copying stuff to device and from device were surprisingly boring so far. Even though one could probably design benchmarks where this becomes more interesting ... and it is a "result" that these really don't matter, I think.

denisalevi:
brian2CUDA would certainly benefit from a solution for `cpp_standalone` that can easily be adapted for `cuda_standalone` and is comparable with brian2GeNN. Our benchmarks currently focus only on the `_last_run_time` measurements, so I will be looking into preparation and after run times in the next two weeks.

mstimberg:
I'm currently working on making all this use the code slots as proposed by Dan earlier, see #1006 .

PR #1003, opened 2018-08-29
[MRG] Remove lastupdate variable, except for event-driven

mstimberg:
Following the discussion in #979 (and as an alternative solution), this PR:
* Only creates a `lastupdate` variable if it is used by Brian automatically, i.e. for `event-driven` synapses.
* Tells the user that they have to add the variable manually, if they try to access it.

This breaks backwards-compatibility for users that manually used `lastupdate`, but I think with the error message, this should be ok. There's rarely a use case for doing this manually instead of using the `event-driven` mechanism.
The check whether a user tries to access `lastupdate` is a bit ugly, directly in `groups.py`, but I think it is worth it, given that we do not want to silently break backwards-compatibility.

### DIFF
--- a/brian2/groups/group.py
+++ b/brian2/groups/group.py
@@ -803,8 +803,21 @@ class Group(VariableOwner, BrianObject):
             if internal_variable is not None:
                 return None
             else:
-                raise KeyError(('The identifier "%s" could not be resolved.') %
-                               (identifier))
+                # Give a more detailed explanation for the lastupdate variable
+                # that was removed with #988
+                if identifier == 'lastupdate':
+                    error_msg = ('The identifier "lastupdate" could not be '
+                                 'resolved. Note that this variable is only '
+                                 'automatically defined for models with '
+                                 'event-driven synapses. You can define it '
+                                 'manually by adding "lastupdate : second" to '
+                                 'the equations and setting "lastupdate = t" '
+                                 'at the end of your on_pre and/or on_post '
+                                 'statements.')
+                else:
+                    error_msg = ('The identifier "%s" could not be resolved.' %
+                                 identifier)
+                raise KeyError(error_msg)
 
         elif len(matches) > 1:
             # Possibly, all matches refer to the same object
--- a/brian2/groups/group.py
+++ b/brian2/groups/group.py
@@ -804,7 +804,7 @@ class Group(VariableOwner, BrianObject):
                 return None
             else:
                 # Give a more detailed explanation for the lastupdate variable
-                # that was removed with #988
+                # that was removed with PR #1003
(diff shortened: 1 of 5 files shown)


mstimberg:
[had to push-force this branch again, since I mistakenly updated the `docs_spinx/resources` submodule to a local commit]

mstimberg:
> Are there any other places we could potentially save space? For example, i, j, etc.?

I think we'll always need the information about `i` and `j` (they are just links to `presynaptic_idx` etc. and not copies), so I don't think we can save anything there. We should encourage users more to use `Synapses(..., delay=...)` for homogeneous delays (including no delay!), because otherwise we have a `delay` variable for each synapse.

denisalevi:
> I think we'll always need the information about i and j (they are just links to presynaptic_idx etc. and not copies), so I don't think we can save anything there. We should encourage users more to use Synapses(..., delay=...) for homogeneous delays (including no delay!), because otherwise we have a delay variable for each synapse.

At least in cpp standalone, you store per synapse a pre and post idx in `_dynamic_array_<synapses>__synaptic_pre/post`. The vector is populated in the `synapses_create_...` templates. In many cases this array is not used at all and it wastes 8 bytes per synapse. I ended up deleting it on the device whenever it is not used. Its another rather hacky "copy everything to the device and delete it later if we don't need it" approach of mine... But well, check it out [here](https://github.com/brian-team/brian2cuda/blob/1b2519f1ddf6e7525f205df0543bacaeaacc8f49/brian2cuda/device.py#L244-L265). Probably easier for you toc atch in brian and not populate the vectors in the first place. The check is simply: Only create the vector when either one of the following is true:
1. There is a variable whos variable_index is `_presynaptic_idx`
2. There is a variable with varname `i`

And the same for `_postsynaptic_idx` and `j`. Or do you see potential problems with this?

mstimberg:
Hi @denisalevi. Not 100% sure I understood correctly, you don't want to store the general `i`/`j` information in the synapses if it is not used afterwards? I don't see how we can do this, given that at the time of the synapse creation we don't know yet whether it will be used afterwards. Even for standalone where in principle we can get this information at the time of the build, this is problematic. `i` and `j` cannot only be used in code, but also for indexing or initialization. In particular, indexing can happen *after* the `build`, e.g. you might be interested in the synaptic weights of all synapses coming from a certain neuron.
For the postsynaptic index, I think it is a rather uncommon use case to have synapses that do not affect their target anyway ;)

denisalevi:
Yeah I see. I didn't think about indexing other arrays with `i` and `j`. That might also be a problem in brian2CUDA for initialization, which is happening on the device. Will check that. Glad we talked about that :) thanks @mstimberg

PR #1006, opened 2018-09-04
[MRG] Standalone code slots

mstimberg:
This adds four new code slots that can be used with `device.insert_code`: `before_start`, `after_start`, `before_end`, `after_end`. With this we can add code before/after things like loading static arrays from disk or writing the results to disk.

This offers as an alternative for benchmarking and replaces #1002.


mstimberg:
[To save a bit of time, I'll merge this right away even though not all tests ran through yet. The tests will run again on the master branch anyway]

PR #1012, opened 2018-10-04
[MRG] Make the exponential_euler state updater handle constant RHS equations

### LINKED ISSUE #1010: ValueError when running SpatialNeuron with method='exponential_euler'
Here's a minimal example:

```
from brian2 import *

eqns = ''' 
Im = amp/meter**2: amp/meter**2
dk/dt = Hz : 1   # breaks exponential_euler
'''

soma = Soma(diameter=15*um)
neuron = SpatialNeuron(morphology=soma, model=eqns, method='exponential_euler')
briannet = Network(neuron)
briannet.run(1*ms)
```

produces

> The expression "Hz", defining the variable k, could not be separated into linear components
> UnsupportedEquationsException: Can only solve conditionally linear systems with this state updater.

Other numerical integration methods work fine. I am running brian2.1.3.1, and can run other examples involving SpatialNeurons and exponential_euler without problems (e.g. [Hodgkin_Huxley](https://brian2.readthedocs.io/en/stable/examples/compartmental.hh_with_spikes.html)).

mstimberg:
In principle the exact state updater should of course also be able to handle such equations, but I have to think a bit more about how to implement this. For now, this solves the original problem of the OP from #1010  who wanted to use `exponential_euler` (in the context of a multicompartment model).

Fixes #1010

### DIFF
--- a/brian2/stateupdaters/exponential_euler.py
+++ b/brian2/stateupdaters/exponential_euler.py
@@ -49,16 +49,20 @@ def get_conditionally_linear_system(eqs, variables=None):
     for name, expr in diff_eqs:
         var = sp.Symbol(name, real=True)
     
-        # Factor out the variable
-        s_expr = sp.collect(str_to_sympy(expr.code, variables).expand(),
-                            var, evaluate=False)
-        
-        if len(s_expr) > 2 or var not in s_expr:
-            raise ValueError(('The expression "%s", defining the variable %s, '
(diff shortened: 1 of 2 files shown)


thesamovar:
Just seeing this now - for some reason I didn't get an email.

mstimberg:
> Looks good to me!

Coolio, gets merged right away :)

PR #1022, opened 2018-11-05
[MRG] Warn when SpikeGeneratorGroup sets spike times < current time 

### LINKED ISSUE #1017: Segfault with SpikeGeneratorGroup
Say I have run my network, with a `sgg = SpikeGeneratorGroup()` for T seconds.
Now if I set the spikes in `sgg` with time instants < T; in the next run, the program crashes with a segmentation fault.
It should ideally throw a runtime error.

mstimberg:
As discussed in #1017 , spikes that fall before the current time (set between runs with `set_spikes`) are now ignored and a warning is issued. To make this work with standalone, I had to make some minor adjustments to the way read-only variables are marked as "dirty" by code writing to them.

Fixes #1017

### DIFF
--- a/brian2/devices/cpp_standalone/device.py
+++ b/brian2/devices/cpp_standalone/device.py
@@ -568,14 +568,27 @@ class CPPStandaloneDevice(Device):
         if self.enable_profiling:
             self.profiled_codeobjects.append(codeobj.name)
 
-        # Mark all the non-read-only used in this code object as "dirty". This
-        # is almost certainly too much, most of these
-        # variables will only be read. However, the templates for synapse
-        # creation will write to read-only variables. This is noted in the
-        # WRITES_TO_READ_ONLY_VARIABLES comment in the template.
+        # We mark all writeable (i.e. not read-only) variables used by the code
+        # as "dirty" to avoid that the cache contains incorrect values. This
+        # might remove a number of variables from the cache unnecessarily,
+        # since many variables are only read in the code.
+        # On the other hand, there are also *read-only* variables that can be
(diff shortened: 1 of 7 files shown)


brian2/devices/cpp_standalone/device.py
```diff
@@ -568,18 +568,30 @@ def code_object(self, owner, name, abstract_code, variables, template_name,
...
+        # write source and target indices, and monitors write the shared values.
+        # To correctly mark these values as changed, templates can include a
+        # "WRITES_TO_READ_ONLY_VARIABLES" comment, stating the name of the
+        # changed variables. For a monitor, this would for example state that
+        # the number of recorded values "N" changes. For the recorded variables,
+        # however, this information cannot be included in the template because
+        # it is up to the user to define which variables are recorded. For such
+        # cases, the "owner" object (e.g. a SpikeMonitor) can define a
+        # "written_readonly_vars" attribute, storing a set of `Variable` objects
+        # that will be changed by the owner's code objects.
```
thesamovar: This is all getting deeply arcane and cements the idea that we need to do a refactoring at some point! :)

mstimberg: I do agree! My long-term plan is that each code has explicit attributes stating which variables it reads and which it writes, which would not only make this stuff here obsolete but could also be used for optimisations such as running independent things in parallel.

thesamovar: Yeah it feels like there's no reason why we shouldn't have all that information available in a clear and straightforward way given the general approach we're taking. Parallel idea is very neat!

Issue #1026, opened 2018-11-10
Building brian2.synapses.cythonspikequeue failed (see error message(s) above) -- pure Python version will be used instead


sanjayankur31:
Hello: I am getting these while building brian2 on Fedora 29. 

```
In file included from /usr/lib64/python3.7/site-packages/numpy/core/include/numpy/ndarraytypes.h:1821,
                 from /usr/lib64/python3.7/site-packages/numpy/core/include/numpy/ndarrayobject.h:18,
                 from /usr/lib64/python3.7/site-packages/numpy/core/include/numpy/arrayobject.h:4,
                 from brian2/synapses/cythonspikequeue.cpp:527:
/usr/lib64/python3.7/site-packages/numpy/core/include/numpy/npy_1_7_deprecated_api.h:15:2: warning: #warning "Using deprecated NumPy API, disable it by " "#defining NPY_NO_DEPRECATED_API NPY_1_7_API_VERSION" [-Wcpp]
 #warning "Using deprecated NumPy API, disable it by " \
  ^~~~~~~
brian2/synapses/cythonspikequeue.cpp: In function ‘void __Pyx__ExceptionSave(PyThreadState*, PyObject**, PyObject**, PyObject**)’:
brian2/synapses/cythonspikequeue.cpp:24115:21: error: ‘PyThreadState’ {aka ‘struct _ts’} has no member named ‘exc_type’; did you mean ‘curexc_type’?
     *type = tstate->exc_type;
                     ^~~~~~~~
                     curexc_type
brian2/synapses/cythonspikequeue.cpp:24116:22: error: ‘PyThreadState’ {aka ‘struct _ts’} has no member named ‘exc_value’; did you mean ‘curexc_value’?
     *value = tstate->exc_value;
                      ^~~~~~~~~
                      curexc_value
brian2/synapses/cythonspikequeue.cpp:24117:19: error: ‘PyThreadState’ {aka ‘struct _ts’} has no member named ‘exc_traceback’; did you mean ‘curexc_traceback’?
     *tb = tstate->exc_traceback;
                   ^~~~~~~~~~~~~
                   curexc_traceback
brian2/synapses/cythonspikequeue.cpp: In function ‘void __Pyx__ExceptionReset(PyThreadState*, PyObject*, PyObject*, PyObject*)’:
[...]

Issue #1020, opened 2018-11-01
Cannot debug on Windows because of py-cpuinfo


Chris-Currin:
py-cpuinfo has an error on Windows, [Broken when using Pyinstaller](https://github.com/workhorsy/py-cpuinfo/issues/80), which was fixed in a PR, but Brian2 uses py-cpuinfo 3.20 which does not have this fix.

Weirdly, this only occurs when debugging (in PyCharms/IntelliJ), running is fine.

This can be reproduced on a Windows machine by running
`> python -c "import cpuinfo; print(cpuinfo.actual_get_cpu_info_from_cpuid())"`
or
`> python -c "from brian2 import *"`
```
File "C:\Users\Chris\Anaconda3\lib\site-packages\cpuinfo\cpuinfo.py", line 560, in _asm_func
    if memmove(address, byte_code, size) < 0:
OSError: exception: access violation writing 0xFFFFFFFF95460000
```

If one then does 
`> pip uninstall py-cpuinfo -y` 
`> pip install py-cpuinfo -y`
```
brian2 2.2 has requirement py-cpuinfo!=4.0.0,>=0.1.6, but you'll have py-cpuinfo 4.0.0 which is incompatible.
```

Then py-cpuinfo v4 should be installed. Running 
`>python -c "import cpuinfo; print(cpuinfo.get_cpu_info())"` works. (command changed with the version)

It seems brian2 code can then be debugged fine, **except** the code runs twice.

The alternative is just to do *only* `> pip uninstall py-cpuinfo -y`  
(note that `> conda uninstall py-cpuinfo ` wants to remove brian2 too)
and then the exception that the module does not exist is caught.

Has this been investigated more?

mstimberg:
Hi. We excluded py-cpuinfo 4.0 because it has this annoying "run scripts twice" problem on Windows (see #990). I'm not quite sure I understand your problem with 3.2 (note that there's also a version 3.3), you say you are seeing it only in debug mode, but why can you reproduce it then with the lines you mention? And what does this have to do with using pyinstaller? You see, I'm confused :) 

Note that as a workaround, you *can* use py-cpuinfo 4.0, but you have to wrap your script with:
```Python
if __name__ == '__main__':
    from brian2 import *
    ...
```

Chris-Currin:
Thanks for the response!

[...]

Issue #990, opened 2018-08-20
Problems with brian2 and py-cpuinfo v4.0 on Windows


thesamovar:
There is a bug in py-cpuinfo 4.0 on Windows (reported at https://github.com/workhorsy/py-cpuinfo/issues/108) that causes issues with Brian on Windows. We should set our version requirement to ``<=3.3.0`` (the previous release) until the bug is fixed in py-cpuinfo.

thesamovar:
I'll leave this issue open until we know what py-cpuinfo will do. It might be that they don't fix it in a future version, in which case we'll need a better workaround.

mstimberg:
Yes, I was maybe a bit too optimistic, I hadn't seen the latest response before making the change. Maybe directly change the dependency to `<4` to be safe?

sdeiss:
Given the error message below, and being new to Brian2 and not a software sleuth, I do not understand the solution that you have suggested for installing.  Can you walk a newbie through how to fix this?

(C:\Users\Steve\Anaconda3) C:\Users\Steve>conda install -c conda-forge brian2
Fetching package metadata .............
Solving package specifications: .

Package plan for installation in environment C:\Users\Steve\Anaconda3:

The following NEW packages will be INSTALLED:

    blas:          1.0-mkl
    brian2:        2.2.2.1-py36he980bc4_0 conda-forge
    certifi:       2019.3.9-py36_0        conda-forge
    gsl:           2.4-hfa6e2cd_1004      conda-forge
    py-cpuinfo:    5.0.0-py_0             conda-forge
    pysocks:       1.6.8-py36_1002        conda-forge
    urllib3:       1.24.2-py36_0          conda-forge
    vc:            14-0                   conda-forge
    win_inet_pton: 1.1.0-py36_0           conda-forge
    wincertstore:  0.2-py36_1002          conda-forge

The following packages will be UPDATED:

    conda:         4.3.21-py36_0                      --> 4.6.14-py36_0           conda-forge
    conda-env:     2.6.0-0                            --> 2.6.0-1                 conda-forge
    cython:        0.25.2-py36_0                      --> 0.29.7-py36h6538335_0   conda-forge
    menuinst:      1.4.7-py36_0                       --> 1.4.16-py36_0           conda-forge
    pycosat:       0.6.2-py36_0                       --> 0.6.3-py36hfa6e2cd_1001 conda-forge
    requests:      2.14.2-py36_0                      --> 2.21.0-py36_1000        conda-forge
[...]

PR #1029, opened 2018-11-14
[MRG] Packaging issues

mstimberg:
This PR fixes two issues:
* we previously shipped the Cython-generated C++ version of the spike queue as part of the source package which had the advantage that a pip install did not need Cython (it still needed a C++ compiler of course). The C++ code generated with older versions of Cython is not compatible with Python 3.7 (see #1026). I feel that it's the easiest to ship the Cython code only, and require a recent version of Cython as an install dependency (we want it for code generation on Python 3.x anyway). All this applies to the source installation only.
* As a first workaround to our issues with `py-cpuinfo` (#990, #1020), I removed it as a dependency for Linux and OS X (where we are not using it anyway), and made sure that it is only imported under Windows.


brian2/codegen/cpp_prefs.py
```diff
@@ -7,31 +7,40 @@
...
+        # Note that this overwrites the arch_flag, i.e. only the best option will
+        # be used
+        if 'sse' in res['flags']:
+            msvc_arch_flag = '/arch:SSE'
+        if 'sse2' in res['flags']:
+            msvc_arch_flag = '/arch:SSE2'
+        if 'avx' in res['flags']:
+            msvc_arch_flag = '/arch:AVX'
+        if 'avx2' in res['flags']:
+            msvc_arch_flag = '/arch:AVX2'
```
thesamovar: I wonder if we could entirely eliminate our dependency on py-cpuinfo by just always having ``/arch:SSE /arch:SSE2 /arch:AVX /arch:AVX2`` in the flags for windows? It'll presumably ignore the ones it doesn't know but might use the last of the flags that are applicable? I haven't tried that though.

setup.py
```diff
@@ -154,23 +154,26 @@ def build_extension(self, ext):
...
                         'pyparsing',
                         'jinja2>=2.7',
-                        'py-cpuinfo>=0.1.6, !=4.0.0',
-                        'setuptools>=6.0'  # FIXME: setuptools>=6.0 is only needed for Windows
+                        'py-cpuinfo>=0.1.6,!=4.0.0;platform_system=="Windows"',
+                        'setuptools>=21'
                        ],
       setup_requires=['numpy>=1.10',
-                      'setuptools>=6.0'
+                      'setuptools>=21'
```
thesamovar: What's this?

brian2/codegen/cpp_prefs.py
mstimberg: I'd be more than happy to do that, but we'd indeed need to test it first. Maybe you can do it on your machine, which I guess has AVX2, and I do it in the VM, which is probably on the lower end of capabilities?

setup.py
thesamovar: OK fine. Weird version numbers - they went all Chrome/Firefox style?

mstimberg:
As just commented in the cpuinfo issues, I added a mini script that we can call to get the cpu flags. Since this script is protected with a `__main__` block, it should work fine with the latest version of cpuinfo.

This is executed once per `brian2` import (on Windows only), I did not bother to write the results to disk or anything like that. I'm not sure it's worth it given the potential complications.

thesamovar:
This looks great, but I'd like to try it on my machine before merging.

How long does the cpuinfo subprocess launch take? If it's just a matter of a few ms then indeed not worth cacheing it.

mstimberg:
> How long does the cpuinfo subprocess launch take? If it's just a matter of a few ms then indeed not worth cacheing it.

I only tested it under Linux, the script takes about 1s to execute. However, this is not the overhead from the subprocess launch but almost completely the time that importing cpuinfo and getting the information via `get_cpu_info` takes. It could be worth caching it, but not more so than before we switched to the subprocess approach.

thesamovar:
If it's 1s I think cache it. If we save in ``~/.brian`` it's pretty safe I think?

mstimberg:
> If it's 1s I think cache it

I think it does quite different things on Linux and on Windows, so it's worth checking it on Windows first (we are not using it at all on Linux/OS X now).

If we do the caching on-demand (as opposed to during installation), it should be pretty straight-forward. But even then, we should probably be careful and store some identifier of the machine or something like that with it. Running things on a university network with different machines but a shared home directory might otherwise be problematic (even though it is a quite low chance that the machines differ in the flags we care about).

thesamovar:
1.4s on my reasonably powerful windows machine, so I think it's worth it. All tests pass too.

thesamovar:
Let me know if/when you'd like me to do a final check.

thesamovar:
OK code looks good and tests pass on my machine so I'm happy with it.

mstimberg:
Great, thanks. I'll wait for the tests on the servers to pass and will then merge.

PR #1034, opened 2018-11-19
[MRG] Add release notes for version 2.2.1

mstimberg:
I'd like to do a new release soon (maybe later today?), given that the annoyances around py-cpuinfo and Cython will lead to more and more user issues as time passes. Not sure whether it should be 2.2.1 or 2.2.0.1, given that there are few big changes.

### DIFF
--- a/docs_sphinx/introduction/release_notes.rst
+++ b/docs_sphinx/introduction/release_notes.rst
@@ -1,6 +1,55 @@
 Release notes
 =============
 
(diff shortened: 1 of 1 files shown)


mstimberg:
> I think 2.2.1 is fine - thinks 2.2.0.1 can be reserved for when we accidentally make a mistake and correct a release the next day sort of thing. ;)

Yea, agree. I'll do a release later today then.

PR #1038, opened 2018-12-06
[MRG] Check range of synaptic targets after applying post-synaptic condition

### LINKED ISSUE #1037: Synapse generator syntax sometimes incorrect with postsynaptic condition
The following code tries to connect to a neuron with target index 10 (which does not exist), triggered by a `rand()` condition in the `if` part. It does not matter whether this is actually doing anything, e.g. `rand() <= 1.0` triggers the same problem.
```Python
source = NeuronGroup(10, '')
target = NeuronGroup(10, '')
syn = Synapses(source, target)
syn.connect(j='k for k in range(i, i+5) if i < 5 and rand() < 0.5')
```
Of course users should rather use `sample(i, i+5, p=0.5)` here, but 1) it should still work and 2) it might be a symptom of a larger issue.

mstimberg:
This should fix #1037 -- note that the relevant (newly added) tests are part of a "long" test, i.e. they will not run by default (including on the test servers).

Oh, and with the change synapse generation is slightly less efficient when using `skip_if_invalid`, because the post-synaptic condition will be evaluated for non-existing targets. However, I think this is enough of a corner case that it does not merit to further complicate the (already quite complicated...) template.

### DIFF
--- a/brian2/codegen/runtime/cython_rt/templates/synapses_create_generator.pyx
+++ b/brian2/codegen/runtime/cython_rt/templates/synapses_create_generator.pyx
@@ -86,12 +86,6 @@ cdef void _flush_buffer(buf, dynarr, int buf_len):
         {% endif %}
 
             {{vector_code['create_j']|autoindent}}
-            if _j<0 or _j>=N_post:
-                {% if skip_if_invalid %}
-                continue
-                {% else %}
-                raise IndexError("index j=%d outside allowed range from 0 to %d" % (_j, N_post-1))
-                {% endif %}
             _raw_post_idx = _j + _target_offset
 
             {% if postsynaptic_condition %}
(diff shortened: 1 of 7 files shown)


mstimberg:
> h, and with the change synapse generation is slightly less efficient when using `skip_if_invalid`, because the post-synaptic condition will be evaluated for non-existing targets.

Um, I just realized that we *have* to do something about this, because if the post-synaptic condition refers to a post-synaptic variable, we certainly do not want to index into it with an invalid index!

mstimberg:
I applied what we discussed in #1037, it needed a few more small changes, but I think it now all makes sense. If you happen to find a corner case that is not covered, let me know!

mstimberg:
[One test timed out on appveyor, I think we can safely ignore this]

mstimberg:
I'd say the major risk is that it could crash (or give an error with a not that helpful error message in Python mode). The worst thing that could potentially happen is that we don't give an error message for an out of bound access (because those are not always triggering segmentation faults). E.g. `... if k<10 and v_post > 0` would check `v_post` for `k>=10` instead of raising an error (assuming that `N_post=10`). The only way I could imagine wrong results would be to have a condition that is incorrect in the first place.

All in all, I don't think there's a higher risk to have "silent wrong results" than before.

thesamovar:
Good to go then!

PR #1042, opened 2019-01-03
[MRG] Fix #1041

### LINKED ISSUE #1041: SpikeGeneratorGroup period checking error
I needed to run a SpikeGeneratorGroup on a period, so I created one using the following command:
`ing = SpikeGeneratorGroup(1, [0, 0, 0, 0, 0], [0, 45, 75, 105, 135]*ms, period=150*ms, name='ing')`.
When running the network, I got this error :  
`NotImplementedError: The period of ing is <ing.period: 150. * msecond>, which is not an integer multiple of its dt of 0.0001.`

I decided to make a barebone example to check it wasn't working, as well print what the SpikeGeneratorGroup.before_run() is testing for:

```python
from brian2 import SpikeGeneratorGroup, ms, run
import numpy as np

period = 150*ms
ing = SpikeGeneratorGroup(1, [0], [0]*ms, period=150*ms, name='ing')
dt = ing.dt

print(str(abs(int(period/dt)*dt - period)) + ' > '
      + str(period * np.finfo(dt.dtype).eps))

run(300*ms)
```
I receive the same error when running this code; the print outputs this : 
`100. us > 33.30669074 as`
I tried again but with a period of 100*ms; the print outputs is the same but this time the network runs. 
It would seem there's a precision problem when checking the period of SpikeGeneratorGroup. 
I think the test should be relaxed to be only as precise as the `dt` itself, to avoid this kind of error.

mstimberg:
This should fix #1041, but the applied heuristic (see code) is certainly debatable!

### DIFF
--- a/brian2/input/spikegeneratorgroup.py
+++ b/brian2/input/spikegeneratorgroup.py
@@ -94,7 +94,8 @@ class SpikeGeneratorGroup(Group, CodeRunner, SpikeSource):
         self.variables = Variables(self)
         self.variables.create_clock_variables(self._clock)
 
-        indices, times = self._check_args(indices, times, period, N, sorted)
+        indices, times = self._check_args(indices, times, period, N, sorted,
+                                          self._clock.dt)
 
         self.variables.add_constant('N', value=N)
         self.variables.add_array('period', dimensions=second.dim, size=1,
@@ -193,8 +194,8 @@ class SpikeGeneratorGroup(Group, CodeRunner, SpikeSource):
             if period_bins > max_int:
                 logger.warn('Periods longer than {} timesteps (={}) are not '
                             'supported, the period will therefore be '
-                            'considered infinite. Set the period to 0*second to'
-                            'avoid this'
+                            'considered infinite. Set the period to 0*second '
+                            'to avoid this '
                             'warning.'.format(max_int, str(max_int*dt*second)),
                             'spikegenerator_long_period')
                 period = period_bins = 0
@@ -242,7 +243,8 @@ class SpikeGeneratorGroup(Group, CodeRunner, SpikeSource):
             your arrays contain large numbers of spikes. Defaults to ``False``.
         '''
 
-        indices, times = self._check_args(indices, times, period, self.N, sorted)
+        indices, times = self._check_args(indices, times, period, self.N,
+                                          sorted, self.dt)
 
         self.variables['period'].set_value(period)
         self.variables['neuron_index'].resize(len(indices))
@@ -254,7 +256,7 @@ class SpikeGeneratorGroup(Group, CodeRunner, SpikeSource):
         self.variables['spike_time'].set_value(times)
         # _lastindex and _timebins will be set as part of before_run
 
-    def _check_args(self, indices, times, period, N, sorted):
+    def _check_args(self, indices, times, period, N, sorted, dt):
         times = Quantity(times)
         if len(indices) != len(times):
             raise ValueError(('Length of the indices and times array must '
@@ -262,9 +264,14 @@ class SpikeGeneratorGroup(Group, CodeRunner, SpikeSource):
                                                         len(times)))
         if period < 0*second:
             raise ValueError('The period cannot be negative.')
-        elif len(times) and period != 0*second and period <= np.max(times):
-            raise ValueError('The period has to be greater than the maximum of '
-                             'the spike times')
+        elif len(times) and period != 0*second:
+            period_bins = np.round(period / dt)
+            # Note: we have to use the timestep function here, to use the same
+            # binning as in the actual simulation
+            max_bin = timestep(np.max(times), dt)
+            if max_bin >= period_bins:
+                raise ValueError('The period has to be greater than the '
+                                 'maximum of the spike times')
         if len(times) and np.min(times) < 0*second:
             raise ValueError('Spike times cannot be negative')
         if len(indices) and (np.min(indices) < 0 or np.max(indices) >= N):
--- a/brian2/input/spikegeneratorgroup.py
+++ b/brian2/input/spikegeneratorgroup.py
@@ -152,8 +152,15 @@ class SpikeGeneratorGroup(Group, CodeRunner, SpikeSource):
                                  'than its dt of %s.' % (self.name,
                                                          self.period[:],
                                                          dt*second))
-            if (abs(int(period/dt)*dt - period) >
-                    period * np.finfo(dt.dtype).eps):
+            # We check that the period is a multiple of dt. Because we cannot do
+            # any exact test, we use the following heuristic: we calculate all
+            # the multiples of the period that fit into 1000s and then check
+            # whether all these times correspond to the same time step as we
+            # obtain by first converting the period to the equivalent number of
+            # timesteps
+            n_periods = int(1000. / period)
+            periods = np.arange(n_periods) * period
+            if not np.all(timestep(periods*second, dt*second) % timestep(period*second, dt*second) == 0):
                 raise NotImplementedError('The period of %s is %s, which is '
                                           'not an integer multiple of its dt '
                                           'of %s.' % (self.name,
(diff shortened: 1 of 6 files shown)


brian2/input/spikegeneratorgroup.py
```diff
@@ -150,15 +150,23 @@ def before_run(self, run_namespace):
...
+            # timesteps
+            if period < 1000:
+                n_periods = int(1000. / period)
+                periods = np.arange(n_periods) * period
+                if not np.all(timestep(periods*second, dt*second) % timestep(period*second, dt*second) == 0):
+                    raise NotImplementedError('The period of %s is %s, which is '
+                                              'not an integer multiple of its dt '
+                                              'of %s.' % (self.name,
+                                                          self.period[:],
+                                                          dt*second))
```
thesamovar: How expensive is this check? In principle we could check just that last 10 or 100 values of ``periods`` and that's probably sufficient?

thesamovar:
In terms of the logic of the check, I didn't fully follow it through. Is it equivalent to asking whether or not it would make any difference over the duration of a 1000s long run? That seems like a reasonable check to me, although if the information about the length of the run actually requested is available we could potentially use that? If not now, it could be a warning issued at the end of the simulation using ``net.t``.

mstimberg:
Actually, I realized that my test did not make much sense. For a period that is between 1 and 2 dt, `timestep(period, dt)` will be 1 and therefore the test will always work. This means, that it would for example accept a period of `0.11ms`... What I'd really wanted to test was that the time between periods is always equivalent to the period (both measured in time steps). But then, for this I can simply test whether `timestep(n_periods * period, dt) == n_periods * timestep(period, dt)`, right? This is also much faster and shouldn't take any significant time.

And yes, the idea was to check whether it makes a difference for a 1000s run. At the moment, we do not have any information about the actual runtime at that point, but we could change `before_run` so that it passes on this information. This would need changes in quite a few places, but AFAICT it would not break anything in brian2genn and would only mean a one-line change for brian2cuda.

It would come with a few inconviences, though, e.g. we often use `run(0*ms)` to trigger all the `before_run` tests. Also, with multiple runs we would not only need to use the runtime, but also the elapsed time, otherwise the error check would not work for many small runs instead of a long run. What do you think – worth the effort?

thesamovar:
I looked at the code in a bit more detail to remind myself of what's going on here, and there are a bunch of things that are odd or confusing.

Firstly, I think the problem in #1041 is simpler to fix, it's just that we use ``int(period/dt)*dt`` as the period when constrained to the dt grid. This is slightly wrong, it should be ``round(period/dt)*dt``. You can see that this is the error in that issue because it gives ``abs(int(period/dt)*dt - period)`` as ``100 us`` which is just ``dt``. In other words, ``150*ms`` is just slightly less than ``1500*dt`` rather than just slightly more, and therefore ``int(period/dt)`` is coming out at 1499.

The condition ``(abs(round(period/dt)*dt - period) > period * np.finfo(dt.dtype).eps)`` should still be fine with that change I think? It's just a standard floating point equality check.

Secondly, on the subject of whether or not the period needs to be a multiple of dt or not. I guess this check has always been in here, so we could just keep things simple for ourselves and leave it like that?

One oddity of the code is that we calculate various things in very different ways. We assume period is a multiple of dt, and we calculate ``timebins`` as part of the check that there are no repeated spikes in a single time period. However, we then discard that and essentially recalculate it in a different way in the template. How about instead of doing that, we store the timebins and then this can massively simplify the templates, we'd just have to compute ``period_n = int(round(period/dt))`` and ``timebins`` as it's currently computed. Then, in the template, we'd use ``n = clock.n % period_n`` and avoid all the floating point stuff entirely. This would clean up some odd corner cases in the current code. For example:

```
src = SpikeGeneratorGroup(1, [0, 0, 0], [0*ms, .9*ms, .99999*ms], period=1*ms)
G = SpikeMonitor(src)
run(2*ms)
print G.it
```

gives no error and prints

```
(<spikemonitor_1.i: array([0, 0, 0, 0])>, <spikemonitor_1.t: array([0. , 0.9, 1. , 1.9]) * msecond>)
```

That's because we check that all spike times are ``<period`` via ``period <= np.max(times)``; and in the template we break from the loop if ``(fabs({{spike_time}}[_idx] - padding_after) < epsilon)``. The first check doesn't take into account that we round to the nearest ``1e-3*dt``, while the second one check does. If the data were correctly sanitised, the ``(fabs({{spike_time}}[_idx] - padding_after) < epsilon)`` check shouldn't even need to be there, it testing true must always be a sign that something went wrong. Switching to just doing a single conversion of spike times to multiples of dt and then always using that would sort all these issues. It wouldalso simplify the data sanitising code quite a bit I think.

What do you think?

mstimberg:
Many good points, I'm also a bit confused why we implemented things like that back in the days... Maybe we did not want to store integer time steps because this would be wasteful memory-wise for big spike patterns? But then, I think most users would prefer here to use a bit more memory to get better performance (I guess your proposed solution would be faster). I'll have a closer look at this tomorrow (or feel free to have a go at this if you are motivated :) ).

thesamovar:
Without delving into the history, I'm guessing this solution evolved over time from something that was initially more time focussed into something more timestep focussed, without a rethink being done at any point.

(I don't think I'll get a chance to work on this any time soon btw, beginning of term and I've got to teach this term...)

mstimberg:
I changed the code so that it now stores the spike times and the period as timebins (i.e. multiples of `dt`), as you said this quite simplifies the code.

There is one minor backward-incompatible change: previously, you could set very high periods (the default value was 1e100*second as a shorthand for "infinity"), now it can be maximally 2^31 time steps (I decided to stay with `int32` for now, as we do this elsewhere as well). I don't think this is problematic, with the default `dt` this corresponds to > 200000s, I don't see many use cases for such long periods. If you try to set a period that is too long, you now get a warning saying that it will be considered infinitely long. The new correct way to set the period to infinity is to set `period=0*second`. This raises an error with previous versions of Brian, so I don't think there is any danger of a "silent" backwards incompatibility, or did I overlook something.

I you are happy with all this, this is ready-to-merge from my side.

thesamovar:
While I check out the code, random thought. Do you think we should have a warning whenever ``t/dt>=2**31`` across all code? e.g. whenever we do a run, we could compute ``end_time/min(all dt values)>=2**31``. It would be a generalised warning that some stuff might not be safe.

brian2/tests/test_spikegenerator.py
```diff
@@ -102,6 +102,34 @@ def test_spikegenerator_period():
...
+@attr('standalone-compatible')
+@with_setup(teardown=reinit_devices)
+def test_spikegenerator_period_rounding():
+    # See discussion in PR #1042
+    src = SpikeGeneratorGroup(1, [0, 0, 0], [0*ms, .9*ms, .99999*ms],
+                              period=1*ms)
+    mon = SpikeMonitor(src)
+    run(2 * ms)
+    assert_allclose(mon.t, [0., 0.9, 1., 1.9] *msecond)
+
```
thesamovar: This test is wrong. Actually, this is an error because the spike at .99999ms is not emitted. So we can keep this test but it ought to raise an error. See the error in logic above.

brian2/input/spikegeneratorgroup.py
```diff
@@ -245,7 +262,7 @@ def _check_args(self, indices, times, period, N, sorted):
                                                         len(times)))
         if period < 0*second:
             raise ValueError('The period cannot be negative.')
-        elif len(times) and period <= np.max(times):
+        elif len(times) and period != 0*second and period <= np.max(times):
```
thesamovar: This is wrong, and the example test below exactly shows why it's wrong. What we want to do, after having computed ``period_bins`` and ``timebins`` is to check that ``period_bins<=max(timebins)``.

brian2/tests/test_spikegenerator.py
mstimberg: Oops, I added this a bit hastily and did somewhat miss the point of this example... It's a bit weird for a test since it tests for expected behaviour that is certainly not expected for the user. But still, definitely better than silently ignoring the spike.

mstimberg:
Corner case should be fixed, It's slightly annoying that the test about the maximum time < period moved into `before_run`, i.e. you'll not get it right at the initialization of the `SpikeGeneratorGroup` but only when you do `run(..)`. But that's consistent with other tests that depend on `dt`, so well.

thesamovar:
Good to go I think! Nice work on this.

mstimberg:
The appveyor runs failed because of timeouts, but apart from that all tests passed. Thanks for the detailed look at this, merging it now.

PR #1044, opened 2019-01-09
[MRG] Expose integer timestep

### LINKED ISSUE #1043: Expose integer timestep to templates
With the recent changes in `SpikeGeneratorGroup`, it seems to be slower than before for the numpy target (see [benchmark](https://brian-team.github.io/brian2_benchmarks/#benchmarks.RuntimeOnlyExamples.track_SpikeGenerator_with_period)). I think this is mostly because we use the `timestep` function to convert the time into an integer timestep. This is of course redundant, as our clock already uses an integer timestep internally. However, only `t` and `dt` are available to code/templates at the moment. We should expose the timestep as well (maybe as a underscore-prefixed name to not clash with any user-defined names?), maybe not as `timestep`/`_timestep` since we use this as the name for the function.

mstimberg:
This should fix #1043. The PR speeds up things, but it seems that the current integer-based code is still slower than the previous float-based code. I don't see any obvious reason for this. Well, it is probably not worth spending much time on this, given that performance is never great with the `numpy` codegen target.

### DIFF
--- a/brian2/codegen/runtime/cython_rt/templates/spikegenerator.pyx
+++ b/brian2/codegen/runtime/cython_rt/templates/spikegenerator.pyx
@@ -1,10 +1,10 @@
 {% extends 'common.pyx' %}
 
 {% block maincode %}
-    {# USES_VARIABLES { _spikespace, t, dt, neuron_index, _timebins, _period_bins, _lastindex, timestep, N } #}
+    {# USES_VARIABLES { _spikespace, neuron_index, _timebins, _period_bins, _lastindex, t_in_timesteps, N } #}
 
     cdef int32_t _the_period    = {{_period_bins}}
-    cdef int32_t _timebin       = _timestep({{t}}, {{dt}})
+    cdef int32_t _timebin       = {{t_in_timesteps}}
     cdef int32_t _cpp_numspikes = 0;
 
     if _the_period > 0:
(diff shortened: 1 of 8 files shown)


thesamovar:
Convert ``t_in_timesteps`` to 32 bit int before doing the searchsorted? Can't imagine why that would make any difference, but maybe there's some weird missing optimisation there and it ends up converting it over and over again rather than just once? Otherwise, code looks good to me.

mstimberg:
Good suggestion, it does not seem to change anything, though. I still don't understand where the time is spent, according to the profiler both versions spend about the same time in the `searchsorted` function, so it must be something else. I guess we could optimize the code further by generating different code for the case with and without period (therefore making the `if period > 0` check unnecessary), but this does not seem to bring the time down enough, either...

thesamovar:
It's weird. How substantial is the difference on what test?

mstimberg:
I used the same test as in the [benchmark](https://brian-team.github.io/brian2_benchmarks/#benchmarks.RuntimeOnlyExamples.track_SpikeGenerator_with_period):
```Python
from brian2 import *
prefs.codegen.target = 'numpy'
s_gen = SpikeGeneratorGroup(100, np.arange(100),
                            np.repeat([0, 50], 50)*ms,
                            period=100*ms)
run(10*second)
print device._last_run_time
```
It gives me about 2.1s – 2.3s (it is quite variable) in this branch, and 1.8s – 2s on version 2.2.1. I don't think this is something to really worry about, it won't be noticeable in actual simulations I'd say. But I expected it to be faster, not slower...
On master (using the `timestep` function) it is 2.8s – 3s, so I think this PR is relevant.

thesamovar:
Is the difference about the time spent in that code block or different initialisation time?

You've removed the ``if not_end_period``, maybe that was helping it to run faster? But I don't see how if it's spending the same time in searchsorted.

mstimberg:
It's the time in the code block. It is still mysterious to me...

mstimberg:
Either way, this PR is a clear improvement, so merging it now.

PR #1047, opened 2019-01-25
[MRG] Allow compiler keywords in function implementations

### LINKED ISSUE #672: Allow FunctionImplementation to declare additional sources, directories, etc.

mstimberg:
When writing functions with target-specific native, non-trivial code (e.g. as in the [OpenCV example](https://brian2.readthedocs.io/en/stable/examples/advanced.opencv_movie.html)), you might have to use additional header files and libraries. We already support this via preferences which is motivated by standalone where you compile everything at once. For runtime targets like weave or Cython, we do not need to compile every code object file with these additional arguments, but only those where the function is used (potentially this could speed up compilation, but not sure it makes a practical difference).
Finally, we currently do not support to put the code in a completely separate file and compile it with the Brian-generated files. With this PR, this is possible by providing the separate source file(s) as an argument to a `sources` keyword in the `@implementation` decorator.

I modified the OpenCV example a bit to use the new keyword arguments, it does not use the separate source file thing, though.

This is basically ready to be merged from my side, the only major thing missing is support for external source files in Cython. Then, I still need to write a bit of documentation and I wanted to have a look whether (or rather how...) this breaks Brian2GeNN and Brian2CUDA.

Closes #672

### DIFF
--- a/brian2/codegen/codeobject.py
+++ b/brian2/codegen/codeobject.py
@@ -128,22 +128,57 @@ def _error_msg(code, name):
     return error_msg
 
 
+def check_compiler_kwds(compiler_kwds, accepted_kwds, target):
+    '''
+    Internal function to check the provided compiler keywords against the list
+    of understood keywords.
+
+    Parameters
+    ----------
+    compiler_kwds : dict
+        Dictionary of compiler keywords and respective list of values.
+    accepted_kwds : list of str
+        The compiler keywords understood by the code generation target
+    target : str
+        The name of the code generation target (used for the error message).
+
+    Raises
+    ------
+    ValueError
+        If a compiler keyword is not understood
+    '''
+    for key, value in compiler_kwds.iteritems():
+        if key not in accepted_kwds:
+            formatted_kwds = ', '.join("'{}'".format(kw)
+                                       for kw in accepted_kwds)
+            raise ValueError("The keyword argument '{}' is not understood by "
+                             "the code generation target '{}'. The valid "
+                             "arguments are: {}.".format(key, target,
+                                                         formatted_kwds))
+
+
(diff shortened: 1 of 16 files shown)


thesamovar:
Nice!

mstimberg:
This is now ready to merge from my side. I've added support for external files in Cython, as well as some tests and documentation. The Cython stuff was a little trickier than I though, Cython is very picky about the exact relative location of all files. I ended up copying all additional source (and header) files to the Cython cache directory which seems not too bad a solution.

thesamovar:
Haven't looked at code yet, so for the moment just a quick question on what you just said. That's for user code in the same directory as their script, right? Presumably for 3rd party libraries you can give an absolute location of header files, etc.?

mstimberg:
To avoid any misunderstandings, this is only for Cython `.pyx` files, not for standalone/weave (or even C headers to be used from a Cython file), for those all the standard stuff applies (e.g. you can set `include_dirs`, `library_dirs`, etc.).

The problem is that Cython files have to be located like equivalent Python modules. For example in our test suite, the test module is `brian2.tests.test_functions`. If you have a Cython `.pyx` file in the same directory, you'd need to also refer to it as `brian2.tests.cython_module`. This all then unfortunately breaks down because we compile and import our generated code from the Cython cache directory. I tried various combinations of doing this, but either it worked during compilation time, or during run time, never both... So the current solution is: you give the name of an external source file `cython_module.pyx` that can be located wherever you want (you give the path). During compilation, this module will be copied to the Cython cache directory and you'll therefore have to import it as `cython_module`, regardless of whether it is somewhere in a Python package structure (as in our test suite) or not. The "Cython header" (`.pxd` file) is supposed to be in the same place as the `.pyx` file and also copied to the Cython cache directory.

PR #1053, opened 2019-02-13
[MRG] Add new ways to create subgroups (e.g. with single numbers or arrays)

mstimberg:
Sometimes, especially in the context of `SpatialNeuron`, you might want to refer to only a single element as a subgroup. For this, you'd previously have to use an awkward slice such as `group[2:3]`. With the changes here, you can use `group[2]` instead. You can also use arrays, e.g. `group[[3, 4, 5]]` instead of `group[3:6]`, which again is something that is useful for `SpatialNeuron`, where you can get such indices from the `Morphology` object (e.g. `morpho.dend.indices[:]` will get the indices of the `dend` section).

It's not a big feature, but it adds a bit of consistency/flexibility, I think. Under the hood, it doesn't change anything, subgroups still have to be contiguous.

### DIFF
--- a/brian2/groups/neurongroup.py
+++ b/brian2/groups/neurongroup.py
@@ -2,6 +2,7 @@
 This model defines the `NeuronGroup`, the core of most simulations.
 '''
 import collections
+import numbers
 import string
 
 import numpy as np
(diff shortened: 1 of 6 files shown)


mstimberg:
[As for the other pull request, the test failure is only about a build timeout, all tests do pass]

mstimberg:
> This is just syntactic sugar for slices right? I don't see any reason why not!

Yes, exactly, no new functionality. It is mostly about no longer having to explain why `group.v[3:4]` and `group[3:4].v` both work, but you can only use `group.v[3]` and not `group[3].v`.

Issue #1009, opened 2018-09-21
run_regularly needs to be explicitly added to Network but not MagicNetwork


thesamovar:
I wonder if this is what we want? It's not what I expected.

thesamovar:
It might be my rather specific setup:

```
... make network ...
run(...)
G.run_regularly(...)
run(...)
```

This didn't work, but it might work if the run_regularly was before the first run?

thesamovar:
I guess the bug then would be that ``contained_objects`` isn't re-checked on re-running.

mstimberg:
Ah, right, I think it is never checked for a run, it is used in `Network.add`.

thesamovar:
I think this counts as a bug right? Not just an unexpected behaviour?

mstimberg:
I guess so, it is logical if you know how things are implemented behind the scenes, but we don't want users to have to think about that. I could have easily made the same mistake.

Vigneswaran-Chandrasekaran:
I made the code, which describes the issue reported.

`#importing necessary libraries
from brian2 import *
from matplotlib import pyplot as plt

#clear any pre-defined brian variables
start_scope()

#define differential equation
tau = 10*ms
diff_eqn = '''dv/dt = (1-v)/tau : 1'''

#define neuron groups
Ng1 = NeuronGroup(10, diff_eqn, threshold='v>0.5', reset='v = 0.32',method='exact')
Ng2 = NeuronGroup(10, diff_eqn, threshold='v>0.8', reset='v = 0.22',method='exact')

#define synaptic weights connecting two neuron groups 
Sy = Synapses(Ng1,Ng2)
Sy.connect(condition='i >= j')    

#define state monitor for each neuron group
M1 = StateMonitor(Ng1, 'v', record=True)
M2 = StateMonitor(Ng2, 'v', record=True)
`
I defined the components and then used magicNetwork to run,

`
#schedule particular run for neuron groups
Ng1.run_regularly('''v = 2*v''', dt=20*ms)
Ng2.run_regularly('''v = 2.5*v''', dt=30*ms)
run(100*ms)
`
When I try to plot, my model
`
plt.plot(M1.t/ms, M1.v[0], label='Neuron 1')
plt.plot(M2.t/ms, M2.v[1], label='Neuron 2')
xlabel('Time (ms)')
ylabel('mV')
plt.show()
`
[...]

PR #1069, opened 2019-04-06
Fix to the issue_#1009

Vigneswaran-Chandrasekaran:
Fix for the issue #1009 
@mstimberg, as you mentioned,

> We'd like to change this as follows:
In Network.add, only the objects provided as the argument should be stored to Network.objects. In other words, Network.add should ignore contained_objects.

     Network.add() only stores the argument passed to Network.object.

>In Network.run, instead of just taking the list of Network.objects, create a new, full list of objects by going through the list and adding all contained_objects recursively. This should happen in an independent list, i.e. not modify the original Network.objects list.

In Network.run() I've traversed the objects added by Network.add() and appended them to a separate list along with their respective contained_objects recursively.
     In Network.run(), I've added a statement,
```python
self.objects = self.allobjects #where allobjects contain contained_objects
```
because self.objects is utilized by many functions of the class. So I felt like changing everywhere to self.allobjects may be unnecessary. 

Kindly give your feedback and correct me, if I made anything wrong.


mstimberg:
Thanks for the changes, I now realized that this is slightly more complicated than I though, as you can see from the failing tests.

> , I've added a statement,

> `self.objects = self.allobjects #where allobjects contain contained_objects`

> because self.objects is utilized by many functions of the class. So I felt like changing everywhere to self.allobjects may be unnecessary.

I would prefer to avoid this, because it changes the state of the `Network` object, depending on whether it has been run or not. For example, if you'd ask for `net.get_states()` before a run, you'd only get the states of the objects that have been explicitly added, after a run you'd also get the states of all child objects.

I think the best solution would be to refactor the `add_contained_objects` method so that it returns a new list of objects (instead of storing them to an attribute of the network). In basically all functions that use `self.objects`, you'd then include something like
```Python
all_objects = self.get_all_objects()  # new version of add_contained_objects
```
and then use `all_objects` instead of `self.objects`.
This changes quite a bit of code, but it should also take care of the test failures.

Vigneswaran-Chandrasekaran:
@mstimberg , Yes noted. I'll change it accordingly. Thanks :+1:

Vigneswaran-Chandrasekaran:
Hi @mstimberg ,

When I started to proceed further, I think I didn't quite understand the approach clearly,

> In basically all functions that use self.objects, you'd then include something like
all_objects = self.get_all_objects()  # new version of add_contained_objects
and then use all_objects instead of self.objects.

The change of *using all_objects instead of self.objects*, should be made in all functions which use self.objects or only those involved in the simulation (running). 
For example:
```python
    @device_override('network_after_run')
    def after_run(self):
        for obj in self.objects:               #here we should make the change from self.objects to all_objects
            if obj.active:                          # to reflect them in runtime.
                obj.after_run()      

        #whereas in a function like:
    def get_states(...):
         for obj in self.objects:               
    """
        changing self.objects to all_objects here would again bring back 
        same problem of sudden network state change after adding         
        conatined_objects.
    """
```
Should we only make changes to the particular function which are used by run() ? 
Because I think changing everywhere to all_objects would again mean the same job of what self.objects was doing.
Kindly correct me, if I understood anything wrong.

mstimberg:
> Should we only make changes to the particular function which are used by run() ?
> Because I think changing everywhere to all_objects would again mean the same job of what self.objects was doing.

I meant indeed to change it everywhere, not just in `run()`. You are right that this does not change anything in many situations, but there is a difference between using `all_objects` which is calculated at that time and using `self.objects`, which is updated only during `run()`. For example, if you access `net['some_object']`, then this goes through `Network.__getitem__` which currently uses `self.objects`. If you try this before a `run()` and therefore before the update of `self.objects`, you cannot access child objects; after `run()`, you can. 

Now one alternative to what I proposed would be to keep the function `add_contained_objects` that you already had, but have it update `self.objects` directly instead of `self.all_objects`. You could then call this function at the beginning of every function that uses `self.objects`, to make sure that `self.objects` is up to date.

However, I prefer the solution with `all_objects` calculated on the spot and not stored in an attribute, because it is clearer and more in common with our general approach in Brian 2 – whenever possible, we try to not change the "state" of objects as a side effects of calling general methods on it. In particular for testing, it would be good to always know what `self.objects` stored, without worrying that it changes with the call to a seemingly unrelated method.

Vigneswaran-Chandrasekaran:
Hi @mstimberg ,

Thanks for the reply. Now I understood clearly. 

> However, I prefer the solution with all_objects calculated on the spot and not stored in an attribute, because it is clearer and more in common with our general approach in Brian 2 – whenever possible, we try to not change the "state" of objects as a side effects of calling general methods on it. In particular for testing, it would be good to always know what self.objects stored, without worrying that it changes with the call to a seemingly unrelated method.

Yes, I agree with this. Maintaining the general way throughout is very important. 

> Now one alternative to what I proposed would be to keep the function add_contained_objects that you already had, but have it update self.objects directly instead of self.all_objects. You could then call this function at the beginning of every function that uses self.objects, to make sure that self.objects is up to date.

Although this solves the issue, we frequently would try to call the function to sync the update of the object, although in user time frame he wouldn't see the change of state, but on back there would be frequent changing of state. So I will stick with the first idea itself. I should finish fixing the issue asap, because it has already taken too much time here. :smile:

mstimberg:
>  I should finish fixing the issue asap, because it has already taken too much time here.

In my experience, most seemingly simple issues end up being not that simple after all :smile:

Vigneswaran-Chandrasekaran:
I guess this commit almost fixed the issue, albeit some test cases are failing.
@mstimberg kindly check the changes and correct me, if I made anything wrong. 
Thanks :)

mstimberg:
You are getting errors in standalone mode (BTW: to run the standalone tests on your own machine, run the `run_nose_tests_standalone.py` script in `dev/tools`). I think this is because you changed the behaviour of `Network.run`/`Network.before_run`, but `CPPStandaloneDevice.network_run` is overwriting the functionality of `Network.run`. The standalone version still expects that everything has been added by `Network.add`, but that is no longer the case. So you need to make a very similar change to what you did in `Network.run`, i.e. when the current code accesses `net.objects` it should instead use something like `all_objects` which you previously filled with `net._sort_objects()`.

The way that `CPPStandaloneDevice.network_run` "overwrites" `Network.run` is a bit obscure, I have to admit... It does not overwrite it via inheritance (`CPPStandaloneDevice` does not inherit from `Network`), but it uses our handcrafted "device override" mechanism: https://brian2.readthedocs.io/en/stable/developer/devices.html#device-override-methods

Vigneswaran-Chandrasekaran:
Thank you :blush: 
Yes, I'll make the necessary changes. :+1:

Vigneswaran-Chandrasekaran:
Thanks for the changes :blush: 

It would be more helpful if you kindly explain the role of the commit,
_Prevent sympy 1.1 from getting installed_

mstimberg:
Sure, I just wanted to wait for the tests to run through to be sure it actually fixes the issue :) The error you were having had nothing to do with your changes, it is a bug in the `sympy` package. We've seen this before (#856), our fix was simply to exclude the buggy version of `sympy` – 1.0.x is fine, 1.1.1 is fine, only 1.1.0 has the bug. When we install the dependencies on the test server we ask for `sympy!=1.1.0`. For some reason, the dependency resolution mechanism now decided to install version `1.1` on the Windows test server. I think `1.1` and `1.1.0` are identical, so I now changed our dependency to `sympy!=1.1.0,!=1.1` to exclude both variants. If that works, I'm going to do the same thing in `setup.py`, just to be sure that no one installs Brian together with one of these `sympy` versions.

Vigneswaran-Chandrasekaran:
Thank you very much for the explanation :)
Let's hope it fixes the issue :smiley:

mstimberg:
I actually figured out that the reason why it installs `sympy` 1.1.0 in the first place is a bug in conda (conda/conda#8588). My fix in this branch only works by accident, but that should be good enough for now :)

I'll review the code in this PR tomorrow!

Vigneswaran-Chandrasekaran:
Surely not an accident.
It is quite wise to locate the exact bug origin :relaxed: 

> I'll review the code in this PR tomorrow!

Yes, Thank you very much :)
I have a couple of concerns about the change,

- sort_object() is called by the functions when they need sorted objects (for Standalonemode its is added with get_all_objects()), but each time the Network's objects and its contained_objects are sorted freshly as the sorted objects are not stored anywhere. Though this doesn't affect anything much, I feel sorting repeatedly whenever the function is called, may bring in some slowness esp. with large networks with many elements. 

-  I made the change all_objects instead of self.objects wherever I encountered the usage of self.objects, but I'm not fully confident that the changes are disspiated throughout the implementation with preserving contextual correctness (like initially we didn't consider the sort_object() function but later it made a big change in network behaviour) in some places where self.objects indirectly plays the role.

Till now, I was ony hoping for the correctness, without considering effeciency as integral part :disappointed: .
So it would be really perfect if you kindly suggest any improvements after reviewing to tailor out unnecessary parts. :smile:

mstimberg:
It is perfectly fine to first work on correctness, and then consider refactoring the code for better performance. In this case, I don't think doing work repeatedly is a problem, even if we had 100s of objects in a `Network`, the time to sort a list would still be negligible. The important thing is that changes in the object list are taken into account everywhere, e.g. when you call a function like `Network.get_states`, then add something to the network, and then call `Network.run`. We could certainly store the results of the sorting procedure, but we'd have to take care to "invalidate" this storage whenever necessary, without forgetting corner cases. I don't think this is worth it. The only place where I could potentially see a benefit of doing this is in `Network.run`/`Network.before_run`/`Network.after_run`, because 1) we are sure that the network does not change between these calls and 2) there is another common operation that we are doing for these three functions, we are filtering out objects where the `active` flag is `False`. So it might be a nice idea to have an instance attribute `_active_objects` in `Network` which contains all objects (i.e. objects + their contained objects) in a sorted order, but note those that are inactive. However, this is not strictly necessary for merging this PR.

In general, I'd say the changes in this PR look good, but I have a few more specific comments, I'll make them directly in the code.

brian2/core/network.py
```diff
@@ -385,38 +385,42 @@ def profiling_info(self):
     _globally_stopped = False
 
     def __getitem__(self, item):
+        all_objects = self.get_all_objects()
```
mstimberg: I'm not 100% sure about this, but I think for `__getitem__`, etc. (i.e. the methods that make it possible to use `Network` like a `dict`) it would be better to only use `self.objects`, since it is a user-facing syntax, and the user is thinking of the network as containing e.g. `NeuronGroup` objects, but not something like `StateUpdater`.

It is a rarely used feature (and if it is used, then mostly for "major objects" such as `NeuronGroup`), but it would break backwards compatibility. @thesamovar : what do you think?

brian2/core/network.py
```diff
@@ -461,6 +464,32 @@ def add(self, *objs):
                         raise TypeError("Can only add objects of type BrianObject, "
                                         "or containers of such objects to Network")
 
+    def get_all_objects(self):
+
+        all_objects = []
+
+        def add_child_objects(*objs):
```
mstimberg: I don't think this function should be defined within the `get_all_objects` function (and therefore redefined at each call)

brian2/core/network.py
```diff
@@ -461,6 +464,32 @@ def add(self, *objs):
...
+            for obj in objs:
+                if isinstance(obj, BrianObject):
+                    if obj not in all_objects:  # Don't include objects twice
+                        all_objects.append(obj)
+                    add_child_objects(obj.contained_objects)
+                else:
+                    # allow adding values from dictionaries
```
mstimberg: This is unecessary here: we want users to be able to add objects from dictionaries when they call `Network(...)` or `Network.add(...)`, but here the function works on an internal attribute, `Network.objects`, which is always a list.

brian2/core/network.py
```diff
@@ -461,6 +464,32 @@ def add(self, *objs):
...
+                    if isinstance(obj, Mapping):
+                        add_child_objects(*obj.values())
+                    else:
+                        try:
+                            for o in obj:
+                                if o is obj:
```
mstimberg: Same comment as above, this cannot happen for `self.objects`.

brian2/core/network.py
```diff
@@ -487,11 +516,12 @@ def remove(self, *objs):
...
 
     def _full_state(self):
+        all_objects = self.get_all_objects()
         state = {}
-        for obj in self.objects:
+        for obj in all_objects:
             if hasattr(obj, '_full_state'):
                 state[obj.name] = obj._full_state()
-        clocks = set([obj.clock for obj in self.objects])
+        clocks = set([obj.clock for obj in all_objects])
```
mstimberg: Minor stylistic point (and this also applies to the code before your change), this would be nicer when written as:
```Python
clocks = {obj.clock for obj in all_objects}
```

brian2/core/network.py
```diff
@@ -526,7 +556,8 @@ def store(self, name='default', filename=None):
         versions. If you are interested in storing the state of a network for
         documentation or analysis purposes use `Network.get_states` instead.
         '''
-        clocks = [obj.clock for obj in self.objects]
+        all_objects = self.get_all_objects()
+        clocks = [obj.clock for obj in all_objects]
```
mstimberg: Again, already a shortcoming of the previous code, but this should be rather a set (`{...}`) instead of a list, we don't need to store the same `Clock` repeatedly.

brian2/core/network.py
```diff
@@ -565,15 +596,16 @@ def restore(self, name='default', filename=None):
...
         '''
+        all_objects = self.get_all_objects()
         if filename is None:
             state = self._stored_state[name]
         else:
             with open(filename, 'rb') as f:
                 state = pickle.load(f)[name]
         self.t_ = state['0_t']
-        clocks = set([obj.clock for obj in self.objects])
+        clocks = set([obj.clock for obj in all_objects])
```
mstimberg: And again: `{...}`

brian2/devices/cpp_standalone/device.py
```diff
@@ -906,14 +907,45 @@ def write_static_arrays(self, directory):
             arr.tofile(os.path.join(directory, 'static_arrays', name))
             static_array_specs.append((name, c_data_type(arr.dtype), arr.size, name))
         self.static_array_specs = static_array_specs
-            
+    
+    def get_all_objects(self, net):
```
mstimberg: This seems to be quite redundant with the code in `Network`, can you not directly call `net._sort_objects()`?

brian2/devices/cpp_standalone/device.py
```diff
@@ -906,14 +907,45 @@ def write_static_arrays(self, directory):
...
+        return(all_objects)
+        
     def find_synapses(self):
         # Write the global objects
         networks = [net() for net in Network.__instances__()
                     if net().name != '_fake_network']
         synapses = []
         for net in networks:
-            net_synapses = [s for s in net.objects if isinstance(s, Synapses)]
+            net_synapses = [s for s in self.get_all_objects(net) if isinstance(s, Synapses)]
```
mstimberg: `Synapses` are never `contained_objects` of other objects, so looking at `net.objects` should be enough here.

brian2/core/network.py
Vigneswaran-Chandrasekaran: >It would be better to only use self.objects, since it is a user-facing syntax, and the user is thinking of the network as containing

Yes, I agree with that. I had the same thought when making the changes and also followed it in functions like __get_state()__ and __remove()__ (uses self.objects still) as they are specific to querying over user-defined objects, but in ____getitem____(), I made myself reluctant thinking it is utilized by modules to manage objects (including __contained_objects__). 
Now it is all clear, I'll make the change.

brian2/devices/cpp_standalone/device.py
Vigneswaran-Chandrasekaran: Yes, I can. Sorry for that poor one.

brian2/core/network.py
Vigneswaran-Chandrasekaran: Yes, but we want __all_objects__ to be calculated whenever function __get_all_objects()__ is called and we don't store it anywhere, along with we want to have the function to be recursive to traverse heirarchically. 
With these needs, I found difficult to place list (all_objects) intialization and to utilize the list to update with object members with function recursion. I'll make the desirable changes.

Vigneswaran-Chandrasekaran:
Thank you very much @mstimberg :blush: 
I'll make the necessary changes.

Vigneswaran-Chandrasekaran:
Thanks @mstimberg, for mentioning the areas, which have to refactor. 
From your points, I've made the changes. Now I think everything working fine.

> I think for __getitem__, etc. (i.e. the methods that make it possible to use Network like a dict) it would be better to only use self.objects

Currently, I made all those functions to use self.objects.

> The only place where I could potentially see a benefit of doing this is in Network.run/Network.before_run/Network.after_run, because 1) we are sure that the network does not change between these calls and 2) there is another common operation that we are doing for these three functions, we are filtering out objects where the active flag is False. 
So it might be a nice idea to have an instance attribute _active_objects in Network which contains all objects (i.e. objects + their contained objects) in sorted order, but note those that are inactive. 

Yes, this is a good approach. But I have got a doubt that won't objects flagged as active=False, would turn active during some time and if so again we need to check & update changes encountered to maintain correctness? And I also think, bringing in Network attribute for objects would again bring back our current problem. Kindly, as usual correct me if I understood things wrongly :blush:. 

I think I resolved the required changes and would be sure if all the test cases passes. But our Appveyor build is been queued for long time (7 hours) and I think its blocked. I tried to search in their forum about this issue, `https://help.appveyor.com/discussions/problems/15716-my-builds-are-queued-for-too-long`, `https://help.appveyor.com/discussions/problems/884-build-queued#comment_34972495`.
It makes me to hold my nerve to check the correctness of the changes :laughing:.

mstimberg:
> Yes, this is a good approach. But I have got a doubt that won't objects flagged as active=False, would turn active during some time and if so again we need to check & update changes encountered to maintain correctness? And I also think, bringing in Network attribute for objects would again bring back our current problem. Kindly, as usual correct me if I understood things wrongly blush.

Objects can only change their active attribute in between runs, and for each run we call `before_run`. If the attribute is only used in `before_run`/`run`/`after_run`, then there should not be any problem.

> But our Appveyor build is been queued for long time (7 hours) and I think its blocked.

Unfortunately, with the free account appveyor does not run any jobs in parallel, so each test run takes several hours to complete. We might switch to Azure Pipelines at some point.

Vigneswaran-Chandrasekaran:
>If the attribute is only used in before_run/run/after_run, then there should not be any problem.

Okay. So I guess I would add a separate object which will only be used by these three functions.

mstimberg:
> Okay. So I guess I would add a separate object which will only be used by these three functions.

Actually, I'd say don't worry about it. The current version does a bit of unnecessary work, but I think it is clearer than having this attribute which is potentially outdated between runs (even though it does not matter at that point). There is less potential for errors this way and repeatedly filtering a list has a negligible performance impact.

mstimberg:
Everything is fine from my side now, I'll shuffle functions around a bit more, but I think it is easier for me to quickly do this myself instead of trying to explain it :smile: After that, I'll wait for the tests to run through and if everything looks fine, we can finally merge this thing! Thanks for your contribution!

Vigneswaran-Chandrasekaran:
>Everything is fine from my side now

Great

> I'll shuffle functions around a bit more, but I think it is easier for me to quickly do this myself instead of trying to explain it smile 

:smile: Yes, I'll try to figure out from the changes.

>Thanks for your contribution!

I should be thankful to be frank :blush:. I learnt a lot from this issue and it gave me a good starting point for Brian. It is also my first PR :laughing:

mstimberg:
The tests successfully ran through, so I'm merging this now, thanks again!

The appveyor tests are not running at the moment, you were right that there is a problem. I opened a [support ticket](https://help.appveyor.com/discussions/problems/23567-builds-no-longer-starting-both-automatic-and-manual-builds) with the appveyor team, let's see what comes out of it. But that shouldn't be an issue for this PR, as it does not touch on parts of the code where differences between UNIX and Windows are relevant.

PR #1071, opened 2019-04-09
Start deprecating brian2.hears

thesamovar:
Just updated warnings and docs to reflect that we will deprecate brian2.hears in favour of brian2hears.

### DIFF
--- a/brian2/hears.py
+++ b/brian2/hears.py
@@ -1,6 +1,9 @@
 '''
 This is only a bridge for using Brian 1 hears with Brian 2.
 
+.. deprecated:: 2.2.2.2
+    Use the `brian2hears <https://github.com/brian-team/brian2hears>`_ package instead.
+
 NOTES:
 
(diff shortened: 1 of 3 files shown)


mstimberg:
> Note: didn't check that the docs build correctly yet or that it passed tests, although I think it should be fine.

Forgot to say: docs build fine and tests pass as well – the test failures on appveyor are a general thing (incompatibility between weave 0.16 and numpy 1.16 which was fixed in weave 0.17, but 0.17 is not packages for 32bit Windows) that I just fixed in master.

thesamovar:
Sure! I went with github because we might conceivably change where we host the docs, but changing the repo is unlikely, but actually it's very unlikely and we can just update the message if we do! Do you want to just go ahead and make that change and then merge or shall I?

mstimberg:
Ok, I'll make the change and merge!

Issue #970, opened 2018-06-12
Do not access Python data structures from weave/cython


mstimberg:
Currently, a major slow down in run time mode is our access of Python data structures from our generated C++ code (weave/Cython). For example, monitors and synapse generation resize dynamical arrays, and spike propagation accesses the `SpikeQueue`. For the latter, we are already using a C implementation by default, but accessing it from C++ code goes via its Python wrapper, adding unnecessary overhead.

I propose that we:
* [ ] write a C++ implementation of the `DynamicArray` class and provide a Cython wrapper around it so that it can still be used from numpy
* [ ] remove the Python implementation of the `SpikeQueue` and always use the Cython version
* [ ] directly access the underlying C++ implementation of the `SpikeQueue` from C++ code

The only disadvantage of this that I am seeing is that you can no longer run Brian from source without compiling the `SpikeQueue` and `DynamicArray`. On the other hand, if you are using the code directly from github instead of from conda or pypi (where we provide binary packages for Windows and OS X already, and can potentially provide compiled packages for Linux as well), you are probably also able to call a command to compile the extensions manually.

I see two big advantages of doing things this way:
* Improved performance for weave/Cython (always good)
* we can use the same set of templates for the code objects in weave and C++ standalone

If, in the longer run, we also get rid of Cython by writing our own weave replacement, this would greatly reduce the number of templates and the code you have to write when implementing a new feature.

thesamovar:
Very good idea. Strictly speaking, we don't need to get rid of our Python implementation: that could still be a fallback. Not sure if that's necessary any longer given how easy it is to install binaries now.

mstimberg:
True, we could keep the Python implementation as a fallback for the `numpy` target. I'm also not sure whether this is worth it, my main argument against it would be that it means more code, more testing, and more possible bugs.

thesamovar:
As a general rule I agree with that, although this particular piece of code is now very stable. I think we can be pretty confident about it?

PR #1073, opened 2019-04-23
[MRG] Python 2/3 compatiblity

mstimberg:
This PR changes the code base into one that can be used with Python 2 or 3, without needing to translate the code with `2to3.py`. I used the `future` package for a few specific changes (use of `basestring`, the `builtins` package), but mostly I only used "modern Python 2" that is also Python 3 compatible. Our previous code sometimes used things like `iteritems()` instead of `items()`, I did not bother keeping this distinction for Python 2, i.e. all code uses the Python 3 variant which is suboptimal on Python 2. However, it is only used on things like dictionary of equations/functions/etc., i.e. on data structures with relatively few elements where efficiency should not matter. We should have a look at the benchmark data after merging, though, I might have overlooked something.


thesamovar:
I haven't looked in detail at the changes - too many of them. I don't think I'll have the time to go through that in detail. However, in general I think this is a great idea. Slight concern about switchings iteritems to items for python 2 is that although they're only used for small dicts, in the codegen stuff they're called a LOT, so small differences could add up. Any way to do the performance comparison before merging? Otherwise, I'm happy for this to happen without having looked in detail at the code assuming it passes all tests.

mstimberg:
> I haven't looked in detail at the changes - too many of them. I don't think I'll have the time to go through that in detail.

Sure, most of the changes are boring automatic changes any way (e.g. `from __future__ import absolute_imports` everywhere).

>  Slight concern about switchings iteritems to items for python 2 is that although they're only used for small dicts, in the codegen stuff they're called a LOT, so small differences could add up.

I don't really think that the microsecond differences for the iteration will be measurable, we are doing way more expensive stuff during code gen. But my intuition might well be wrong, it would be certainly good to measure it!

> Any way to do the performance comparison before merging?

Yes, I think I can run the benchmarks also on a different branch, I'll run them overnight.

mstimberg:
The benchmark results are not the most readable ones, but I don't see any slow-down with the changes in this branch: https://brian-team.github.io/brian2_benchmarks/

However, not related to this PR, Cython compilation times went up quite dramatically recently (see e.g. the [CUBA benchmark](https://brian-team.github.io/brian2_benchmarks/#benchmarks.FullExamples.time_CUBA_example?branch=master&p-target='cython'&commits=d71a9993-c062d072)), I'll try to figure out what is going on there.

mstimberg:
> However, not related to this PR, Cython compilation times went up quite dramatically recently (see e.g. the [CUBA benchmark](https://brian-team.github.io/brian2_benchmarks/#benchmarks.FullExamples.time_CUBA_example?branch=master&p-target='cython'&commits=d71a9993-c062d072)), I'll try to figure out what is going on there.

Um, I think it is just that the older benchmarks were taken when the benchmarking script did not delete the Cython/weave cache before running each test :roll_eyes: I'll have to run them all again, I guess.

I did not realize how slow Cython is, though... On my machine, the CUBA example (not a very complex one in the first place) takes ~2s of compilation with weave and ~13s with Cython! Given that most new users will use Python 3 and therefore Cython, this is not great. I'll have a look at whether there is something obvious we can do to make Cython compilation faster. Otherwise, we should maybe prioritize #970 and get started with our own `weave` replacement (which would be much simpler than `weave`, rather like http://pyinline.sourceforge.net/).

mstimberg:
For the record: I updated the [benchmark results](https://brian-team.github.io/brian2_benchmarks/#/). I also ran the benchmarks for the `2to3` branch and did not notice any significant difference.

mstimberg:
> Otherwise, I'm happy for this to happen without having looked in detail at the code assuming it passes all tests.

Tests are passing now (there were some conda issues unrelated to the changes in this PR), and my benchmarks did not show any noticeable change in performance. I'm therefore going ahead merging this!

thesamovar:
Nice!

PR #1074, opened 2019-04-25
[WIP] Compile Cython in parallel

mstimberg:
As mentioned in #1073, Cython takes a long time to compile, starting a simple example like CUBA from scratch (i.e. with an empty cache) takes about 13s on my machine. Fairly complex templates like the one for synapse generation take ~1s to get turned into C++ code, and a further ~3s to compile. There is not that much we can do about, maybe compiling regularly used "support functions" (e.g. `rand`) only once into a library that then gets linked and referred to from the Cython code would help a bit, but this would be a major change and come with all kind of new potential bugs.

There is one pretty straight forward thing, though, that I've been thinking about for a while but never got around implementing: at a start of a run, we call `compile()` on all `CodeObject`s, before they get used during the run. This is done sequentially, but could also be done in parallel. This is what this PR does: `compile()` calls `create_extension`, but if the Cython module is not yet available in the disk or memory cache it does not compile it and wait for completion. Instead, it fires of a `Process` and returns the `Process` object. Later, when the module is actually needed (i.e. in `CodeObject.run`), we join this process to make sure that it has finished. We then get the module and carry on.

This slightly complicates the logic for file locking (which now becomes necessary even when we don't run multiple Cython scripts in parallel), but it is not too bad either. I'd say we remove/deprecate the preference that allows the user to switch off lock file usage (don't think any user ever did this anyway), this simplifies things a bit.

I implemented these changes in this PR, apart from some documentation and additional testing it seems to work on my machine, the Cython test suite (which runs tests in parallel and is therefore quite suited for finding problems with file locking) runs fine.

Oh, and I forgot: for the CUBA example, it brings down the compilation time from ~13s to ~7s, this is quite significant! Did I overlook something (potentially the test suite will tell...)?

### DIFF
--- a/brian2/codegen/generators/cython_generator.py
+++ b/brian2/codegen/generators/cython_generator.py
@@ -236,12 +236,15 @@ class CythonCodeGenerator(CodeGenerator):
                             "global _namespace_num{var_name}",
                             "cdef _numpy.ndarray[{cpp_dtype}, ndim=1, mode='c'] _buf_{var_name} = _namespace['{var_name}']",
                             "_namespace{var_name} = <{cpp_dtype} *> _buf_{var_name}.data",
-                            "_namespace_num{var_name} = len(_namespace['{var_name}'])"
+                            "_namespace_num{var_name} = _buf_{var_name}.shape[0]"
                         ]
-                        support_code.append(
+                        support_code.extend([
                             "cdef {cpp_dtype} *_namespace{var_name}".format(
                                 cpp_dtype=get_cpp_dtype(ns_value.dtype),
-                                var_name=ns_key))
+                                var_name=ns_key),
+                            "cdef int _namespace_num{var_name}".format(
+                                var_name=ns_key)
+                        ])
 
                     else:  # e.g. a function
                         newlines = [
@@ -326,7 +329,7 @@ class CythonCodeGenerator(CodeGenerator):
                                 "cdef {cpp_dtype} * {array_name} = <{cpp_dtype} *> _buf_{array_name}.data"]
 
                 if not var.scalar:
-                    newlines += ["cdef int _num{array_name} = len(_namespace['{array_name}'])"]
+                    newlines += ["cdef int _num{array_name} = _buf_{array_name}.shape[0]"]
 
                 if var.scalar and var.constant:
                     newlines += ['cdef {cpp_dtype} {varname} = _namespace["{varname}"]']
--- a/brian2/codegen/runtime/cython_rt/cython_rt.py
+++ b/brian2/codegen/runtime/cython_rt/cython_rt.py
@@ -109,6 +109,9 @@ class CythonCodeObject(NumpyCodeObject):
         self.libraries = (list(prefs['codegen.cpp.libraries']) +
                           compiler_kwds.get('libraries', []))
         self.sources = compiler_kwds.get('sources', [])
+        self.compiled_code = None
+        self.build_process = None
+        self.module_name = None
 
     @classmethod
     def is_available(cls):
@@ -134,7 +137,7 @@ class CythonCodeObject(NumpyCodeObject):
             return False
 
     def compile(self):
-        self.compiled_code = cython_extension_manager.create_extension(
+        self.module_name, self.build_process = cython_extension_manager.create_extension(
             self.code,
             define_macros=self.define_macros,
             libraries=self.libraries,
@@ -145,9 +148,14 @@ class CythonCodeObject(NumpyCodeObject):
             compiler=self.compiler,
             owner_name=self.owner.name+'_'+self.template_name,
             sources=self.sources
-            )
+        )
         
     def run(self):
+        if self.compiled_code is None:
+            if self.build_process is not None:
+                self.build_process.join()
+                self.build_process = None
+            self.compiled_code = cython_extension_manager.get_module(self.module_name)
         return self.compiled_code.main(self.namespace)
 
     # the following are copied from WeaveCodeObject
--- a/brian2/codegen/runtime/cython_rt/extension_manager.py
+++ b/brian2/codegen/runtime/cython_rt/extension_manager.py
@@ -56,6 +56,36 @@ def get_cython_extensions():
             '.dll', '.obj', '.exp', '.lib'}
 
 
+def assure_lib_dir():
+    lib_dir = get_cython_cache_dir()
+    if '~' in lib_dir:
+        lib_dir = os.path.expanduser(lib_dir)
+    try:
+        os.makedirs(lib_dir)
+    except OSError:
+        if not os.path.exists(lib_dir):
+            raise IOError(
+                "Couldn't create Cython cache directory '%s', try setting the "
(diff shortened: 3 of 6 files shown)


thesamovar:
I'm getting a lot of errors on my local machine, like this (plus many warnings):

```
ERROR: brian2.tests.test_poissongroup.test_time_dependent_rate
    self.test(*self.arg)
  File "D:\programming\brian\brian2\brian2\tests\test_poissongroup.py", line 64, in test_time_dependent_rate
    run(2*ms)
  File "D:\programming\brian\brian2\brian2\units\fundamentalunits.py", line 2375, in new_f
    result = f(*args, **kwds)
  File "D:\programming\brian\brian2\brian2\core\magic.py", line 371, in run
  File "D:\programming\brian\brian2\brian2\core\magic.py", line 231, in run
  File "D:\programming\brian\brian2\brian2\core\base.py", line 276, in device_override_decorated_function
    return func(*args, **kwds)
  File "D:\programming\brian\brian2\brian2\units\fundamentalunits.py", line 2375, in new_f
    result = f(*args, **kwds)
  File "D:\programming\brian\brian2\brian2\core\network.py", line 951, in run
    self.before_run(namespace)
  File "D:\programming\brian\brian2\brian2\core\base.py", line 276, in device_override_decorated_function
    return func(*args, **kwds)
  File "D:\programming\brian\brian2\brian2\core\network.py", line 843, in before_run
  File "D:\programming\brian\brian2\brian2\core\network.py", line 841, in before_run
    obj.before_run(run_namespace)
  File "D:\programming\brian\brian2\brian2\groups\group.py", line 1128, in before_run
    codeobj_class=self.codeobj_class
  File "D:\programming\brian\brian2\brian2\codegen\codeobject.py", line 416, in create_runner_codeobj
    compiler_kwds=compiler_kwds
  File "D:\programming\brian\brian2\brian2\devices\device.py", line 326, in code_object
    codeobj.compile()
  File "D:\programming\brian\brian2\brian2\codegen\runtime\cython_rt\cython_rt.py", line 150, in compile
    sources=self.sources
  File "D:\programming\brian\brian2\brian2\codegen\runtime\cython_rt\extension_manager.py", line 216, in create_extension
    p.start()
  File "C:\Anaconda\lib\multiprocessing\process.py", line 130, in start
    self._popen = Popen(self)
  File "C:\Anaconda\lib\multiprocessing\forking.py", line 258, in __init__
    cmd = get_command_line() + [rhandle]
  File "C:\Anaconda\lib\multiprocessing\forking.py", line 358, in get_command_line
    is not going to be frozen to produce a Windows executable.''')
RuntimeError: 
            Attempt to start a new process before the current process
            has finished its bootstrapping phase.

            This probably means that you are on Windows and you have
            forgotten to use the proper idiom in the main module:

                if __name__ == '__main__':
                    freeze_support()
                    ...

            The "freeze_support()" line can be omitted if the program
            is not going to be frozen to produce a Windows executable.

Error encountered with object named "poissongroup_1_thresholder".
    self.test(*self.arg)

An error occurred when preparing an object. RuntimeError: 
            Attempt to start a new process before the current process
            has finished its bootstrapping phase.

            This probably means that you are on Windows and you have
            forgotten to use the proper idiom in the main module:

                if __name__ == '__main__':
                    freeze_support()
                    ...

            The "freeze_support()" line can be omitted if the program
            is not going to be frozen to produce a Windows executable.
```

mstimberg:
Ugh, let's reconsider dropping Windows support ;) What script did you run to run the tests? Maybe the fix is something as trivial as adding the `if __name__ == '__main__'` line to `run_nose_tests_cython.py` and other script that do not have it? The script `run_nose_tests.py`  already has it, though.

thesamovar:
That's no good though because users shouldn't have to write if name=main, so everything should work on the assumption that it's not run like that.

thesamovar:
But yes it was from ``run_nose_tests_cython.py`` which doesn't have an if name=main block. Trying again now using that.

thesamovar:
Still getting a lot of these warnings which are presumably either mistakes or should be caught?

```
Running tests that do not use code generation
..................S.................S........WARNING    Integrating equations with GSL is still considered experimental [brian2.stateupdaters.GSL]
...................WARNING    The expression "i1 / 1" divides two integer values. In previous versions of Brian, this would have used either an integer ("flooring") or a floating point division, depending on the Python version and the code generation target. In the current version, it always uses a floating point division. Explicitly ask for an  integer division ("//"), or turn one of the operands into a floating point value (e.g. replace "1/2" by "1.0/2") to no longer receive this warning. [brian2.parsing.bast.floating_point_division]
....................................................................................................................WARNING    The SpikeGeneratorGroup contains spike times earlier than the start time of the current run (t = 5. ms), these spikes will be ignored. [brian2.input.spikegeneratorgroup.ignored_spikes]
.............WARNING    The expression "(i + 1) / N" divides two integer values. In previous versions of Brian, this would have used either an integer ("flooring") or a floating point division, depending on the Python version and the code generation target. In the current version, it always uses a floating point division. Explicitly ask for an  integer division ("//"), or turn one of the operands into a floating point value (e.g. replace "1/2" by "1.0/2") to no longer receive this warning. [brian2.parsing.bast.floating_point_division]
........................................................................................
```

thesamovar:
OK putting an if name=main block leads to just as many errors, but now a different one:

```
EERROR      Brian 2 encountered an unexpected error. If you think this is bug in Brian 2, please report this issue either to the mailing list at <http://groups.google.com/group/brian-development/>, or to the issue tracker at <https://github.com/brian-team/brian2/issues>. Please include this file with debug information in your report: c:\users\dgoodman\appdata\local\temp\brian_debug_pg8gp_.log  Additionally, you can also include a copy of the script that was run, available at: c:\users\dgoodman\appdata\local\temp\brian_script_ciss_3.py Thanks! [brian2]
  File "<string>", line 1, in <module>
  File "C:\Anaconda\lib\multiprocessing\forking.py", line 381, in main
    self = load(from_parent)
  File "C:\Anaconda\lib\pickle.py", line 1384, in load
    return Unpickler(file).load()
  File "C:\Anaconda\lib\pickle.py", line 864, in load
    dispatch[key](self)
  File "C:\Anaconda\lib\pickle.py", line 886, in load_eof
    raise EOFError
EOFError
```

mstimberg:
> Still getting a lot of these warnings which are presumably either mistakes or should be caught?

The warnings have been there for a long time (i.e. this is not specific to this PR), but I agree that they are quite ugly...

Windows is really not fun for multiprocessing in Python! I changed the code to use threads instead of processes, that should avoid this specific problem at least. Under Linux, the performance gain is less impressive (compilation time around 10s now), but it still seems to be useful.

I'll have to clean things up and document after my vacation, though.

thesamovar:
I think threads has very little effect in Python because of the GIL right?

mstimberg:
> I think threads has very little effect in Python because of the GIL right?

In general yes, but in this case a lot of time should be taken by the compilation which is done outside of Python and therefore not limited by the GIL. Also, the compilation process is at least partly limited by the speed of IO which also releases the GIL.

mstimberg:
Closing this since the long-term plan is to get rid of Cython.

Issue #853, opened 2017-06-07
store / restore for cpp_standalone


otdewiljes:
Any idea whether plans exist to implement store / restore for C++ standalone mode?

Having the ability to save and load intermediate/branching stages of a long project would greatly improve the applicability of Python-orchestrated C++ standalone mode.

mstimberg:
I just realized that we never replied to this issue :worried:... We are indeed interested in bringing this functionality to standalone, but it is unfortunately not trivial. The main problem is integrating this with the way standalone mode executes code.

Right now, in standalone mode, when you do:
```Python
set_device('cpp_standalone', build_on_run=False)
# set up network
run(...)
# change something
run(...)
device.build()
```
It will take in all the information until the `device.build()` line, generate code and execute it.
Now, with store/restore, things rather look like this:
```Python
# set up network
store()
run(...)
# do something with the results
restore()
# change something
run(...)
```
This means that at the `#do something with the results` place we already need to have run the standalone simulation (otherwise we wouldn't have access to the results), and the following run would somehow have to start another simulation, but one that doesn't do exactly the same thing.

I think this could be a very powerful feature, but there are many decisions to take about how it would look like both from the user's and from the implementation perspective.

otdewiljes:
Well, it's great to see that this issue is still on the radar. For my use case, it would be rendered moot if ever something GPU-accelerated became available as backend/device (brian2cuda? I'd imagine it would be easier to use tensorflow as a backend, though) WITH store/restore.

PR #1086, opened 2019-06-05
[MRG] Correctly store/restore internal `SpikeGeneratorGroup` state

### LINKED ISSUE #1084: SpikeGeneratorGroup not correctly working after restore
A user report an issue on the [mailing list](https://groups.google.com/d/msg/briansupport/SSdLkRigOJU/br__yN0IBAAJ). The following code leads to a broadcasting error with the `numpy` target (and leads to incorrect values on Cython/weave):
```Python
nb_sim = 2                           # number of successive simulations
time_simulation = 100*ms             # time of each simulation          
N_in = 3                             # number of input neurons
freq = 8                             # frequence

times = array([10*k + (1000/freq)*n for n in range(5) for k in range(5) for _ in range(N_in)])*ms
indices = array([k for _ in range(round(times.size/N_in)) for k in range(N_in)])
G_in = SpikeGeneratorGroup(N_in, indices, times)

spikemon_in = SpikeMonitor(G_in)
net = Network(G_in, spikemon_in)

net.store('initialized network')

for k in range(nb_sim):
    net.restore('initialized network')
    net.run(time_simulation)
    
    print(spikemon_in.t)
    print(spikemon_in.i)
    
    print('Epoch {} completed'.format(k + 1))
```
The error is:
```
ValueError: could not broadcast input array from shape (75) into shape (4)
...
The error was raised in the following line:
_array_spikegeneratorgroup__spikespace[:_n_spikes] = _indices
```
This suggests that the `SpikeGeneratorGroup` tries to emit all spikes at once, which seems similar to #1017, but here we do not have any spike times "in the past".

A workaround is to use `SpikeGeneratorGroup.set_spikes` after the `restore` call, i.e. in this example:
```Python
for k in range(nb_sim):
    net.restore('initialized network')
    G_in.set_spikes(indices, times)
    net.run(time_simulation)
    ...
```

mstimberg:
In `SpikeGeneratorGroup`, we use variables directly stored in the object (i.e. not as `Variables`) to decide whether we need to rebuild the integer timestep representation of spike times (e.g. when `dt` has changed between runs). In principle, we *could* store them as "proper" variables, but that would also be somewhat inconvenient: e.g. we currently set `_previous_dt = None` to note that the simulation has not been run yet; if it were a variable using our full machinery, then it would necessarily have to store a value in seconds,  so we'd need some workaround like `_previous_dt = 0*second` to denote "not run yet".

With this PR I opted for the straightforward solution to fix #1084: we now manually store/restore the two internal variable values.

Using proper variables might have been a cleaner option, but it would also have been a more fundamental change and potentially introduce new problems (e.g. for C++ standalone or Brian2GeNN).

### DIFF
--- a/brian2/input/spikegeneratorgroup.py
+++ b/brian2/input/spikegeneratorgroup.py
@@ -145,6 +145,20 @@ class SpikeGeneratorGroup(Group, CodeRunner, SpikeSource):
 
         self.variables['period'].set_value(period)
 
+    def _full_state(self):
+        state = super(SpikeGeneratorGroup, self)._full_state()
+        # Store the internal information we use to decide whether to rebuild
+        # the time bins
+        state['_previous_dt'] = self._previous_dt
+        state['_spikes_changed'] = self._spikes_changed
+        return state
+
+    def _restore_from_full_state(self, state):
+        state = state.copy()  # copy to avoid errors for multiple restores
+        self._previous_dt = state.pop('_previous_dt')
+        self._spikes_changed = state.pop('_spikes_changed')
+        super(SpikeGeneratorGroup, self)._restore_from_full_state(state)
+
     def before_run(self, run_namespace):
         # Do some checks on the period vs. dt
         dt = self.dt_[:]  # make a copy
--- a/brian2/tests/test_spikegenerator.py
+++ b/brian2/tests/test_spikegenerator.py
@@ -8,7 +8,7 @@ import tempfile
 
 from nose import with_setup
 from nose.plugins.attrib import attr
-from numpy.testing.utils import assert_raises, assert_equal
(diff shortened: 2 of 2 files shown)


thesamovar:
Does ``store()``/``restore()`` still not work in standalone? That's the only obvious thing missing here. If that's still not generally implemented then I don't see any issue with this solution.

mstimberg:
> Does `store()`/`restore()` still not work in standalone? That's the only obvious thing missing here. If that's still not generally implemented then I don't see any issue with this solution.

No, we still don't have store/restore functionality in standalone. It's issue #853, and it is unfortunately not trivial. I will reply to that issue, just noticed that no one ever did...

thesamovar:
In that case, good to merge this I think.

PR #1088, opened 2019-06-07
Remove import print_function for compatibility with binder setup

bdevans:
The line "from __future__ import print_function" occurring after the magic "%matplotlib notebook" (added as part of the brian2-binder preparation) and before "from brian2 import *" results in this bug: https://github.com/ipython/ipython/issues/11366

Since Python 2 is soon to be retired, it is probably simplest just to remove the offending line.

### DIFF
--- a/examples/frompapers/Destexhe_et_al_1998.py
+++ b/examples/frompapers/Destexhe_et_al_1998.py
@@ -30,7 +30,6 @@ Notes
 * Other small discrepancies with the paper -- values from the NEURON code were
   used whenever different from the values stated in the paper
 '''
-from __future__ import print_function
 from brian2 import *
 from brian2.units.constants import (zero_celsius, faraday_constant as F,
                                     gas_constant as R)


mstimberg:
I don't want to break the Python 2 compatibility for our example because of a corner case. It is true that Python 2 is only supported until the end of the year, but about 50% of our conda downloads are still the Python 2 version, so this is far from being a small minority. I just opened a "counter pull request" at brian-team/brian2-binder#5 which should deal with this by moving the `from __future__` imports before the magic line. I did not test it very extensively, though.

bdevans:
Hi Marcel,
I've edited the PR branch to hopefully catch an edge case but it otherwise
looks ok. (Also this is the first PR I've dealt with so please correct me
if I'm doing something wrong!) I'm not sure what the need for the clause
where I have added a comment, but otherwise happy to merge.
Best,
Ben

mstimberg:
> I'm not sure what the need for the clause where I have added a comment, but otherwise happy to merge.

Sorry, what comment are you referring to?

bdevans:
> Sorry, what comment are you referring to?
Sorry that wasn't clear - I was referring to your counter PR: [brian-team/brian2-binder#5](https://github.com/brian-team/brian2-binder/pull/5)

mstimberg:
> Sorry that wasn't clear - I was referring to your counter PR: [brian-team/brian2-binder#5](https://github.com/brian-team/brian2-binder/pull/5)

Apologies again, but I'm still confused: you said "where I have added a comment", but I do not see any comment by you in that PR...

bdevans:
> Apologies again, but I'm still confused: you said "where I have added a comment", but I do not see any comment by you in that PR...

I was commenting on the following lines in the PR
```
if len(future_imports):
    # force an empty line
    future_imports.append('')
```
"Is this necessary? If so, should it not also be done for `docstring` (and possibly `other_code`)?"

mstimberg:
Ah, ok, I do not see this comment in the PR, maybe you started a review (where you can give multiple comments, but then need to finalize the review before they are all published at once). Either way, I think the lines are correct because the problem is specific to the combination of the `future_imports` string with the `magic` line: If there are `from __future__` imports, then we need a line after them before the `%matplotlib notebook` line, but if there are none, then we do not want an empty line before `%matplotlib notebook`. We do not have the same issue for `docstring` (because it is in a separate cell) or for `other_code` (because there is nothing coming after it).

If you are happy with everything, feel free to merge :) And thanks for the review/fix!

bdevans:
> If you are happy with everything, feel free to merge :) And thanks for the review/fix!

Will do and you're welcome - it's been a useful exercise! :)

Issue #973, opened 2018-06-14
Add option to standalone to delete entire directory after run


thesamovar:
Useful for saving space on big runs

mstimberg:
We have an explicit command for this now, see #1104 .

PR #1104, opened 2019-07-01
[MRG] Add function to delete standalone code/results

mstimberg:
This PR adds a `Device.delete_data` method that can be used to delete the code+results directory in standalone mode (it doesn't do anything for runtime). This is a partial fix for #973, but it needs a manual call instead of e.g. an option to `set_device` or `device.build`. I am hesitant to add the latter, since it would have to be something like an exit hook, and I can imagine all kind of problems with this in multiprocessing settings (not sure, though). With the current approach, the user would be responsible for calling the clean up manually, but it is still be a bit nicer than manually asking for `device.project_dir` and deleting it.

The PR also changes the test suite to do this automatically after each test (this leads to a lot of trivial changes, because I did not want to change the semantics of `reinit_devices` that might have been used outside of the test suite). I also gave a `brian_standalone` prefix to the automatically generated temp directories, so that it is clearer for users where those directories filling up `/tmp` come from...


mstimberg:
[As you can see from the commits, there were again a few issues with the builds on travis/appveyor. It should all work now finally]

thesamovar:
One concern here is that this function deletes the entire directory, including any files the user might have put in there. I think that's OK if it's clearly stated, but the name should probably reflect that. So I might make a change like ``delete_data`` -> ``delete_project_directory`` and ``reinit_and_delete`` -> ``reinit_and_delete_project_directory``. It's longer, but less likely that people will make a catastrophic mistake with some warning in the name of the function itself. What do you think?

mstimberg:
Hm, maybe. Not sure what the use case for putting things into the project directory is, though. I think I am slightly hesitant about changing the name to something with `_project_dir`, because I added this as a function of `Device`, and it could potentially have some meaning for devices that do not have a project directory (e.g. freeing up all memory in runtime mode?). If we rename it to `_project_dir`, then I'd move it to `CPPStandaloneDevice` (with our device proxy mechanism, you could still leave `device.delete_project_dir()` in your script when you use runtime mode).

I don't really think it is necessary to rename `reinit_and_delete`, it's really for the test suite only.

mstimberg:
[To make things easier with the PRs going on in parallel, I pushed all the changes for travis/appveyor to master and rebased this PR so that it now only contains the changes that are about the new `delete_data` function]

thesamovar:
I think the ``_project_directory`` name is OK. It will make sense for most devices, and for those it doesn't make any sense for it can just do nothing.

OK for ``reinit_and_delete`` (although if it's for internal use only, maybe ``_reinit_and_delete``?).

Finally, instead of ``brian_standalone`` prefix for temp dirs, maybe have a ``brian_standalone/`` subdirectory? Like that it's easy to delete them all in one go?

mstimberg:
I somewhat forgot about this PR, but since it changes the test suite in many places, it would be good to merge it soon.

I thought a bit about the `delete_data` vs. `delete_project_directory` again, and I think I am still in favour of the former. The main reason is that I feel that

```Python
set_device('cpp_standalone')
...
device.delete_data()
plt.plot(mon.t/ms, mon.v.T/mV)  # raises an error
```
is more obvious to raise an error than
```Python
...
device.delete_project_directory()
plt.plot(mon.t/ms, mon.v.T/mV)  # raises an error
```
I'm not sure that most user's understand that deleting the project directory makes it impossible to access any data from the simulation. Of course, if they decide to delete the project directory, then they probably have a reason for it, but still, they might think the directory just contains code.

> Finally, instead of `brian_standalone` prefix for temp dirs, maybe have a `brian_standalone/` subdirectory? Like that it's easy to delete them all in one go?

In general I'd say, sure, why not, but this is actually not trivial to do. We are using Python's standard `tempfile.mkdtemp` function which does not allow for this, and it isn't straightforward to replace it (the location of the temp directory can be changed by environment variables, is different on different OSes, and so on). Not sure this is worth the effort.

thesamovar:
I see your point but I think the converse also holds: writing ``delete_data()`` may make you think it'll delete the "data" but nothing else (code, other files). How about ``delete_files()``? That feels like more or less the same but the name gives you the alarming sense that you probably shouldn't be doing it unless you know what you're doing.

thesamovar:
And for temp directory that's fine.

mstimberg:
> I see your point but I think the converse also holds: writing `delete_data()` may make you think it'll delete the "data" but nothing else (code, other files). How about `delete_files()`?

I'm still not quite sure what the "other files" that the user could have put into the project directory are? But I see your point. What about going all the way with `delete_code_and_data` ?

thesamovar:
If they write ``dir='.'`` for example?

I'm happier with ``delete_code_and_data`` except that it still doesn't say that it will delete the entire directory.

mstimberg:
> If they write `dir='.'` for example?

Woah, who'd do that :astonished: ? 

But I see what you mean. Maybe it would be actually better to invest some more time and upgrade the function. Since we know which files we generate, we could explicitly delete those. And then we could optionally delete the directory as well, and maybe even require a `force` option if files are present that we did not put there. Something like this:
```Python
device.delete(data=True, code=False)  # only delete data but keep code
device.delete(directory=True)  # delete everything, but don't delete the dir if other files are there
device.delete(directory=True, force=True)  # just delete everything
```

thesamovar:
Well who'd type ``sudo rm -rf /``? ;)

I like your new solution, but there's a slight issue which is that sometimes you change something which changes a file name, and only later update your code to do the deleting part, and then you get some orphaned files left over. But your example syntax above more or less covers that, i.e. allowing for an explicit delete everything (even if it's not generated by this script) option.

mstimberg:
Ok, I pushed a new version which implements what I proposed earlier. The tests are quite basic but should cover most of the relevant use cases. Explanations of the options and their interactions are in the docstring, but if you are fine with everything, I'll also write something for the general documentation.

One non-obvious thing: files in the `static_arrays` folder (used when you do e.g. `group.v = np.random.randn(len(group))*mV` in standalone mode) are deleted with the `code` and not with the `data` option. This might look wrong at the first sight (since these are numpy arrays with values), but the idea is that when you keep the `code`, you can still compile + run everything – running the `main` binary would fail if you deleted the values in `static_arrays`.

mstimberg:
I added some documentation in the standalone section, merging it now!

Issue #636, opened 2016-01-26
Automatic segmentation in multicompartmental models


romainbrette:
When creating a multicompartmental model, a basic question is how many compartments we need to have. This issue also arises when loading files coming from reconstructions, which often have a huge number of compartments, and merging is desired.

One method is to split a branch evenly in units of space length. But the problem is the space length is calculated for DC inputs, and it depends on channel conductances. At high frequencies (transients), the space length is smaller, and if frequency is high enough, it doesn't depend on channel conductances anymore:

lambda(f) = 1/2 * (d/(pi*f*Ri*Cm))^(1/2)

This is what Neuron uses, with f = 100 Hz (which is of course arbitrary) and standard compartment length 0.1*lambda(f). What we notice is that in both cases (lambda(0) and lambda(f)), the space length scales as sqrt(d), ie, like axial resistance. Cm is almost always a constant (except for myelin), of around 1 µF/cm^2; Ri is normally in the order of 100 ohm.cm. Therefore, we can simply consider that, for the purpose of resegmentation, lambda(f) = k*sqrt(d), where k is a constant.

In summary, I propose that automatic resegmentation is based on splitting a branch into compartments of equal length, in units of sqrt(d). One could specify a total number of compartments, or a precision (dimensionless number). A cylinder could also be initialized without specifying the number of compartments, or by specifying a precision value.

There is one issue, which is to deal with truncated cones. Initially, it may be fine to simply consider the average diameter (it won't make much of a difference anyway). Otherwise, we can do a more specific calculation.

rcaze:
We had a long discussion on segmenting with @mstimberg in issue #569 and the segmeter function is segmenting a morphological object
```python
import unittest
import numpy as np
from itertools import izip
import brian2 as br2


def dist2p(pt1,  pt2):
    """Compute the Euclidean distance between two points"""
    return np.sqrt(np.sum([(a-b)**2 for a, b in izip(pt1, pt2)]))


def segmenter(sec, nseg=2):
    """Segment the coordinate of all segments given the coordinates of
    the section and the number of segments
    """
    # Give the distance between all the points of the section and onvert in m
    lengths = br2.asarray(sec.length)*1e6
    # Compute the total distance after each points
    distances = [np.sum(lengths[:i+1]) for i in range(len(lengths))]
    # Give the total length
    tot_length = distances[-1]
    # Give the distances after each segments
    seg_dists = [0.] + [(i*tot_length)/float(nseg)
[...]

PR #1108, opened 2019-07-05
Compartment condensation

chbehrens:
Hi all,

I drafted an implementation of compartment condensation for extended morphologies. Realistic morphologies e.g. based on EM data and loaded via .swc files often contain thousands of nodes and thus by default thousands of compartments in brian2. In order to speed up simulations it is useful to be able to reduce the number of compartments while still maintaining the overall structure. So I added a `condense` function to the `Morphology` class to condense compartments within sections. 

The level of condensation can be controlled by a parameter lambda that is in principle related to the ratio of membrane conductivity to internal conductivity of a compartment - though these are not yet defined at the level of the morphology, so here area and cross section divided by length (resp. `r_length_1/2`) are used instead.

Any suggestions on how to improve this and make it more useful are welcome!
Thanks,
Christian

### DIFF
--- a/brian2/spatialneuron/morphology.py
+++ b/brian2/spatialneuron/morphology.py
@@ -1913,7 +1913,8 @@ class Section(Morphology):
         self._y = np.delete(self._y, i + 1)
         self._z = np.delete(self._z, i + 1)
         self._diameter = np.delete(self._diameter, i + 1) * meter
-        self._length = np.concatenate((self._length[:i], [self._length[i] + self._length[i + 1]], self._length[i + 2:]))
+        self._length = np.concatenate((self._length[:i], [self._length[i] + self._length[i + 1]],
+                                       self._length[i + 2:])) * meter
         self._n = self._n - 1
 
     @property
--- a/brian2/spatialneuron/morphology.py
+++ b/brian2/spatialneuron/morphology.py
@@ -1092,6 +1092,25 @@ class Morphology(object):
             Morphology._replace_three_point_soma(child,
                                                  all_compartments)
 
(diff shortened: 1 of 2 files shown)


mstimberg:
Hi @chbehrens, apologies for not commenting on this for such a long time... In any case, a PR is the perfect way of discussing this.

I'll be away again soon for a week, but here is some quick feedback:

As you say, in principle we would like to have conductivity information that is not available when dealing with a `Morphology`. I therefore wonder whether it would not make sense to somehow integrate this into `SpatialNeuron`. On the other hand, I'd say it feels natural to have this of part of the `Morphology`, there are other things besides this condensation that you might do to prepare the morphology before you create a `SpatialNeuron` object. Maybe it would actually make sense to make `Ri` and `Cm` part of the `Morphology` definition instead of the `SpatialNeuron` definition (we'd still keep it backwards compatible, though)? Or doing it based on length/diameter as you do currently is actually enough? I think this is what @romainbrette was proposing in #636 ? 

One other thought: in your current implementation, the `condense` method changes the `Morphology` in place, I think it might be better if it returned a new object instead (potentially something like a `CondensedMorphology` with an additional attribute for the index mapping?)

Somewhat related: you currently calculate things like the per compartment area once and store it into an attribute instead of recalculating it on-demand. I certainly get the reasoning behind it, but we made this choice on purpose because we found that changing values like the diameter or length of a compartment after creation could be useful. A typical use case would be to correct/change some values in a morphology of a loaded SWC file.

CC @rcaze who was interested in (and started working on) all this quite some time ago...

Issue #1107, opened 2019-07-05
Random numbers in equations


chbehrens:
Hi all,

it would be a great enhancement to be able to draw random numbers from common distributions inside equations. This would for example allow to implement synaptic release noise by drawing the actual number of released vesicles from a Poisson distributions.

mstimberg:
This has been open for a while, but just for reference in case it got lost: we have actually included a new `poisson` function with [Brian 2.3](https://brian2.readthedocs.io/en/stable/introduction/release_notes.html#brian-2-3), so this is partly done. There's still a question of how to more easily add other distributions.

Issue #788, opened 2016-11-28
Improve decision process for function implementations


mstimberg:
The current algorithm to determine which function implementation to use is kind of convoluted:

1. Check for an implementation for the respective `CodeObject` class (e.g. `CPPCodeObject`)
2. Check for an implementation for the `CodeGenerator` class of the `CodeObject` class
3. Repeat the above for all super class of the `CodeObject`/`CodeGenerator`

In addition, you can either register a function for a specific `CodeObject`/`CodeGenerator` class, or for a string name that is compared to the `class_name` attribute of the respective class.

This has been problematic for `brian2cuda` and `brian2genn` which want to inherit from `CPPCodeGenerator` (because most of the code generation is identical), but don't want `"cpp"` function implementations to be used (since they are not defining kernels with `__device__`). For the built-in functions it is enough to provide their own implementation, but this will break for user-defined functions or if we decide to add a new function to Brian 2.

A quick&simple solution would be to have another class attribute, e.g. `inherit_function_implementations` which would be `True` for `WeaveCodeGenerator` but `False` for `CUDACodeGenerator`.

thesamovar:
I'd say that brian2cuda and brian2genn are precisely what we're aiming at with this mechanism, so it definitely makes sense to design the system to fit their use case. I don't have a strong feeling about the implementation, but I do feel like the way you'd define a function for CUDA is not very different to the CPU way. Perhaps one could imagine a refactor that takes that into account? Maybe not. Not sure.

PR #1111, opened 2019-08-07
[MRG] Revisit use of vectorisation_idx in function + add poisson function

mstimberg:
This will need a bit of explanation...

When the user includes a call to `rand()` or `randn()` in the code, the generated code will be actually `rand(vectorisation_idx)`. This is necessary for the numpy target, because we have to know how many random numbers the function call should return. `vectorisation_idx` will be an array of all neuron indices e.g. in the state updater, but only a subset of the indices when used as part of e.g. the reset. In C-based code generation targets, `vectorisation_idx` will always be a single (e.g. neuron) index, which could be potentially useful to generate the same random numbers independent of execution order in OpenMP. This is not implemented yet, though, so C-based targets simply ignore this argument.

Note that an alternative solution to this issue would have been to require calls such as `rand(i)` (where `i` refers to the neuron index) instead of `rand()`, but this would have been ugly, confusing and prone to errors, and also not straightforward to do in synaptic contexts (there is no global "synapse index").

So what we did was instead to translate all calls to function without arguments, to one with the `vectorisation_idx` argument. This worked ok, but was also constraining: it was impossible to have functions that return arrays of values (in numpy), without at the same time having at least one argument that was a vector. For most functions that wouldn't be an issue, because only scalar arguments would also mean that the return value (vector or not) would be identical for all neurons. However, there is a very relevant category of functions where it *is* a problem, i.e. random number generators that take an additional argument. For binomial functions, we somewhat worked around it by having the user create a function object outside of the code with fixed distribution parameters, this special `my_binomial` function could then be called without parameters in the code. This of course only works if the parameters can be fixed in advance (as we require for `PoissonInput`).

In this PR, I implemented a slightly more general but generally backwards-compatible system (Brian2GeNN needs a small fix, though): `Function` objects have an additional attribute `auto_vectorize`, which states that `vectorisation_idx` should be added as an additional argument. This can be used for argumentless functions like `rand`, but also for functions that take an argument.

As part of this PR, I also added a `poisson` (cf. #1107) function which makes use of this system. The user writes `poisson(lambda_value)`, but the generated code will be `poisson(lambda_value, vectorisation_idx)`. The implementation for numpy uses numpy's function, and the C-based target basically use a copy&pasted version of numpy's underlying C code. I briefly considered using the GSL instead, but I did not quite see how to integrate it with our system for setting seeds. Maybe let's discuss about this aspect and which other random distributions we should support in #1107?

### DIFF
--- a/brian2/core/network.py
+++ b/brian2/core/network.py
@@ -865,10 +865,7 @@ class Network(Nameable):
 
         for obj in all_objects:
             if obj.active:
-                try:
-                    obj.before_run(run_namespace)
-                except Exception as ex:
-                    raise brian_object_exception("An error occurred when preparing an object.", obj, ex)
+                obj.before_run(run_namespace)
 
         # Check that no object has been run as part of another network before
         for obj in all_objects:
--- a/brian2/core/network.py
+++ b/brian2/core/network.py
@@ -865,7 +865,10 @@ class Network(Nameable):
 
         for obj in all_objects:
             if obj.active:
-                obj.before_run(run_namespace)
+                try:
+                    obj.before_run(run_namespace)
+                except Exception as ex:
+                    raise brian_object_exception("An error occurred when preparing an object.", obj, ex)
 
         # Check that no object has been run as part of another network before
         for obj in all_objects:
--- a/docs_sphinx/advanced/functions.rst
+++ b/docs_sphinx/advanced/functions.rst
@@ -22,9 +22,11 @@ Default functions
 The following functions (stored in the `DEFAULT_FUNCTIONS` dictionary) are
 ready for use:
 
-* Random numbers: ``rand()``, ``randn()`` (Note that these functions should be
-  called without arguments, the code generation process will take care of
-  generating an array of numbers for numpy).
+* Random numbers: ``rand`` (random numbers drawn from a uniform distribution
+  between 0 and 1), ``randn`` (random numbers drawn from the standard normal
+  distribution, i.e. with mean 0 and standard deviation 1),
+  and ``poisson` (discrete random numbers from a Poisson distribution with rate
+  parameter :math:`\lambda`)
 * Elementary functions: ``sqrt``, ``exp``, ``log``, ``log10``, ``abs``, ``sign``
 * Trigonometric functions: ``sin``, ``cos``, ``tan``, ``sinh``, ``cosh``,
   ``tanh``, ``arcsin``, ``arccos``, ``arctan``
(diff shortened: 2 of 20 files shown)


mstimberg:
> The only thing I would say is that we could do with some documentation for it maybe?

This is a very valid point, I added some documentation, please have a look. Writing the documentation I also realized that 

1. Defining functions that use `auto_vectorise` was too difficult to get right, because the function wrapper could not deal with the fact that the number of arguments is different in the function and in the user code calling it (you'd always get an error about incorrect numbers of arguments for one of them, depending on whether you included `_vectorisation_idx` in the list of arguments you defined units for)
2. Function dependencies (e.g. reusing `rand()` in another function) required that you know about the actual name of the function (e.g. the C++ implementation of `rand` is actually called `_rand`)

Both of these issues should be fixed now as well.

I think it's still not perfect, but good enough for now, bigger fixes could be done as part of #788.

brian2/core/network.py
```diff
@@ -865,10 +865,7 @@ def before_run(self, run_namespace):
 
         for obj in all_objects:
             if obj.active:
-                try:
-                    obj.before_run(run_namespace)
-                except Exception as ex:
-                    raise brian_object_exception("An error occurred when preparing an object.", obj, ex)
```
thesamovar: What's the thinking here?

docs_sphinx/advanced/functions.rst
```diff
@@ -22,9 +22,11 @@ Default functions
...
 ready for use:
 
-* Random numbers: ``rand()``, ``randn()`` (Note that these functions should be
-  called without arguments, the code generation process will take care of
-  generating an array of numbers for numpy).
+* Random numbers: ``rand`` (random numbers drawn from a uniform distribution
+  between 0 and 1), ``randn`` (random numbers drawn from the standard normal
+  distribution, i.e. with mean 0 and standard deviation 1),
+  and ``poisson` (discrete random numbers from a Poisson distribution with rate
+  parameter :math:`\lambda`)
```
thesamovar: Typo here around poisson (start double backticks end single backtick)

thesamovar:
Made a couple of little comments but looks good now!

brian2/core/network.py
mstimberg: Oops, I did not mean to commit this change... I used it during debugging because this way the debugger stops at the actual exception and I have access to the full stack frame.

mstimberg:
Good to merge now, I guess?

thesamovar:
Yep!

Issue #1114, opened 2019-08-22
Unexpected exception in last cell of the first tutorial


rgerkin:
In the first tutorial (`brian2/tutorials/1-intro-to-brian-neurons.ipynb`) on the master branch (cloned a few minutes ago), I get the following exception for the last cell (unexpected; I recognize that some of the exceptions raised in earlier cells are by design):

```
---------------------------------------------------------------------------
ValueError                                Traceback (most recent call last)
<ipython-input-26-d3a3c36b1c36> in <module>
     29 run(duration)
     30 
---> 31 _ = hist(spikemon.t/ms, 100, histtype='stepfilled', facecolor='k', weights=ones(len(spikemon))/(N*defaultclock.dt))
     32 xlabel('Time (ms)')
     33 ylabel('Instantaneous firing rate (sp/s)');

/opt/conda/lib/python3.7/site-packages/matplotlib/pyplot.py in hist(x, bins, range, density, weights, cumulative, bottom, histtype, align, orientation, rwidth, log, color, label, stacked, normed, data, **kwargs)
   2634         align=align, orientation=orientation, rwidth=rwidth, log=log,
   2635         color=color, label=label, stacked=stacked, normed=normed,
-> 2636         **({"data": data} if data is not None else {}), **kwargs)
   2637 
   2638 

/opt/conda/lib/python3.7/site-packages/matplotlib/__init__.py in inner(ax, data, *args, **kwargs)
   1587     def inner(ax, *args, data=None, **kwargs):
   1588         if data is None:
-> 1589             return func(ax, *map(sanitize_sequence, args), **kwargs)
   1590 
   1591         bound = new_sig.bind(ax, *args, **kwargs)

/opt/conda/lib/python3.7/site-packages/matplotlib/axes/_axes.py in hist(self, x, bins, range, density, weights, cumulative, bottom, histtype, align, orientation, rwidth, log, color, label, stacked, normed, **kwargs)
   6690 
   6691         if len(w) != nx:
-> 6692             raise ValueError('weights should have the same shape as x')
   6693 
   6694         for xi, wi in zip(x, w):

ValueError: weights should have the same shape as x
```

Vigneswaran-Chandrasekaran:
Hi, I am newbie to Brian. As each cell in the notebook contains `start_scope()` line in the top, I guess there wouldn't be any previous cells's influence, I'm afraid.

rgerkin:
[...]

PR #1115, opened 2019-08-23
Fix to tutorial

rgerkin:
Required for tutorial to work in Matplotlib 3.1.0, and at least some versions before it.


rgerkin:
See #1114 for details.

PR #1127, opened 2019-11-06
[MRG] switch testing infrastructure to pytest

mstimberg:
This (unfortunately quite massive...) PR switches our test infrastructure from nosetests to pytests, since as [nosetests documentation](https://nose.readthedocs.io/en/latest/) states:
> Nose has been in maintenance mode for the past several years and will likely cease without a new person/team to take over maintainership.

In its current state, this PR should basically make everything that worked with nosetests before work with pytest as well, including stuff like parallel test execution and coverage measurements (the measured coverage dropped from 86% to 82%, I'll try to figure out why this is later). Right now, this does not improve much over the previous state apart from using a maintained software and potentially making it easier for contributors which nowadays are probably more familiar with pytest. However, there is a small improvement for the standalone tests: we now do stuff like resetting the device automatically for every test (which should not take any measurable time for the runtime device), therefore the somewhat [complicated instructions](https://brian2.readthedocs.io/en/stable/developer/guidelines/testing.html#summary) whether you had to add a teardown function or not are now obsolete.

There are also a few things that we can do now to improve the test suite, I'll open a new issue about this.


thesamovar:
OK I don't think I'm going to go through this line by line, but I'm in favour of the principle of it. I'm going to try to run it on my Windows machine now. Is there any particular bit of the code you would like me to look at?

thesamovar:
OK first issue is that it is showing plots while I run the test suite. Could be that I'm on an old version of matplotlib, is that likely to be the issue?

thesamovar:
Lots of these errors too:

```
=================================== ERRORS ====================================
_______________ ERROR collecting brian2/codegen/optimisation.py _______________
..\..\brian2\tests\__init__.py:29: in collect
    while full_name[-1] != os.path.splitext(self.name)[0]:
E   IndexError: list index out of range
________ ERROR collecting brian2/codegen/runtime/numpy_rt/numpy_rt.py _________
..\..\brian2\tests\__init__.py:29: in collect
    while full_name[-1] != os.path.splitext(self.name)[0]:
E   IndexError: list index out of range
```

Could be again another out of date package. I'll try installing in a venv and re-running.

thesamovar:
But that will have to wait to tomorrow.

thesamovar:
Rest of the tests seem to be running OK.

thesamovar:
Is there a neat conda/pip trick for creating and installing all of the required packages for a branch of a repo into a new virtual environment? I guess that's pretty much what we do in the CI servers?

mstimberg:
> OK I don't think I'm going to go through this line by line, but I'm in favour of the principle of it. I'm going to try to run it on my Windows machine now. Is there any particular bit of the code you would like me to look at?

I don't think there's anything really interesting in the code as such, the most important is that it actually runs... Maybe have a look at the `developer/testing` documentation?

> OK first issue is that it is showing plots while I run the test suite. Could be that I'm on an old version of matplotlib, is that likely to be the issue?

This is odd, we are not plotting anything as part of the tests. How exactly did you run the test suite?

> Lots of these errors too:
> [...]

This looks weird, too. This is an admittedly somewhat hacky part of the test run, but if I understand the error message correctly it means that it is trying to collect tests from a file without file extension...?

> Is there a neat conda/pip trick for creating and installing all of the required packages for a branch of a repo into a new virtual environment? I guess that's pretty much what we do in the CI servers?

We do everything manual on the CI servers, actually. But just doing `pip install .` in the source directory should install all mandatory requirements (i.e. not matplotlib, for example).

thesamovar:
OK so I created a fresh py3.7 environment, did ``pip install .`` and ``pip install pytest`` then ``python -c "import brian2; brian2.test()" and get the following errors:

```
======================================================= ERRORS ========================================================
____________________________________ ERROR collecting brian2/sphinxext/briandoc.py ____________________________________
brian2\sphinxext\briandoc.py:24: in <module>
    from docutils import statemachine
E   ModuleNotFoundError: No module named 'docutils'
___________________________________ ERROR collecting brian2/sphinxext/docscrape.py ____________________________________
brian2\sphinxext\docscrape.py:13: in <module>
    from sphinx.pycode import ModuleAnalyzer
E   ModuleNotFoundError: No module named 'sphinx'
________________________________ ERROR collecting brian2/sphinxext/docscrape_sphinx.py ________________________________
brian2\sphinxext\docscrape_sphinx.py:3: in <module>
    import sphinx
E   ModuleNotFoundError: No module named 'sphinx'
________________________________ ERROR collecting docs_sphinx/auto_examples/COBAHH.py _________________________________
docs_sphinx\auto_examples\COBAHH.py:83: in <module>
    plot(trace.t/ms, trace[1].v/mV)
E   NameError: name 'plot' is not defined
```

Seems to be collecting docstrings from the ``docs_sphinx/auto_examples`` directory?

thesamovar:
Also sphinx and docutils not installed by ``pip install .`` I guess.

It's also raising some other weird errors, like in the numpy only tests it is trying to use weave sometimes which raises an error on python 3 obviously.

thesamovar:
More calls to weave from the cython test set, and also all the gsl stuff failed. Will try to fix some of these and run again, but looks like there might be some issues.

mstimberg:
> Also sphinx and docutils not installed by `pip install .` I guess.

Yes, sorry, you can use the slightly cryptic `pip install .[test,docs]` to install sphinx and pytest, etc. as well.

> Seems to be collecting docstrings from the docs_sphinx/auto_examples directory?

What is `docs_sphinx/auto_examples`, it's not part of Brian...?

mstimberg:
Not sure why it calls `weave`, it should set the preferences (it works on my machine and on appveyor...). You do not use a Brian preference file or something like that?

The GSL is a bit tricky, it needs the GSL libraries but they are not a dependency (except for the conda package, where we can install them easily). Maybe we should skip these tests instead of failing if the GSL is not installed? But we didn't do this before either, so maybe we deal with this as part of #1129?

thesamovar:
Oh maybe the auto_examples is some old thing? Will try deleting that directory and see if that sorts it.

Yeah the weave is there because of a preference file, running again with that removed. But in any case, shouldn't come up when testing cython right?

thesamovar:
OK with auto_examples deleted, doctests now passes except for:

```
====================================================== FAILURES =======================================================
___________________________________________ [doctest] variables_indices.rst ___________________________________________
012 includes `NeuronGroup`, `Synapses`, `PoissonGroup` and everything else that has
013 state variables). The `variables` attribute can be accessed as a (read-only)
014 dictionary, mapping variable names to `Variable` objects storing the
015 information about the respective variable. However, it is not a simple
016 dictionary but an instance of the `Variables` class. Let's have a look at its
017 content for a simple example::
018
019     >>> tau = 10*ms
020     >>> G = NeuronGroup(10, 'dv/dt = -v / tau : volt')
021     >>> for name, var in sorted(G.variables.items()):
Differences (unified diff with -expected +actual):
    @@ -1,3 +1,3 @@
    -N : <Constant(dimensions=Dimension(),  dtype=int64, scalar=True, constant=True, read_only=True)>
    +N : <Constant(dimensions=Dimension(),  dtype=int32, scalar=True, constant=True, read_only=True)>
     dt : <ArrayVariable(dimensions=second,  dtype=float, scalar=True, constant=True, read_only=True)>
     i : <ArrayVariable(dimensions=Dimension(),  dtype=int32, scalar=False, constant=True, read_only=True)>
```

thesamovar:
OK to be honest though, these now all look like minor things. Still, it is a shame it's now so hard for me to run the test suite on my own machine.

mstimberg:
> Yeah the weave is there because of a preference file, running again with that removed. But in any case, shouldn't come up when testing cython right?

It is a bit surprising, because we are setting the preference to use whatever codegen target we are testing, and this should happen after taking into account a preference file. Maybe we should specifically deal with this situation in the test suite.

The `int32` vs. `int64` issue in the doctest is surprising, there is this issue that the `long` dataype is `int32` on Windows but `int64` on Linux which would explain this test failure, but then why does it work correctly on appveyor :confused: ?

mstimberg:
> It is a bit surprising, because we are setting the preference to use whatever codegen target we are testing, and this should happen after taking into account a preference file. Maybe we should specifically deal with this situation in the test suite.

Actually there was a genuine bug related to the preference setting. It should now work even if you had a preference file (if you do not explicitly set `reset_preferences` to `False`, tests always start from the default preferences)

mstimberg:
Ok, I'm merging this now. There are certainly things to improve, but I'd say it is more or less working as well as the previous solution based on nose at least.

Issue #111, opened 2013-09-10
Abstract code functions


mstimberg:
The current code has some support for defining functions in abstract code, see `brian2.parsing.functions`. Defining functions in abstract code can be useful for making code more readable and has the advantage over standard functions that it automatically works for all target languages.

There's not much programming necessary to make this work, it's mostly about deciding the user interface and syntax.

PR #1133, opened 2019-11-14
[MRG] Support `expm1`, `log1p`, `exprel`

### LINKED ISSUE #1117: Add more functions (e.g. expm1)
We might want to add more predefined functions, in particularly those that are part of the C++ standard (and therefore need basically no work to add). I'm mostly thinking about `expm1` (which is `exp(x) -1`, but more accurate for x close to 0), and potentially `log1p` (`log(1 + x)`) for completeness. Then, we should probably also add something like `exprel` (`expm1(x) / x`, but returning 1 for x=0), which is easily implemented if we have `expm1` (see e.g. [scipy's implementation in Cython](https://github.com/scipy/scipy/blob/0a5bb3ba367d34a93bbd6b6afe886fb842911a3a/scipy/special/_exprel.pxd#L7-L13)).

One thing I am wondering though is: in rate equations, what you often use is:
`x/(exp(x/y) - 1)`. Now, with `exprel`, you can express this in a safe way by writing `y/exprel(x/y)`, but I am not sure this is obvious to most users. Many NEURON models instead define a `vtrap` function, i.e. `vtrap(x, y) = x/(exp(x/y) - 1)`, but this name seems very NEURON-specific to me. Maybe there is some other standard name for this, though?@thesamovar, @romainbrette any thoughts?

Just to show that having such a function could have practical relevance: For a typical rate equation from the HH model like this
```
beta_m = 0.28*(mV**-1)*(V+23*mV)/
        (exp((V+23*mV)/(5*mV))-1)/ms : Hz
```
we will have the obvious problem that for V=-23mV, this will give NaN – but of course with floating point values getting *exactly* to -23mV is rather unlikely. However. if we are very close to this number, we already get inaccurate results due to catastrophic cancellation. This plot shows the values from the above equation in blue and the corrected solution with `exprel` in orange:

accuracy

Of course you need to be really close (the axis shows the difference to -23mV in **nano**volt) and the absolute differences are rather small, but still...

mstimberg:
This fixes #1117 by introducing the functions `expm1`, `log1p`, and `exprel`

As discussed in #1117, the `exprel` function can be useful for the rate equations in HH equations, and I used it in our examples wherever it applied – maybe as a first step this is enough to gently nudge users towards using it?

While having something like `vtrap` would make the equations yet a bit more readable, I think in the long run implementing functions (or maybe call them "macros") in the abstract code would be the best solution. We discussed this in #111 many years ago, but maybe we should think about it again. I think my ideal solution would be to make it possible to replace something like this:
```Python
rate_eqs = '''
alpha_m = 0.32*(mV**-1)*4*mV/exprel((13*mV-v+VT)/(4*mV))/ms : Hz
beta_m = 0.28*(mV**-1)*5*mV/exprel((v-VT-40*mV)/(5*mV))/ms : Hz
alpha_h = 0.128*exp((17*mV-v+VT)/(18*mV))/ms : Hz
beta_h = 4./(1+exp((40*mV-v+VT)/(5*mV)))/ms : Hz
alpha_n = 0.032*(mV**-1)*5*mV/exprel((15*mV-v+VT)/(5*mV))/ms : Hz
beta_n = .5*exp((10*mV-v+VT)/(40*mV))/ms : Hz
'''
```

by something like this (did not verify whether this is actually correct, but you get the idea):
```Python
rate_eqs = '''
alpha_m = 0.32*(mV**-1)*exp_linear(-v+VT, 13*mV, 4*mV)/ms : Hz
beta_m = 0.28*(mV**-1)*exp_linear(v-VT, -40*mV, 5*mV)/ms : Hz
alpha_h = 0.128*exp((17*mV-v+VT)/(18*mV))/ms : Hz
beta_h = 4./(1+exp((40*mV-v+VT)/(5*mV)))/ms : Hz
alpha_n = 0.032*(mV**-1)*exp_linear(-v+VT, 15*mV, 5*mV)/ms : Hz
beta_n = .5*exp((10*mV-v+VT)/(40*mV))/ms : Hz
exp_linear(x, midpoint, scale) = scale*(x - midpoint) / exprel((x - midpoint) / scale)
'''
```

But well, that's independent of this PR, so let's discuss this somewhere else, e.g. in #111.


mstimberg:
> Maybe it would be nice to have a simple example showing the difference in accuracy of using these functions?

Agreed, I'll try to come up with something.

mstimberg:
Superseded by #1141 .

PR #1135, opened 2019-11-20
[MRG] Improve Visual Studio support on Windows

mstimberg:
While debugging an error that only occurred under Windows (more on this later) I noticed that our technique for finding `vcvarsall.bat`(via `msvc9compiler.find_vcvarsall`)  does not work for Visual Studio 2019. There is a new function in setuptools that is meant to be used for more recent Visual Studio versions (from 2015 on, I think). This function works a bit differently, it does not return the `vcvarsall.bat` filename but instead directly a dictionary with all environment variables. With this PR, we support this way of configuring the compiler environment as well. The code is currently a bit long and messy, but we could potentially simplify it when we drop Python 2.7 support and no longer need to support Visual Studio 2008. 

There are a few other minor standalone-related changes:
* We no longer delete the standalone directory for failed tests (to make debugging easier)
* I fixed wrong parameters to switch off code optimization on msvc (not sure this was always wrong, maybe I introduced this problem during the test refactoring)

All this work is basically a side project of an endless debugging session in my Windows VM, where I tried to figure out why [a test](https://ci.appveyor.com/project/brianteam/brian2/builds/28973532/job/qb1bw798ogdsceae) crashed with an "Access violation reading violation" with Visual Studio 2015 and single precision floats... I did not manage to get to the bottom of it – I could reproduce it in the VM, but it disappeared when I added print statements in certain places, and when I switched off optimization to get debugging information :roll_eyes: I suspected it had something to do with the `__restrict` definition (that's why I switched them off for arrays of length 1 – since such arrays are equivalent to a single primitive pointer, it is a bit unclear whether you are allowed to annotate them), but I did not manage to fix it. The error now no longer occurrs on the test server because this PR switches off the optimization, so I'll give up for now. If you happen to be motivated to investigate this further, the failing test was `test_linked_variables_scalar` in `test_neurongroup` (with the standalone device and Visual Studio 2015, and `default_float_dtype=float32`)...

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -8,6 +8,13 @@ environment:
     CMD_IN_ENV: "cmd /E:ON /V:ON /C .\\dev\\continuous-integration\\appveyor\\run_with_env.cmd"
 
   matrix:
+    - PYTHON: "C:\\Miniconda37-x64"
+      PYTHON_VERSION: "3.7"
+      PYTHON_ARCH: "64"
+      platform: x64
+      STANDALONE: "TRUE"
+      FLOAT_DTYPE_32: "TRUE"
+
     - PYTHON: "C:\\Miniconda37"
       PYTHON_VERSION: "3.7"
       PYTHON_ARCH: "32"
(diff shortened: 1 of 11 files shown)


mstimberg:
@thesamovar, sorry for nagging, but do you think you could have a look at this PR rather soonish? Together with #1133 this is the last thing I'd like to merge before making a release.

thesamovar:
It's on the list! (Unfortunately quite a few other urgent things on the list right now, but I will get to it I promise!)

mstimberg:
> OK, tests now passing for me and code looks good (although I have to admit I don't 100% follow what is going on).

It'll be much simpler when we stop supporting Python 2 and Visual Studio <= 2015 :smile:

PR #1138, opened 2019-12-03
[MRG] Unsigned data type index

### LINKED ISSUE #1136: Segmentation fault when monitoring more than 2**31 neuron states
The `StateMonitor` example from the brian2 [user’s guide](https://brian2.readthedocs.io/en/stable/user/recording.html) segfaults

```
N=215
G = NeuronGroup(N, '''dv/dt = (1-v)/(10*ms) : 1
                      v_th : 1''',
                threshold='v > v_th',
                # randomly change the threshold after a spike:
                reset='''v=0
                         v_th = clip(v_th + rand()*0.2 - 0.1, 0.1, 0.9)''')
G.v_th = 0.5
M = StateMonitor(G, 'v', record=True)
run(1000*second)
```
when run for 1000 seconds and with `N=215`. `N=214` finishes without problem and we have `M.v.shape == (214, 10000000)`, which is just below the maximum value of an int32, 2,147,483,647.

With `prefs.codegen.target = 'numpy'` it runs without any problem.

mstimberg:
This PR replaces `int` by `size_t` in several places where a variable is used for indexing. This should fix  #1136, i.e. avoid crashes for indexes bigger than 2^31. Note that the main focus here is on getting rid of the 32bit limitation for variables in monitors, where this issue can appear comparatively easily. It does *not* lift the limitation of a maximum of 2^31 neurons or synapses, for example. These would be more difficult to handle, since we do not want in general to use 64bit integers for neuron indices, for example (this would significantly increase the memory needed to store synapses).

Note that in C++ standalone mode, loops that we parallelize with OpenMP cannot use `size_t`, since OpenMP requires the use of signed integer types (at least on Windows, but potentially also on Linux with older OpenMP versions).

@Debilski Can you confirm that this branch (`index_data_type`) fixes the issue that you were seeing in #1136?

### DIFF
--- a/brian2/codegen/generators/cython_generator.py
+++ b/brian2/codegen/generators/cython_generator.py
@@ -336,7 +336,7 @@ class CythonCodeGenerator(CodeGenerator):
                                 "cdef {cpp_dtype} * {array_name} = <{cpp_dtype} *> _buf_{array_name}.data"]
 
                 if not var.scalar:
-                    newlines += ["cdef int _num{array_name} = len(_namespace['{array_name}'])"]
+                    newlines += ["cdef size_t _num{array_name} = len(_namespace['{array_name}'])"]
 
                 if var.scalar and var.constant:
                     newlines += ['cdef {cpp_dtype} {varname} = _namespace["{varname}"]']
@@ -348,7 +348,7 @@ class CythonCodeGenerator(CodeGenerator):
                                        numpy_dtype=get_numpy_dtype(var.dtype),
                                        pointer_name=pointer_name,
                                        array_name=array_name,
-                                       varname=varname,
+                                       varname=varname
                                        )
                     load_namespace.append(line)
                 handled_pointers.add(pointer_name)
@@ -516,6 +516,7 @@ ctypedef fused _to_sign:
     char
     short
     int
+    long
     float
     double
 
@@ -531,6 +532,7 @@ ctypedef fused _to_clip:
     char
     short
     int
+    long
     float
     double
 
@@ -546,7 +548,7 @@ DEFAULT_FUNCTIONS['clip'].implementations.add_implementation(CythonCodeGenerator
                                                              name='_clip')
 
 timestep_code = '''
-cdef int _timestep(double t, double dt):
+cdef int64_t _timestep(double t, double dt):
     return <int64_t>((t + 1e-3*dt)/dt)
 '''
 DEFAULT_FUNCTIONS['timestep'].implementations.add_implementation(CythonCodeGenerator,
--- a/brian2/codegen/runtime/weave_rt/templates/synapses_create_array.cpp
+++ b/brian2/codegen/runtime/weave_rt/templates/synapses_create_array.cpp
@@ -19,13 +19,13 @@ _arg_tuple_new[0] = _new_num_synapses;
 int *_synaptic_pre_data = (int*)(((PyArrayObject*)(PyObject*){{_dynamic__synaptic_pre}}.attr("data"))->data);
 int *_synaptic_post_data = (int*)(((PyArrayObject*)(PyObject*){{_dynamic__synaptic_post}}.attr("data"))->data);
 
-for (int _idx=0; _idx<_numsources; _idx++) {
+for (size_t _idx=0; _idx<_numsources; _idx++) {
     {# After this code has been executed, the arrays _real_sources and
        _real_variables contain the final indices. Having any code here it all is
        only necessary for supporting subgroups #}
     {{vector_code|autoindent}}
-    _synaptic_pre_data[_idx + _old_num_synapses] = _real_sources;
-    _synaptic_post_data[_idx + _old_num_synapses] = _real_targets;
+    _synaptic_pre_data[_idx + {{N}}] = _real_sources;
+    _synaptic_post_data[_idx + {{N}}] = _real_targets;
 }
 
 // now we need to resize all registered variables and set the total number of
@@ -33,6 +33,6 @@ for (int _idx=0; _idx<_numsources; _idx++) {
 _owner.mcall("_resize", _arg_tuple_new);
 
 // And update N_incoming, N_outgoing and synapse_number
-//_arg_tuple[0] = _old_num_synapses;
+//_arg_tuple[0] = {{N}};
 _owner.mcall("_update_synapse_numbers", _arg_tuple_old);
 {% endblock %}
--- a/brian2/codegen/runtime/weave_rt/templates/synapses_create_generator.cpp
+++ b/brian2/codegen/runtime/weave_rt/templates/synapses_create_generator.cpp
@@ -13,18 +13,18 @@
     int *const _prebuf = new int[_buffer_size];
     int *const _postbuf = new int[_buffer_size];
     int _curbuf = 0;
-    const int _N_pre = {{constant_or_scalar('N_pre', variables['N_pre'])}};
-    const int _N_post = {{constant_or_scalar('N_post', variables['N_post'])}};
+    const size_t _N_pre = {{constant_or_scalar('N_pre', variables['N_pre'])}};
+    const size_t _N_post = {{constant_or_scalar('N_post', variables['N_post'])}};
     int _raw_pre_idx, _raw_post_idx;
     const int oldsize = {{_dynamic__synaptic_pre}}.size();
 
     // scalar code
-    const int _vectorisation_idx = 1;
+    const size_t _vectorisation_idx = 1;
     {{scalar_code['setup_iterator']|autoindent}}
     {{scalar_code['create_j']|autoindent}}
     {{scalar_code['create_cond']|autoindent}}
     {{scalar_code['update_post']|autoindent}}
-    for(int _i=0; _i<_N_pre; _i++)
+    for(size_t _i=0; _i<_N_pre; _i++)
     {
         bool __cond, _cond;
         _raw_pre_idx = _i + _source_offset;
@@ -66,7 +66,7 @@
             {% endif %}
         }
         {% if iterator_func=='range' %}
-        for(int {{iteration_variable}}=_uiter_low; {{iteration_variable}}<_uiter_high; {{iteration_variable}}+=_uiter_step)
+        for(long {{iteration_variable}}=_uiter_low; {{iteration_variable}}<_uiter_high; {{iteration_variable}}+=_uiter_step)
         {
         {% elif iterator_func=='sample' %}
         if(_uiter_p==0) continue;
@@ -77,19 +77,19 @@
         else
             _log1p = 1.0; // will be ignored
         const double _pconst = 1.0/_log1p;
-        for(int {{iteration_variable}}=_uiter_low; {{iteration_variable}}<_uiter_high; {{iteration_variable}}++)
+        for(long {{iteration_variable}}=_uiter_low; {{iteration_variable}}<_uiter_high; {{iteration_variable}}++)
         {
             if(_jump_algo) {
                 const double _r = _rand(_vectorisation_idx);
                 if(_r==0.0) break;
-                const int _jump = floor(log(_r)*_pconst)*_uiter_step;
+                const size_t _jump = floor(log(_r)*_pconst)*_uiter_step;
                 {{iteration_variable}} += _jump;
                 if({{iteration_variable}}>=_uiter_high) continue;
             } else {
                 if(_rand(_vectorisation_idx)>=_uiter_p) continue;
             }
         {% endif %}
-            long __j, _j, _pre_idx, __pre_idx;
+            size_t __j, _j, _pre_idx, __pre_idx;
             {
                 {{vector_code['create_j']|autoindent}}
                 __j = _j; // pick up the locally scoped _j and store in __j
@@ -136,7 +136,7 @@
             {% endif %}
             {{vector_code['update_post']|autoindent}}
 
-            for (int _repetition=0; _repetition<_n; _repetition++) {
+            for (size_t _repetition=0; _repetition<_n; _repetition++) {
                 _prebuf[_curbuf] = (int)_pre_idx;
                 _postbuf[_curbuf] = (int)_post_idx;
                 _curbuf++;
--- a/brian2/devices/cpp_standalone/templates/synapses.cpp
+++ b/brian2/devices/cpp_standalone/templates/synapses.cpp
@@ -10,42 +10,42 @@
 
 {% block maincode %}
 
-	// This is only needed for the _debugmsg function below	
-	{# USES_VARIABLES { _synaptic_pre } #}	
+    // This is only needed for the _debugmsg function below
+    {# USES_VARIABLES { _synaptic_pre } #}
 
-	// scalar code
-	const int _vectorisation_idx = -1;
-	{{scalar_code|autoindent}}
+    // scalar code
+    const size_t _vectorisation_idx = -1;
+    {{scalar_code|autoindent}}
 
-	{{ openmp_pragma('parallel') }}
-	{
-	std::vector<int> *_spiking_synapses = {{pathway.name}}.peek();
-	const unsigned int _num_spiking_synapses = _spiking_synapses->size();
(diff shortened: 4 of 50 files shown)


Debilski:
Thanks. I’ll have a look tomorrow.

Debilski:
My minimal test is working fine now. Thanks.

brian2/codegen/generators/cython_generator.py
```diff
@@ -546,7 +548,7 @@ def determine_keywords(self):
                                                              name='_clip')
 
 timestep_code = '''
-cdef int _timestep(double t, double dt):
+cdef int64_t _timestep(double t, double dt):
```
thesamovar: Any reason we use ``int64_t`` in some places and ``size_t`` in others? Actually it's just here - because ``_timestep`` should always have been 64-bit I guess?

brian2/codegen/runtime/weave_rt/templates/synapses_create_array.cpp
```diff
@@ -19,20 +19,20 @@ _arg_tuple_new[0] = _new_num_synapses;
...
-for (int _idx=0; _idx<_numsources; _idx++) {
+for (size_t _idx=0; _idx<_numsources; _idx++) {
     {# After this code has been executed, the arrays _real_sources and
        _real_variables contain the final indices. Having any code here it all is
        only necessary for supporting subgroups #}
     {{vector_code|autoindent}}
-    _synaptic_pre_data[_idx + _old_num_synapses] = _real_sources;
-    _synaptic_post_data[_idx + _old_num_synapses] = _real_targets;
+    _synaptic_pre_data[_idx + {{N}}] = _real_sources;
+    _synaptic_post_data[_idx + {{N}}] = _real_targets;
```
thesamovar: What's the reason for this change? (Used again later.)

brian2/codegen/runtime/weave_rt/templates/synapses_create_generator.cpp
```diff
@@ -66,7 +66,7 @@
             {% endif %}
         }
         {% if iterator_func=='range' %}
-        for(int {{iteration_variable}}=_uiter_low; {{iteration_variable}}<_uiter_high; {{iteration_variable}}+=_uiter_step)
+        for(long {{iteration_variable}}=_uiter_low; {{iteration_variable}}<_uiter_high; {{iteration_variable}}+=_uiter_step)
```
thesamovar: What's the difference between long and int here? Used a few times throughout. I think long is not guaranteed to be 64 bit?

brian2/devices/cpp_standalone/templates/synapses.cpp
```diff
@@ -10,51 +10,51 @@
...
+    {{scalar_code|autoindent}}
 
-	{{ openmp_pragma('parallel') }}
-	{
-	std::vector<int> *_spiking_synapses = {{pathway.name}}.peek();
-	const unsigned int _num_spiking_synapses = _spiking_synapses->size();
+    {{ openmp_pragma('parallel') }}
+    {
+    std::vector<int> *_spiking_synapses = {{pathway.name}}.peek();
+    const int _num_spiking_synapses = _spiking_synapses->size();
```
thesamovar: Removed unsigned for consistency, right?

brian2/codegen/generators/cython_generator.py
mstimberg: Right, I fixed this along the way, this should have been `int64_t` for a while. It is somewhat unrelated to the other changes.

brian2/codegen/runtime/weave_rt/templates/synapses_create_array.cpp
mstimberg: It's just that previously we set `_old_num_synapses = {{N}}` but never change either of them, so I thought we could just use `{{N}}` directly.

brian2/codegen/runtime/weave_rt/templates/synapses_create_generator.cpp
mstimberg: Yes, I stopped short of fixing this in any fundamental way. It just seemed odd that we use `int` here when `_uiter_low` was defined as a `long` earlier.

brian2/devices/cpp_standalone/templates/synapses.cpp
mstimberg: Exactly (we cannot use it in the for loop that is parallelized with OpenMP)

thesamovar:
Actually, will just check this on a Windows machine. I see that appveyor is failing?

thesamovar:
Am seeing this error a lot:

```
..\..\brian2\tests\__init__.py:31: in collect
    while full_name[-1] != os.path.splitext(self.name)[0]:
E   IndexError: list index out of range
```

mstimberg:
> Actually, will just check this on a Windows machine. I see that appveyor is failing?

Please do. The appveyor failures are annoying but they are either timeouts, or an issue that has been fixed in master already (which is not merged into this branch).

mstimberg:
About the `IndexError`: how are you running the tests? Are you using the `run_tests.py` script? The errors could happen if it tried to collect tests from files without extension, but pytest shouldn't collect such files in the first place, I think? You had the same problem at some point in #1127, I don't quite remember what made them go away though....

thesamovar:
That was something else I think, but might be an old pytest version? 3.5.1.

mstimberg:
> That was something else I think, but might be an old pytest version? 3.5.1.

Indeed, I can reproduce the problem with this old version. The fix seems straightforward, I'll go ahead – we could also require a more recent pytest version, but we shouldn't put too many obstacles into the way either.

thesamovar:
OK I'll wait for the fix then.

thesamovar:
OK seems to work now. One test failure:

```
================================== FAILURES ===================================
_______________________ [doctest] variables_indices.rst _______________________
012 includes `NeuronGroup`, `Synapses`, `PoissonGroup` and everything else that has
013 state variables). The `variables` attribute can be accessed as a (read-only)
014 dictionary, mapping variable names to `Variable` objects storing the
015 information about the respective variable. However, it is not a simple
016 dictionary but an instance of the `Variables` class. Let's have a look at its
017 content for a simple example::
018 
019     >>> tau = 10*ms
020     >>> G = NeuronGroup(10, 'dv/dt = -v / tau : volt')
021     >>> for name, var in sorted(G.variables.items()):
Differences (unified diff with -expected +actual):
    @@ -1,3 +1,3 @@
    -N : <Constant(dimensions=Dimension(),  dtype=int64, scalar=True, constant=True, read_only=True)>
    +N : <Constant(dimensions=Dimension(),  dtype=int32, scalar=True, constant=True, read_only=True)>
     dt : <ArrayVariable(dimensions=second,  dtype=float, scalar=True, constant=True, read_only=True)>
     i : <ArrayVariable(dimensions=Dimension(),  dtype=int32, scalar=False, constant=True, read_only=True)>

E:\Dan\programming\brian2\docs_sphinx\developer\variables_indices.rst:21: DocTestFailure
1 failed, 49 passed in 4.24 seconds
```

mstimberg:
I pushed a fix that skips this doctest (which is due to int=32bit on Windows) – the print output is quite fragile anyway.

thesamovar:
That works. I was just seeing some Cython fails, but might be from old cruft. Deleted my .cython directory and re-running now.

mstimberg:
Oh, and I think the reason why we are not seeing this on appveyor is that we are probably running the tests on the package (like calling `brian2.test()`) – this has access to the doctests in the module itself, but not to the doctests in the `docs_sphinx` rst files. On the Linux test server we are doing this differently.

thesamovar:
OK all tests pass except Cython which has some GSL errors. Is there any reason these aren't being skipped with Cython but are for weave?

mstimberg:
> OK all tests pass except Cython which has some GSL errors. Is there any reason these aren't being skipped with Cython but are for weave?

They shouldn't be skipped for either, actually. I think it skips them if there is a `NotImplementedError` which is the case in `numpy` (where we don't support GSL at all). Maybe `weave` raises this error "by accident"?

thesamovar:
Ah no, it's that the weave tests didn't run because I didn't install scipy in that environment!

Can you install GSL on Windows? If so, how do you do it? If not, perhaps it should skip those tests on Windows?

mstimberg:
You can, we are testing it on appveyor. If you are already using conda, then it should be just a `conda install gsl`.

thesamovar:
Just in the middle of trying that but my conda has gone berserk... will get back to you shortly.

thesamovar:
My conda seems to have been utterly destroyed by something, no idea what.

thesamovar:
Oooh I worked it out - brian-team channel was still in there messing things up.

thesamovar:
Oh, even weirder it's a bug in venv support in pycharm that appears to be specific to Python 3.7...

thesamovar:
Guess that's what you get for using a 2016 version of pycharm.

thesamovar:
OK all tests pass now, this can merge.

thesamovar:
Actually, passes for py3, but for py2 still get a few of these:

```
================================== FAILURES ===================================
_______________________________ test_refractory _______________________________

    def test_refractory():
        # Compare integration with and without the addition of refractoriness --
        # note that the cell here is not spiking, so it should never be in the
        # refractory period and therefore the results should be exactly identical
        # with and without (unless refractory)
        eqs_base = 'dv/dt = -v/(10*ms) : 1'
        for method in ['linear', 'exact', 'independent', 'euler', 'exponential_euler', 'rk2', 'rk4']:
            G_no_ref = NeuronGroup(10, eqs_base, method=method)
            G_no_ref.v = '(i+1)/11.'
            G_ref = NeuronGroup(10, eqs_base + '(unless refractory)',
                                refractory=1*ms, method=method)
            G_ref.v = '(i+1)/11.'
            net = Network(G_ref, G_no_ref)
>           net.run(10*ms)

..\..\brian2\tests\test_stateupdaters.py:709: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _
..\..\brian2\core\base.py:280: in device_override_decorated_function
    return func(*args, **kwds)
..\..\brian2\units\fundamentalunits.py:2392: in new_f
    result = f(*args, **kwds)
..\..\brian2\core\network.py:1012: in run
    self.before_run(namespace)
..\..\brian2\core\base.py:280: in device_override_decorated_function
    return func(*args, **kwds)
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _

self = <Network at time t=0. s, containing objects: NeuronGroup(clock=Clock(dt=100. *...name='defaultclock'), when=groups, order=0, name='neurongroup_1_stateupdater')>
run_namespace = {'@py_builtins': <module '__builtin__' (built-in)>, '@pytest_ar': <module '_pytest.assertion.rewrite' from 'C:\Anacond...-packages\_pytest\assertion\rewrite.pyc'>, 'ALLOW_THREADS': 1, 'Annotation': <class 'matplotlib.text.Annotation'>, ...}

    @device_override('network_before_run')
    def before_run(self, run_namespace):
        '''
        before_run(namespace)
    
        Prepares the `Network` for a run.
    
        Objects in the `Network` are sorted into the correct running order, and
        their `BrianObject.before_run` methods are called.
    
        Parameters
        ----------
        run_namespace : dict-like, optional
            A namespace in which objects which do not define their own
            namespace will be run.
        '''
        all_objects = self.sorted_objects
        prefs.check_all_validated()
    
        # Check names in the network for uniqueness
        names = [obj.name for obj in all_objects]
        non_unique_names = [name for name, count in Counter(names).items()
                            if count > 1]
        if len(non_unique_names):
            formatted_names = ', '.join("'%s'" % name
                                        for name in non_unique_names)
            raise ValueError('All objects in a network need to have unique '
                             'names, the following name(s) were used more than '
                             'once: %s' % formatted_names)
    
        # Check that there are no SummedVariableUpdaters targeting the same
        # target variable
        _check_multiple_summed_updaters(all_objects)
    
        self._stopped = False
        Network._globally_stopped = False
    
        device = get_device()
        if device.network_schedule is not None:
            # The device defines a fixed network schedule
            if device.network_schedule != self.schedule:
                # TODO: The human-readable name of a device should be easier to get
                device_name = list(all_devices.keys())[list(all_devices.values()).index(device)]
                logger.warn(("The selected device '{device_name}' only "
                             "supports a fixed schedule, but this schedule is "
                             "not consistent with the network's schedule. The "
                             "simulation will use the device's schedule.\n"
                             "Device schedule: {device.network_schedule}\n"
                             "Network schedule: {net.schedule}\n"
                             "Set the network schedule explicitly or set the "
                             "core.network.default_schedule preference to "
                             "avoid this warning.").format(device_name=device_name,
                                                           device=device,
                                                           net=self),
                            name_suffix='schedule_conflict', once=True)
    
        logger.debug("Preparing network {self.name} with {numobj} "
                     "objects: {objnames}".format(self=self,
                        numobj=len(all_objects),
                        objnames=', '.join(obj.name for obj in all_objects)),
                     "before_run")
    
        self.check_dependencies()
    
        for obj in all_objects:
            if obj.active:
                try:
                    obj.before_run(run_namespace)
                except Exception as ex:
>                   raise brian_object_exception("An error occurred when preparing an object.", obj, ex)
E                   BrianObjectException: Original error and traceback:
E                   Traceback (most recent call last):
E                     File "E:\Dan\programming\brian2\brian2\core\network.py", line 901, in before_run
E                       obj.before_run(run_namespace)
E                     File "E:\Dan\programming\brian2\brian2\groups\group.py", line 1133, in before_run
E                       codeobj_class=self.codeobj_class
E                     File "E:\Dan\programming\brian2\brian2\codegen\codeobject.py", line 309, in create_runner_codeobj
E                       for v, u_v in zip(code.values(), user_code.values()):
E                   AttributeError: 'unicode' object has no attribute 'values'
E                   
E                   Error encountered with object named "neurongroup_1_stateupdater".
E                   Object was created here (most recent call only, full details in debug log):
E                     File "C:\Anaconda2\envs\brian2py2\lib\site-packages\_pytest\python.py", line 174, in pytest_pyfunc_call
E                       testfunction(**testargs)
E                   
E                   An error occurred when preparing an object. AttributeError: 'unicode' object has no attribute 'values'
E                   (See above for original error message and traceback.)
```

Possibly for the other PR too. Any idea?

thesamovar:
OK the weird thing about this is that it's consistently the same test (``test_refractory``) for numpy/weave/cython on a fresh Python 2 environment with all packages up to date, but it doesn't happen when you run just that test directly. Any ideas?

mstimberg:
This is quite confusing, I cannot reproduce it on either Linux nor Windows, and we don't see this error on the test server either.  The error message implies that either `code` or `user_code` is a string instead of a dictionary, but that would seem like a quite fundamental error that should occur everywhere. That you only see this as part of the test suite indicates that it should have something to do with the test setup or for example the preference settings, but I am having a hard time seeing this leading to the error you are seeing. Do you still get the error if you run only this test via `pytest`?
```
pytest brian2/tests/test_stateupdaters.py::test_refractory -c brian2/tests/pytest.ini
```
(you might need to adapt the paths, here this is from brian2's main directory)

For a more useful error report by pytest it might also be good to remove the `try`/`catch` here and just let it raise the error normally:
https://github.com/brian-team/brian2/blob/ecb3df376f2a24ae70437d71f4b20133595b95af/brian2/core/network.py#L900-L903

mstimberg:
Oh, also: you are only seeing this error in the `index_data_type` branch, right?

thesamovar:
Also seeing it in the other branch you already merged, but that's OK, it seems to be just some sort of weird testing bug. Would be nice if it didn't happen though. Will try as soon as I can, which might not be until Wednesday afternoon.

mstimberg:
Ok in that case I am going to merge this branch now. It would still be nice to get to the bottom of this, I'd also much prefer if we didn't have these odd bugs. Luckily most of the recent problems were on Python 2 :)

PR #1141, opened 2019-12-09
[MRG] Add support for new functions, and check C99 support

mstimberg:
This PR includes the same main change as #1133, i.e. support for `exprel`, `expm1`, and `log1p`. I later realized that with Python 2 on Windows we need to support Visual Studio 2008 and this does not yet have C99 support, so just adding these functions would lead to nasty error messages on Python 2 for Windows users.

Therefore this PR not only adds support for the new functions, but also adds a new mechanism to the `FunctionImplementation` mechanism, allowing it to check at runtime whether a feature is actually supported. This new mechanism is now only used to do a check for C99 support, but could be useful in the future for more things (say, enabling support for a function based on the CUDA version). The C99 check uses the preprocessor of `gcc` or `msvc` to check the compiler version which should be rather robust (and quick) compared to something like parsing the output of `gcc --version`. Note that for `msvc`, we accept Visual Studio versions starting from 2013, even though that version did not strictly speaking fully support C99. It does support the new header files/functions that we are interested in (https://devblogs.microsoft.com/cppblog/c99-library-support-in-visual-studio-2013/).


brian2/codegen/cpp_prefs.py
```diff
@@ -236,3 +239,87 @@ def update_for_cross_compilation(library_dirs, extra_compile_args,
...
+                cmd = 'cl /E mytest.cpp > NUL 2>&1'
+            return_value = os.system(cmd)
+            _compiler_supports_c99 = return_value == 0
+            os.remove(tmp_file)
+        else:
+            cmd = ('echo "#if (__STDC_VERSION__ < 199901L)\n#error\n#endif" | '
+                  'cc -xc -c - > /dev/null 2>&1')
+            return_value = os.system(cmd)
+            _compiler_supports_c99 = return_value == 0
+    return _compiler_supports_c99
```
thesamovar: I guess this check doesn't take that long, but it always makes me a little uncomfortable having these sorts of things running every time.

mstimberg: It is cached in the global `_compiler_supports_c99` variable.

thesamovar: But not between runs?

mstimberg: Ah, no, indeed. But then I don't know how to cache it between invocations of Python without making things complicated to detect changes in the compiler setup. The part that potentially takes some non-negligible time is finding/executing `vcvarsall.bat`, but we'd have to do this every time to notice a compiler update I guess?

mstimberg:
> Code looks fine, only thing is the comment about efficiency, and wondering if it is a good idea that some Brian code won't run on particular versions of Windows.

Yes, I agree that this is not great, but I think it would be even worse if we had different numerical results on different  versions, even with a warning. Given that we can probably drop support for old Visual Studio versions (which aren't compatible with Python 3) in a few weeks from now, I don't think it is worth doing much about it. If that weren't the case, I'd probably say that the best solution would be to have a potentially slower but equally accurate fallback function implementation that we write ourselves.

brian2/codegen/cpp_prefs.py
thesamovar: Executing ``vcvarsall.bat`` is probably slow, but maybe we can just use its exact path as a key? If that path hasn't changed you don't need to do the rest to check the C99 compatibility?

thesamovar:
OK fair enough. Probably same comment applies to the cacheing idea actually. Good to merge (although I'd like to test on Windows again first actually).

brian2/codegen/cpp_prefs.py
mstimberg: I think finding `vcvarsall.bat` in the first place is slow... Also, the "modern version" of doing it (which is needed for the newer Visual Studio versions) doesn't actually return the path for the file but a dictionary with all the environment variables.

thesamovar:
OK the same int32/int64 doctest thing comes up here again, but I guess that'll go once we merge.

thesamovar:
I also get one Cython error in ``test_external_function_cython`` that looks genuine.

```
________________________ test_external_function_cython ________________________

self = <distutils._msvccompiler.MSVCCompiler object at 0x0000023FF53F5198>
target_desc = 'shared_object'
objects = ['C:\\Users\\Dan\\.cython\\brian_extensions\\Users\\Dan\\.cython\\brian_extensions\\func_def_cython.obj']
output_filename = 'C:\\Users\\Dan\\.cython\\brian_extensions\\func_def_cython.cp37-win_amd64.pyd'
output_dir = 'C:\\Users\\Dan\\.cython\\brian_extensions', libraries = []
library_dirs = ['C:\\Anaconda2\\envs\\brian2pytest\\libs', 'C:\\Anaconda2\\envs\\brian2pytest\\PCbuild\\amd64', 'C:\\Program Files (x...64', 'C:\\Program Files (x86)\\Microsoft Visual Studio\\2017\\BuildTools\\VC\\Tools\\MSVC\\14.14.26428\\lib\\x64', ...]
runtime_library_dirs = [], export_symbols = ['PyInit_func_def_cython']
debug = None, extra_preargs = None, extra_postargs = []
build_temp = 'C:\\Users\\Dan\\.cython\\brian_extensions\\Users\\Dan\\.cython\\brian_extensions'
target_lang = 'c'

    def link(self,
             target_desc,
             objects,
             output_filename,
             output_dir=None,
             libraries=None,
             library_dirs=None,
             runtime_library_dirs=None,
             export_symbols=None,
             debug=0,
             extra_preargs=None,
             extra_postargs=None,
             build_temp=None,
             target_lang=None):
    
        if not self.initialized:
            self.initialize()
        objects, output_dir = self._fix_object_args(objects, output_dir)
        fixed_args = self._fix_lib_args(libraries, library_dirs,
                                        runtime_library_dirs)
        libraries, library_dirs, runtime_library_dirs = fixed_args
    
        if runtime_library_dirs:
            self.warn("I don't know what to do with 'runtime_library_dirs': "
                       + str(runtime_library_dirs))
    
        lib_opts = gen_lib_options(self,
                                   library_dirs, runtime_library_dirs,
                                   libraries)
        if output_dir is not None:
            output_filename = os.path.join(output_dir, output_filename)
    
        if self._need_link(objects, output_filename):
            ldflags = self._ldflags[target_desc, debug]
    
            export_opts = ["/EXPORT:" + sym for sym in (export_symbols or [])]
    
            ld_args = (ldflags + lib_opts + export_opts +
                       objects + ['/OUT:' + output_filename])
    
            # The MSVC linker generates .lib and .exp files, which cannot be
            # suppressed by any linker switches. The .lib files may even be
            # needed! Make sure they are generated in the temporary build
            # directory. Since they have different names for debug and release
            # builds, they can go into the same directory.
            build_temp = os.path.dirname(objects[0])
            if export_symbols is not None:
                (dll_name, dll_ext) = os.path.splitext(
                    os.path.basename(output_filename))
                implib_file = os.path.join(
                    build_temp,
                    self.library_filename(dll_name))
                ld_args.append ('/IMPLIB:' + implib_file)
    
            if extra_preargs:
                ld_args[:0] = extra_preargs
            if extra_postargs:
                ld_args.extend(extra_postargs)
    
            output_dir = os.path.dirname(os.path.abspath(output_filename))
            self.mkpath(output_dir)
            try:
                log.debug('Executing "%s" %s', self.linker, ' '.join(ld_args))
>               self.spawn([self.linker] + ld_args)

C:\Anaconda2\envs\brian2pytest\lib\distutils\_msvccompiler.py:520: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _

self = <distutils._msvccompiler.MSVCCompiler object at 0x0000023FF53F5198>
cmd = ['C:\\Program Files (x86)\\Microsoft Visual Studio\\2017\\BuildTools\\VC\\Tools\\MSVC\\14.14.26428\\bin\\HostX86\\x64\\link.exe', '/nologo', '/INCREMENTAL:NO', '/LTCG', '/DLL', '/MANIFEST:EMBED,ID=2', ...]

    def spawn(self, cmd):
        old_path = os.getenv('path')
        try:
            os.environ['path'] = self._paths
>           return super().spawn(cmd)

C:\Anaconda2\envs\brian2pytest\lib\distutils\_msvccompiler.py:543: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _

self = <distutils._msvccompiler.MSVCCompiler object at 0x0000023FF53F5198>
cmd = ['C:\\Program Files (x86)\\Microsoft Visual Studio\\2017\\BuildTools\\VC\\Tools\\MSVC\\14.14.26428\\bin\\HostX86\\x64\\link.exe', '/nologo', '/INCREMENTAL:NO', '/LTCG', '/DLL', '/MANIFEST:EMBED,ID=2', ...]

    def spawn(self, cmd):
>       spawn(cmd, dry_run=self.dry_run)

C:\Anaconda2\envs\brian2pytest\lib\distutils\ccompiler.py:909: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _

cmd = ['"C:\\Program Files (x86)\\Microsoft Visual Studio\\2017\\BuildTools\\VC\\Tools\\MSVC\\14.14.26428\\bin\\HostX86\\x64\\link.exe"', '/nologo', '/INCREMENTAL:NO', '/LTCG', '/DLL', '/MANIFEST:EMBED,ID=2', ...]
search_path = 1, verbose = 0, dry_run = 0

    def spawn(cmd, search_path=1, verbose=0, dry_run=0):
        """Run another program, specified as a command list 'cmd', in a new process.
    
        'cmd' is just the argument list for the new process, ie.
        cmd[0] is the program to run and cmd[1:] are the rest of its arguments.
        There is no way to run a program with a name different from that of its
        executable.
    
        If 'search_path' is true (the default), the system's executable
        search path will be used to find the program; otherwise, cmd[0]
        must be the exact path to the executable.  If 'dry_run' is true,
        the command will not actually be run.
    
        Raise DistutilsExecError if running the program fails in any way; just
        return on success.
        """
        # cmd is documented as a list, but just in case some code passes a tuple
        # in, protect our %-formatting code against horrible death
        cmd = list(cmd)
        if os.name == 'posix':
            _spawn_posix(cmd, search_path, dry_run=dry_run)
        elif os.name == 'nt':
>           _spawn_nt(cmd, search_path, dry_run=dry_run)

C:\Anaconda2\envs\brian2pytest\lib\distutils\spawn.py:38: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _

cmd = 'C:\\Program Files (x86)\\Microsoft Visual Studio\\2017\\BuildTools\\VC\\Tools\\MSVC\\14.14.26428\\bin\\HostX86\\x64\\link.exe'
search_path = 1, verbose = 0, dry_run = 0

    def _spawn_nt(cmd, search_path=1, verbose=0, dry_run=0):
        executable = cmd[0]
        cmd = _nt_quote_args(cmd)
        if search_path:
            # either we find one or it stays the same
            executable = find_executable(executable) or executable
        log.info(' '.join([executable] + cmd[1:]))
        if not dry_run:
            # spawn for NT requires a full path to the .exe
            try:
                rc = os.spawnv(os.P_WAIT, executable, cmd)
            except OSError as exc:
                # this seems to happen when the command isn't found
                if not DEBUG:
                    cmd = executable
                raise DistutilsExecError(
                      "command %r failed: %s" % (cmd, exc.args[-1]))
            if rc != 0:
                # and this reflects the command running but failing
                if not DEBUG:
                    cmd = executable
                raise DistutilsExecError(
>                     "command %r failed with exit status %d" % (cmd, rc))
E               distutils.errors.DistutilsExecError: command 'C:\\Program Files (x86)\\Microsoft Visual Studio\\2017\\BuildTools\\VC\\Tools\\MSVC\\14.14.26428\\bin\\HostX86\\x64\\link.exe' failed with exit status 1104

C:\Anaconda2\envs\brian2pytest\lib\distutils\spawn.py:81: DistutilsExecError

    def test_external_function_cython():
        if prefs.codegen.target != 'cython':
            pytest.skip('Cython-only test')
    
        this_dir = os.path.abspath(os.path.dirname(__file__))
        @implementation('cython', 'from func_def_cython cimport foo',
                        sources=[os.path.join(this_dir, 'func_def_cython.pyx')])
        @check_units(x=volt, y=volt, result=volt)
        def foo(x, y):
            return x + y + 3*volt
    
        G = NeuronGroup(1, '''
                           func = foo(x, y) : volt
                           x : volt
                           y : volt''')
        G.x = 1*volt
        G.y = 2*volt
        mon = StateMonitor(G, 'func', record=True)
        net = Network(G, mon)
>       net.run(defaultclock.dt)

..\..\brian2\tests\test_functions.py:617: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _
..\..\brian2\core\base.py:280: in device_override_decorated_function
    return func(*args, **kwds)
..\..\brian2\units\fundamentalunits.py:2392: in new_f
    result = f(*args, **kwds)
..\..\brian2\core\network.py:1012: in run
    self.before_run(namespace)
..\..\brian2\core\base.py:280: in device_override_decorated_function
    return func(*args, **kwds)
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _

self = <Network at time t=0. s, containing objects: StateUpdater(clock=Clock(dt=100. * usecond, name='defaultclock'), when=gr...ame='defaultclock'), when=start, order=0, name='neurongroup_1'), <StateMonitor, recording ['func'] from neurongroup_1>>
run_namespace = {'@py_builtins': <module 'builtins' (built-in)>, '@pytest_ar': <module '_pytest.assertion.rewrite' from 'C:\\Anaconda2...\lib\\site-packages\\_pytest\\assertion\\rewrite.py'>, 'ALLOW_THREADS': 1, 'AxisError': <class 'numpy.AxisError'>, ...}

    @device_override('network_before_run')
    def before_run(self, run_namespace):
        '''
        before_run(namespace)
    
        Prepares the `Network` for a run.
    
        Objects in the `Network` are sorted into the correct running order, and
        their `BrianObject.before_run` methods are called.
    
        Parameters
        ----------
        run_namespace : dict-like, optional
            A namespace in which objects which do not define their own
            namespace will be run.
        '''
        all_objects = self.sorted_objects
        prefs.check_all_validated()
    
        # Check names in the network for uniqueness
        names = [obj.name for obj in all_objects]
        non_unique_names = [name for name, count in Counter(names).items()
                            if count > 1]
        if len(non_unique_names):
            formatted_names = ', '.join("'%s'" % name
                                        for name in non_unique_names)
            raise ValueError('All objects in a network need to have unique '
                             'names, the following name(s) were used more than '
                             'once: %s' % formatted_names)
    
        # Check that there are no SummedVariableUpdaters targeting the same
        # target variable
        _check_multiple_summed_updaters(all_objects)
    
        self._stopped = False
        Network._globally_stopped = False
    
        device = get_device()
        if device.network_schedule is not None:
            # The device defines a fixed network schedule
            if device.network_schedule != self.schedule:
                # TODO: The human-readable name of a device should be easier to get
                device_name = list(all_devices.keys())[list(all_devices.values()).index(device)]
                logger.warn(("The selected device '{device_name}' only "
                             "supports a fixed schedule, but this schedule is "
                             "not consistent with the network's schedule. The "
                             "simulation will use the device's schedule.\n"
                             "Device schedule: {device.network_schedule}\n"
                             "Network schedule: {net.schedule}\n"
                             "Set the network schedule explicitly or set the "
                             "core.network.default_schedule preference to "
                             "avoid this warning.").format(device_name=device_name,
                                                           device=device,
                                                           net=self),
                            name_suffix='schedule_conflict', once=True)
    
        logger.debug("Preparing network {self.name} with {numobj} "
                     "objects: {objnames}".format(self=self,
                        numobj=len(all_objects),
                        objnames=', '.join(obj.name for obj in all_objects)),
                     "before_run")
    
        self.check_dependencies()
    
        for obj in all_objects:
            if obj.active:
                try:
                    obj.before_run(run_namespace)
                except Exception as ex:
>                   raise brian_object_exception("An error occurred when preparing an object.", obj, ex)
E                   distutils.errors.LinkError: command 'C:\\Program Files (x86)\\Microsoft Visual Studio\\2017\\BuildTools\\VC\\Tools\\MSVC\\14.14.26428\\bin\\HostX86\\x64\\link.exe' failed with exit status 1104

..\..\brian2\core\network.py:903: LinkError
```

error message from console:

```
func_def_cython.c
LINK : fatal error LNK1104: cannot open file 'C:\Users\Dan\.cython\brian_extensions\func_def_cython.cp37-win_amd64.pyd'
```

But this file does exist, so I'm not sure what the problem was. Re-running just that test seems to work, so might not be a reproducible error?

thesamovar:
Re-running the suite worked fine so maybe just some odd anomaly.

Good to merge.

PR #1144, opened 2019-12-19
[MRG] Release notes 2.3

mstimberg:
These are the release notes for the release which I'd like to do tomorrow to get it out before the holidays (and the EOL for Python 2!).  I went with 2.3 as the version number, even though it could also be 2.2.3, given that most of the changes are somewhat "behind the scenes". 

The PR also fixes a number of other smallish things in the documentation and the README, but nothing to controversial I think :smile:


thesamovar:
Agreed!

PR #1149, opened 2020-01-23
[MRG] Update Binder links to point to the right place

thesamovar:
Current Binder links in sphinx docs point to the wrong place, this fixes that.

### DIFF
--- a/docs_sphinx/index.rst
+++ b/docs_sphinx/index.rst
@@ -9,7 +9,7 @@ highly flexible and easily extensible.
 
 To get an idea of what writing a simulation in Brian looks like, take a look
 at :doc:`a simple example </examples/CUBA>`, or run our
-`interactive demo <http://mybinder.org/v2/gh/brian-team/brian2-binder/master?filepath=index.ipynb>`_.
+`interactive demo <http://mybinder.org/v2/gh/brian-team/brian2-binder/master?filepath=demo.ipynb>`_.
 
 .. only:: html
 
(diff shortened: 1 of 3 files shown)


docs_sphinx/index.rst
```diff
@@ -9,7 +9,7 @@ highly flexible and easily extensible.
 
 To get an idea of what writing a simulation in Brian looks like, take a look
 at :doc:`a simple example </examples/CUBA>`, or run our
-`interactive demo <http://mybinder.org/repo/brian-team/brian2-binder/notebooks/demo.ipynb>`_.
+`interactive demo <http://mybinder.org/v2/gh/brian-team/brian2-binder/master?filepath=index.ipynb>`_.
```
mstimberg: This should be `filepath=demo.ipynb` I guess?

thesamovar: Yes! Was that the only mistake I made? :)

thesamovar:
Should be good to go now assuming docs are building correctly. Remind me, does that appear immediately or not until we do a new release? It would be good to have that fixed when we launch the new website if possible (but not worth doing a release for).

mstimberg:
I'll check whether everything builds correctly in a minute. It kind of appears immediately, but only on https://brian2.readthedocs.io/en/latest/ -- the main page https://brian2.readthedocs.io by default redirects to https://brian2.readthedocs.io/en/stable , i.e. the documentation of the last release (which makes sense because otherwise users would see documentation for unreleased features).

mstimberg:
The links in the tutorials themselves are correct now, but the tutorial index page still uses the old links:
https://github.com/brian-team/brian2/blob/8f76c22556b3e788eb78b019ed5e1b1a11b713eb/dev/tools/docs/build_tutorials.py#L124

thesamovar:
OK fixed that one too!

thesamovar:
Can we explicitly update stable to point to that version until a new release comes out?

mstimberg:
All looking good to me now.

PR #1156, opened 2020-02-10
[MRG] Use filelock module

### LINKED ISSUE #1154: Investigate issues with parallel processes using Cython
According to a report on the [development mailing list](https://groups.google.com/d/topic/brian-development/FkeRd7BP9yM/discussion), running many simulations in parallel (with Cython) can sometimes lead to processes freezing. I'd say the most likely cause is a deadlock with the lock files we use to protect processes from trying to write to the same source files. In the [`filelock` branch](https://github.com/brian-team/brian2/tree/filelock), I replaced our own locking system by https://github.com/benediktschmitt/py-filelock , which seems to be a bit more mature and well-tested.
However, this does not seem to fix all issues.

mstimberg:
Even though it did not fix issue #1154 after all, I still think it is preferable to use the `filelock` module from https://github.com/benediktschmitt/py-filelock instead of our own solution. The approaches are basically the same, but the `filelock` module makes more effort to release locks in case of exceptions, etc. I included it directly in our utils package, since it is a single file and released into the Public Domain, so this seemed easier than adding a new dependency (it also allows us to downgrade the log level, otherwise we'd get messages about acquired/released locks by default...). 

This PR also documents the issue discussed in #1154, and I'd therefore say it closes #1154.

### DIFF
--- a/brian2/codegen/runtime/cython_rt/extension_manager.py
+++ b/brian2/codegen/runtime/cython_rt/extension_manager.py
@@ -14,12 +14,6 @@ import sys
 import time
 
 try:
-    import msvcrt
-except ImportError:
-    msvcrt = None
-    import fcntl
-
-try:
     import hashlib
 except ImportError:
     import md5 as hashlib
@@ -39,6 +33,7 @@ except ImportError:
 from brian2.codegen.cpp_prefs import update_for_cross_compilation
 from brian2.utils.logger import std_silent, get_logger
 from brian2.utils.stringtools import deindent
+from brian2.utils.filelock import FileLock
 from brian2.core.preferences import prefs
 
 __all__ = ['cython_extension_manager']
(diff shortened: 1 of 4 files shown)


thesamovar:
Does including a copy mean that we won't benefit from any updates? Maybe preferable to have it as a dependency if it's pure Python?

mstimberg:
> Does including a copy mean that we won't benefit from any updates? Maybe preferable to have it as a dependency if it's pure Python?

I agree in principle, but there is the issue of pylock's logger that logs acquire/release messages as INFO messages (which we display by default). We'd have to somehow work around this issue if we don't use our own version. The last commit was 9 month ago, so it does not seem to change a lot anyway.

thesamovar:
Yeah and looking at the code it's pretty short. Maybe set up an alert for when they release new versions so we can remember to include changes?

thesamovar:
Although actually, is there no way to capture logs from another package and redefine them at a different level? I would have thought you could do that.

mstimberg:
>  Maybe set up an alert for when they release new versions so we can remember to include changes?

Yep, I activate the "watch releases" option on their repository :eyeglasses:

mstimberg:
> Although actually, is there no way to capture logs from another package and redefine them at a different level? I would have thought you could do that.

You can easily filter log messages from other packages to suppress them, but I'm not sure how you'd change their log level. Probably you could write your own logging handler or something like that. I think we have the problem mostly because we catch all log messages (to format them consistently and send them to the log file), otherwise `filelock` would still use the default display level – we display info messages but the default is to only display warning messages.

Don't think it is worth messing with all this...

thesamovar:
Fine by me! :)

Issue #528, opened 2015-08-04
better options for record keyword in StateMonitor


thesamovar:
For example, ``record=True`` should work for Synapses, as well as things like ``record='abs(i-j)<5'``, etc.

mstimberg:
> For example, record=True should work for Synapses,

It works for synapses, just not in standalone mode (because it converts True --> [0, 1, 2, ...] at creation time and the number of synapses is not yet known at this point).

> as well as things like record='abs(i-j)<5', etc.

This is slightly tricky as we need a check for whether only constants are involved in the expression. And even constants can change between runs so we'd have to deal with this as well.

To make all this work properly, we'd have to move the index calculation out of the Python code into a new template (or an initializer section of a template as proposed in #390). Not necessarily very complicated, but it means more target-dependent code.

All in all, I think this is useful but a non-trivial amount of work.

thesamovar:
Ah I see. Might it be worth special casing record=True?

mstimberg:
record=True is probably important enough to be special cased, but it requires some non-trivial changes to the templates. And we'd have to decide what to do when the number of synapses changes between two runs and how to detect this efficiently (in standalone particularly)

Issue #961, opened 2018-05-28
StateMonitor doesn't check if synapse indices passed through `record=` parameter exist


denisalevi:
When recording variables from a `Synapses` object and passing the synapse indices as array to the `StateMonitors` `record` parameter, there is no check if the indices are in the range of created synapses.  In standalone mode, this is not possible, since the synapses are not created before the network is run. But its also the case in runtime mode.
If an index is passed that does not have an associated synapse with it, a random memory location seems to be returned in the monitor.

Example:
```python
from brian2 import *                                       

N = NeuronGroup(10, 'v:1', threshold='True')

syn = Synapses(N, N, 'w:1', on_pre='w+=1')
syn.connect(p=0.1)

mon = StateMonitor(syn, 'w', record=[0, 200])

run(10*defaultclock.dt)

print(mon.w)
```
Output:
```
[[  0.00000000e+000   1.00000000e+000   2.00000000e+000   3.00000000e+000
    4.00000000e+000   5.00000000e+000   6.00000000e+000   7.00000000e+000
    8.00000000e+000   9.00000000e+000]
 [  6.93157187e-310   6.93157187e-310   6.93157187e-310   6.93157187e-310
    6.93157187e-310   6.93157187e-310   6.93157187e-310   6.93157187e-310
    6.93157187e-310   6.93157187e-310]]
```

And in the figure below is an example where things got really confusing. These are traces of some Synapses variables (weights in this case), which are all initialised with 10 and should not change for the first 10000 ms (x-axis). But the indices passed as `record` are too high, giving some traces at 0 (closer inspection shows some `e-310` values). But on top of that the yellow trace even seems to converge so some steady state value. This is not my own simulation, so I don't know if there aren't some other effects here, but I was told and it looked to me as if the weights should clearly not behave as the yellow trace does. It looks like the monitors out of range synapse weight index indexes some other dynamic variables memory that was initialized elsewhere (there are other dynamic variables in the Synapses object that might be initialized just after the weight values?).

![weights](https://user-images.githubusercontent.com/11474000/40619912-2faa029a-6297-11e8-8620-cc06b18c70ca.png)

So I guess it would make sense to have some kind of check or report for this? Some suggestions:

1. Before the run statement, one could check if the indices are at least in the range of the maximum number of synapses, that could be created. In the example above that would be 100 synapses, so a `record=[200]` should maybe throw an error.
2. In runtime mode, it could be checked directly if the recorded indices are in the range of the number of created synapses.
[...]

PR #1157, opened 2020-02-21
[MRG] Move preparation into templates

### LINKED ISSUE #390: Deal with preparation in templates

mstimberg:
As discussed in #390, it would be more consistent if we moved stuff that we run only once in `before_run` (spike queue initialization and preparation of `SpatialNeuron`) into the templates. This also opens the possibility to fix issues #528 and #961 by moving the interpretation of `StateMonitor`'s `record` argument into the template.

This PR lays the foundation for future changes. In addition to the `maincode` block, templates now also have a `before_code` and `after_code` block. If they are present, they get compiled into separate files and executed during `before_run`/`after_run`. Since we certainly do not want to compile a "do nothing" before/after code block for every code object, we check for the presence of a special `EMPTY_CODE_BLOCK` comment in the code (set in the `common_groups` template). If it is present it means that the child template did not overwrite the `before_code`/`after_code` block and we therefore do not need to compile anything. We cannot simply check for an empty code block since the main template already defines a lot of boilerplate code for each of these blocks.

For now, we only use `before_code` for the spike queue initialization and the `SpatialNeuron` precalculation. There's no facility to have the user put code directly into these blocks, but I think we shouldn't worry about this before we have a clear use case. There are also small inconveniences (e.g. the `USES_VARIABLES` declaration applies to the whole template and not individual blocks), but I'd say it's "good enough" and an improvement over the previous state.

Closes #390


brian2/spatialneuron/spatialneuron.py
```diff
@@ -600,24 +600,12 @@ def __init__(self, group, method, clock, order=0):
...
-                                                          additional_variables=self.variables,
-                                                          run_namespace=run_namespace)
-        self._prepare_codeobj()
+        super(SpatialStateUpdater, self).before_run(run_namespace)
         # Raise a warning if the slow pure numpy version is used
-        # For simplicity, we check which CodeObject class the _prepare_codeobj
-        # is using, this will be the same as the main state updater
         from brian2.codegen.runtime.numpy_rt.numpy_rt import NumpyCodeObject
-        if isinstance(self._prepare_codeobj, NumpyCodeObject):
+        if isinstance(self.code_objects[0], NumpyCodeObject):
```
thesamovar: This ``self.code_objects[0]`` makes me slightly nervous. Can the different items in the ``code_objects`` list have different types or are they guaranteed to all have the same type? If the former, then hard coding the 0 to refer to some particular thing feels like it might break later. If the latter, it's OK but perhaps we should think of adding some assertions for that?

brian2/synapses/synapses.py
```diff
@@ -344,8 +337,8 @@ def before_run(self, run_namespace):
                                                              needed_variables=needed_variables,
                                                              template_kwds=template_kwds,
                                                              run_namespace=run_namespace)
-
-        self._code_objects.insert(0, weakref.proxy(self._pushspikes_codeobj))
+        self.code_objects[:] = [weakref.proxy(self._pushspikes_codeobj),
+                                weakref.proxy(self.create_default_code_object(run_namespace))]
```
thesamovar: Didn't check the logic here carefully, but is this always OK to replace ``code_objects`` rather than insert a new item?

brian2/spatialneuron/spatialneuron.py
mstimberg: In general there's no guarantee. This could potentially be an issue in situations where we allow the user to specify a `codeobj_class` manually (which we do for `NeuronGroup` but not in `SpatialNeuron` – not sure we had a reason for that). Here, we only have a single object in the list, so there's no potential for confusion.

brian2/synapses/synapses.py
mstimberg: In this example we are explicitly overwriting the standard logic to sneak in an additional code object, and we cannot create this other object after the standard code object creation has been done. This is all quite convoluted and difficult to follow – definitely a prime candidate for some refactoring. For now, I don't see how to do it differently without doing massive changes.

To make this a bit clearer: in `Synapses.create_code_objects`, we overwrite the default procedure from `CodeRunner.create_code_objects`, where `self.code_objects` is replaced:
https://github.com/brian-team/brian2/blob/bf5f4bc3971c463e470dcad890d2e6c13b6d584e/brian2/groups/group.py#L1132-L1139

mstimberg:
I'll go ahead with merging this then, right?

thesamovar:
Yep!

Issue #1162, opened 2020-03-06
Improve report function in C++ standalone mode


mstimberg:
[Note that this issue is "reserved" for prospective GSoC students and probably there will be multiple pull requests open at the same time. No pull request will be merged before the end of the application period.]

When you use `report='text'` with a run (see the [documentation on progress reports](https://brian2.readthedocs.io/en/stable/user/running.html#progress-reporting) and [custom reports](https://brian2.readthedocs.io/en/stable/advanced/scheduling.html#id1)), you get a standard progress report that tells you how much of the simulation has been already done, how much time was spent, and an estimate of the remaining time. In runtime mode, the simulated and estimated time are formatted in hours/minutes/seconds for readability:
```
18.2041 s (1%) simulated in 1m 0s, estimated 53m 56s remaining.
```
In standalone mode, the display simply shows the times in seconds:
```
37.4416 s (3%) simulated in 70 s, estimated 1800 s remaining.
```

**It would be nice to be consistent here and show the time nicely formatted in standalone mode as well.**

*Bonus* improvements:

* the [documentation](https://brian2.readthedocs.io/en/stable/advanced/scheduling.html#custom-progress-reporting) currently does not have any interesting example for a custom progress report function in C++ standalone. It would be nice to add one, e.g. one that writes the progress to a file or one that displays a simple text-based progress bar on the terminal.
* we might want to implement some simple(!) approach to improve the estimates of the remaining time (this would also apply to runtime mode). See e.g. https://stackoverflow.com/questions/933242/smart-progress-bar-eta-computation for some pointers.

rahuliitg:
When I'm clicking on the "Edit on Github" link in the documentation page, I'm getting a 404 error. @mstimberg can you please guide me about how to edit the documentation in this case.

rahuliitg:
Hi, @mstimberg I am getting initial checks failed because I'm using the std::string in my solution. Should I include corresponding libraries while generating coder in cpp or try out a solution using char*. The latter solution makes code cumbersome.

mstimberg:
> When I'm clicking on the "Edit on Github" link in the documentation page, I'm getting a 404 error. @mstimberg can you please guide me about how to edit the documentation in this case.

[...]

PR #1165, opened 2020-03-09
[GSoc2020] Improve report function in C++ standalone mode #1162

zeph1yr:
Added a time_format function to device.py and included <string> library to network.cpp that solves this issue.

### DIFF
--- a/brian2/codegen/cpp_prefs.py
+++ b/brian2/codegen/cpp_prefs.py
@@ -110,7 +110,7 @@ prefs.register_preferences(
         '''
         ),
     extra_compile_args_gcc=BrianPreference(
-        default=['-w', '-O3', '-ffast-math', '-fno-finite-math-only', '-march=native'],
+        default=['-w', '-O3', '-ffast-math', '-fno-finite-math-only', '-march=native', '-std=c++11'],
         docs='''
         Extra compile arguments to pass to GCC compiler
         '''
--- a/brian2/devices/cpp_standalone/device.py
+++ b/brian2/devices/cpp_standalone/device.py
@@ -1381,14 +1381,49 @@ class CPPStandaloneDevice(Device):
 
         # Code for a progress reporting function
         standard_code = '''
+                       std::string time_format(float time_in_seconds)
+                       {
+                                std::string s;
+                                if(time_in_seconds<1)
+                                {
+                                    s = ">1s";
+                                    return s;
+                                }
+
+                                int day = time_in_seconds / (24 * 3600);
+
+                                if(day!=0) s+= std::to_string(day)+"d ";
+
+                                time_in_seconds = fmod(time_in_seconds, (24 * 3600));
+
+                                int hour = time_in_seconds / 3600;
+
+                                if(hour!=0)s+= std::to_string(hour)+"h ";
+
+                                time_in_seconds = fmod(time_in_seconds,3600);
+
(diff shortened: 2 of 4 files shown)


zeph1yr:
This commit fixes #1162

mstimberg:
@zeph1yr  Thanks for the pull request, please let me know with a short comment whether you are still working on it or whether you consider it finished from your side. If it is basically finished, I'd start reviewing it.

zeph1yr:
@mstimberg I have tested the implemented function and consider it finished from my side. Eagerly waiting it to be reviewed.

mstimberg:
I tried out your code and noticed that the time formatting is only applied to the elapsed time, not to the estimated remaining time. Could you change that? I'll have a closer look at the details afterwards.

zeph1yr:
@mstimberg A new PR is on it's way.

zeph1yr:
@mstimberg Also there was a need to implement our own '''std::to_string''' function because "to_string not being part of std" was a known bug under MinGW which was later fixed and it depends on the compiler version. After trying to debug failed Travis CI build implementing my own function seemed to be a better choice.

mstimberg:
> Also there was a need to implement our own '''std::to_string''' function because "to_string not being part of std" was a known bug under MinGW which was later fixed and it depends on the compiler version. 

This is interesting to know and would certainly have been a good reason for that approach in other contexts. However, we do not support MinGW under Windows, in our code we always assume the use of Visual Studio. 

>After trying to debug failed Travis CI build implementing my own function seemed to be a better choice.

Do you mean Appveyor? Travis CI is running under Linux and OS X. But even Appveyor should not use MinGW as explained above.

mstimberg:
See my comment [here](https://github.com/brian-team/brian2/pull/1171#issuecomment-598303539), `std::to_string` should work correctly with older `gcc` versions if we pass `-std=c++11`.

zeph1yr:
> Do you mean Appveyor? Travis CI is running under Linux and OS X. But even Appveyor should not use MinGW as explained above.
@mstimberg If I remember correctly, it was Travis CI build that failed and the environment was linux. The error pointed to to_string function and implementing my own function resolved the issue. 
An updated PR is on it's way.

zeph1yr:
> See my comment [here](https://github.com/brian-team/brian2/pull/1171#issuecomment-598303539), `std::to_string` should work correctly with older `gcc` versions if we pass `-std=c++11`.

@mstimberg An updated PR is on it's way.


> Do you mean Appveyor? Travis CI is running under Linux and OS X. But even Appveyor should not use MinGW as explained above.

If I remember correctly it was Travis CI build that failed and the environment was Linux, also the error pointer to to_string function and implementing my own function resolved the issue. Anyways, thank you for the insight.

mstimberg:
> If I remember correctly it was Travis CI build that failed and the environment was Linux, also the error pointer to to_string function and implementing my own function resolved the issue.

No, never mind, it did indeed fail on Linux (at least for some g++ versions). I was just confused because you mentioned MinGW which is a Windows-only software.

mstimberg:
Apologies for the long delay in reviewing. I did not want to give further comments on PRs made as part of GSoC applications after the application deadline, to not give participants the false impression that this work was still taken into account for the application. I am going to close all "GSoC pull request" now, but of course we'd be more than happy to see you as a contributor in the future. Thanks again for your contribution!

PR #1174, opened 2020-03-13
[WIP] Resting state integration

### LINKED ISSUE #1064: Automatic calculation of resting state

Vigneswaran-Chandrasekaran:
This PR is the newer version of PR #1123 
To follow up please refer to the older one ( #1123 ) conversations for the discussions and comments.

#### Brief description:
Calculation of resting-state of the model, using the `resting_state` function of `NeuronGroup` class. 

##### Usage syntax:
```python
group = NeuronGroup(...)
equilibrium_point = group.resting_state(initial_guess)
```
##### Simple example:
```python
tau = 10*ms
eqs = '''dv/dt = (1-v)/tau : 1'''
group = NeuronGroup(1, eqs, method='exact')
result = group.resting_state({'v':  0})    # giving initial guess
```

Fixes #1064

### DIFF
--- a/brian2/groups/neurongroup.py
+++ b/brian2/groups/neurongroup.py
@@ -16,10 +16,12 @@ from brian2.core.preferences import prefs
 from brian2.core.spikesource import SpikeSource
 from brian2.core.variables import (Variables, LinkedVariable,
                                    DynamicArrayVariable, Subexpression)
+from brian2.core.namespace import get_local_namespace
 from brian2.equations.equations import (Equations, DIFFERENTIAL_EQUATION,
                                         SUBEXPRESSION, PARAMETER,
                                         check_subexpressions,
-                                        extract_constant_subexpressions)
+                                        extract_constant_subexpressions,
+                                        SingleEquation)
 from brian2.equations.refractory import add_refractoriness
 from brian2.parsing.expressions import (parse_expression_dimensions,
                                         is_boolean_expression)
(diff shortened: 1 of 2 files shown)


Vigneswaran-Chandrasekaran:
I've made the necessary changes for integrating Jacobian to calculate resting state as discussed [previously](https://github.com/brian-team/brian2/pull/1123#issuecomment-598340469). The new structure now looks like,

- The NeuronGroup's member function `resting_state()` prepares the initial guess of state variables (default to `zero` if not given by the user). 
- Further, `resting_state()` calls two helper functions, `_rhs_equation` and `_jacobian_equation`, which returns the `NeuronGroup` objects with RHS and Jacobian equations respectively. 
- `rhs_group` and `jac_group` acts as bare-skeleton NeuronGroup objects with defined governing equations. The above groups are passed to `_wrapper()`.
- `_wrapper()` calls `evaluate_states(group, values, states)` function, which sets/gets the values/states of the corresponding group (i.e.  `rhs_group` or `jac_group`)
- Once, the solution is found by `root()`, the Jacobin matrix is prepared for the solution, using the same helper functions and stability is analyzed

Everything works fine and solutions are accurate, I also added a test case for stability check. However, a couple of doubts including,
- The `fjac` of the result of `root()` is different from actual Jacobin of the result
- Number of `iter` with `jac = True` and `jac = False` are not much different

PR #1177, opened 2020-03-15
[Gsoc 2020] New approach to estimate the remaining time

### LINKED ISSUE #1173: Improve remaining time estimation
Taking this out of #1162 since it is not only about standalone mode, quoting from there:

We might want to implement some simple(!) approach to improve the estimates of the remaining time (this would also apply to runtime mode). See e.g. https://stackoverflow.com/questions/933242/smart-progress-bar-eta-computation for some pointers.

rahuliitg:
This commit estimates the remaining time using the exponential moving average method. It uses the dynamic value of "smoothing factor (alpha)" which is calculated by detecting slow end or slow start and works well even when we have varying or constant progress.

@mstimberg this is what I got after using a new approach for estimation (data is taken from this file -> https://gist.github.com/0f11f1e15c949e1c254afd2b57d35cdf )
image

Fixes #1173

### DIFF
--- a/brian2/core/network.py
+++ b/brian2/core/network.py
@@ -87,6 +87,37 @@ def _format_time(time_in_s):
 
     return text
 
+class progressData(object):
+    '''
+        object to store data about progress history
+        time_elapsed_arr stores time_elapsed till the computation started
+        completed_arr stores how much computation completed till now
+
+    '''
+    def __init__(self):
+        self.index = -1
+        self.time_elapsed_arr = []
+        self.completed_arr = []
+        self.alpha = 0.6
+        self.speed = 0
+
+    def estimate_remaining_time(self,completed, time_elapsed):
+        self.index += 1
+        index = self.index
+        if(self.index == 0):
+            self.speed = completed/time_elapsed
+        else:
+            currSpeed = (completed-self.completed_arr[index-1])/(time_elapsed-self.time_elapsed_arr[index-1])
+            if(index >= 3):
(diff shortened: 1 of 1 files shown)


mstimberg:
This looks good in general! I wonder how well it works in practice, as there are a number of hardcoded heuristics in there (the alpha values, for example). My gist was an attempt at capturing some typical situations, but we should test this more widely to make sure it does not make the estimations worse in other circumstances.

rahuliitg:
@mstimberg yes I was also thinking about that. Is there any dataset from real simulations on which I can test.

mstimberg:
I guess you could run some of the more complex [examples](https://brian2.readthedocs.io/en/stable/examples/index.html). But you'd have to store data for later analysis yourself. I'm afraid I don't have any other data set handy at the moment.

mstimberg:
Apologies for the long delay in reviewing. I did not want to give further comments on PRs made as part of GSoC applications after the application deadline, to not give participants the false impression that this work was still taken into account for the application. I am going to close all "GSoC pull request" now, but of course we'd be more than happy to see you as a contributor in the future. Thanks again for your contribution!

This PR was a promising start but it would still need a bit of testing and polishing before it can get integrated. If you are interested in continuing to work on this, please feel free to reopen the PR! But please take into account that I am still not 100% sure whether we will find a solution that is a clear improvement over the current state. There is therefore a risk that we will abandon the idea at some point :-/

Issue #1179, opened 2020-04-06
biexp synapse documentation not matching implementation?


felix11h:
A biexponential synapse, as per the Brian 2 documentation, should implement a current/conductance as 
`g(t)=(tau2/(tau2-tau1))*(exp(-t/tau1)-exp(-t/tau2))`

When playing around with the biexponential synapse concept, I found that maybe the above description might need to be updated. Consider

``` python
import matplotlib as mpl
mpl.use('Agg')
import pylab as pl

import numpy as np
from brian2 import *

fig, ax = pl.subplots()

xmin, xmax = 0,100
xs = np.linspace(xmin, xmax, 10000)

# biexp as implemented in library/synapses.py
model = '''
        dx/dt = (invpeak*y-x)*invtau1 : 1
        dy/dt = -y*invtau2            : 1
        '''

tau1, tau2 = 1.5*ms, 10*ms

invtau1, invtau2 = 1./tau1, 1./tau2
invpeak = (tau2 / tau1) ** (tau1 / (tau2 - tau1))

N = NeuronGroup(1, model)
N.y = 1

mon = StateMonitor(N, 'x', record=True)
run(xmax*ms)
ax.plot(mon.t/ms, mon.x.T)


# --------------------------

# g(t) of biexp as explained in documentation
def g(t, tau1, tau2):
    return (tau2/(tau2-tau1))*(np.exp(-t/tau1)-np.exp(-t/tau2))

ax.plot(xs, g(xs, tau1/ms, tau2/ms))

fig.savefig('figure.png')
```

which gives 

![figure](https://user-images.githubusercontent.com/610633/78569168-7f1cbe00-7823-11ea-9889-fad94753cdf8.png)

I expected to see identical curves. 

Note that it doesn't simply seem to be a matter of a missing minus sign as the amplitudes also don't seem to match. Any thoughts on this? Am I missing something? Thanks!

mstimberg:
[...]

PR #1181, opened 2020-04-15
Biexponential synapse documentation updated.

Syed-Osama-Hussain:
Attempt to solve #1179.


mstimberg:
Hi @Syed-Osama-Hussain, apologies for leaving this unresolved for [checks notes] 5 years! Your proposed changes weren't quite correct (there was some LaTeX issue, but also it sounded as if the normalization factor should apply to the code, where it was meant for the documented equation), and since I did not want to bother you with updating this PR after such a long time, I went ahead and fixed the issue myself in #1722. Apologies again, and we'd be glad to have you contribute in the future – I promise I won't take another five years to get back to you!

PR #1193, opened 2020-06-04
Use rotating log handler to keep debug log from growing out of bounds.

Debilski:
Instead of a plain `FileHandler` the debug log is now logged using the `RotatingFileHandler` which allows us to specify a maximum file size. (Hard coded to 10MB but if needed, I can change the value or make it configurable.)
Note that the `RotatingFileHandler` currently needs to create at least one backup file for the mechanism to work, which means that we have to account for this in the cleanup function.

### DIFF
--- a/brian2/utils/logger.py
+++ b/brian2/utils/logger.py
@@ -91,6 +91,20 @@ if 'logging' not in prefs.pref_register:
             to a file. The log level can be set via the `logging.file_log_level`
             preference.
             '''),
+        file_log_max_size=BrianPreference(
+            default=10000000,
+            docs='''
+            The maximum size for the debug log before it will be rotated.
+
+            If set to any value ``> 0``, the debug log will be rotated once
+            this size is reached. Rotating the log means that the old debug log
+            will be moved into a file in the same directory but with suffix ``".1"``
+            and the a new log file will be created with the same pathname as the
+            original file. Only one backup is kept; if a file with suffix ``".1"``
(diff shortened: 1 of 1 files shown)


mstimberg:
Thanks, looking great! I think I'd prefer if the value were configurable, the cleanest would be to add a preference (e.g. `file_log_max_size`) to the preferences defined near the top of the file. Even better would be if setting this preference to 0 switched the feature completely off, i.e. used a `FileHandler` instead of a `RotatingFileHandler`. Simply because we never know whether log file rotation might not lead to problems on some architectures/file systems/... at some point (although unlikely for a well-tested part of the standard library), and it would be a shame if users could not switch it off in such cases.
But this is all a bonus, let us know if you don't have the time to implement it. Thanks again!

mstimberg:
Great, thanks. If `RotatingFileHandler` already switches everything off for `maxBytes` ≤ 0, this should be good enough. I don't have a strong opinion on the value of 10MB, @thesamovar what do you think? I am not worried about the test suite, I don't think I ever used the log file for errors triggered as part of it. In principle we could even switch off file logging completely when running the test suite, the main reason for leaving it enabled is that it could not catch errors with the logging itself otherwise.

mstimberg:
Ok, let's go with the 10MB for now, we can always change the default later. I tested it and the rotation works fine and happens at the expected size. Setting it to 0 seems to work as well.

PR #1194, opened 2020-06-05
[MRG] Minor compatibility fixes

mstimberg:
Should get rid of any `DeprecationWarning` or `SyntaxWarning` triggered by our code.


mstimberg:
> Seems good to me - why remove test_unitarray?

This file was a bit odd. I used in originally while developing the new version of the unit support, but it consisted of print statements and not of actual tests. Due to its name it gets imported (and therefore executed) when looking for tests, but it does not have any actual tests in it. It wasn't doing much harm, but it started to raise deprecation warnings for calling scipy functions that are no longer supported or in different places now. I could have just remove the scipy part, but I felt it made more sense to simply get rid of the whole file. If we wanted to keep it, it could go into `dev`. But then, who would run it :)?

thesamovar:
That's fine - it can go! :-)

PR #1196, opened 2020-06-12
Brian obj exception

### LINKED ISSUE #1183: Exception chaining instead of BrianObjectException?
Now that Brian is going to be Python 3 only, should we use exception chaining instead of BrianObjectException?

mstimberg:
Fixes #1183 
This replaces our hand-crafted `BrianObjectException` (which inherits from both `BrianObjectException` and the actual exception that was raised, and includes the stacktrace of the original exception) by exception chaining. In Python 2, we did not have another choice, but I think in Python 3 this is clearly an improvement, because:
1. Our exception type sometimes made problems during development (I don't remember the details, but working on Brian itself sometimes lead to errors where the `BrianObjectException` could not be constructed, completely hiding the actual problem).
2. In Python 3 our previous solution leads to a lot of redundancy in the exception traceback (see example below).

The are two open points before merging:
1. The current implementation uses exception chaining in the standard way, but this means that the actual error is not stated in the end, but only higher up in the traceback. The advantage is that it is not repeated, but I think it is properly better to have it clearly stated in the end (when I'm teaching Python, I'm usually telling students to "scroll to the end of the error message").
2. There's a little backwards incompatibility when the code uses `try`/`except`, because the `BrianObjectException` that gets raised is no longer a subclass of the actual error (`DimensionMismatchError`, `TypeError`, ....). I don't think it is a big issue, and I added a section to our compatibility docs to state that users shouldn't rely on `try`/`except` for program logic in Brian code. Not that I'm expecting many users to do, but still... It makes a small difference in our test cases, but that's ok for me. For situations where `BrianObjectException` is raised, i.e. when the error is detected at the start of a run, our tests previously used:
```Python
with pytest.raises(DimensionMismatchError): 
    run(0*ms)
```
Now, it should read:
```Python
with pytest.raises(BrianObjectException) as exc: 
    run(0*ms)
    assert exc.errisinstance(DimensionMismatchError)
```
### Example
For the following code:
```Python
from brian2 import *
group = NeuronGroup(1, 'dv/dt = -v/tau : volt')
tau = 10 # missing units
run(100*ms)
```
<details>
<summary>Error in master</summary>

```pytb
ERROR      Brian 2 encountered an unexpected error. If you think this is bug in Brian 2, please report this issue either to the mailing list at <http://groups.google.com/group/brian-development/>, or to the issue tracker at <https://github.com/brian-team/brian2/issues>. Please include this file with debug information in your report: /tmp/brian_debug_r8d17r7s.log  Additionally, you can also include a copy of the script that was run, available at: /tmp/brian_script_zcmaqkdd.py Thanks! [brian2]
  File "/home/marcel/programming/brian2/brian2/equations/equations.py", line 956, in check_units
    check_dimensions(str(eq.expr), self.dimensions[var] / second.dim,
  File "/home/marcel/programming/brian2/brian2/equations/unitcheck.py", line 45, in check_dimensions
    fail_for_dimension_mismatch(expr_dims, dimensions, err_msg)
  File "/home/marcel/programming/brian2/brian2/units/fundamentalunits.py", line 184, in fail_for_dimension_mismatch
    raise DimensionMismatchError(error_message, dim1)
brian2.units.fundamentalunits.DimensionMismatchError: Expression -v/tau does not have the expected unit metre ** 2 * kilogram * second ** -4 * amp ** -1 (unit is V).

  File "/home/marcel/programming/brian2/brian2/core/network.py", line 897, in before_run
    obj.before_run(run_namespace)
  File "/home/marcel/programming/brian2/brian2/groups/neurongroup.py", line 884, in before_run
    self.equations.check_units(self, run_namespace=run_namespace)
  File "/home/marcel/programming/brian2/brian2/equations/equations.py", line 959, in check_units
    raise DimensionMismatchError(('Inconsistent units in '
brian2.units.fundamentalunits.DimensionMismatchError: Inconsistent units in differential equation defining variable v:
Expression -v/tau does not have the expected unit metre ** 2 * kilogram * second ** -4 * amp ** -1 (unit is V).

  File "/home/marcel/.config/JetBrains/PyCharm2020.1/scratches/scratch_10.py", line 4, in <module>
    run(100*ms)
  File "/home/marcel/programming/brian2/brian2/units/fundamentalunits.py", line 2385, in new_f
    result = f(*args, **kwds)
  File "/home/marcel/programming/brian2/brian2/core/magic.py", line 373, in run
    return magic_network.run(duration, report=report, report_period=report_period,
  File "/home/marcel/programming/brian2/brian2/core/magic.py", line 231, in run
    Network.run(self, duration, report=report, report_period=report_period,
  File "/home/marcel/programming/brian2/brian2/core/base.py", line 278, in device_override_decorated_function
    return func(*args, **kwds)
  File "/home/marcel/programming/brian2/brian2/units/fundamentalunits.py", line 2385, in new_f
    result = f(*args, **kwds)
  File "/home/marcel/programming/brian2/brian2/core/network.py", line 1008, in run
    self.before_run(namespace)
  File "/home/marcel/programming/brian2/brian2/core/base.py", line 278, in device_override_decorated_function
    return func(*args, **kwds)
  File "/home/marcel/programming/brian2/brian2/core/network.py", line 899, in before_run
brian2.core.base.BrianObjectException: Original error and traceback:
  File "/home/marcel/programming/brian2/brian2/equations/equations.py", line 956, in check_units
    check_dimensions(str(eq.expr), self.dimensions[var] / second.dim,
  File "/home/marcel/programming/brian2/brian2/equations/unitcheck.py", line 45, in check_dimensions
    fail_for_dimension_mismatch(expr_dims, dimensions, err_msg)
  File "/home/marcel/programming/brian2/brian2/units/fundamentalunits.py", line 184, in fail_for_dimension_mismatch
    raise DimensionMismatchError(error_message, dim1)
brian2.units.fundamentalunits.DimensionMismatchError: Expression -v/tau does not have the expected unit metre ** 2 * kilogram * second ** -4 * amp ** -1 (unit is V).

  File "/home/marcel/programming/brian2/brian2/core/network.py", line 897, in before_run
    obj.before_run(run_namespace)
  File "/home/marcel/programming/brian2/brian2/groups/neurongroup.py", line 884, in before_run
    self.equations.check_units(self, run_namespace=run_namespace)
  File "/home/marcel/programming/brian2/brian2/equations/equations.py", line 959, in check_units
    raise DimensionMismatchError(('Inconsistent units in '
brian2.units.fundamentalunits.DimensionMismatchError: Inconsistent units in differential equation defining variable v:
Expression -v/tau does not have the expected unit metre ** 2 * kilogram * second ** -4 * amp ** -1 (unit is V).

Error encountered with object named "neurongroup".
  File "/home/marcel/.config/JetBrains/PyCharm2020.1/scratches/scratch_10.py", line 2, in <module>
    group = NeuronGroup(1, 'dv/dt = -v/tau : volt')

An error occurred when preparing an object. brian2.units.fundamentalunits.DimensionMismatchError: Inconsistent units in differential equation defining variable v:
Expression -v/tau does not have the expected unit metre ** 2 * kilogram * second ** -4 * amp ** -1 (unit is V).
```
</details>

<details>
<summary>Error in this PR</summary>

```pytb
  File "/home/marcel/programming/brian2/brian2/equations/equations.py", line 956, in check_units
    check_dimensions(str(eq.expr), self.dimensions[var] / second.dim,
  File "/home/marcel/programming/brian2/brian2/equations/unitcheck.py", line 45, in check_dimensions
    fail_for_dimension_mismatch(expr_dims, dimensions, err_msg)
  File "/home/marcel/programming/brian2/brian2/units/fundamentalunits.py", line 184, in fail_for_dimension_mismatch
    raise DimensionMismatchError(error_message, dim1)
brian2.units.fundamentalunits.DimensionMismatchError: Expression -v/tau does not have the expected unit metre ** 2 * kilogram * second ** -4 * amp ** -1 (unit is V).

The above exception was the direct cause of the following exception:

  File "/home/marcel/programming/brian2/brian2/core/network.py", line 897, in before_run
    obj.before_run(run_namespace)
  File "/home/marcel/programming/brian2/brian2/groups/neurongroup.py", line 884, in before_run
    self.equations.check_units(self, run_namespace=run_namespace)
  File "/home/marcel/programming/brian2/brian2/equations/equations.py", line 959, in check_units
    raise DimensionMismatchError(('Inconsistent units in '
brian2.units.fundamentalunits.DimensionMismatchError: Inconsistent units in differential equation defining variable v:
Expression -v/tau does not have the expected unit metre ** 2 * kilogram * second ** -4 * amp ** -1 (unit is V).

The above exception was the direct cause of the following exception:

  File "/home/marcel/.config/JetBrains/PyCharm2020.1/scratches/scratch_10.py", line 4, in <module>
    run(100*ms)
  File "/home/marcel/programming/brian2/brian2/units/fundamentalunits.py", line 2385, in new_f
    result = f(*args, **kwds)
  File "/home/marcel/programming/brian2/brian2/core/magic.py", line 373, in run
    return magic_network.run(duration, report=report, report_period=report_period,
  File "/home/marcel/programming/brian2/brian2/core/magic.py", line 231, in run
    Network.run(self, duration, report=report, report_period=report_period,
  File "/home/marcel/programming/brian2/brian2/core/base.py", line 278, in device_override_decorated_function
    return func(*args, **kwds)
  File "/home/marcel/programming/brian2/brian2/units/fundamentalunits.py", line 2385, in new_f
    result = f(*args, **kwds)
  File "/home/marcel/programming/brian2/brian2/core/network.py", line 1008, in run
    self.before_run(namespace)
  File "/home/marcel/programming/brian2/brian2/core/base.py", line 278, in device_override_decorated_function
    return func(*args, **kwds)
  File "/home/marcel/programming/brian2/brian2/core/network.py", line 899, in before_run
    raise BrianObjectException("An error occurred when preparing an object.", obj) from ex
brian2.core.base.BrianObjectException: Error encountered with object named "neurongroup".
  File "/home/marcel/.config/JetBrains/PyCharm2020.1/scratches/scratch_10.py", line 2, in <module>
    group = NeuronGroup(1, 'dv/dt = -v/tau : volt')

An error occurred when preparing an object.(See above for original error message and traceback.)
```
</details>

### DIFF
--- a/brian2/core/base.py
+++ b/brian2/core/base.py
@@ -294,10 +294,6 @@ class BrianObjectException(Exception):
     `BrianObject` was defined to better enable users to locate the source of
     problems.
 
-    You should use the `brian_object_exception` function to raise this, and
-    it should only be raised in an ``except`` block handling a prior
-    exception.
-
     Parameters
     ----------
 
@@ -308,26 +304,20 @@ class BrianObjectException(Exception):
     original_exception : Exception
         The original exception that was raised.
     '''
-    def __init__(self, message, brianobj, original_exception):
+    def __init__(self, message, brianobj):
         self._brian_message = message
         self._brian_objname = brianobj.name
-        self._brian_origexc = '\n'.join(traceback.format_exception_only(type(original_exception),
-                                                                        original_exception))
-        self._brian_origtb = traceback.format_exc()
         self._brian_objcreate = brianobj._creation_stack
         logger.diagnostic('Error was encountered with object "{objname}":\n{fullstack}'.format(
             objname=self._brian_objname,
             fullstack=brianobj._full_creation_stack))
 
     def __str__(self):
-        return ('Original error and traceback:\n{origtb}\n'
-                'Error encountered with object named "{objname}".\n'
+        return ('Error encountered with object named "{objname}".\n'
                 '{objcreate}\n\n'
-                '{message} {origexc}'
+                '{message}'
                 '(See above for original error message and traceback.)'
-                ).format(origtb=self._brian_origtb,
-                         origexc=self._brian_origexc,
-                         objname=self._brian_objname, message=self._brian_message,
+                ).format(objname=self._brian_objname, message=self._brian_message,
                          objcreate=self._brian_objcreate)
 
 
@@ -341,16 +331,6 @@ def brian_object_exception(message, brianobj, original_exception):
 
     See `BrianObjectException` for arguments and notes.
     '''
-    DerivedBrianObjectException = type('BrianObjectException',
-                                       (BrianObjectException, original_exception.__class__),
-                                       {})
-    new_exception = DerivedBrianObjectException(message, brianobj, original_exception)
-    # Copy over all exception attributes
-    for attribute in dir(original_exception):
-        if attribute.startswith('_'):
-            continue
-        try:
-            setattr(new_exception, attribute, getattr(original_exception, attribute))
-        except AttributeError:  # some attributes cannot be set
-            pass
-    return new_exception
+
+    raise NotImplementedError('The brian_object_exception function is no longer used. '
+                              'Raise a BrianObjectException directly.')
--- a/brian2/core/base.py
+++ b/brian2/core/base.py
@@ -315,7 +315,7 @@ class BrianObjectException(Exception):
     def __str__(self):
         return ('Error encountered with object named "{objname}".\n'
                 '{objcreate}\n\n'
-                '{message}'
+                '{message} '
                 '(See above for original error message and traceback.)'
                 ).format(objname=self._brian_objname, message=self._brian_message,
                          objcreate=self._brian_objcreate)
--- a/brian2/core/base.py
+++ b/brian2/core/base.py
@@ -316,6 +316,7 @@ class BrianObjectException(Exception):
         return ('Error encountered with object named "{objname}".\n'
                 '{objcreate}\n\n'
                 '{message}'
+                '(See above for original error message and traceback.)'
                 ).format(objname=self._brian_objname, message=self._brian_message,
                          objcreate=self._brian_objcreate)
 
--- a/brian2/core/base.py
+++ b/brian2/core/base.py
@@ -316,7 +316,6 @@ class BrianObjectException(Exception):
         return ('Error encountered with object named "{objname}".\n'
                 '{objcreate}\n\n'
                 '{message}'
-                '(See above for original error message and traceback.)'
                 ).format(objname=self._brian_objname, message=self._brian_message,
                          objcreate=self._brian_objcreate)
 
--- a/brian2/tests/test_stateupdaters.py
+++ b/brian2/tests/test_stateupdaters.py
@@ -766,8 +766,8 @@ def test_check_for_invalid_values_linear_integrator():
                 assert 'invalid_values' in repr(clog)
             else:
                 assert G.x[0] != 0
-        except BrianObjectException:
-            pass
+        except BrianObjectException as exc:
+            assert isinstance(exc.__cause__, UnsupportedEquationsException)
 
 
 if __name__ == '__main__':
--- a/brian2/tests/test_stateupdaters.py
+++ b/brian2/tests/test_stateupdaters.py
@@ -170,39 +170,44 @@ def test_multiplicative_noise():
     Eq1 = Equations('dv/dt = v*xi*(5*ms)**-0.5 :1')
     group1 = NeuronGroup(1, Eq1, method='euler')
     net1 = Network(group1)
-    with pytest.raises(UnsupportedEquationsException):
+    with pytest.raises(BrianObjectException) as exc:
         net1.run(0*ms)
+        assert exc.errisinstance(UnsupportedEquationsException)
 
     # Noise is multiplicative (multiplied with time)
     Eq2 = Equations('dv/dt = (t/ms)*xi*(5*ms)**-0.5 :1')
     group2 = NeuronGroup(1, Eq2, method='euler')
     net2 = Network(group2)
-    with pytest.raises(UnsupportedEquationsException):
+    with pytest.raises(BrianObjectException) as exc:
         net2.run(0*ms)
+        assert exc.errisinstance(UnsupportedEquationsException)
 
     # Noise is multiplicative (multiplied with time-varying variable)
     Eq3 = Equations('''dv/dt = w*xi*(5*ms)**-0.5 :1
                        dw/dt = -w/(10*ms) : 1''')
     group3 = NeuronGroup(1, Eq3, method='euler')
     net3 = Network(group3)
-    with pytest.raises(UnsupportedEquationsException):
+    with pytest.raises(BrianObjectException) as exc:
         net3.run(0*ms)
+        assert exc.errisinstance(UnsupportedEquationsException)
 
     # One of the equations has multiplicative noise
     Eq4 = Equations('''dv/dt = xi_1*(5*ms)**-0.5 : 1
                        dw/dt = (t/ms)*xi_2*(5*ms)**-0.5 :1''')
     group4 = NeuronGroup(1, Eq4, method='euler')
     net4 = Network(group4)
-    with pytest.raises(UnsupportedEquationsException):
+    with pytest.raises(BrianObjectException) as exc:
         net4.run(0*ms)
+        assert exc.errisinstance(UnsupportedEquationsException)
 
     # One of the equations has multiplicative noise
     Eq5 = Equations('''dv/dt = xi_1*(5*ms)**-0.5 : 1
                        dw/dt = v*xi_2*(5*ms)**-0.5 :1''')
     group5 = NeuronGroup(1, Eq5, method='euler')
     net5 = Network(group4)
-    with pytest.raises(UnsupportedEquationsException):
+    with pytest.raises(BrianObjectException) as exc:
         net5.run(0*ms)
+        assert exc.errisinstance(UnsupportedEquationsException)
 
 
 def test_pure_noise_deterministic(fake_randn_randn_fixture):
@@ -643,8 +648,9 @@ def test_locally_constant_check():
         else:
             # This should not
             with catch_logs():
-                with pytest.raises(UnsupportedEquationsException):
-                    net.run(0*ms)
+                with pytest.raises(BrianObjectException) as exc:
+                    net.run(0 * ms)
+                    assert exc.errisinstance(UnsupportedEquationsException)
 
         # multiplicative
         G = NeuronGroup(1, 'dv/dt = -v*ta(t)/(10*ms) : 1',
@@ -656,42 +662,48 @@ def test_locally_constant_check():
         else:
             # This should not
             with catch_logs():
-                with pytest.raises(UnsupportedEquationsException):
+                with pytest.raises(BrianObjectException) as exc:
                     net.run(0*ms)
+                    assert exc.errisinstance(UnsupportedEquationsException)
 
     # If the argument is more than just "t", we cannot guarantee that it is
     # actually locally constant
     G = NeuronGroup(1, 'dv/dt = -v*ta(t/2.0)/(10*ms) : 1',
                         method='exact', namespace={'ta': ta0})
     net = Network(G)
-    with pytest.raises(UnsupportedEquationsException):
+    with pytest.raises(BrianObjectException) as exc:
         net.run(0*ms)
+        assert exc.errisinstance(UnsupportedEquationsException)
 
     # Arbitrary functions are not constant over a time step
     G = NeuronGroup(1, 'dv/dt = -v/(10*ms) + sin(2*pi*100*Hz*t)*Hz : 1',
                     method='exact')
     net = Network(G)
-    with pytest.raises(UnsupportedEquationsException):
+    with pytest.raises(BrianObjectException) as exc:
         net.run(0*ms)
+        assert exc.errisinstance(UnsupportedEquationsException)
 
     # Stateful functions aren't either
     G = NeuronGroup(1, 'dv/dt = -v/(10*ms) + rand()*Hz : 1',
                     method='exact')
     net = Network(G)
-    with pytest.raises(UnsupportedEquationsException):
+    with pytest.raises(BrianObjectException) as exc:
         net.run(0*ms)
+        assert exc.errisinstance(UnsupportedEquationsException)
 
     # Neither is "t" itself
     G = NeuronGroup(1, 'dv/dt = -v/(10*ms) + t/second**2 : 1', method='exact')
     net = Network(G)
-    with pytest.raises(UnsupportedEquationsException):
+    with pytest.raises(BrianObjectException) as exc:
         net.run(0*ms)
+        assert exc.errisinstance(UnsupportedEquationsException)
 
     # But if the argument is not referring to t, all should be well
     G = NeuronGroup(1, 'dv/dt = -v/(10*ms) + sin(2*pi*100*Hz*5*second)*Hz : 1',
                     method='exact')
     net = Network(G)
     net.run(0*ms)
+
 
 def test_refractory():
     # Compare integration with and without the addition of refractoriness --
@@ -754,7 +766,7 @@ def test_check_for_invalid_values_linear_integrator():
                 assert 'invalid_values' in repr(clog)
             else:
                 assert G.x[0] != 0
-        except UnsupportedEquationsException:
+        except BrianObjectException:
             pass
 
 
--- a/brian2/codegen/codeobject.py
+++ b/brian2/codegen/codeobject.py
@@ -396,7 +396,7 @@ def create_runner_codeobj(group, code, template_name,
                 check_units_statements(c, variables)
             except (SyntaxError, ValueError) as ex:
                 error_msg = _error_msg(c, name)
-                raise ValueError(error_msg + str(ex))
+                raise ValueError(error_msg) from ex
 
     all_variable_indices = copy.copy(group.variables.indices)
     if additional_variables is not None:
@@ -423,7 +423,7 @@ def create_runner_codeobj(group, code, template_name,
                     value.implementations.add_numpy_implementation(value.pyfunc)
                 else:
                     raise NotImplementedError(('Cannot use function '
-                                               '%s: %s') % (varname, ex))
+                                               '%s: %s') % (varname, ex)) from ex
 
     # Gather the additional compiler arguments declared by function
     # implementations
--- a/brian2/codegen/runtime/cython_rt/cython_rt.py
+++ b/brian2/codegen/runtime/cython_rt/cython_rt.py
@@ -4,6 +4,7 @@ import sys
 
 import numpy
 
+from brian2.core.base import BrianObjectException
 from brian2.core.variables import (DynamicArrayVariable, ArrayVariable,
                                    AuxiliaryVariable, Subexpression)
 from brian2.core.functions import Function
@@ -155,7 +156,12 @@ class CythonCodeObject(NumpyCodeObject):
     def run_block(self, block):
         compiled_code = self.compiled_code[block]
         if compiled_code:
-            return compiled_code.main(self.namespace)
+            try:
+                return compiled_code.main(self.namespace)
+            except Exception as exc:
+                message = ('An exception occured during the execution of the {} '
+                           'block of code object {}.\n').format(block, self.name)
+                raise BrianObjectException(message, self.owner) from exc
 
     def _insert_func_namespace(self, func):
         impl = func.implementations[self]
--- a/brian2/codegen/runtime/numpy_rt/numpy_rt.py
+++ b/brian2/codegen/runtime/numpy_rt/numpy_rt.py
@@ -255,7 +255,7 @@ class NumpyCodeObject(CodeObject):
             _, _, tb = sys.exc_info()
             tb = tb.tb_next  # Line in the code object's code
             message += lines[tb.tb_lineno - 1] + '\n'
-            raise BrianObjectException(message, self.owner)
+            raise BrianObjectException(message, self.owner) from exc
         # output variables should land in the variable name _return_values
         if '_return_values' in self.namespace:
             return self.namespace['_return_values']
--- a/brian2/codegen/runtime/numpy_rt/numpy_rt.py
+++ b/brian2/codegen/runtime/numpy_rt/numpy_rt.py
@@ -7,7 +7,7 @@ from collections.abc import Iterable
 
 import numpy as np
 
-from brian2.core.base import brian_object_exception
+from brian2.core.base import BrianObjectException
 from brian2.core.preferences import prefs, BrianPreference
 from brian2.core.variables import (DynamicArrayVariable, ArrayVariable,
                                    AuxiliaryVariable, Subexpression)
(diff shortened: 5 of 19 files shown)


thesamovar:
Cool, will probably review in 1.5 weeks time because I'd like to think carefully about this one (and I'm actually on holiday next week!).

thesamovar:
Actually have some time today.

I agree with your point (1), if we can it would be good to put back the string representation of the original error so you can get a quick glance of the problem by looking at the end.

For point (2) I think it's OK since it's unlikely people have been relying on this and those that have been doing so should find it very easy to adapt. That said, does it conflict with our newly added policy on backwards incompatible changes being announced in advance, etc.?

I have no issue with changing the tests of course.

Actually, could we generate a new exception class by generating a new class deriving from both the original exception class and ``BrianObjectException``? It's a bit hacky perhaps, but would be backwards compatible and sort of neat in some way.

mstimberg:
> I agree with your point (1), if we can it would be good to put back the string representation of the original error so you can get a quick glance of the problem by looking at the end.

It feels like bad practice to include this kind of information in the exception object (same for the "(See above for original error message and traceback.)"), since it is only for presentation. I just realized that we could instead configure the console logging handler to format the exception in any way we want. We could even (optionally) get rid of most of the traceback – most Brian users are not really interested in what happened outside of their own code.

> That said, does it conflict with our newly added policy on backwards incompatible changes being announced in advance, etc.?

Well, since we have not published this policy yet, it cannot conflict with it :) as part of the PR I added a section about exceptions that explicitly states that you should not rely on them.



> Actually, could we generate a new exception class by generating a new class deriving from both the original exception class and `BrianObjectException`? It's a bit hacky perhaps, but would be backwards compatible and sort of neat in some way.

Well, that was the previous approach. As I mentioned, I ran into issues with it every once in a while, and I'm not sure it is a good practice in general. The information about the original exception class is still there (in the exception object's `__cause__` attribute), so it feels redundant to have it inherit from the class as well.

mstimberg:
I played around with the log formatter a bit, and here's the output in the most simplified version:
```pytb
ERROR      Brian 2 encountered an unexpected error. If you think this is bug in Brian 2, please report this issue either to the mailing list at <http://groups.google.com/group/brian-development/>, or to the issue tracker at <https://github.com/brian-team/brian2/issues>. Please include this file with debug information in your report: /tmp/brian_debug_yneum6z_.log  Additionally, you can also include a copy of the script that was run, available at: /tmp/brian_script_a0666428.py Thanks! [brian2]
  File "/home/marcel/.config/JetBrains/PyCharm2020.1/scratches/scratch_10.py", line 4, in <module>
    run(100*ms)
brian2.core.base.BrianObjectException: Error encountered with object named "neurongroup".
  File "/home/marcel/.config/JetBrains/PyCharm2020.1/scratches/scratch_10.py", line 2, in <module>
    group = NeuronGroup(1, 'dv/dt = -v/tau : volt')

An error occurred when preparing an object.

This was directly caused by the following exception:

brian2.units.fundamentalunits.DimensionMismatchError: Inconsistent units in differential equation defining variable v:
Expression -v/tau does not have the expected unit metre ** 2 * kilogram * second ** -4 * amp ** -1 (unit is V).
```

This only keeps the first line of the traceback for the `BrianObjectException` (because this is the line in the user code that actually triggered the exception, the rest is rather boring), and does not show at all the traceback for the  underlying exceptions (because these are all within Brian code). It also only shows the last of the underlying exceptions, since it typically contains all the information (but we don't have many situations where exceptions get chained in the first place).

I think for most users, this shortened exception would be much more useful than what we usually display. For user-reported exceptions that are due to Brian bugs it makes things potentially a bit more difficult, but the full exception info is still in the log file which therefore becomes more relevant.

mstimberg:
I have now implemented this and made it configurable. I quite like it!
The idea is that by default it removes most of the exception chaining and traceback. It now should only show traceback lines that are of direct relevance to the user (lines in user code triggering an exception), and no internal Brian code lines.

I am sure some people are going to hate that we are messing with Python internals here, but I think this could be a real improvement for the average users who are having a hard time to parse the full traceback output. Of course, this is configurable so you can always go back to the full output if you want. As a side effect, this makes the log file more useful, since if you are interested in the full output (e.g. for our debugging), this is where you find it.

This obviously needs more testing to make sure we are not hiding relevant information (also I did not yet test this across ipython, jupyter notebook, etc.).

thesamovar:
I wonder if we should consult with the community on this one? It feels like a major change in some ways.

Does this have an effect only on Brian code or does it call some sort of hook so that whenever Brian is imported (even if not used) it modifies Python behaviour?

Does the configurability let you turn it off completely?

mstimberg:
> Does this have an effect only on Brian code or does it call some sort of hook so that whenever Brian is imported (even if not used) it modifies Python behaviour?

This is a good point. It currently affects everything, but we should probably at least restrict it to our own exception types (`DimensionMismatchError`, `EquationError`, ...). This has always been the case for our "Brian has encountered an unexpected exception" message, though. As soon as you `import brian2`, it will be active.

> Does the configurability let you turn it off completely?

Kind off, I mean it still goes through our logging infrastructure but then it tries does only the default operations and calls the default exception hook. All this shows that Brian's hybrid status as library and "scripting environment" is a bit shaky. If we consider it a library, then we are doing it all wrong: we shouldn't add any handlers to the logging facility, we should not add an except hook, etc. But then, our approach has always been that all that the user needs to do is put an `from brian2 import *` in the beginning (with all the ugliness of the numpy/pylab import). We could revisit this approach (not for the import, I think it's too late for that), by requiring the user to write something like
```Python
from brian2 import *
activate_brian_logging()
```
which would switch on the log files and our custom exception handling. I think many more advanced users would appreciate this change, but it would also mean that most beginners would never benefit from our improvements since they do not switch it on.  But potentially we could deal with this by extensive advertising (e.g. putting the line into every example). There could also be a preference that would lead to switching it on automatically on import.

thesamovar:
OK so how about we restrict this to our own exception types and allow a config file way of switching it off (with it on by default)? And the same config option to switch off the existing except hook.

thesamovar:
I would deal with it mostly by config file rather than by putting that function call in every example! I guess 90% of users will just leave it on, but it's good to have a way to turn it off for expert users.

mstimberg:
I have thought about this a bit more, but I am still not sure what the best approach is given user's expectations and Brian's history. What I am seeing as more and more problematic is that we add logging and an expecthook *by default*. This is clearly wrong and against all best practices if we consider `brian2` a library. It does not really help that you can (or will be able to) switch it off via a configuration file. From the "best practice" point of view it would be best to turn the logic around, i.e. that you have to switch on the feature, but then most users would never notice it (for the logging, I'm not sure that would be a big loss, to be honest).

I wonder if we can find a pragmatic middleground solution where:
1. You can switch off these features on/off with a preference which forces them to be on or off, but
2. By default they are on "automatic", which will only switch them on when the user imports brian2 via  `from brian2 import *`

The reasoning behind 2 is that packages that use `brian2` as a library (say, https://github.com/VisualNeuroscience-UH/CxSystem2) will probably not use a wildcard import, while most "naive" users who would benefit from e.g. simplified exception tracebacks probably do.

What do you think about that approach? Too magic and surprising from the user's point of view?

I also wonder whether we should not get rid of our biggest historical sin at some point and at least not make `from brian2 import *` execute `from pylab import *`, but that discussion is for another time :wink:

thesamovar:
I like this solution actually, it feels like a good compromise. Can we tell if Brian is being wildcard imported? Doesn't the same code get executed if you write ``from brian2 import *`` and ``from brian2 import NeuronGroup`` or just ``import brian2 as br2``?

I think the ``pylab`` one is very tricky because it would break a lot of code! I kind of agree that it might be better if we hadn't done that, but at this point I think it will just create too many headaches.

mstimberg:
> I like this solution actually, it feels like a good compromise. Can we tell if Brian is being wildcard imported? Doesn't the same code get executed if you write `from brian2 import *` and `from brian2 import NeuronGroup` or just `import brian2 as br2`?

This certainly needs more testing, but I think it works by creating a new `_wildcard.py` module and including it in `__all__`. This module will be executed with `from brian2 import *`, but not in any other case (except if someone does `from brian2 import _wildcard`, but well).

thesamovar:
Huh - I learn a new Python thing. Can't find any docs on this.

mstimberg:
I think roughly speaking, `from brian2 import *` is equivalent to 
```python
for x in brian2.__all__:
   import x
```
so the approach should work.

thesamovar:
Well I guess it should be easy to do a quick test to check.

mstimberg:
I had another look at this. My proposed `__all__` approach does work in principle, but it would mean that we define a proper `__all__` in `brian2/__init__.py` with *everything* that should get imported... For the Brian part that's fine, since we already define a `__all__` in `brian2/only.py` and we could simply reuse it. But it would mean that we write an `__all__`  for everything that we import by default (numpy, matplotlib, ...), and that is not maintainable in the long term.

So this means we are back at square one... I'd still like to have a way for libraries to switch off Brian's logging/excepthook facilities completely, but there are drawbacks to every solution. I see two options:
1. Use an approach similar to Brian1's approach to switch off units, i.e. a separate package that can be imported as a marker before the actual Brian2 import (something like `import brian2_no_logging`).
2. A simple function to switch off the features (`import brian2 as b2; b2.disable_logging()`).

Both approaches have the disadvantage that they somewhat rely on the import order, which is never great. E.g. if a library builds upon Brian and uses one of the mechanisms and then gets used in a script that also imports `brian2`, then the order of imports decides whether Brian's custom logging is used.

thesamovar:
I think both the global approaches have the big problem of unexpected interactions, e.g. if you import a library that switches off the logging it will screw up your expectations. I suppose we could potentially deal with that by requesting libraries use a context manager that temporarily disables it, but that feels really awkward.

Could we have a keyword added to ``BrianObject`` that disables the logging stuff for that object? And maybe a context manager that temporarily sets the default of that keyword for the duration of that context?

How about this insane hack?

We include in ``from brian2 import *`` some oddly named variable like ``_this_is_a_crazy_brian_hack`` and take decisions based on whether or not this variable name is found in any of the calling frames?

Separate question:

Am I right that a library couldn't avoid whatever we do in ``brian2.__init__`` by just calling ``import brian2.only as br2``?

mstimberg:
> We include in `from brian2 import *` some oddly named variable like `_this_is_a_crazy_brian_hack` and take decisions based on whether or not this variable name is found in any of the calling frames?

Oooh, I see... I guess for an exception this could work, and this is a rare event so going through the stack searching for something shouldn't be relevant from the performance point of view. Actually we only need to do it once I guess.
For switching on/off all logging this is a bit trickier (and there are also some more subtle issues, e.g. if we switch off our logging you won't see INFO messages for the chosen integration method).

> Am I right that a library couldn't avoid whatever we do in `brian2.__init__` by just calling `import brian2.only as br2`?

No, unfortunately not. Importing `brian2.only` necessarily goes through `brian2.__init__.py`.

The only "clear" solution I'm seeing would be to have two user-facing packages: `brian2` (everything as it is now), `brian2_only` or maybe `brian2_lib` or something like that, and both would then import from `_brian2` where all the actual code is. We would deprecate `brian2.only` and direct users to `brian2_lib` instead. Maybe this is actually not too bad? A lot of potential for things to go wrong, though, and not really making development more straightforward :grimacing:

thesamovar:
Another idea. Most of the errors where we use ``BrianObjectException`` happen as part of a ``run()`` command, right? There's no particular need for us to raise one otherwise because the main point is to show you which object caused the problem. So, how about we make it so that we only raise ``BrianObjectException`` if you're running via a magic network? We could then tell people to explicitly construct Network objects if they want to avoid the issue. Could that resolve the problem?

I kind of like the multiple package thing, but doesn't it have the same problem that it's about global state again? If a user imports Brian and then imports the library, there will still be problems.

mstimberg:
> Another idea. Most of the errors where we use `BrianObjectException` happen as part of a `run()` command, right? There's no particular need for us to raise one otherwise because the main point is to show you which object caused the problem. So, how about we make it so that we only raise `BrianObjectException` if you're running via a magic network? We could then tell people to explicitly construct Network objects if they want to avoid the issue. Could that resolve the problem?

For the exceptions (maybe we leave the larger logging issue for later), this could indeed work. They are all raised as part of `run` (or mostly `before_run`), i.e. from the network. We could still make it configurable with on/off/automatic settings, where automatic is the default and switches it on for magic networks and off for Network objects.  This behaviour will certainly be surprising if you happen to switch between the two approaches in a script, but we have to make a compromise somewhere.



> I kind of like the multiple package thing, but doesn't it have the same problem that it's about global state again? If a user imports Brian and then imports the library, there will still be problems.

I think it is subtly different. The whole problem is that for a nice solution we'd need some explicit call for "convenience stuff" that is independent from importing Brian, and given that we don't want to have most users add such a call, my previous proposals only considered the opposite ("switch off convenience stuff"). But if we separate the packages, we'd now have a specific marker for "i want convenience" – importing `brian2`. We could therefore have `import brian2` switch on the convenience-mode (logging + exceptions), regardless of whether `_brian2` has already been imported by `brian2_lib`, and importing `brian2_lib` would not switch off the mode (`brian2_lib` would be kind of like `brian2.only` without going through `brian2`).

mstimberg:
I am starting to like the separate package approach :thinking: 

I created a prototype repository here. It only includes Brian's unit system, the logger, and the preferences. The adjustments I had to do were minimal (replace `brian2` by `_brian2` in a few places), but it seems to work as intended. If you run `script_convenience.py` (which imports `brian2`), you will get our "Brian has encountered an unexpected error" message. If you run `script_advanced.py` (which imports `brian2lib`), you get the vanilla Python error message. I just realized we don't need a separate `brian2lib` package actually, this could be the same as `_brian2`.

mstimberg:
...where "here" was supposed to link to: https://github.com/mstimberg/TEST_restructured_brian2

thesamovar:
I haven't had a chance to play around with the repo you shared yet, but my feeling is more and more that I love the idea of having a ``brian2lib`` package that advanced users and developers could use, and a ``brian2`` package that would be more what we'd expect from standard users. And, as long as the only thing that changes is the way exceptions are displayed and what goes in the log, then I think we're good. ``brian2lib`` would be minimal, and wouldn't import ``matplotlib``, ``numpy``, etc. I don't really see a downside UNLESS there are people relying on current behaviour, i.e. that you would expect ``NeuronGroup`` to be defined in ``brian2`` not ``brian2lib``.

mstimberg:
I'm starting to really like the idea, too. This seems like a much better solution that the `brian2.only` clutch  – if only I've though of that approach back in the days... 


> . I don't really see a downside UNLESS there are people relying on current behaviour, i.e. that you would expect `NeuronGroup` to be defined in `brian2` not `brian2lib`.

Yes, we need to look carefully at things like that. With the approach in my toy repository, you could still use either `from brian2lib import NeuronGroup` or `from brian2 import NeuronGroup`. But if you used something like `from brian2.groups.neurongroup import NeuronGroup`, this would fail.

mstimberg:
As I mentioned before, I think the best approach is to first merge this PR (which does not do any kind of restructuring) and do a release rather soon with all the other bug fixes that piled up.

I opened a new issue at #1217 to discuss the restructuring. We need to be very careful about doing the transition, so this is not something that we are going to rush.

thesamovar:
Agreed. Can you remind me quickly what is changed in this before I review?

mstimberg:
The main change is that it now correctly uses exception chaining, which avoids a weird duplication in the stack trace on Python 3 (see the before/after example in the first post: https://github.com/brian-team/brian2/pull/1196#issue-433788752). This unfortunately makes some of the tests that test for exception types slightly more complex, but the necessary change is quite straightforward (again, example in the first post: https://github.com/brian-team/brian2/pull/1196#issue-433788752).

There are two more changes apart from documentation:
1. before, there was a minor difference in how numpy and Cython handled errors during a run (e.g. because of a synapse creation statement that created illegal synapses): numpy wrapped the error in a  `BrianObjectException` whereas Cython did not. Now both do.
2. There is a new preference to make it possible to switch off the "Brian encountered an unexpected error" message.

The change in 2 is not really related to `BrianObjectException`, but I don't see any downside to it and it has to do with the error display in the larger sense.

thesamovar:
OK I think I like the new system better, but does it mean also that

```
try:
    some brian code
except DimensionMismatchError:
   handle it
```

will no longer work? I guess that's a bit of a shame, but probably worth it.

brian2/core/base.py
```diff
@@ -308,26 +304,20 @@ class BrianObjectException(Exception):
...
-                'Error encountered with object named "{objname}".\n'
+        return ('Error encountered with object named "{objname}".\n'
                 '{objcreate}\n\n'
-                '{message} {origexc}'
+                '{message}'
                 '(See above for original error message and traceback.)'
-                ).format(origtb=self._brian_origtb,
-                         origexc=self._brian_origexc,
-                         objname=self._brian_objname, message=self._brian_message,
+                ).format(objname=self._brian_objname, message=self._brian_message,
```
thesamovar, line 320: The spacing is not handled correctly in here, and leads to ``An error occurred when preparing an object.(See above for original error message and traceback.)``. Maybe just add a space before the ``(See above for``?

brian2/core/base.py
```diff
@@ -341,16 +331,6 @@ def brian_object_exception(message, brianobj, original_exception):
...
-        if attribute.startswith('_'):
-            continue
-        try:
-            setattr(new_exception, attribute, getattr(original_exception, attribute))
-        except AttributeError:  # some attributes cannot be set
-            pass
-    return new_exception
+
+    raise NotImplementedError('The brian_object_exception function is no longer used. '
+                              'Raise a BrianObjectException directly.')
```
thesamovar, line 336: This is just to remind ourselves not to do this in future, right?

brian2/tests/test_stateupdaters.py
```diff
@@ -754,7 +766,7 @@ def test_check_for_invalid_values_linear_integrator():
                 assert 'invalid_values' in repr(clog)
             else:
                 assert G.x[0] != 0
-        except UnsupportedEquationsException:
+        except BrianObjectException:
```
thesamovar, line 769: Should this be made more specific or unnecessary do you think?

thesamovar (approved):
OK, happy for this to be merged, but see a couple of comments above.

mstimberg:
> OK I think I like the new system better, but does it mean also that
> 
> ```
> try:
>     some brian code
> except DimensionMismatchError:
>    handle it
> ```
> 
> will no longer work? I guess that's a bit of a shame, but probably worth it.

It depends, code like
```Python
try:
   x +=  y
except DimensionMismatchError:
    # do something
```
will still work, but basically everything that is raised on a call to run is a `BrianObjectException` now and needs to be caught as such.

brian2/core/base.py
mstimberg: Yea, I wasn't sure about that. It's for the odd chance that a user (or a creator of a library) used `brian_object_exception` somewhere.

brian2/tests/test_stateupdaters.py
mstimberg: I added an assertion which checks the "cause" of the ecxeption.

thesamovar:
It's ever so slightly a shame, but as you say it's not really very safe to use exceptions for program logic anyway, so not the worst thing in the world.

thesamovar:
So, I think this is ready to merge and we can do a new minor release, right? Or do you think this needs a bigger release number?

mstimberg:
> It's ever so slightly a shame, but as you say it's not really very safe to use exceptions for program logic anyway, so not the worst thing in the world.

Yes, I agree. This is also why I put this point into the "Compatibility and reproducibility" notes. 

> So, I think this is ready to merge and we can do a new minor release, right? Or do you think this needs a bigger release number?

Yes, I'll prepare the release notes and have a look at two final issues where I'm not sure whether they are easy to fix or not. The release contains mostly bugfixes but there are some minor new features (e.g. new preferences), so it should probably become the 2.4 release. But I'd be fine with 2.3.1 as well if you prefer.

PR #1197, opened 2020-06-15
Make GSL tests optional

mstimberg:
Users running `brian2.test()` will see many errors if the GSL is not correctly installed. It gets automatically installed with the conda package, but if you install via pip it is up to you to install it. While the error message clearly states that you need to install the GSL, it's not great for users to see tens of tests failing for an optional dependency that only a few of them will ever use.

With this PR, all GSL tests are marked with a pytest `gsl` attribute and excluded if you run `brian2.test()`, you'll have to explicitly ask for them with `brian2.test(test_GSL=True)`. Nothing changes on the test servers, we simply switch on the `test_GSL` option in the `run_test_suite.py` script that we use there.

### DIFF
--- a/brian2/tests/__init__.py
+++ b/brian2/tests/__init__.py
@@ -79,14 +79,20 @@ def clear_caches():
     make_statements._cache.clear()
 
 
-def make_argv(dirnames, markers=None, doctests=False):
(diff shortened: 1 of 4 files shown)


mstimberg:
[Note that there's an unrelated build failure for conda packages which is already fixed in `master`]

PR #1210, opened 2020-08-02
Make RNG seed test standalone compatible

denisalevi:
Hey there,

I just adapted one of the brian2 tests to make it standalone compatible. I don't think this changes anything for what is tested here. In general, I sometimes come across a brian2 test that could be made standalone compatible. In the past I just added those tests to the brian2cuda tests. But I could also just add them to brian2 as here, would you want that?

And I am currently adding more tests for setting the RNG seed. In brian2cuda the random number generation is quite different depending on where it is used (synapse connection conditions use host side RNG, stateupdaters etc. use a device side random number buffer and binomial uses on the fly device side RNG). Do you want such tests in the brian2 test suite or should I keep them in brian2cuda? I guess brian2 uses the same RNG mechanism everywhere? I can post the tests I added here once they are pushed.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -98,9 +98,9 @@ install:
        conda install --yes --quiet anaconda-client conda-build jinja2;
     fi
   - if [[ $MINIMAL_VERSIONS == 'yes' ]]; then
-    PYTHON_PACKAGES="python=$PYTHON numpy==1.15 scipy==0.16 libgfortran==1 pytest "pytest-xdist<2" sphinx==1.8 ipython sympy==1.2 jinja2==2.7 pyparsing setuptools coverage<5 gsl>1.15 cython";
+    PYTHON_PACKAGES="python=$PYTHON numpy==1.15 scipy==0.16 libgfortran==1 pytest pytest-xdist<2 sphinx==1.8 ipython sympy==1.2 jinja2==2.7 pyparsing setuptools coverage<5 gsl>1.15 cython";
     else
-    PYTHON_PACKAGES="python=$PYTHON numpy pytest "pytest-xdist<2" sphinx>=1.8 ipython sympy>=1.2 pyparsing jinja2 setuptools coverage<5 gsl>1.15 cython";
+    PYTHON_PACKAGES="python=$PYTHON numpy pytest pytest-xdist<2 sphinx>=1.8 ipython sympy>=1.2 pyparsing jinja2 setuptools coverage<5 gsl>1.15 cython";
     fi
   - PYTHON_PACKAGES="$PYTHON_PACKAGES scipy";
   - if [[ $REPORT_COVERAGE == 'yes' ]]; then PYTHON_PACKAGES="$PYTHON_PACKAGES pytest-cov coveralls"; fi
--- a/.travis.yml
+++ b/.travis.yml
@@ -98,9 +98,9 @@ install:
        conda install --yes --quiet anaconda-client conda-build jinja2;
     fi
   - if [[ $MINIMAL_VERSIONS == 'yes' ]]; then
-    PYTHON_PACKAGES="python=$PYTHON numpy==1.15 scipy==0.16 libgfortran==1 pytest pytest-xdist sphinx==1.8 ipython sympy==1.2 jinja2==2.7 pyparsing setuptools coverage<5 gsl>1.15 cython";
+    PYTHON_PACKAGES="python=$PYTHON numpy==1.15 scipy==0.16 libgfortran==1 pytest "pytest-xdist<2" sphinx==1.8 ipython sympy==1.2 jinja2==2.7 pyparsing setuptools coverage<5 gsl>1.15 cython";
     else
-    PYTHON_PACKAGES="python=$PYTHON numpy pytest pytest-xdist sphinx>=1.8 ipython sympy>=1.2 pyparsing jinja2 setuptools coverage<5 gsl>1.15 cython";
+    PYTHON_PACKAGES="python=$PYTHON numpy pytest "pytest-xdist<2" sphinx>=1.8 ipython sympy>=1.2 pyparsing jinja2 setuptools coverage<5 gsl>1.15 cython";
     fi
   - PYTHON_PACKAGES="$PYTHON_PACKAGES scipy";
   - if [[ $REPORT_COVERAGE == 'yes' ]]; then PYTHON_PACKAGES="$PYTHON_PACKAGES pytest-cov coveralls"; fi
--- a/brian2/tests/test_neurongroup.py
+++ b/brian2/tests/test_neurongroup.py
@@ -204,9 +204,9 @@ def test_variableview_inplace_calculations():
 
     # Floor division
     G.x //= 2
-    G.y //= 0.001
+    G.y //= 0.01
     assert_allclose(G.x[:], x_vals // 2)
-    assert_allclose(G.y[:], y_vals // 0.001)
+    assert_allclose(G.y[:], y_vals // 0.01)
     with pytest.raises(DimensionMismatchError):
         G.x //= 2*mV
     with pytest.raises(DimensionMismatchError):
@@ -1628,6 +1628,7 @@ def test_random_values_fixed_seed():
 @pytest.mark.standalone_compatible
 @pytest.mark.multiple_runs
 def test_random_values_fixed_and_random():
+    prefs.core.default_float_dtype = np.float32
     G = NeuronGroup(10, 'dv/dt = -v/(10*ms) + 0.1*xi/sqrt(ms) : 1')
     mon = StateMonitor(G, 'v', record=True)
 
--- a/brian2/tests/test_neurongroup.py
+++ b/brian2/tests/test_neurongroup.py
@@ -6,14 +6,14 @@ import uuid
 import numpy as np
 import pytest
 import sympy
-from numpy.testing import assert_equal
+from numpy.testing import assert_equal, assert_raises
 
 from brian2.core.clocks import defaultclock
 from brian2.core.magic import run
 from brian2.core.network import Network
 from brian2.core.preferences import prefs
 from brian2.core.variables import linked_var
-from brian2.devices.device import seed
+from brian2.devices.device import seed, device
 from brian2.equations.equations import Equations
 from brian2.groups.group import get_dtype
 from brian2.groups.neurongroup import NeuronGroup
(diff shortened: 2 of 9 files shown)


mstimberg:
Hi Denis, thanks for the test! It does seem to be failing currently for the `seed()` part, I'll have a look into that. In general we'd be very happy to make more tests "standalone-compatible", so feel free to open PRs for that.

In Brian 

> And I am currently adding more tests for setting the RNG seed. In brian2cuda the random number generation is quite different depending on where it is used (synapse connection conditions use host side RNG, stateupdaters etc. use a device side random number buffer and binomial uses on the fly device side RNG). Do you want such tests in the brian2 test suite or should I keep them in brian2cuda? I guess brian2 uses the same RNG mechanism everywhere?

Yes, Brian 2 uses the same RNG everywhere, but I'd be happy to include such tests in the Brian test suite. How do you handle seeds in brian2cuda if you have multiple RNGs?

mstimberg:
It turned out that the failing tests were for `float32` and due to a bug in our `assert_allclose` method (which your test uses for the first time in an inverted fashion, assuming that it *fails*). We had the issue that we needed to use a different error tolerance for `float64` vs. `float32` and solved this by using a tolerance value that gets multiplied with the machine epsilon. The default tolerance value is 4.5e8, but we use values as high as 1e14 in some places. With `float64`'s machine epsilon of ~1e-16 this leads to `rtol` values of 1e-7 for the default, and 2e-2 for the 1e14 value. I only now realized (:facepalm:) that it does not make any sense to multiply these values with `float32`'s epsilon value which is ~1e-7 which leads to `rtol` values much bigger than one. This is why your new test failed here, even if the random numbers are different, the values are still close enough for the crazily high `rtol` value...

I changed the approach in a way that leaves the `rtol` values for `float64` unchanged, and tried some kind of "logarithmic" approach. Not sure it makes any numerical sense, but it should work for now and is definitely better than what we had before. The approach is: if `rtol` is provided as 1e8, this means that it about halves the number of digits compared to the machine epsilon (final `rtol` is ~1e-8 for `float64` with a machine epsilon of ~1e-16). So we do the same for `float32` and use about half the digits of 1e-7, turning the adjusted `rtol` into something between 1e-3 and 1e-4. The "worst" value for a specified `rtol` of 1e14 is then  ~0.2, which still seems to high but is not completely meaningless either. @thesamovar Any thoughts on this?

denisalevi:
Hi Marcesl,

thanks for looking into the failing test! I saw that some checks failed but I couldn't directly spot wether it was an issue with CI or my changes :P 

> Yes, Brian 2 uses the same RNG everywhere, but I'd be happy to include such tests in the Brian test suite. How do you handle seeds in brian2cuda if you have multiple RNGs?

I've been spending quite some time to get the brian seed machenism working with the brian2cuda RNGs. When you use `brian2.seed()`, all seeds are set for all RNG mechanisms, such that results should be reproducable with the same seed set in brian2. But there is still something failing where I couldn't figure out yet why the random numbers are not the same. Secretly hoping it has to do with the tolerance issues you described above 🤞. I will look into it next week.

mstimberg:
> I've been spending quite some time to get the brian seed machenism working with the brian2cuda RNGs. When you use `brian2.seed()`, all seeds are set for all RNG mechanisms, such that results should be reproducable with the same seed set in brian2.

Will it use the same seed for all RNGs, or some kind of derived seed that is different for all of them? It would be reproducible in both cases, but it would be kind of odd if the random numbers generated in different places were exactly identical :upside_down_face:

denisalevi:
> Will it use the same seed for all RNGs, or some kind of derived seed that is different for all of them? It would be reproducible in both cases, but it would be kind of odd if the random numbers generated in different places were exactly identical :upside_down_face: 

Yeah, good point! There are actually only two seeds being set, one for cuRAND host side API calls and one for cuRAND device side API calls. The first is set to the seed passed as argument to `brian2.seed()`. The latter is set to that seed `+1`. I did see something similar in brian2 at some point and I tried to find out if this could end up generating not so random number. Any thoughts on that? Alternatively, I could just set the cpp seed and generate two random numbers using cpp `rand()` as seeds for cuRAND.

In general, I do have a little bit of a uneasy feeling about all this RNG since it is not really tested anywhere. I was thinking to add some tests that check the statistics of random numbers. At least for brian2CUDA this might make sense since I am using parallelized device side random number generation. I did follow the cuRAND documentation, using e.g. the same seed for each parallel thread with a different RNG sequence offset. But if there is somewhere a bug that results in e.g. same sequence offset for a few threads, we would not catch that anywhere in our tests.

mstimberg:
The whole seed issue is quite tricky. We are using the + thread_ID approach for the seeds in the OpenMP parallelization, but this is not really the correct way to do it. I think for a few threads it is still fine, but I wouldn't use this approach for parallelization on a GPU. But cuRAND is using RNGs adapted to that task (and not Mersenne-Twister), so if you follow their recommendations all should be fine.

The testing issue is difficult: in general testing anything that is random is tricky, since normally we want tests to be deterministic. Our current tests are very basic, e.g. testing that numbers that are supposed to be random are not all the same (even though there is a tiny chance that this could actually happen). But for your concerns we'd actually have to have much more complex tests, e.g. we'd like to catch if two random number streams are just offset versions of each others. I'd be happy to add more tests in that direction, but I'm not sure whether putting a lot of effort into this is worth it.

mstimberg:
Apart from a number of time outs, all tests now seem to pass. I think we can therefore merge this. @thesamovar what do you think of the approach described in https://github.com/brian-team/brian2/pull/1210#issuecomment-672143822 ?

denisalevi:
> The whole seed issue is quite tricky. We are using the + thread_ID approach for the seeds in the OpenMP parallelization, but this is not really the correct way to do it. I think for a few threads it is still fine, but I wouldn't use this approach for parallelization on a GPU. But cuRAND is using RNGs adapted to that task (and not Mersenne-Twister), so if you follow their recommendations all should be fine.

Yeah, I am using the cuRAND default generator (Now that I think about it, the default can change depending on the cuda version and possibly hardware. So there might be different generators on different installations. But since reproducibility is only guaranteed on a single machine, I guess that doesn't matter, right?).
cuRAND has also other generator options, including two Mersenne-Twister algorithm approaches (which limit the number of parallel threads if I remember correctly). The cuRAND generator type is exposed to brian2cuda via `brian2.prefs`. But I don't check whether the chosen generator type is compatible with the parallelization. Could probably easily be done: If you choose the Mersenn-Twister generator, we limit the number of threads. But I guess this is not relevant at all and if anybody ever wanted this, it is easy to implement.

> The testing issue is difficult: in general testing anything that is random is tricky, since normally we want tests to be deterministic. Our current tests are very basic, e.g. testing that numbers that are supposed to be random are not all the same (even though there is a tiny chance that this could actually happen). But for your concerns we'd actually have to have much more complex tests, e.g. we'd like to catch if two random number streams are just offset versions of each others. I'd be happy to add more tests in that direction, but I'm not sure whether putting a lot of effort into this is worth it.
Yes, that is why I haven't really done any testing of this sort yet. I will leave it for now I guess.

thesamovar:
@mstimberg I have to say I'm not sure I really understood it I'm afraid. Could you write down the maths of it in a comment?

mstimberg:
But writing the maths will make it look stupid :laughing: ! I'll do it in a sec... I'm really open to another suggestion, but I'd like something where 1) nothing changes for `float64` and 2) something that scales reasonably with the provided `rtol` argument is used for `float32`.

mstimberg:
Ok, here's the old algorithm:
CodeCogsEqn

The new algorithm is something like this:
CodeCogsEqn(2)

Where 𝜺 is the eps value for the respective dtype, 𝜺_64 is epsilon for float64, and `rtol` is the value provided by the user (defaults to 4.5e8).

thesamovar:
I don't get the maths - why these functions?

If I understand correctly, the issue is that if you use a multiple of machine epsilon as the relative tolerance, then if you set that number well for double it's poor for float and vice versa. I think that just means that the errors are not a function of the errors introduced by arithmetic operations, instead the errors are introduced by the randomness and have size independent of machine precision. So we should just a fixed constant for the tolerance for both 32 and 64 bits, no? Unless we're modelling the errors as the sum of arithmetic precision errors and randomness, but I assume that randomness leads to much bigger differences, so a simple fixed value seems the right approach here.

Am I missing something?

mstimberg:
Sorry, I have to run, but just quickly: you are talking about the specific RNG test now? I was talking about a more general issue where multiplying the machine epsilon by a fixed value leads to meaningless high values (>>1) for `float32`.

But I consider getting rid of all this, and simply not test such examples for `float32`. Simulating `SpatialNeuron` with 32 bit is just too inaccurate.

thesamovar:
Well generally, if the errors appear because of arithmetic operations, then it makes sense to have a multiple of machine epsilon, but if they appear because of randomness it makes sense to have a fixed value that doesn't depend on machine epsilon. In the case where you have a multiple of machine epsilon it should be the appropriate machine epsilon for float or double values of course. If we're using a constant multiple of epsilon that doesn't work for both, then it's probably because the errors aren't coming from arithmetic, right? So we should use a constant that doesn't depend on epsilon at all.

mstimberg:
What I was worried of is that the machine epsilon is the error for a single arithmetic operation, but there is no guarantee that these errors simply add up (in particular with things like thresholds). But I realized that this is probably not the issue here. It seems that least for one of the tests where we needed a very high tolerance for float64 (which lead to a uselessly high tolerance for float32), you are right about the fact that the main error comes from somewhere else (it's a test where we compare to a theoretical result that assumes an infinitely long cable). So there it would make sense to have a big *absolute* tolerance which is independent of eps, and leave the relative tolerance as it is (i.e., multiplied by eps). Our old code multiplied the provided absolute tolerance by the machine epsilon as well which I think does not make sense (eps is about relative error), but this did not matter much since we almost always use an absolute tolerance of 0.

I'll have a closer look at the tests and see whether it makes sense to have a non-zero absolute tolerance for all of them.

thesamovar:
Right I agree with multiple operations things get complicated, although I think this is in general too hard a problem to come up with an answer to here (for example, at a point with sensitive dependence on initial conditions it's going to be hopeless). My guess would be that for some tests, it will be determined largely by arithmetic, e.g. exact integration of a stable differential equation. In those tests we should use a multiple of machine epsilon. For other tests, with randomness for example, it makes more sense to just have an absolute value. The right tolerance depends on the test.

mstimberg:
Some tests are timing out (which we should probably investigate at some point...), but otherwise everything passes now. @thesamovar What do you think, I guess this is ready to merge?

thesamovar:
Sorry a bit rushed to look at the code in detail right now. Did you implement absolute tol for tests where it makes sense and rel tol for the others as discussed? If so - yep, this can go ahead.

mstimberg:
Yes, that's basically it. The most fundamental change is that our `assert_allclose` function takes a `rtol` that will be multiplied with eps (as before), but `atol` will not be multiplied with it. Some tests use them both, i.e. `atol` for the part that is not about numerics, and `rtol` for the part that is. The only small problem I am seeing is that this does not deal well with situations where the desired result is 0, but it does not seem to be a problem in our test suite.

thesamovar:
Yeah from what I have read, you just need a different check to compare to 0.

mstimberg (approved):

PR #1215, opened 2020-08-26
Make TimedArray's implementations extensible

mstimberg:
This just rearranges the code so that other code generations can add implementations for `TimedArray` (same approach that we added for `BinomialFunction` to make it possible to support it in `brian2genn`).

See brian-team/brian2cuda#47


thesamovar (approved):
Seems uncontroversially good to me! :-)

PR #1220, opened 2020-08-27
Remove unused get_array_name argument

denisalevi:
Just a wrong function argument that had no effect other than confusing me for a bit.

### DIFF
--- a/brian2/codegen/generators/cpp_generator.py
+++ b/brian2/codegen/generators/cpp_generator.py
@@ -248,7 +248,7 @@ class CPPCodeGenerator(CodeGenerator):
             else:
                 line = ''
             line = line + self.c_data_type(var.dtype) + ' ' + varname + ' = '
-            line = line + self.get_array_name(var, self.variables) + '[' + index_var + '];'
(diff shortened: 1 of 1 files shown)


denisalevi:
Oh BTW, am I using up CI resources or such when making a tiny PR like this one? Or does that not matter? I guess for a trivial fix like this one not all tests need to run? :)

mstimberg:
Uh, that's an odd one, thanks. It accidentally works because the `self.variables` dictionary is not empty and evaluates to `True` :astonished: 

> Oh BTW, am I using up CI resources or such when making a tiny PR like this one? Or does that not matter? I guess for a trivial fix like this one not all tests need to run? :)

It does use up resources, but that's not a big deal. There's rarely an urgency to merge PRs right away and this change *could* potentially break something (as opposed to e.g. a change in the documentation), so I'd rather run the CI here.

denisalevi:
Yep, I guess if it didn't evaluate to True, it would have tipped of some tests. :)

mstimberg (approved):

Issue #1187, opened 2020-05-20
Allow cyclical subexpressions if marked "constant over dt"


mstimberg:
Brian does not allow subexpressions that contain cycles. However, in the case of subexpressions that are marked as "constant over dt" this is not really an issue since they are implemented as a combination of a state variable and a `run_regulary` operation updating the state variable. The `Equations` object does not make this distinction, though, and therefore raises an error. This should be quite straightforward to fix in `Equations`.

See for a use case of a recursive definition in [this thread](https://groups.google.com/d/msg/briansupport/eD-_Dl3XYDo/1gc9mCarAwAJ) on the mailing list. In principle, the "constant over dt" mechanism could also be used for difference equations and very abstract models ("x_new = x_old + 1"), but I don't find "constant over dt" an intuitive term to use for those. Not sure it is worth renaming the flag (or adding an alternative name) for this use case, though.

nishantraghuvanshi:
Hi @mstimberg 

I'm working on this issue. My approach is to exclude "constant over dt" subexpressions from the cycle detection in _sort_subexpressions(), since they're evaluated once per timestep (like parameters) and don't need strict acyclic dependencies.

Does this sound like the right direction?

PR #1223, opened 2020-08-28
Exclude "constant over dt" subexpressions from cycle checking

mstimberg:
See #1187

### DIFF
--- a/brian2/tests/test_equations.py
+++ b/brian2/tests/test_equations.py
@@ -486,6 +486,20 @@ def test_extract_subexpressions():
 
 
 @pytest.mark.codegen_independent
+def test_cyclical_subexpressions():
+    with pytest.raises(ValueError):
+        # dependency cycle
+        Equations('''dv/dt = (-v + s1)/ (10*ms) : 1
+                     s1 = 2 * s2 : 1
+                     s2 = s1/2 : 1''')
+
+    # With constant over dt, the cycle disappears
+    Equations('''dv/dt = (-v + s1)/ (10*ms) : 1
+                 s1 = 2 * s2 : 1
+                 s2 = s1/2 : 1 (constant over dt)''')
+
+
+@pytest.mark.codegen_independent
 def test_repeated_construction():
     eqs1 = Equations('dx/dt = x : 1')
     eqs2 = Equations('dx/dt = x : 1', x='y')
--- a/brian2/equations/equations.py
+++ b/brian2/equations/equations.py
@@ -901,14 +901,18 @@ class Equations(Hashable, Mapping):
         static_deps = {}
         for eq in self._equations.values():
             if eq.type == SUBEXPRESSION:
-                static_deps[eq.varname] = [dep for dep in eq.identifiers if
-                                           dep in self._equations and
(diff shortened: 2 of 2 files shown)


brian2/tests/test_equations.py
```diff
@@ -485,6 +485,20 @@ def test_extract_subexpressions():
...
+    with pytest.raises(ValueError):
+        # dependency cycle
+        Equations('''dv/dt = (-v + s1)/ (10*ms) : 1
+                     s1 = 2 * s2 : 1
+                     s2 = s1/2 : 1''')
+
+    # With constant over dt, the cycle disappears
+    Equations('''dv/dt = (-v + s1)/ (10*ms) : 1
+                 s1 = 2 * s2 : 1
+                 s2 = s1/2 : 1 (constant over dt)''')
```
thesamovar, line 499: This feels very unclear to me how to interpret. If s2 is constant over dt but s1 is not, but s2 depends on s1.... wha? Argh! Seriously though, I don't understand what this is supposed to do, and it makes me think we shouldn't be allowing it.

mstimberg: This is consistent with our use of "constant over dt" elsewhere and I don't see a reason for not allowing it. It's equivalent to
`s2 : 1`
together with
`...run_regularly('s2 = s1/2')`
i.e. it refers to the value of `s1` at the beginning of the time step.

thesamovar: Can you remind me what's the primary use case of constant over dt? This feels wrong to me, and I feel like the run_regularly solution would be better in most cases because it forces people to be explicit about what it means, but if the idea was always to freeze a value with its value at the start of the window (for approximations and numerical integration reasons) then it makes sense.

mstimberg: I think originally we introduced it for expressions involving things like `rand()`, but you can also use it e.g. to approximate a slow non-linearity as constant over a time step so that you can use exact integration. That it is called "constant over dt" is not great for all situations. But I prefer it over a generic "`variable : unit`" in the equations which needs the additional `run_regularly` operation to make sense.

thesamovar: I think I'd prefer a new syntax for that use case then, and have cycle checking on for constant over dt and off for the new syntax? maybe something like "force constant over dt"?

thesamovar: Sorry I'm being a bit annoying, but I'm just worried that we're storing up problems for ourselves here.

mstimberg: But does cycle checking for a situation where it is not a problem make sense? And wouldn't a new syntax make things even more complicated? Happy to postpone the decision on this PR for a bit. It's not a very important issue and if it involves changing syntax, we should make sure we get it right.

thesamovar: Yeah let's postpone for the moment.

How about this, we allow cycles but we raise a warning to explain to users what it means, only if they have a cycle? If they use ``(force constant over dt)`` we assume they know what they're doing and don't raise a warning. The ``force`` makes it clear that you're asking Brian to do something that isn't straightforward and taking responsibility for it.

PR #1225, opened 2020-08-28
Synapse information

mstimberg:
See my comment in this [briansupport thread](https://groups.google.com/g/briansupport/c/aJK29Kiloxw/m/T1eadGyFBQAJ). This adds support for:
* `len(S[0, :])` and similar expressions (before only e.g. `len(S[0, :].i)` worked)
* support for `N_outgoing_pre` and `N_incoming_post` as attributes of `Synapses`, which give access to the same information as `N_incoming` and `N_outgoing`, but on a per-neuron instead of on a per-synapse basis. The implementation is simple – in principle we could also use something fancier using a reference and indicies, but then the variables would be available to use in code where you cannot use them correctly (synaptic code will always loop over synapses, not over neurons)

### DIFF
--- a/brian2/synapses/synapses.py
+++ b/brian2/synapses/synapses.py
@@ -490,6 +490,8 @@ class SynapticSubgroup(object):
                                 'synaptic subgroup has been created'))
         return self._stored_indices
 
+    def __len__(self):
+        return len(self._stored_indices)
 
     def __repr__(self):
         return '<%s, storing %d indices of %s>' % (self.__class__.__name__,
--- a/brian2/synapses/synapses.py
+++ b/brian2/synapses/synapses.py
@@ -943,6 +943,13 @@ class Synapses(Group):
         # Activate name attribute access
         self._enable_group_attributes()
 
+    N_outgoing_pre = property(fget= lambda self: self.variables['N_outgoing'].get_value(),
+                              doc='The number of outgoing synapses for each neuron in the '
+                                  'pre-synaptic group.')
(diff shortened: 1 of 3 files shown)


thesamovar (approved):
Code is fine, but I find the documentation a little confusing. Is this just about how they are indexed or are they different numbers? It's not entirely clear.

mstimberg:
> Code is fine, but I find the documentation a little confusing. Is this just about how they are indexed or are they different numbers? It's not entirely clear.

I agree it is a bit confusing, but I am not quite sure what to do about it. It's the same numbers, but indexed differently.
Part of the problem is that normally, variables without `pre`/`post` suffix are variables that can be different for each synapse, and `_pre`/`_post` variables are variables of the neurons, but indexed differently. Following this logic, it would make sense to have `N_incoming`/`N_outgoing` variables *in the `NeuronGroup`*, and then `N_incoming_post` and `N_outgoing_pre` as references to them, indexed per synapse. But we cannot really have the `Synapses` object store something in the `NeuronGroup`s connected to it (and it would also be unclear what it means for several `Synapses` objects connected to the same `NeuronGroup`). Also, we'd break backwards compatibility if the current `N_incoming` and `N_outgoing` changed names.

mstimberg:
I added an example to the docs, hopefully that makes things a bit clearer...

mstimberg:
I'll merge this for now, we can always improve the documentation. Syntax-wise I don't think we have that many options for the names since we want to keep backwards compatibility.

thesamovar:
Example is good.

Issue #1008, opened 2018-09-14
SpatialNeuronGroup


psipeter:
Dear Brian developers,

My name is Peter, and I'm a PhD student at the University of Waterloo in Chris Eliasmith's lab. My work involves incorporating biologically-detailed neuron models into the Neural Engineering Framework (and its associated software suite nengo), building medium-sized networks from these models, and investigating various neurocognitive phenomenon. Up to this point I've used NEURON to model compartmental neurons, but I've become disenchanted with this language and I'm looking for an alternative approach.

I recently discovered that Brian2 supports compartmental modelling, and I am excited about interfacing your software with ours, with the goal of eventually building functional and biologically-realistic neural networks! I've already done some preliminary work and have successfully built nengo networks populated by NeuronGroups, as well as individual by SpatialNeurons. Now I'm at the point where I want to simulate groups of these SpatialNeurons. Previous posts have described how to extract the equations from a SpatialNeuron and instantiate a NeuronGroup with those equations, but this is somewhat laborious from the user's perspective, and may not take full advantage of optimization or multi-threading.

I'd like to rekindle the development of a SpatialNeuronGroup class and associated methods.  Are there members of the development team that would be willing to add support for SpatialNeuronGroups? It seems to me like this would be a powerful tool with numerous applications in computational neuroscientists, and it's development would be a natural extension of Brian's core goals.

Thanks for your consideration.

 - Peter

thesamovar:
I think this could potentially be a nice feature to have, but at least from my point of view not extremely high priority. So, I'd be happy for us to support you in doing this and potentially merge it into Brian once it's finished, but realistically I won't have a lot of time. Not sure what @mstimberg thinks.

The main challenge is that different SpatialNeurons may have different numbers and structure of compartments, so there is no way to vectorise the code across neurons. Essentially, a SpatialNeuronGroup would just have to be a wrapper around a list of SpatialNeurons. I wonder how helpful it is to add specific syntax in this case?

The other alternative is that oyu have a SpatialNeuronGroup where the compartmental structure is forced to be the same across neurons. This is certainly feasible and would lead to potential vectorisation benefits, but I'm not sure how useful that is to modellers?

These are the questions that have stopped us from making moves towards implementing a SpatialNeuronGroup in the past. What are your thoughts?

psipeter:
[...]

PR #1229, opened 2020-09-14
Non-contiguous subgroups

mstimberg:
Brian has always allowed subgroups, but these needed to use slicing syntax. While working on the `SpatialNeuronGroup` (see e.g. #1008) I found that quite limiting, e.g. it would not be possible to have a subgroup representing the somata of several neurons in the group. To a lesser extent, the same problem already exists for `SpatialNeuron`, e.g. you could not have a subgroup for all distal dendrite compartments.
In the current version, you can work around this limitation by using indices directly. For example, this would work:
```Python
distal = neuron.indices['distance>100*um']
neuron.g_na[distal] = ...
neuron.g_kd[distal] = ...
neuron.g_Ca[distal] = ...
```
This is ok, but it has the disadvantage of being quite redundant, and in general we try to avoid that the user needs to know about indices in the first place. With this PR, you can write instead:
```Python
distal = neuron['distance > 100*um']
distal.g_na = ...
distal.g_kd = ...
distal.g_Ca = ...
```
There's also an argument to make for consistency. Currently:
```pycon
>>> neurons = NeuronGroup(10, 'v : volt')
>>> neurons.v[:5] = -70*mV
>>> neurons[:5].v = -70*mV  # same effect
>>> neurons.v['i>=5'] = -60*mV
>>> neurons['i>=5'].v = -60*mV
...
TypeError: Subgroups can only be constructed using slicing syntax, a single index, or an array of contiguous indices.
```
With the new support, the last line would work as well.

Finally, another use case is to record a non-contiguous subset of neurons with a `SpikeMonitor`. For example to record a random subset of neurons in a non-random network (i.e. where you cannot simply take any subset).

Now, the main arguments against supporting this feature are performance and code complexity. It is true that for `SpikeMonitor` and `PopulationRateMonitor` the code is now a bit more complex to also support non-contiguous subgroups, but performance-wise it should still be ok by requiring that subgroup indices are ordered. Of course, if you want to record the first and the last neuron of a large group, this will be quite inefficient.

For now, you cannot create `Synapses` based on a non-contiguous subgroup, because this complicates the indexing quite a bit. It's perfectly doable, but I wanted to get the basics working first.

Oh, while thinking about the consistency of the different syntaxes of indexing things, I also made two diagrams (the same thing with and without units). I guess it can be useful to explain the difference between `neurons[:5].v` and `neurons.v[:5]`, but not sure whether it is too detailed for the documentation. Might also go into a blog post one day.

Here they are for reference: https://gist.github.com/mstimberg/25b5058bc85a1f1fbaff958741629ee8

### DIFF
--- a/brian2/codegen/runtime/cython_rt/templates/ratemonitor.pyx
+++ b/brian2/codegen/runtime/cython_rt/templates/ratemonitor.pyx
@@ -1,31 +1,50 @@
-{# USES_VARIABLES { N, t, rate, _clock_t, _clock_dt, _spikespace,
-                    _num_source_neurons, _source_start, _source_stop } #}
+{# USES_VARIABLES { N, t, rate, _clock_t, _clock_dt, _spikespace} #}
 {% extends 'common_group.pyx' %}
 
 {% block maincode %}
 
     cdef size_t _num_spikes = {{_spikespace}}[_num{{_spikespace}}-1]
-    
+    {% if subgroup and not contiguous %}
+    # We use the same data structure as for the eventspace to store the
+    # "filtered" events, i.e. the events that are indexed in the subgroup
+    cdef int[{{source_N}} + 1] _filtered_events
+    cdef size_t _source_index_counter = 0
+    _filtered_events[{{source_N}}] = 0
+    {% endif %}
+    {% if subgroup %}
     # For subgroups, we do not want to record all spikes
+    {% if contiguous %}
     # We assume that spikes are ordered
-    cdef int _start_idx = -1
-    cdef int _end_idx = -1
-    cdef size_t _j
+    _start_idx = _num_spikes
+    _end_idx = _num_spikes
     for _j in range(_num_spikes):
         _idx = {{_spikespace}}[_j]
         if _idx >= _source_start:
             _start_idx = _j
             break
-    if _start_idx == -1:
-        _start_idx = _num_spikes
-    for _j in range(_start_idx, _num_spikes):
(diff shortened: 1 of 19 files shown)


mstimberg:
@thesamovar Did you have a chance to look at this?

thesamovar:
No, will be too busy for a little while I'm afraid with neuromatch and teaching, and this feels like a substantial feature that will require some discussion.

mstimberg:
Ok, sure thing. Was just making sure that you had seen it.

thesamovar:
I haven't forgotten! 😉 But, this is a big one so I need to allocate some substantial time to it...

mstimberg:
Hi @thesamovar . I updated this PR with latest `master`, would be great if you could have a look at it. See my [initial post](https://github.com/brian-team/brian2/pull/1229/#issue-701179867) for more details about the reasoning. The main functionality is there, the reason why this is still marked as a draft is because it is lacking more tests and documentation before it can be merged.

thesamovar (approved):
OK I read the reasoning and a quick look at the code. I think realistically I'm not going to have time to go through the code in detail any time soon, so this might have to do.

I'm slightly reluctant because of potential inefficiencies. Could we potentially raise a warning if users try to record from such a thing that it might be inefficient? Not sure if that's a good idea or not.

Also, is there anything that users would reasonably expect to be able to do with this that wouldn't work either silently (big problem of course) or with an error (ok if it's clear, bad if it's unclear)?

Overall I'm going with approve because I think this is a feature that would be convenient in many circumstances.

PR #1235, opened 2020-09-21
Function unit check

### LINKED ISSUE #1234: Function clip does not check for units when used in code
If you call Brian's wrapper around the `clip` function directly in a Python script, it will check for units:
```pycon
>>> from brian2 import *
>>> clip(1*mV, 0, 1)
...
brian2.units.fundamentalunits.DimensionMismatchError: clip (units are V and 1).
```
but in code strings or equations it does not:
```pycon
>>> group = NeuronGroup(1, 'v : volt')
>>> group.v = 'clip(-70*mV, -0.2, 50*nS)'
>>> group.v
<neurongroup.v: array([-70.]) * mvolt>
```
It would raise an error if the first argument did not have the unit expected by `v`, since the "return unit" of `clip` is set to the unit of the first argument.

Currently, the `_arg_unit` argument of `clip` is `[None, None, None]` which means that any unit is accepted for each of the arguments.

Discovered by @kernfel over at the [Brian discourse forum](https://brian.discourse.group/t/stdp-weights-converging-to-zero/146/7).

mstimberg:
This fixes #1234. Functions now have three options to specify the arguments of their units.
1. State a specific unit that is expected (e.g. the `timestep` function needs units of time, and e.g. `exp` needs a dimensionless value)
2. State that any value is ok (e.g. for `sqrt`)
3. State that several arguments need to have the same units (e.g. for `clip`)

Number 3 is new, the syntax is to define an (arbitrary) string instead of the unit. For example:
```Python
@check_units(a='first', b='first', c='second', d='second')
def foo(a, b, c, d):
    return (a+b) * (c+d)
```
would require that both `a` and `b`, as well as `c` and `d` have the same units. So far, `clip` is the only function provided by Brian making use of this feature.


thesamovar:
I think the syntax needs work. At the moment, ``c='second'`` would likely make anyone think that c was in seconds, i.e. had units of time!

How about ``@check_units(same=(('a', 'b'), ('c', 'd')))``? If there were only one group you could do ``@check_units(same=('a', 'b'))``.

I'm not 100% happy with that either.

mstimberg:
Not quite sure whether this is really a problem... I agree that my example wasn't chosen very well, but you could write
```Python
@check_units(a='A', b='A', c='B', d='B')
```
instead. 
 I did not want to add new arguments to `check_units`, because each argument limits the names you can take for arguments, and it becomes quickly confusing what is an argument of the function and what an argument of `check_unit`.

Here's another proposal for the function thing as above:
```Python
@check_units(a='b', b='a', c='d', d='c')
```
We could later extend this to arbitrary expressions that refers to the dimensions of the other arguments. This would allow something like:
```Python
@check_units(a=None, b=None, c='a*b'):
def foo(a, b, c):
    return a*b + c
```
This could also replace the mechanism for the return argument where you'd currently have to specify a function.

thesamovar:
OK that makes sense. Actually I really like your suggestion of leaving it open for having functions of the arguments. How about just ``@check_units(b='a', d='c')``? There's something slightly uncomfortable about the circularity of ``@check_units(a='b', b='a', c='d', d='c')`` 😉.

mstimberg:
>  How about just `@check_units(b='a', d='c')`? There's something slightly uncomfortable about the circularity of `@check_units(a='b', b='a', c='d', d='c')` wink.

Right, this is indeed possible, it simply checks what you give. Specifying both directions is certainly redundant and does not help. I think writing `@check_units(a=None, b='a', c=None, d='c')` instead of `@check_units(b='a', d='c')` is a bit clearer, but both are interpreted exactly the same.

All this is now implemented in the latest version, let me know what you think.

thesamovar (approved):

PR #1242, opened 2020-10-06
Point unresolved error message to discourse forum [ci skip]

kernfel:
Given [@mstimberg's comment](https://groups.google.com/g/brian-development/c/N61MUSEfU64) on the brian-development list, I'm guessing error messages should also point to discourse.


kernfel:
Hm. Note to self, [ci skip] does not seem to work in the pull request title.

mstimberg (approved):

mstimberg:
Yes, definitely. Many thanks.

> Hm. Note to self, [ci skip] does not seem to work in the pull request title.

Indeed, it has to be in the commit message. I usually put it in its own line so it does not appear in short commit summaries.

PR #1248, opened 2020-10-26
Synapses is not resolving dt correctly.

jangmarker:
I've committed a test case that demonstrates the issue:
```python
source_dt = 0.1*ms
synapse_dt = 1*ms
source_target = NeuronGroup(1, 'v:1', dt=source_dt, threshold='False')
synapse = Synapses(source_target, source_target, 'dw/dt = 1/ms : 1 (clock-driven)', 
                   dt=synapse_dt, method='euler')
synapse.connect()
synapse.w = 0
run(1*ms)

assert synapse.w[0] == 1
```

The issue is that `synapse.w[0]` will be `0.1` because `Synapses` uses the `NeuronGroup`'s `dt` for scaling the constant `1` in the equation, but `Synapses` uses its own `dt` for the actual timestep.

I have found the source of the issue, the `SynapseGroup`'s `dt` variable is being prefixed with `_clock_`, why that is I am not sure of.
Removing the prefix fixes my issue and still all tests in `test_synapse.py` pass, but I do not understand the code well enough to be sure that this is the correct fix.

### DIFF
--- a/brian2/synapses/synapses.py
+++ b/brian2/synapses/synapses.py
@@ -1057,8 +1057,7 @@ class Synapses(Group):
         self.variables.add_dynamic_array('_synaptic_post', size=0,
                                          dtype=np.int32, constant=True,
                                          read_only=True)
-        self.variables.create_clock_variables(self._clock,
-                                              prefix='_clock_')
+        self.variables.create_clock_variables(self._clock)
         if '_offset' in self.target.variables:
             self.variables.add_reference('_target_offset', self.target,
                                          '_offset')
--- a/brian2/tests/test_synapses.py
+++ b/brian2/tests/test_synapses.py
@@ -1135,6 +1135,19 @@ def test_clocks():
     assert synapse.post._clock.dt == target_dt
     assert synapse._clock.dt == synapse_dt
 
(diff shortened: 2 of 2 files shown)


mstimberg:
Woah, that's a major bug that has been around for ... 6 years! I agree with the analysis and the fix. I have to say I don't remember why we prefixed the clock variables in that code. Normally that should have been necessary only for monitors, where we need to make a distinction between the current time `t` (then called `_clock_t`) and the array of recorded times `t`. It might have just been a copy&paste error, because I added a similar line in many places. The reason why this did not trigger any issues is that in the absence of a `dt` variable, synaptic code defaults to using `dt_post` (which of course has the same value in most simulations).
Thanks again, I'll merge the PR now.
If you need to make your current code work with the buggy version of Brian (until we make a new release), adding these two lines should work around the issue (and still work with the fix):
```Python
if '_clock_dt' in synapse.variables:
    synapse.variables._variables['dt'] = synapse.variables['_clock_dt']
```

mstimberg (approved):

jangmarker:
Hi Marcel, thanks for responding so quickly! I highly appreciate you posting the workaround :-)

PR #1251, opened 2020-11-24
Improve error message for multiple uses of `xi` and add link to docs

mstimberg:
As reported in the [discourse forum](https://brian.discourse.group/t/multiple-independent-sources-of-noise-in-single-neuron-model/212), the error message for multiple uses of `xi` is not very helpful, and the main documentation does not directly point to the solution either. This PR makes the error message quite verbose (but hopefully clear) and adds a link to the documentation pointing to the place where this issue is discussed.


mstimberg:
I'll go ahead with the merge without review, given that this is a small change that only affects documentation and error message.

PR #1254, opened 2020-11-27
Cleanup

mstimberg:
Not the most exciting PR, but there were a couple of unused variables/imports and things for compatibility with older Python versions that I finally wanted to get rid of.

### DIFF
--- a/brian2/codegen/translation.py
+++ b/brian2/codegen/translation.py
@@ -343,12 +343,9 @@ def make_statements(code, variables, dtype, optimise=True, blockname=''):
     # none are yet defined (or declared)
     subdefined = dict((name, None) for name in subexpressions)
     for line in lines:
-        stmt = line.statement
-        read = line.read
-        will_write = line.will_write
         # update/define all subexpressions needed by this statement
         for var in sorted_subexpr_vars:
-            if var not in read:
+            if var not in line.read:
                 continue
 
             subexpression = subexpressions[var]
@@ -362,7 +359,7 @@ def make_statements(code, variables, dtype, optimise=True, blockname=''):
                 op = ':='
                 # check if the referred variables ever change
                 ids = subexpression.identifiers
-                constant = all(v not in will_write for v in ids)
+                constant = all(v not in line.will_write for v in ids)
                 subdefined[var] = 'constant' if constant else 'variable'
 
             statement = Statement(var, op, subexpression.expr, comment='',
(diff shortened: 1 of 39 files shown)


brian2/codegen/translation.py
```diff
@@ -297,10 +297,6 @@ def make_statements(code, variables, dtype, optimise=True, blockname=''):
         elif not variables[stmt.var].scalar:
             scalar_write_done = True
 
-    # all variables which are written to at some point in the code block
-    # used to determine whether they should be const or not
-    all_write = set(line.write for line in lines)
-
```
thesamovar, line 303: Is there a case for leaving this here purely as a form of documentation?

brian2/codegen/translation.py
```diff
@@ -349,8 +345,6 @@ def make_statements(code, variables, dtype, optimise=True, blockname=''):
     for line in lines:
         stmt = line.statement
         read = line.read
-        write = line.write
-        will_read = line.will_read
         will_write = line.will_write
```
thesamovar, line 348: And same here.

brian2/spatialneuron/morphology.py
```diff
@@ -782,7 +790,8 @@ def start_z(self):
         '''
         return _to_meters(self.start_z_)
 
-    @abc.abstractproperty
+    @property
+    @abc.abstractmethod
```
thesamovar, line 794: What's the effect of this change? (Just so I can understand.)

thesamovar (approved):
Looks good to me, but a couple of minor points that are not essential.

brian2/codegen/translation.py
mstimberg: I'd find it a bit confusing, especially since it says that it uses this to determine the `const` modifier which is not true (it uses `will_write` for that).

thesamovar: Fair enough! :-)

brian2/codegen/translation.py
mstimberg: It's true that there was more consistency here before... I wonder whether it wouldn't be clearer after all to only keep `stmt = line.statement` (because this is used several times and would get unwieldy otherwise), and use `line.read` and `line.will_write` directly, since they are only used once or twice.

brian2/spatialneuron/morphology.py
mstimberg: There's no real effect. Before Python 3.3, you could not use `@abc.abstractmethod` with a property and you'd have to use `abc.abstractproperty` instead. Since Python 3.3, you can combine `abstractmethod` with `property`, and `abstractproperty` has therefore been deprecated. It is still present in Python 3.9, though.

If I understand correctly, there's actually one advantage over `abstractproperty`: you could make only either the property setter or getter abstract. But we are not using this.

brian2/codegen/translation.py
thesamovar: Maybe, although I don't have very strong feelings about it - so whatever you prefer!

mstimberg: I went ahead with the change. Not the most important change ever, though.

PR #1257, opened 2020-12-21
Fix C99 Windows check

kjohnsen:
I kept getting errors about C99 compiler support for trying different numerical integration methods and I tracked down the problem until I found this bug. The test on Windows for C99 support consists of trying to compile a test `cpp` file. The problem is, the file is created as a temp file, but the test on lines 251 and 254 refer to `mytest.cpp` instead of the tempfile.


kjohnsen:
Hold on, there's another problem: the bash syntax `> NUL 2>&1` seems to be causing problems since `cl` is trying to compile `>`, `NUL`, and `2>&1` as files to compile in addition to the test file.

mstimberg:
Many thanks for the report and the fix. This looks certainly reasonable, but I will do more testing next week/year. Most importantly to figure out while this has not been caught by our test suite (I don't have access to a Windows machine myself at the moment).

mstimberg:
Thanks again for the report and the fix, I am still a bit confused about how this bug made it into our actual code in the first place... I changed the test suite so that it now includes a basic test of the C99 compatibility check. On master this fails for Windows (as expected) whereas it passes on this PR. So all looks good, and I'm going ahead with the merge now.

PS: The failed tests here are unrelated to the issue, they are because our new testing framework fails to upload coverage data to coveralls.io for PRs based on forks. This is due to security restrictions, PRs from forks do not have access to the token we use to identify ourselves. I'll fix this in master right after the merge.

kjohnsen:
No problem. Thanks for testing it out.

PR #1258, opened 2021-01-06
Minor changes to the Refractoriness file in the User's guide

rohithvarma3000:
I have made small changes to the Refractoriness file in the User's guide.
I have changed the word 'changing' to 'varying' because I felt that it was more appropriate in the context.
I have also changed the 'to prevent to count' to 'to prevent the count of'.


mstimberg (approved):
The changes look very reasonable to me, thanks!

PR #1259, opened 2021-01-06
Use Azure pipelines for Linux + OS X as well

mstimberg:
With the transition of travis-ci.org to travis-ci.com, projects by default only have 10000 free "credits", which as it turns out will be used by a single run of our full test suite... I have therefore moved our Linux and OS X tests to Azure Pipelines which we were already using for the Windows tests. Getting this to work with the coverage measurements was a bit tricky, but it seems to work nicely now.


mstimberg:
I'll go ahead and merge this myself as this is a pure infrastructure change and relatively urgent in the sense that we don't have any Linux/OS X tests at the moment.

Issue #1102, opened 2019-06-27
Input() function overwritten


Marghepano:
When importing brian2 it is no longer possible to use the built-in function input(), to get values from the user. 
I.e.
![image](https://user-images.githubusercontent.com/52278396/60254145-ad04c780-98cd-11e9-874c-f9abc450b49a.png)
I would advise to rename the module 'input' (encompassing binomial, poissongroup, poissoninput etc.)

mstimberg:
Hi, thanks for the bug report. This is something that we should indeed fix, until then you can either use
```
from brian2.only import *
```
(see https://brian2.readthedocs.io/en/stable/user/import.html)
or explicitly import the input function again after importing Brian:
```
from brian2 import *
from builtins import input
```

Issue #1217, opened 2020-08-26
Restructure package to allow clean import in libraries


mstimberg:
See discussion in #1196 

In short, importing `brian2` currently activates the logging system and the except hook which is what we want for most users but not if `brian2` is used as part of another library. Putting Brian's code into a `brian2lib` package (that can be imported by libraries without triggering the logging system, etc.), and making `brian2` a module that imports everything from `brian2lib` (+ `numpy`, etc.), and activates the logging system would solve this problem.

The repository at https://github.com/mstimberg/TEST_restructured_brian2 demonstrates the general approach.

We'll need to look carefully at different use cases (e.g. importing single names `from brian2`) and make sure that they don't break.

PR #1261, opened 2021-01-09
Modifying Module Name input To Prevent Overwriting of input() function

rohithvarma3000:
Hi,
I have seen issue #1102 where the input () function has been overwritten due to the input module in the brian2 folder which is causing some inconvenience to the users. So I have modified the name of the input module to input_b. I have also modified the same in the only.py folder as well.
Apart from this, I have also observed that input has been overwritten in two other python files (dev/benchmarks/openmp/STDP_standalone.py and brian2/tests/test_GSL.py) as well. I have the modified the name in those files as well to avoid other issues in the future.


rohithvarma3000:
Do you think there are other files where this file name change should be made to get to ensure that code works??

rohithvarma3000:
Do you think there are other files where this file name change should be made to ensure that code works??

rohithvarma3000:
Hello @mstimberg,
I have identified more places where 'input' is being used as file, folder and variable name. I have appropriately modified all of those that I could find. Do you think there are other places that I need to make a change to ensure that the code continues to work?

mstimberg:
Apologies for being a bit slow responding to this. We need to be careful with this issue because it fixes something that is not a very big problem: if you want to use a shadowed `input()` you will get an error, and there are straightforward workarounds. On the other hand, renaming the `input` package can potentially break a user's script if they do something like `from brian2.input import PoissonGroup` which would be a violation of our [Compatibility policy](https://brian2.readthedocs.io/en/stable/introduction/compatibility.html). So it might actually be the best solution to not fix this issue but instead to clearly document it (and to recommend using `from brian2.only *` for users worried about shadowing things). Again, @thesamovar might have an opinion on this.

Note that we will probably do a bigger rework of the import framework as part of #1217

rohithvarma3000:
Hello @mstimberg,
I understand that backward compatibility will be a huge problem if we try to change the name of the folder. I think it will be better to avoid fixing this issue and rather document this behavior. Where do you think we should add this in the documentation ?
I have also observed that `input` is being used as a variable in some python files. I think that this could lead to confusion to new users and developers as well. Do think it would be a good idea to change these variable names to avoid this confusion? (I don't think changing these variable names will affect the backward compatibility of brian2)

mstimberg:
I think the page about "importing Brian" would be the best place in the documentation: https://brian2.readthedocs.io/en/stable/user/import.html

> I have also observed that `input` is being used as a variable in some python files. I think that this could lead to confusion to new users and developers as well. Do think it would be a good idea to change these variable names to avoid this confusion? (I don't think changing these variable names will affect the backward compatibility of brian2)

I'd be happy with any changes that changes the name of `input` to something else in the code (could even be something somewhat ugly like `inp`, but in cases where it makes sense could also be a more meaningful name about the type of input, say `spike_input`). If it is only used internally in a function (or in an example script) and not e.g. an argument of a function or an object attribute, then it should not be a problem for backwards compatibility.

rohithvarma3000:
Ok, I will work on these two parts.

mstimberg:
Superseded by #1270 and #1271.

PR #1265, opened 2021-01-17
Examples

Ziaeemehr:
Hi,
This is my first PR here, I tried to be consistent with the style.
I have added some examples in multiprocessing, standalone mode and synapse and tried to keep them simple.
Please check and let me know if I need to revise something and whether they are proper to be added to the main repository. please feel free to revise remove whatever.
best,
Abolfazl


mstimberg:
Many thanks for the PR, I'll hopefully find some time to review it in detail soon. As a quick first comment: it would be great if we could have a little explanation in the beginning  what the example is doing in addition to the link to the discourse forum. This should be a docstring (i.e. start and end with triple quotes) instead of a comment, this way it gets rendered nicely on the website.

Ziaeemehr:
OK fine I come back and add some description to each one. 👍

Ziaeemehr:
Hi and sorry for delay, I had a tough reviewer. 
I added some description to each example. 
For me the most suitable case to reduce the running time of the simulations is using `examples/multiprocessing/04_standalone_joblib_avoid_recompiling.py`.  Please check any thing make sense.

I wish we had an option just to compile the generated C++ code ones, have `.so` file and use that many times. I think `ANNarchy` is doing this, but I am not familiar with the internal process what avoid having this.

best,
Abolfazl

mstimberg:
Hi, thank you for adding the descriptions, I will hopefully review everything soon.

> I wish we had an option just to compile the generated C++ code ones, have `.so` file, change a parameter and use that again many times. I think `ANNarchy` is doing this, but I am not familiar with the internal process what avoid having this.

Yes, we are aware of this usage pattern and would really like to improve support for it, it is mostly a lack of human resources that prevents us from doing it. See https://github.com/brian-team/brian2/projects/1 for some related issues. The approach in `ANNarchy` is a bit different than ours, e.g. there is no distinction between runtime and standalone mode (their operating mode is somewhere in between). In their approach, you are slightly more explicit about what is going on under the hood  (you first compile the code then you run it), whereas we tried with Brian to make it possible for users to not know about these things. But of course, in some situations like the one you are interested in, their approach is preferable. We'll hopefully get there at some point as well (as an option, keeping our current operating mode intact), but it will still take some work...

mstimberg:
Apologies for having taken so long to review things, but I finally got around doing it. The examples all look great to me, I think they will be very useful for other users. I removed one example, the `04_...` example that "avoids recompilation", because I didn't really see the difference to the other ones which are already avoiding unnecessary recompilation. I made a number of simplifications to the other examples to make them focus a bit more on the specific point they are making. I also decided to remove some details like the `system('rm -rf standalone*')` call because it would fail on Windows.

Are you happy with the modified examples or do you feel that I made any of them less useful?

Ziaeemehr:
Great, thank you Marcel for your time and modifications,
I looks fine to me. 👍

Cheers

mstimberg:
The examples are now rendered as part of the "latest" documentation (with the next release, they will be part of the "stable" documentation we display by default): https://brian2.readthedocs.io/en/latest/

PR #1267, opened 2021-01-29
Use Versioneer

mstimberg:
This updates Brian to use [versioneer](https://github.com/python-versioneer/python-versioneer) to determine its version instead of relying on us setting it manually with a script on each release. If you are running Brian from the github repository, it determines the version from the tags, but it in releases (tar balls, wheels, etc.) it includes the version explicitly.
The main reason for doing this change is that it helps us to automatize the release process further. It should soon be possible to simply tag a release (which you can also do directly on github via the release mechanism), and all the rest should be taken care of. The other advantage of this solution is that in between releases, we have meaningful version numbers and don't have to manually set the version to something like `2.4+git` for development.

The versioneer system is used by big packages like matplotlib, so it seems to be reasonable robust.
There are two situations where it does not work correctly, unfortunately:
1. When you do a shallow clone of the repository that does not include any tags (e.g. with `git clone --depth 1 ...`). This is what is often done during automatic testing, but it is of course easy to fix.
2. When you dowload the zip file created by  github when you chose Code → Download ZIP. Unfortunately this zip is created manually by github and does not use `git archive`, therefore it does not include the version information.

The second point means that we can no longer ask users to install a development version for testing via
```console
$ pip install https://github.com/brian-team/brian2/archive/master.zip
```
This is unfortunate, but mostly something that we have to be aware of. There are two alternatives that we can use:
1. Recommend installation with `pip` + `git`: `pip install git+https://github.com/brian-team/brian.git` (however, this requires a working `git` installation)
2. Ask for an installation from test-pypi where we now upload development packages (mostly for testing the release mechanism). This looks like ` pip install -i https://test.pypi.org/simple/ Brian2==2.4.2.post0.dev55`

Versioneer offers several options for what the versions in between releases look like, the one that I have chosen looks a bit weird with the `post0.dev` part, but it contains some usable information (in the above example, 55 commits after the 2.4.2 release), and it correctly sorts with respect to the actual releases. In principle, I'd prefer another format which also includes the git hash, but that isn't compatible with test-pypi.

Oh, and just to make this clear: this is a change that is mostly affecting "power users". If you install the package normally via conda or pip, it will use a hardcoded version as it does currently.


thesamovar (approved):
This seems great. Is it a problem that we have to include a copy of versioneer code and therefore presumably have to manually update that with new versions, etc.?

mstimberg:
I don't see it as much of a problem. It was designed to be used this way and we avoid another dependency and problems if it updates. Actually, I used their current development version because they fixed a bug that was relevant to our use case and did not make a new release yet. As long as it works, we won't have to update it.

PR #1268, opened 2021-01-29
Remove duplicated profiling code in codegen.

leoschwarz:
Hello,

I've encountered an issue with `profile=True` in standalone mode where the generated code would fail to compile. It turned out, that the code generator duplicated the profiling lines for `synapses_push_spikes` which led to a duplicated variable definition. Removing them once fixed the issue for me.

The same code is already included at the end of templates/common_group.cpp,
and currently prevents compiling code involving synapses with profile=True.

The unittest failed before, but should be passing now.


mstimberg:
Many thanks for the PR, I think I even came across this issue myself at some point. It all looks good to me at the first sight, but I will run a few more tests next week. Please ping me if I don't get back to you during the week!

mstimberg (approved):
Looking good and works as advertised. Thanks again!

PR #1270, opened 2021-01-31
Modified variables with the name input

rohithvarma3000:
Hello @mstimberg,
As discussed in #1261, I have modified the names of variables whose name was `input` to prevent confusion amongst users and new developers.


mstimberg (approved):
All looks good to me, thanks!

PR #1271, opened 2021-01-31
Update import.rst

rohithvarma3000:
Hello @mstimberg,
Based on our discussion in the pull request #1261 , I have documented the overshadowing of the input function in the `import.rst` file.


mstimberg (approved):
Thanks again, looks good to me (I made a minor formatting change so that `input` does not link to the Python documentation of the builtin function when we talk about the Brian package).

Issue #1255, opened 2020-12-04
Improvement in PoissonInput documentation


johnpi:
It might be useful to add in the documentation of the PoissonInput class something like "Each of the N inputs is connected to all neurons of the target Group but each neuron receives independent realizations of Poisson spike trains." Currently, the text in the documentation of this class does not clarify if it uses an all-to-all connectivity or not and how the value of N inputs affects the target neurons. The suggested change might help users.

mstimberg:
Sorry for the late reply! Yes, I agree that it would be useful to make this clearer. Feel free to propose a pull request with a change to the documentation :smile:

For changes to the documentation, you don't have to jump through all the git hoops to do this. You can simply go to https://github.com/brian-team/brian2/edit/master/docs_sphinx/user/input.rst, click :pencil2: ("Edit this file"), make the changes and then use the option to create a pull request based on the changes.

mstimberg:
Closed via #1272.

PR #1272, opened 2021-02-02
Update input.rst

rohithvarma3000:
Hello @mstimberg,
I have improved the documentation of Poisson Input as requested in #1255 and fixed the example code in the file as well.

### DIFF
--- a/docs_sphinx/user/input.rst
+++ b/docs_sphinx/user/input.rst
@@ -21,10 +21,6 @@ be used, e.g.::
     S = Synapses(P, G, on_pre='v+=0.1')
     S.connect(j='i')
 
-Here, each input of the PoissonGroup is connected to all the neurons of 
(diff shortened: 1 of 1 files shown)


mstimberg (changes requested):
Thanks again for the PR. It does not quite address #1255, though: the issue was about `PoissonInput`, not about `PoissonGroup` (admittedly, it does not help that the section title is "Poisson inputs"...).

rohithvarma3000:
Hello @mstimberg,
Sorry for placing the statement in the wrong location (I was misled by the heading). I hope that this commit puts the statement in the right place.

mstimberg (approved):
Thanks again. I reverted a change in the PoissonGroup code example which broke the rate definition and made two other minor tweaks to the text (making `PoissonInput` and `NeuronGroup` link to the respective documentation), but it now looks good to go.

PR #1273, opened 2021-02-05
No longer use deprecated numpy type aliases

mstimberg:
Numpy type aliases are deprecated since numpy 1.20 and using e.g. `np.bool` over `bool` does not add anything. This PR therefore removes all unnecessary type aliases. Note that types like `np.int32` or `np.float32` are still useful (and not deprecated), since Python does not have any builtin equivalents.

See https://numpy.org/devdocs/release/1.20.0-notes.html#deprecations


thesamovar (approved):
This all seems unobjectionable to me.

Issue #1263, opened 2021-01-13
Missing warning/error when referring to event-driven var from summed var


mstimberg:
In `Synapses`, we raise an error if a clock-driven variable refers to an event-driven variable. In contrast, we do not raise an error (or a warning) if a summed variable refers to an event-driven variable, even though this is probably a mistake as well. A summed variable gets evaluated at every time step and therefore all variables it refers to need to be updated at every time step as well.

The following is an example of an equation that should raise an error (or a warning?)
```Python
syn = Synapses(inp, out, '''ds/dt = -s/(3*ms) : 1 (event-driven)
                            x_post = s : 1 (summed)''', on_pre='s += 1')
```

rohithvarma3000:
Hello @mstimberg,
I have gone through the code and found two locations where the check could be placed potentially. We could place it in the `check_flags` function under the Equations class or in the `__init__` method of the Synapses class. Since these flags may be used simultaneously by the Synapses objects only, I think it will be better to put in the `__init__` method of the Synapses class. What do you think?
Also, do you think we need to raise an error or a warning for this?

mstimberg:
Yes, I agree that `Synapses.__init__` would be the best place. We'll have to make sure that the error is only raised when necessary, i.e. when the right-hand-side of the `summed` variable depends on an event-driven equation. Note that this could happen indirectly, like in the following (completely made up) example:
```Python
syn = Synapses(inp, out, '''ds/dt = -s/(3*ms) : 1 (event-driven)
                            t = s : 1
                            x_post = t : 1 (summed)''', on_pre='s += 1')
```
You should be able to use `Equations.get_substituted_expressions()` to deal with this issue. I don't really see a legitimate use case for the `summed` + `event-driven` combination, therefore I think that it will be used by mistake only. So I'd rather think it should be an error. Maybe @thesamovar has an opinion on this?

mstimberg:
Closed via #1274

PR #1274, opened 2021-02-09
Error/Warning for Summed Variable Pointing to Event Variable

rohithvarma3000:
Hello @mstimberg,
As discussed in #1263, I have added the check for cases where the summed variables are referring to event-driven variables. Here I have used a recursive function to check for the additional cases that you pointed out instead of using the `get_substitued_expression()` function because the `get_substitued_expression()` function was not giving me the desired result. Also, I have defined the recursive function inside the `__init__` because this function will not be needed anywhere outside the `__init__` method. What are your thoughts?
Since we have not yet decided whether we need an error or a warning, I have placed a temporary print statement in its place for now. I will modify it as soon as we reach a decision.

### DIFF
--- a/brian2/synapses/synapses.py
+++ b/brian2/synapses/synapses.py
@@ -795,9 +795,10 @@ class Synapses(Group):
         
         # Checking whether a summed variable is referring to an event-driven variable
         for eq in event_driven:
-            if self._recur_check_event_summed(eq.varname,model):
+            val = self._recur_check_event_summed_clock(eq.varname,model)
+            if val == 1:
                 raise EquationError(f"A summed variable should not refer the event-driven variable {eq.varname}")
-            if self._recur_check_event_clock(eq.varname,model):
+            if val == 2:
                 raise EquationError(f"A clock-driven equation should not refer to the event-driven variable {eq.varname}")
 
         if len(event_driven):
@@ -947,10 +948,10 @@ class Synapses(Group):
         # Activate name attribute access
         self._enable_group_attributes()
     
-    def _recur_check_event_summed(self,var,eqs):
+    def _recur_check_event_summed_clock(self,var,eqs):
         """
-            Recursive Function Used to identify whether a summed variable is 
-            referring to an event-driven variable.
+            Recursive Function Used to identify whether a summed variable 
+            or a clock driven equation is referring to an event-driven variable.
 
             Parammeters
             ----------
@@ -961,51 +962,22 @@ class Synapses(Group):
 
             Returns
             ----------
-            val : boolean
-                Return True if a summed variable is referring to an 
-                event-driven variable, otherwise False or continues recursion
+            val : int
+                Return 1 if a summed variable is referring to an 
+                event-driven variable, returns 2 if a clock driven equation
+                is referring to an event-driven variable, otherwise 0 or continues recursion
 """
-        val = False
         for eq in eqs.values():
             if var in eq.identifiers and eq.varname != var:
-                if val == True:
-                    break
-                elif 'summed' in eq.flags:
-                    val = True
-                    break
-                else:
-                    val = self._recur_check_event_summed(eq.varname,eqs)
-        return val
-    
-    def _recur_check_event_clock(self,var,eqs):
-        """
-            Recursive Function Used to identify whether a clock-driven equation is 
-            referring to an event-driven variable.
-
-            Parammeters
-            ----------
-            var : str
-                variable that is required for checking
-            eqs : Equations Object
-                Equations object in which we need to check
-
-            Returns
-            ----------
-            val : boolean
-                Return True if a clock-driven equation is referring to an 
-                event-driven variable, otherwise False or continues recursion
-"""
-        val = False
-        for eq in eqs.values():
-            if var in eq.identifiers and eq.varname != var:
-                if val == True:
-                    break
+                if 'summed' in eq.flags:
+                    return 1
                 elif eq.type == DIFFERENTIAL_EQUATION and 'event-driven' not in eq.flags:
-                    val = True
-                    break
+                    return 2
                 else:
-                    val = self._recur_check_event_clock(eq.varname,eqs)
-        return val
+                    val = self._recur_check_event_summed_clock(eq.varname,eqs)
+                    if val != 0:
+                        return val
+        return 0
 
     N_outgoing_pre = property(fget= lambda self: self.variables['N_outgoing'].get_value(),
                               doc='The number of outgoing synapses for each neuron in the '
--- a/brian2/synapses/synapses.py
+++ b/brian2/synapses/synapses.py
@@ -815,7 +815,7 @@ class Synapses(Group):
             """
             val = False
             for eq in eqs.values():
-                if var in eq.identifiers:
+                if var in eq.identifiers and eq.varname != var:
                     if val == True:
                         break
                     elif 'summed' in eq.flags:
--- a/brian2/synapses/synapses.py
+++ b/brian2/synapses/synapses.py
@@ -957,6 +957,13 @@ class Synapses(Group):
                 variable that is required for checking
             eqs : Equations Object
                 Equations object in which we need to check
+            orig_var : str
+                Carries the name of the original event-driven variable
+                Default Value - None
+            intermediate_vars : list
+                Carries the names of all the intermediate variables that have referred 
+                to the event-driven variable.
+                Default Value - []
 
             Returns
             ----------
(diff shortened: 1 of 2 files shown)


mstimberg:
Many thanks yet again. Sorry for being a bit slow reviewing things lately, I hope I can look into this in detail tomorrow.

mstimberg:
I had a closer look now and I agree that `get_substituted_expressions` does not help you as much as I thought it would. Having a little function to go down recursively is fine, I'd probably pull it out of `__init__`, though. By giving it a name starting with an underscore we'd make clear that it is only used internally. I realized that – contrary to what I though – we are not raising an error when we refer to an `event-driven` variable from a `clock-driven` equation, even though our [documentation](https://brian2.readthedocs.io/en/stable/user/synapses.html) states
> These equations must also be independent of the other ones, that is, a differential equation that is not event-driven cannot depend on an event-driven equation (since the values are not continuously updated). 

It would be great if we could fix this here as well since it is a very similar issue. I'd say we raise an `EquationError` in both cases,  as I commented in https://github.com/brian-team/brian2/issues/1263#issuecomment-766931741 I think it is always an unintentional error by the user.

I'll make a few more detailed comments directly in the code.

brian2/synapses/synapses.py
```diff
@@ -794,6 +795,41 @@ def __init__(self, source, target=None, model=None, on_pre=None,
...
+            val = False
+            for eq in eqs.values():
+                if var in eq.identifiers and eq.varname != var:
+                    if val == True:
+                        break
+                    elif 'summed' in eq.flags:
+                        val = True
+                        break
+                    else:
+                        val = recur_check_event_summed(eq.varname,eqs)
```
mstimberg, line 825: This could be quite a bit simpler without changing the logic:
```Python
            for eq in eqs.values():
                if var in eq.identifiers and eq.varname != var:
                    if 'summed' in eq.flags:
                        return True
                    else:
                        return recur_check_event_summed(eq.varname, eqs)
            return False
```

brian2/synapses/synapses.py
```diff
@@ -794,6 +795,41 @@ def __init__(self, source, target=None, model=None, on_pre=None,
...
+                    
+            return val
+        # Checking whether a summed variable is referring to an event-driven variable
+        if len(event_vars):
+            for v in event_vars:
+                if recur_check_event_summed(v,model):
+                    print("Event-Driven Variable " + v + " is being summed")
```
mstimberg, line 832: This could also be slightly simpler, no need for `if len(event_vars)`, the loop will simply do nothing if it is empty.

mstimberg (changes requested):
Again thanks for your work on this!

rohithvarma3000:
Hello @mstimberg,
I have removed the `event_vars` and the if condition as requested. The simpler recursive function that you have provided has a flaw in it and it is not giving the expected output. So I will stick with the recursive function that I have made. I have placed the function outside the `__init__` function and made it a protected member.
I have looked into the clock-driven equation issue as well and I have built a similar recursive function called `_recur_check_event_clock` to cater to that error as well. Please verify whether I have put the condition for this case correctly in the recursive function.

mstimberg:
Right, sorry about the mistake in my function, but I still think your version is somewhat complicated and a bit confusing. I might be mistaken again, but I think replacing
```Python
return recur_check_event_summed(eq.varname, eqs)
```
```Python
if recur_check_event_summed(eq.varname,eqs):
   return True
```
in my version should fix the logic. Either way, I don't think we need a second function to go through all the equations again. We need to go through the equations in the same way in both cases, so we can combine the two checks (`if 'summed' in eq.flags`...`elif eq.type == ...`.)

rohithvarma3000:
Hello @mstimberg,
Your logic in the last message had fixed the recursive function.
I was initially hesitant to write both the checks put together in a single function because in that case, we cannot pass a simple boolean value. I thought it may complicate the logic. But now I have worked out the logic in putting both the checks in the same equation. Please check it and provide you feedback.

mstimberg:
Just as a quick comment (will have a closer look tomorrow): maybe everything would be simpler if the recursive function did not return any value at all but directly raised the `EquationError` itself?

rohithvarma3000:
I had not thought about it, but I think that will make the function less complex. I will try this and let you know.

rohithvarma3000:
Hello @mstimberg,
If we try to raise the error directly in the recursive function, we will not be able to notify the user which event-driven variable is facing the issue. In that case, I think the user may have a tough time trying to identify which variable is being misused which is not a good idea.  I feel that if we can provide more information to the user regarding the error, it will be easier for the user to debug the code. What do you think?

rohithvarma3000:
Hello @mstimberg,
Raising the error has greatly simplified the code. I have raised both the errors inside the recursive function itself and updated the documentation of the function accordingly.

mstimberg:
Thanks for the changes, this is almost perfect now. I'd pull the method out of the `Synapses` class, and make it into a simple function since it does not need the `self` reference. For nicer error message we might want to keep the information about the variable it was looking at and maybe even the name of the intermediate equations that referred to it (in principle, we could have a chain of references, even though I don't think this is common in practice). I think something along these lines should work (I did not test it, though):
```Python
def _recur_check_event_summed_clock(var, eqs, orig_var=None, itermediate_vars=None):
    if orig_var is None:
        orig_var = var
    if intermediate_vars is None:
        intermediate_vars = []
    if len(intermediate_vars):
        via_str = ' (via ' + ', '.join(f'\'{v}\'' for v in intermediate_vars) + ')'
    for eq in eqs.values():
        if var in eq.identifiers and eq.varname != var:
            if 'summed' in eq.flags:
                raise EquationError(f"The summed variable \'{eq.varname}\' refers to the event-driven variable \'{orig_var}\'{via_str}")
            elif eq.type == DIFFERENTIAL_EQUATION and 'event-driven' not in eq.flags:
                raise EquationError(f"The clock-driven equation for \'{eq.varname}\' refers to the event-driven variable \'{orig_var}\'{via_str}")
            else:
                if orig_var != var:
                    intermediate_vars.append(var)
                self._recur_check_event_summed_clock(eq.varname, eqs, orig_var=orig_var, intermediate_vars=intermediate_vars)
```
I hope that makes the idea clear (again, not sure the code actually works :) ). Then all that is missing would be to have a few tests to make sure everything works as expected.

rohithvarma3000:
Hello @mstimberg,
I would suggest that we make the function a static method within the class instead of making it a simple function. What do you think about that?

mstimberg:
> I would suggest that we make the function a static method within the class instead of making it a simple function. What do you think about that?

I'm happy with this solution as well, I don't have a strong preference.

rohithvarma3000:
Hello @mstimberg,
I have made the function static and changed it to cater to the improved error message. I have also improved the documentation accordingly as well.

mstimberg (commented):
Thanks, this all looks good at the first sight. I will test out things in detail next week. Testing is an important point actually: do you want to add tests to the test suite as well? Otherwise I can do it next week.

rohithvarma3000:
Hello @mstimberg ,
I have gone through the `test_synapses.py` file and added the test for summed variable referring to an event-driven variable in the function `test_summed_variable_errors()`. I am not sure under which function I should add the test for the clock-driven variable referring to the event-driven variable case. Could you suggest where I should add it?

mstimberg:
I think the best solution would be to add a new test function. Actually it could be `test_event_driven_dependency_error3` and copy the format of the two earlier ones. Which makes me wonder, we should probably also use `UnsupportedEquationsException` instead of `EquationError` for the two new checks that your PR introduces.

rohithvarma3000:
Hello @mstimberg,
I have added the test for the clock-driven equation under the `test_event_dependency_error3()` function as you have suggested.
I am not very clear regarding the difference between the `UnsupportedEquationException` and the `EquationError`.

mstimberg:
Great, looking good. The difference between `UnsupportedEquationException` and `EquationError` is that the former is rather used to signal that we cannot manage a certain type of equations (e.g. a numerical algorithm cannot deal with a stochastic equation) while the latter is used to signal that the user made a mistake (e.g. a missing parentheses in an equation). I actually changed my mind on this and think that the current usage makes sense, so please forget about it :slightly_smiling_face: .

I have only two minor suggestions before I'll be happy to merge the PR:
* I'd prefer to not use `intermediate_vars=[]` with an empty list as a default argument. This is usually discouraged because the default argument will change if we append to the list. You deal with this by making a copy of the list, but I'd still rather use `intermediate_vars=None` and `if intermediate_vars is None: intermediate_vars = []` to deal with this in a non-ambiguous way.
* There are a number of minor stylistic things that are not in line with our default style (see https://brian2.readthedocs.io/en/stable/developer/guidelines/style.html), in particular please use a space after comma, and do not use spaces around the equals sign for keyword arguments in functions.

rohithvarma3000:
Hello @mstimberg,
I have changed the default argument as you have suggested and reformatted the code according to the guidelines.

mstimberg (approved):
Thanks again for all the work. I made a few more minor changes and will merge as soon as the test suite passes.

PR #1275, opened 2021-02-11
Build binary wheels

mstimberg:
This PR makes github action build binary wheels for all platforms and uploads them to PyPI for tagged releases (and to TestPyPI for all commits). This makes for a rather impressive list of files, including wheels for the brand-new Apple machines with an ARM chip :), see https://test.pypi.org/project/Brian2/2.4.2.post0.dev90/#files

This seems to work well, and will installation much better/quicker for users that install with `pip` (actually, we might even reconsider our recommendation to use `conda`). There's one thing that is left to do which is only tangentially related to building the wheels: on some OS X installations, `/usr/bin/gcc` is actually a symlink to the `clang` compiler. This is fine in general, as both are mostly compatible. However, this fails with our default `gcc` arguments which include `-march=native` which is not supported by `clang`. In the current PR, I work around this by trying all the flags and check whether compilation works for them, and then filtering the list of flags accordingly. This does quite a bit of unnecessary compilation, but seems to be fast enough to not be really noticeable. We could of course speed this up by caching the results, but we might also consider having different defaults for `gcc` and `clang` and treating them separately in general. All this is a bit complicated by the fact that I don't have access to an OS X machine for testing, I can only observe the results on the CI server... But maybe just using `clang` on a Linux machine would be enough.

### DIFF
--- a/.github/workflows/publish_to_pypi.yml
+++ b/.github/workflows/publish_to_pypi.yml
@@ -19,12 +19,15 @@ jobs:
       - name: Install cibuildwheel & dependencies
         run: |
           python -m pip install --upgrade pip
-          python -m pip install cibuildwheel==1.9.0 "cython>=0.29" numpy>=1.15 setuptools
+          python -m pip install cibuildwheel==1.9.0
       - name: Build source tarball (only on Linux)
         run: python setup.py sdist --formats=gztar --with-cython --fail-on-error
         if: ${{ startsWith(matrix.os, 'ubuntu-') }}
       - name: Build wheels
         run: python -m cibuildwheel --output-dir dist
+        env:
+          CIBW_BEFORE_BUILD: pip install "cython>=0.29" numpy>=1.15 setuptools
+          CIBW_BUILD_VERBOSITY: 1
       - name: Publish distribution 📦 to Test PyPI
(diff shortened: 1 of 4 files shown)


thesamovar:
Cool! I agree with your solution. I think I'd put the cacheing in there but it's no big deal.

I'm finding lately that I'm using pip more and more and conda less actually. Seems like it has improved a lot lately.

mstimberg:
I inserted a bit of caching, but tried to be cautious so that it still works if you change the flags or compiler during a session between runs – certainly not something that is often done, but it would be a shame to break it in a hard-to-debug way.
I tried to extend the check to MSVC, but it would need a different approach: the MSCV compiler does not fail if you provide it with an unknown flag, it only produces a warning that the flag was ignored. I did not want to get into checking the output, after all the reason for adding this feature were differences between gcc and clang.

thesamovar (approved):
I had a quick look at the code and it seems fine, but I think I'll have to take most of it on trust that it works because it's stuff I'm not familiar with and don't have time at the moment to do any testing myself. In general though, happy with this approach and I think it would be good to get this merged.

mstimberg:
> In general though, happy with this approach and I think it would be good to get this merged.

Good enough for me right now :)

PR #1277, opened 2021-02-19
Provide fair defaults build flags to several archs

emollier:
Greetings,

While integrating the package _brian_ in the Debian project with @mr-c, we tried to get the package to build and test seamlessly on all architectures possible supported by the project without the need for users to change the default build options in _brian2/codegen/cpp_prefs.py_.  You can see for instance a summary of the [latest _brian_ build and test results per architecture during the upload to _experimental_](https://buildd.debian.org/status/package.php?p=brian&suite=experimental) on the build servers status page.

Would you be interested to have such a default build options selector upstreamed?

The present version only matches the architectures on which I believe it might be possible to do some testing on real hardware; thus no SuperH, nor x32, due to missing build dependencies for instance.  Also, there are still some results missing due to the build queue depth on older CPU architectures at the moment, hence the draft status.

Kind Regards,
Étienne.

Edited to add: since upload to _sid_, build results for _experimental_ are not available anymore.  [Build results for _sid_ are available though.](https://buildd.debian.org/status/package.php?p=brian&suite=sid)

### DIFF
--- a/brian2/codegen/cpp_prefs.py
+++ b/brian2/codegen/cpp_prefs.py
@@ -96,11 +96,11 @@ else:
                              '-fno-finite-math-only', '-march=native',
                              '-std=c++11']
     elif re.match('^(alpha|ppc.*|sparc.*)$', machine):
-	default_buildopts = ['-w', '-O3', '-ffast-math',
-			     '-fno-finite-math-only', '-mcpu=native',
+        default_buildopts = ['-w', '-O3', '-ffast-math',
+                             '-fno-finite-math-only', '-mcpu=native',
                              '-mtune=native', '-std=c++11']
     elif re.match('^(parisc.*|riscv.*)$', machine):
-	default_buildopts = ['-w', '-O3', '-ffast-math',
+        default_buildopts = ['-w', '-O3', '-ffast-math',
                              '-fno-finite-math-only', '-std=c++11']
     else:
         default_buildopts = ['-w']
--- a/brian2/codegen/cpp_prefs.py
+++ b/brian2/codegen/cpp_prefs.py
@@ -88,19 +88,22 @@ if platform.system() == 'Windows':
         if 'avx2' in flags:
             msvc_arch_flag = '/arch:AVX2'
 
-# Optimized default build options for a range a CPU architectures
-machine = os.uname().machine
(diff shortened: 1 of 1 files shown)


mstimberg:
Hi. Thanks that would be a very welcome addition (and great to have Brian in Debian!). We added one feature recently that is somewhat related: before the first compilation we now do a number of quick test compilations to filter out unsupported flags. The only thing I am a bit worried about is we don't have the opportunity to run tests on this hardware ourselves. On the other hand, this situation does not really worsen with your addition.

More related to the debian package: I've had a quick look at the test log and saw that it deletes many tests and does not quite run the tests in the correct way, e.g. we now use `pytest` instead of `nosetest`. But then, our full test suite with all bells and whistles might take an hour, maybe this is not what you want. Is there a place to discuss this kind of question or should this go via Debian's mail-based bug tracker? I'd also be happy to discuss this somewhere here on github or on our discourse forum.

emollier:
Hi Marcel,

Marcel Stimberg, on 2021-02-22 02:41:29 -0800:
> More related to the debian package: I've had a quick look at the test log and saw that it deletes many tests and does not quite run the tests in the correct way, e.g. we now use `pytest` instead of `nosetest`. But then, our full test suite with all bells and whistles might take an hour, maybe this is not what you want. Is there a place to discuss this kind of question or should this go via Debian's mail-based bug tracker? I'd also be happy to discuss this somewhere here on github or on our discourse forum.

Yes, the best place to discuss issues in Debian packages would be on bugs.debian.org, and its (arguably dreaded :) Debbugs server.  I took the liberty to open an item about the test suite[1], in which I refer to the present merge request.  From my quick try, just swapping commands is not obvious, and I don't necessarily have personally the whole history behind the skips.

[1]: https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=983411

If you find it more convenient, then we can still continue the discussion some place around here of course.

In any case, thanks for pointing this out!

Cheers,
-- 
Étienne Mollier <etienne.mollier@mailoo.org>
Fingerprint:  8f91 b227 c7d6 f2b1 948c  8236 793c f67e 8f0d 11da
Sent from /dev/pts/2, please excuse my verbosity.

mstimberg:
Great, I will do my best to interact with debbugs :) For this PR, all looks good to me and I'm happy to merge it. Thanks again for the contribution.

mstimberg (approved):

PR #1279, opened 2021-02-24
Add constant definitions instead of replacing them in the code for standalone mode

### LINKED ISSUE #1276: Constant replacements in C++ standalone can go wrong
As discovered by @wxie2013, the following rather trivial example leads to a compilation error:
```Python
set_device('cpp_standalone')
x = 42
G = NeuronGroup(1, 'y : 1')
G.y = 'x'
run(0*ms)
```
The reason is that we quite naively replace `x` in the code with 42 in our "this is a bit of a hack" [`freeze`](https://github.com/brian-team/brian2/blob/20e120921ab237c5a0f6ea04db766c9ca052e811/brian2/devices/cpp_standalone/device.py#L232) function. Unfortunately we use the name `x` quite often, e.g. in macro definitions like `#define _brian_pow(x, y) (pow((x), (y)))` or as the argument of functions. I think the fix is rather straightforward (short of dealing with this as part of a bigger refactoring): the constant replacement should only affect the user code, not all the support code around it.

mstimberg:
This fixes #1276, the problem that external constants were unconditionally search&replaced in the final generated code which could lead to unexpected results for external constants named `x` for example (which we use in several places as function arguments).

At first, I thought this would be easy to fix by just replacing constants in the abstract code instead of in the final code which includes all the function definitions. But I then realized that we use the `freeze` mechanism also to replace constants outside of user code, e.g. templates might refer to `N` as the number of neurons which is represented as an "external constant" as well. The easiest way to fix it that I found was to use the mechanism that we already use for constants such as `_num_arrayname` that we add to give the code access to the length of arrays, i.e. the `%CONSTANTS%` block. This means, instead of replacing the constant with its value in the code, we add something like `const int N = 10` to this block. Potentially this could mean slower code if the compiler does not optimize code based on `const` definitions as much as numerical literals, but I doubt that this would make any measurable difference.

A minor point: the change means we can remove the `CPPStandaloneDevice.freeze` function, but I leave it in for now as it would break `Brian2CUDA` and potentially other projects that we don't know of (`Brian2GeNN` copied the code instead, so it won't be affected). I added a warning for now, we can remove the code at a later point.

To make the difference between the old and the new approach clear, consider the stateupdater code for this example:
```Python
tau = 10*ms
eqs = '''
dv/dt = (1-v)/tau : 1
'''
```

Before, we generated the following code where `N` and `tau` were replaced directly:
```C++
    ///// CONSTANTS ///////////
    //...
    //// MAIN CODE ////////////
    //...
    const double _lio_1 = 1.0 - exp(1.0f*(- dt)/0.01);
    const double _lio_2 = exp(1.0f*(- dt)/0.01);

    const int _N = 10;
    
    for(int _idx=0; _idx<_N; _idx++)
    {
       // ...
    }
```
With the new approach, we have them as constants in the code:
```C++
    ///// CONSTANTS ///////////
    const double tau = 0.01;
    const int64_t N = 10;
    //...

    //// MAIN CODE ////////////
    //...
    const double _lio_1 = 1.0 - exp(1.0f*(- dt)/tau);
    const double _lio_2 = exp(1.0f*(- dt)/tau);

    const int _N = N;
    
    for(int _idx=0; _idx<_N; _idx++)
    {
        //...
    }
```

### DIFF
--- a/brian2/devices/cpp_standalone/device.py
+++ b/brian2/devices/cpp_standalone/device.py
@@ -230,6 +230,11 @@ class CPPStandaloneDevice(Device):
         return None  # handled differently
 
     def freeze(self, code, ns):
+        # TODO: Remove this function at some point
+        logger.warn('The CPPStandaloneDevice.freeze function should no longer '
+                    'be used, add constant definitions directly to the '
+                    'code in the "CONSTANTS" section instead.',
+                    name_suffix='deprecated_freeze_use', once=True)
         # this is a bit of a hack, it should be passed to the template somehow
         for k, v in ns.items():
             if (isinstance(v, Variable) and
@@ -755,7 +760,8 @@ class CPPStandaloneDevice(Device):
         writer.write('main.cpp', main_tmp)
 
     def generate_codeobj_source(self, writer):
-                # Generate data for non-constant values
+        # Generate data for non-constant values
+        renderer = CPPNodeRenderer()
         code_object_defs = defaultdict(list)
         for codeobj in self.code_objects.values():
             lines = []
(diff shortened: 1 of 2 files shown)


thesamovar (approved):
In favour of this. Compilers really ought to work well with this. I can't remember if we're benchmarking regularly these days to pick up on performance regressions? Might be handy to do this occasionally.

mstimberg:
>  I can't remember if we're benchmarking regularly these days to pick up on performance regressions? Might be handy to do this occasionally.

I've been running benchmarks on my work machine, but it's been a while since I did it for the last time: https://brian-team.github.io/brian2_benchmarks/#/   Maybe I could set it up as a regular cron job. Some things about the setup are a bit tricky, in particular with which version of dependencies it installs. But it should be good enough to detect big performance regressions, we don't really need an accurate picture of the whole performance history.

PR #600, opened 2015-11-24
[MRG] Efficient synapse creation

thesamovar:
This branch is for work on issue #65.

PR #1280, opened 2021-02-25
Fixed size synapses

### LINKED ISSUE #664: Implement fixed size random samples
That is the `sample(N, size=...)` generator syntax for synapse construction.

This is straightforward for numpy, but more complicated for weave/cython/standalone.

mstimberg:
This will fix #664 by finally implementing the `sample(..., size=...)` syntax in synapse generators.
This basically works already, but I'm still marking this as a draft because there are a couple of open questions:
1. how do we handle various types of errors? For example, if you ask for a bigger size than possible, should this raise an error? Or should we handle this via the `skip_if_invalid` option, i.e. silently generate fewer samples than the requested `size`?
2. The current algorithm is probably inefficient for small sizes (it scales with the total number of post-synaptic neurons) which are probably a quite common use case. Maybe it would make sense to use a simple "draw `size` random numbers and re-draw numbers for non-unique items" in these cases but we should probably run some benchmarking first.
3. I realized that our current syntax only allows to draw a fixed number of synapses per pre-synaptic neuron, i.e. to keep the "outgoing" synapses constant. A (more?) common use case would be to keep the number if *incoming* synapses constant. Maybe we could without too much work make our generator syntax work in the two directions, i.e. also support `connect(i='... for .... in ...')`?

Finally, we need more tests and documentation, but this should be straightforward.

### DIFF
--- a/brian2/codegen/runtime/cython_rt/templates/common_group.pyx
+++ b/brian2/codegen/runtime/cython_rt/templates/common_group.pyx
@@ -17,6 +17,9 @@ from libc.stdlib cimport abs  # For integers
 from libc.math cimport abs  # For floating point values
 from libc.limits cimport INT_MIN, INT_MAX
 from libcpp cimport bool
+from libcpp.set cimport set
+from cython.operator cimport dereference as _deref, preincrement as _preinc
+cimport cython as _cython
 
 _numpy.import_array()
 cdef extern from "numpy/ndarraytypes.h":
--- a/brian2/codegen/runtime/cython_rt/templates/synapses_create_generator.pyx
+++ b/brian2/codegen/runtime/cython_rt/templates/synapses_create_generator.pyx
@@ -88,14 +88,14 @@ cdef void _flush_buffer(buf, dynarr, int buf_len):
         if _iter_size > _n_total:
             _iter_size = _n_total
         if _selection_algo:
-            {{iteration_variable}} = _iter_low
+            {{iteration_variable}} = _iter_low - _iter_step
         else:
             # For the tracking algorithm, we have to first create all values
             # to make sure they will be iterated in sorted order
             _selected_set.clear()
             while _n_selected < _iter_size:
                 _r = <int> (_rand(_vectorisation_idx) * _n_total)
-                while not _selected_set.insert(_r).second:
+                while not _selected_set.insert(_r).second:  # .second will be False if duplicate
                     _r = <int> (_rand(_vectorisation_idx) * _n_total)
                 _n_selected += 1
             _n_selected = 0
@@ -103,13 +103,13 @@ cdef void _flush_buffer(buf, dynarr, int buf_len):
 
         while _n_selected < _iter_size:
             if _selection_algo:
+                {{iteration_variable}} += _iter_step
                 # Selection sampling technique
                 # See section 3.4.2 of Donald E. Knuth, AOCP, Vol 2,
                 # Seminumerical Algorithms
                 _n_dealt_with += 1
                 _U = _rand(_vectorisation_idx)
                 if (_n_total - _n_dealt_with) * _U >= _iter_size - _n_selected:
-                    {{iteration_variable}} += _iter_step
                     continue
             else:
                 {{iteration_variable}} = _iter_low + _deref(_selected_it)*_iter_step
--- a/brian2/codegen/runtime/cython_rt/templates/synapses_create_generator.pyx
+++ b/brian2/codegen/runtime/cython_rt/templates/synapses_create_generator.pyx
@@ -41,6 +41,9 @@ cdef void _flush_buffer(buf, dynarr, int buf_len):
     {% if iterator_func=='sample' %}
     cdef int _iter_sign
     {% if iterator_kwds['sample_size'] == 'fixed' %}
+    cdef bool _selection_algo
+    cdef set[int] _selected_set = set[int]()
+    cdef set[int].iterator _selected_it
     cdef int _n_selected
     cdef int _n_dealt_with
     cdef int _n_total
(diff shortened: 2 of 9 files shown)


thesamovar:
Quick comments:

1. I like using skip_if_invalid in the same way - coherent and logical to me.
2. I think at some point I wrote some code comparing a number of different algorithms for different sizes and indeed different algorithms are better at different sizes. Maybe this is somewhere in either the Brian 1 or 2 dev directories? Can't look right now but should be relatively easy (search for sampling without replacement I guess).
3. I like this idea - handle it in a separate PR though?

mstimberg:
Agreed with 1 and 3.

>     2\. I think at some point I wrote some code comparing a number of different algorithms for different sizes and indeed different algorithms are better at different sizes. Maybe this is somewhere in either the Brian 1 or 2 dev directories? Can't look right now but should be relatively easy (search for sampling without replacement I guess).

Are you sure this  wasn't for sampling with a probability instead of with a fixed size? We do use two different algorithms there with a heuristic to chose between them. In Brian 1, the algorithm simply used Python's `sample` function in a loop. We had some discussion about the efficiency of different algorithms, starting here: https://github.com/brian-team/brian2/pull/600#issuecomment-177645032 But from a cursory look it seems this was mostly about binomially sized samples, e.g. deciding whether it is more efficient to do `k = binomial(p*n); sample_without_replacement(k, n)` or just check `rand() < p` for all pairs.

In this discussion we considered scikit-learn's `sample_without_replacement` function. This is what their documentation says for the automatic selection mechanism:
>         If ratio is between 0 and 0.01, tracking selection is used.
>        If ratio is between 0.01 and 0.99, numpy.random.permutation is used.
>        If ratio is greater than 0.99, reservoir sampling is used.

Tracking selection is what I meant by the simple approach (you just draw random numbers and redraw them if they are not unique) and reservoir sampling should be less efficient than the method I implemented for small ratios, but probably more efficient for big ratios (due to the smaller number of random numbers that are drawn). I'll benchmark the three algorithms and see what it gives us. I'll probably only do it for standalone and Cython, in numpy `numpy.choice` should do this kind of decision for us.

Curiosly, scikit-learn's *code* does not seem to implement what their documentation says. There is no method using `numpy.random.permutation` and the cutoff between tracking selection and reservoir sampling is 0.2:
```Cython
        # The value 0.2 has been determined through benchmarking.
        if ratio < 0.2:
            return _sample_without_replacement_with_tracking_selection(
                n_population, n_samples, random_state)
        else:
            return _sample_without_replacement_with_reservoir_sampling(
                n_population, n_samples, random_state)
```
https://github.com/scikit-learn/scikit-learn/blob/94abe05b4b96de2ca30d998fb9adb2fbd3eb1bde/sklearn/utils/_random.pyx#L293-L299

thesamovar:
That's what I was (mis)remembering! Agree with your approach.

mstimberg:
Ok, here some benchmark results for the three algorithms (+ `numpy.choice` for reference). These are just the run times without code generation / compilation (but measured in a somewhat crude way by recording the total time and subtracting the total time for `size=0` which is handled as a special case). The absolute numbers are obviously not very important (this is for 1000 "source" and 100000 "target" neurons). The x axis shows the proportion of targets that are chosen, i.e. p=0.1 corresponds to `syn.connect(j='k for k in sample(100000, size=10000)')`. 

benchmarks



As expected, the tracking algorithm (simply drawing random numbers and redrawing them if they are already used) is very fast for low numbers and very slow for high numbers. After running the benchmarks I realized that my C++ implementation is probably not the most efficient as I use `set::find` to check whether a number has already been used, but `set::insert` already does this check. So maybe the real values should be a bit lower for standalone. The difference between the sorted and unsorted columns is that in the sorted column, we sort everything in the same way we do in all other synapse generation: first by pre-synaptic then by post-synaptic index. I don't think we strictly guarantee this, but I have mentioned this property to users who wanted to set/get variables manually in standalone mode and needed to know about the order. I guess it could be useful to have this property. For the selection sampling algorithm this would be an advantage, since it always generates values in order. For the two other algorithms (and `numpy.choice`), we have to add an explicit sorting step. This sorting could probably be improved for the reservoir algorithm by choosing an optimized algorithm for "almost sorted" arrays. But on the other hand, why bother if the selection sampling algorithm already works well (as a side note: the advantage of reservoir sampling is that you do not need to know the size of the total population in advance, but we do not make use of this property). So my proposal would be:
1. Use the tracking algorithm for p < 0.1 and the sampling selection algorithm otherwise
2. Always sort the result (i.e. sort the results of the tracking algorithm)

We should not forget that using random fixed size with p > 0.1 will be pretty rare anyway, at least for big networks where the run time becomes relevant.

thesamovar:
I'm happy with this reasoning!

mstimberg:
Ok, this is now mostly done, the only thing that is missing is documentation. I changed the algorithm implementations slightly (e.g. I did not realize earlier that `std::set` keeps things sorted, so no need to sort things afterwards) and run another more detailed benchmark. The new cutoff is 0.06, see the dotted line below. This seems to work nicely for both Cython and C++ standalone.

new_benchmarks

This PR also extends the test suite quite a bit and fixes #1281 .

thesamovar:
Let me know when the docs are done and I can review. I should probably go through this one quite carefully I guess.

mstimberg:
I added the docs, but only a few lines since there is not much difference with the previous `sample` syntax. I also fixed a little issue with our `device_override` decorator, it turns out that doctests in functions using that decorator were not executed. Review-ready from my side :)

thesamovar (approved):
Good to go!

thesamovar:
I reviewed this, and merged latest commits from master. Will merge once tests complete OK.

mstimberg:
Thanks for the review. All tests pass, merging now.

PR #1283, opened 2021-03-18
change docstring of NeuronGroup so that model isn't a tuple

chaddy1004:
Issue:

Because of the parenthesis in the docstring, PyCharm was expecting model to be a Tuple(str, Equation).

image

Removal of the said parenthesis resolves the issue.

### DIFF
--- a/brian2/groups/neurongroup.py
+++ b/brian2/groups/neurongroup.py
@@ -404,7 +404,7 @@ class NeuronGroup(Group, SpikeSource):
     ----------
     N : int
         Number of neurons in the group.
-    model : (str, `Equations`)
+    model : str, `Equations`
         The differential equations defining the group
     method : (str, function), optional
         The numerical integration method. Either a string with the name of a


thesamovar:
I'll wait for @mstimberg to comment on this and not merge immediately, because I don't know if other environments rely on it being a tuple, etc. I can see we have tuples in other places.

mstimberg (approved):
Thanks, this had been bugging me as well, but I somehow never got around to actually fix it. There are many other places where we are doing similar imprecise type declarations in docstrings, would be great fixing these as well some day...

mstimberg:
Just a thought: maybe there's a tool around (e.g. a plugin for `pytest`) that records the type of all arguments in function calls and automatically checks them against the docstring? This could be very useful and shouldn't be *that* difficult to write.

chaddy1004:
> Just a thought: maybe there's a tool around (e.g. a plugin for `pytest`) that records the type of all arguments in function calls and automatically checks them against the docstring? This could be very useful and shouldn't be _that_ difficult to write.

Thanks for the suggestion! I will try to look for them and share if i get something! Thanks for reviewing :)

PR #1286, opened 2021-04-02
Fix flag_file load encoding error for Python 3.9

kjohnsen:
Try opening flag_file with encoding. As opposed to json.load. Was breaking for Python 3.9


mstimberg (approved):
Great, thanks again!

PR #1294, opened 2021-04-29
Synapse generator expressions i='...'

mstimberg:
As briefly discussed in #1280 , the main interest of "random fixed size sampling" is typically used to have a fixed number of *inputs* per target neuron. In contrast, our current code only allows its use in `j='...'` expressions, i.e. sampling a post-synaptic neurons for each pre-synaptic neurons.

This PR adds the missing feature of also using `i='...'` generator expressions, which together with the previous changes finally makes a fixed number of incoming connections possible. The implementation is rather straightforward: it replaces hardcoded names like `j` and `post_idx` in the templates by placeholders that get filled in depending on the way we iterate. Still, everything gets a bit more complicated... I tried to at least refactor `Synapses.connect` a bit which had a really hard-to-follow logic before.

Everything seems to basically work, but I'm still considering this a draft since testing is incomplete, and I did not document the new feature at all.

### DIFF
--- a/brian2/codegen/runtime/cython_rt/templates/synapses_create_generator.pyx
+++ b/brian2/codegen/runtime/cython_rt/templates/synapses_create_generator.pyx
@@ -54,17 +54,23 @@ cdef void _flush_buffer(buf, dynarr, int buf_len):
     cdef size_t _jump
     {% endif %}
     {% endif %}
+
+    {# For a connect call j='k+i for k in range(0, N_post, 2) if k+i < N_post'
+    "j" is called the "target variable" (and "_post_idx" the "target array", etc.)
+    "i" is called the "base variable" (and "_pre_idx" the "base array", etc.)
+    "k" is called the iteration variable #}
+
     # scalar code
     _vectorisation_idx = 1
     {{scalar_code['setup_iterator']|autoindent}}
-    {{scalar_code['create_j']|autoindent}}
+    {{scalar_code['generator_expr']|autoindent}}
     {{scalar_code['create_cond']|autoindent}}
-    {{scalar_code['update_post']|autoindent}}
+    {{scalar_code['update']|autoindent}}
 
-    for _i in range(N_pre):
-        _raw_pre_idx = _i + _source_offset
+    for _{{base_var}} in range({{base_var_size}}):
+        _raw{{base_array}} = _{{base_var}} + {{base_offset}}
 
-        {% if not postsynaptic_condition %}
+        {% if not target_condition %}
         {{vector_code['create_cond']|autoindent}}
         if not _cond:
             continue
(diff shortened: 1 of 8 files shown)


thesamovar (approved):
This looks good to me. There's a lot of code and I haven't gone through it line by line in detail, and honestly I'm not sure I'll have the time to do that, so I don't think we should wait. Thoughts?

mstimberg:
Yeah, the PR changes quite a lot. But the changes are all quite straightforward after all. In the templates, it replaces the hardcoded variables `i`, `j`, `_pre_idx`, etc. by template variables `{base_var}`, `{target_var}`, etc. Finally, there is some reorganization of `Synapses.connect` because it became really unreadable with all the `if`/`elif` cascades, but this mostly really moves things around. Then there are of course tests, but these again are just basically copy of the existing ones, going in the other direction. So I'm not *too* worried about merging this.

Maybe one thing to review would be just the basic naming that I put as a comment on top of the templates:
```
    {# For a connect call j='k+i for k in range(0, N_post, 2) if k+i < N_post'
    "j" is called the "target variable" (and "_post_idx" the "target array", etc.)
    "i" is called the "base variable" (and "_pre_idx" the "base array", etc.)
    "k" is called the iteration variable #}
```
I'm not 100% sure that the "taget", "base", "iteration" names are very clear, but I could not think of any obvious alternative either.

thesamovar:
Yeah, I wasn't too sure about the names either, but as long as it's commented I think it's fine. Normally we'd use source/target but I guess we don't want to confuse with the source/target groups? But then I guess that's still a problem for base/target.

Could be independent / dependent variable? But it's not great either since the "dependent" variable is not fully dependent.

Or since there is an implicit loop, inner/outer variable? I'm not a big fan of that either.

mstimberg:
Yeah, I wanted to avoid `source`/`target` because it is a bit odd for `i='j'` if the source index is the "target" and the target index is the "source". But you are right that I currently still use "base" and "target" which is not much better... I though a bit about references to loops, but the confusing thing is that we also have "k" as the "iteration variable" :-/
Dependent and independent actually sounds quite reasonable, in what way would the dependent variable not be fully dependent? Another alternative could be "input variable" and "output variable" (which apparently are sometimes used in statistics as well: https://en.wikipedia.org/wiki/Dependent_and_independent_variables#Statistics_synonyms)?

mstimberg:
Not quite sure I get that. Both are varying, but for independent/dependent variables, both are varying as well it is just that one is varying as a function of the other. Ah, but maybe you mean that it is not simply `i=f(j)` because it results in multiple values for `i`? That indeed stretches the analogy quite far... What do you think about "input" and "output" variable? It is relatively general but it should be pretty clear that the "output" is calculated as a function of the "input".

thesamovar:
How about for my example we would have that j would be the outer index, k would be the inner index and i the result index?

Input/output also feels a bit like it implies 1-1 to me. But I'm also fine with independent/dependent. I think it makes sense too, and maybe less disruptive.

mstimberg:
> How about for my example we would have that j would be the outer index, k would be the inner index and i the result index?

Ok, I think I can live with that :) The most important thing is that it is consistent and documented anyway.

PR #1298, opened 2021-05-03
Testsuite In Github Actions

rohithvarma3000:
I have been working on writing a Github actions workflow that will help to replace the azure pipelines in the future. I am initially working on making this work only ubuntu-20.04 and Python version 3.8. I am marking this as a draft pull request because this file is not working yet due to some issues that need to be resolved.


rohithvarma3000:
Screenshot from 2021-05-04 00-54-32
I am getting this error when trying to run the source activate $HOME/test_env. Does anyone have any idea how to solve this?

rohithvarma3000:
Also, I did not understand the purpose of the echo command in the Add conda to PATH step. Could someone help me out with that as well?

rohithvarma3000:
I have done some testing and I have identified that the error that is appearing above occurs when the `source activate $HOME/test_env` command is run and the coverage command in the Run tests part. For a detailed view, feel free to look into the actions tab in my fork.

rohithvarma3000:
Also, you need to add the coveralls repo token in the secrets tab so that it can be used by GitHub actions to upload the coverage data to coveralls. When you add this token, please update me so that I can name the variable accordingly.

rohithvarma3000:
I have also observed that GitHub actions is slower than Azure DevOps in running the tests.

mstimberg:
Hi Rohith, thanks for working on this and sorry for the late reply! 

Regarding the test failure: this was a bit of a tricky issue. The test checks whether an incorrect compiler flag raises a warning, but internally this check is cached to not check a compiler flag mulltiple times. Only the first failed test will raise a warning. In the Azure test suite, the tests are separated by code generation target, i.e. either only numpy, only Cython, or only C++ standalone tests are run. Each of these test suites will run the test, and since they are independent, they will each start afresh with an empty cache and trigger your warning. In your current github action, you run both numpy and Cython tests, the `test_cpp_flags_support` will therefore be run twice and the second run does not trigger the warning anymore, leading to a failed test.
This is not really your fault, the test should work each time, even if run several times in a row. I just fixed this with a commit in your branch. In the long run, it would be better to run tests separately/in parallel as in the previous test suite, though, since this should lead to a faster test run in total.

Regarding the coveralls secret: I think on github you don't need this, you can directly use the general `GITHUB_TOKEN` as `secrets.GITHUB_TOKEN` (see https://github.com/marketplace/actions/coveralls-github-action).

Note that I removed the Azure configuration from your branch, this is just to free up time on the Azure test server – since we still run the test suite for other PRs on Azure, this avoids that this branch unnecessarily runs its tests over there as well.

I did not yet have time to look at your workflow file in detail, but at a cursory look it looks good. Maybe you could use the coveralls github action that I linked earlier instead of doing it manually with the coveralls command (the `CIRCLE_CI` stuff shouldn't be necessary when running things with Github Action)? And instead of `source activate`, I think nowadays you should use `conda activate` – probably `source activate` will stop working at some point.

Thanks again for working on this!

rohithvarma3000:
Hi @mstimberg ,

Thank you for modifying the test that was failing. I am not sure how to make the tests run in parallel (I am guessing that we can do that by changing the SPLIT_RUN variable).

Regarding `conda activate`, I am not really sure but there seems to be some major problem using `conda activate` instead of `source activate` in bash scripting. It gives a command not found error and suggests including the `conda init` (Refer workflow of Added coveralls action and changed source activate to conda activate). I tried to add a step with the suggested command but it did not fix the issue (Refer workflow of Add conda init step). I looked into the internet but I could not find a satisfactory solution that we can use for Github actions. So for now, I will leave it as `source activate`. If you have any good suggestions regarding how to solve this conda activate issue, I am all ears.

The action that you have suggested for coveralls data can't be used in our case. The coveralls Github actions is used to transport only lcov coverage data. But on the other hand, the `.coverage` that we generate is a SQLite database file. So

rohithvarma3000:
- [ ]

rohithvarma3000:
Hi @mstimberg ,

Thank you for modifying the test that was failing. I am not sure how to make the tests run in parallel (I am guessing that we can do that by changing the `SPLIT_RUN` variable).

Regarding conda activate, I am not really sure but there seems to be some major problem using `conda activate` instead of `source activate` in bash scripting. It gives a command not found error and suggests including the `conda init` (Refer workflow of *Added coveralls action and changed source activate to conda activate*). I tried to add a step with the suggested command but it did not fix the issue (Refer workflow of *Add conda init step*). I looked into the internet but I could not find a satisfactory solution that we can use for Github actions. So for now, I will leave it as `source activate`. If you have any good suggestions regarding how to solve this conda activate issue, I am all ears.

The action that you have suggested for coveralls data can't be used in our case. The coveralls Github actions is used to transport only lcov coverage data. But on the other hand, the `.coverage` generate that we generate is an SQLite database file. So, the action was unable to parse the data (Refer workflow of *Fixed Typo*). So I would suggest using the circle ci support that we are using currently or using a different action. 

Currently, the tests are running. But for some reason, the coverage data is returning that the coverage is zero for all the files. In the workflow run *Testing coverage*, I have output the coverage data in the *Run Tests* step. I don't know how to solve this issue. 

Apart from that, the Upload coverage step is working now and it is able to send the data to coveralls. But since I have not used coveralls before, I don't know if the data has been sent correctly or not. Please check this link (https://coveralls.io/jobs/80143056) for example and verify whether you are getting the expected result (excluding the zero coverage part. the zero coverage is an issue coming from the coverage command).
 Also while bringing the pipeline from azure to Github actions, I have tried my best to use the closest context variables as possible. So I request you to go through the coveralls step and verify wether I have used the correct context variables.

mstimberg:
The devil is really in the details, getting everything exactly right is unfortunately a long and laborious task...

> Thank you for modifying the test that was failing. I am not sure how to make the tests run in parallel (I am guessing that we can do that by changing the `SPLIT_RUN` variable).

In the end, we want to run the tests on different OSes and different Python versions via a [matrix](https://docs.github.com/en/actions/reference/workflow-syntax-for-github-actions#jobsjob_idstrategymatrix) configuration. The default tests also do not run the "standalone" configuration, so this would have to be a separate test run as well (with `STANDALONE: true` in the environment variables). The runtime tests could in principle run all sequentially (for a single Python version and OS), the only reason we have split them up with the `SPLIT_RUN` variable is that otherwise, test runs took too long and ran into timeouts  – we might not actually need this with GitHub Actions any more.


> Regarding conda activate, I am not really sure but there seems to be some major problem using `conda activate` instead of `source activate` in bash scripting. It gives a command not found error and suggests including the `conda init` (Refer workflow of _Added coveralls action and changed source activate to conda activate_). I tried to add a step with the suggested command but it did not fix the issue (Refer workflow of _Add conda init step_). I looked into the internet but I could not find a satisfactory solution that we can use for Github actions. So for now, I will leave it as `source activate`. If you have any good suggestions regarding how to solve this conda activate issue, I am all ears.

This is unfortunately a tricky issue, the shell in GitHub Action does not look at the normal bash configuration file that stores the configuration performed by `conda init`. This can be fixed by using `bash -l {0}` as the default shell. I ended up using this together with the following action, and it seems to work: https://github.com/conda-incubator/setup-miniconda

> The action that you have suggested for coveralls data can't be used in our case. The coveralls Github actions is used to transport only lcov coverage data. But on the other hand, the `.coverage` generate that we generate is an SQLite database file. So, the action was unable to parse the data (Refer workflow of _Fixed Typo_). So I would suggest using the circle ci support that we are using currently or using a different action.

I did not actually mean to suggest that specific GitHub Action (but I also wasn't aware that it was only working for lcov data), I just referred to it to say that you do not need any manual token on GitHub (in contrast to Azure Pipelines). I cleaned things up a bit and it should work now, see e.g. https://coveralls.io/builds/39996561 Note that the paths to the files are not displayed nicely. I think this is fixable, but let's not worry about this for now.

> Currently, the tests are running. But for some reason, the coverage data is returning that the coverage is zero for all the files. In the workflow run _Testing coverage_, I have output the coverage data in the _Run Tests_ step. I don't know how to solve this issue.

This was indeed tricky, I think there was some confusion between the directory `brian2` (i.e. the checked out github directory) and the installed package `brian2` used for running the test. I now used `source_pkgs` instead of `source` in the `.coveragerc` configuration and it seems to work correctly.

rohithvarma3000:
I agree with you, It takes a lot of effort to arrive at a working workflow file.

I am happy to know the new miniconda action is able to fix the `conda activate` issue. Now that the conda issue and coverage issues are fixed, What should be my next step. Shall I start working on the job matrix part? Or is there any other critical issue left that we need to look into?

mstimberg:
> What should be my next step. Shall I start working on the job matrix part? Or is there any other critical issue left that we need to look into?

I think the job matrix is the only major thing left, so please go for it!

rohithvarma3000:
I forgot to mention something. To create a separate file for the commands alone, we need to create a composite run steps GitHub action (similar to template in Azure) in a separate repository. So, we need to make a choice here. If we want a two file system similar to the Azure Pipelines, we need to create a GitHub action. If we want to go with a single file, then we need to put all the OS configurations(Ubuntu, Windows and Mac) inside the same job or we can one job for each OS but in that case we need to copy paste the commands 3 times in the workflow file. What are your thoughts?

Also, I had one question. Should we run the `Indicate completion to coveralls.io` job only once in the entire workflow after the completion of testing of all OS and Python Versions or should we run it after each matrix run?

rohithvarma3000:
@mstimberg I have added the matrix for the different variations of Ubuntu. I need your thoughts on the two files and one file system that I had mentioned above to proceed further.

rohithvarma3000:
@mstimberg I have been experimenting with creating an action for the testing. But since the composite run steps actions is still under development, we are allowed to use only certain keywords. We are not allowed to use `uses`, `if`, and `secrets`. So we will be able to put only the `Install Brian and dependencies` and `Run tests` in the composite run steps Github actions. So I think it will be pointless to have a composite GitHub action right now for only two steps. Maybe once the composite Github actions is improved, we can use it in the project. Apart this the docker github action can be used only in Linux and the javascript GitHub action is slightly hard to implement. So I think implementing the two file system will be hard.  What do you think?

mstimberg:
Hi. I agree we don't need the composite actions if it complicates things. I think we can reduce a lot of copy-and-pasting by making better use of the [matrix feature](https://docs.github.com/en/actions/reference/workflow-syntax-for-github-actions#jobsjob_idstrategymatrix) -- the main combinations should be run via the matrix combinations, and then we only need to add/remove some jobs via the `include`/`exclude` feature (in contrast to our current approach where we define everything manually via `include`). So I think the main "matrix multiplication" should do the 3 OS × `STANDALONE` false and true × 32 bit false and true. Let's ignore the `SPLIT_RUN` thing for now, as I mentioned earlier this was only to avoid time-outs and is maybe not necessary any more. All this should run with our "default" Python version which is 3.8 at the moment.
Then, with the `include` part we only add additional tests for one OS (I think Linux is the best choice) with `STANDALONE` false and true and Python versions 3.7 and 3.9. Or maybe it is easier to separate this into a new job with its own matrix? 
BTW, as a minor side-note: I don't quite remember why we introduced a variable for the script name, but I don't think it makes things easier. I'd propose to remove the variable and to just hardcode the script name where it is used.


> Also, I had one question. Should we run the `Indicate completion to coveralls.io` job only once in the entire workflow after the completion of testing of all OS and Python Versions or should we run it after each matrix run?

Actually, merging the coverage information becomes a mess when it spans several operating systems or Python versions. We should therefore only run the coveralls upload step for one Python/OS version, I'd propose to use Python 3.8 and Linux. The final notification therefore only needs to wait for the Linux tests, but I'm not sure if this is easy to implement with the matrix approach. It's not a big deal if it waits for everything before notifying the completion.

rohithvarma3000:
Hi, I have removed `SPLIT_RUN` as you have requested and I have hardcoded the script name as well. But I have left the script name in the env so that in the future if you need to change the script that needs to be run for a particular job, it can be modified easily.

I have made the matrix as you had suggested above. I have included the matrix of python 3.7 and 3.9 in the job itself. The reason, why I prefer to have a single job, is because it is easier to make changes when necessary. If we have two jobs and copy-paste the statements of jobs into another, then whenever we make a change in one job the same change has to be made in the other job as well. Also, I have excluded the case where both `standalone` and `float_dtype_32` are true because it was not present in the Azure Pipelines. If you want that case as well, then we can just remove the exclude statements in the matrix.

Also, I have ensured that the coverage data is uploaded when the OS is Ubuntu, Python Version is 3.8 and float_dtype_32 is false. And as you have mentioned, it would be slightly harder if we have to write a logic such that the second job runs after the Linux jobs are run. So even I suggest we leave it as it now.

I think the work on this workflow file is done. I will remove the draft status. Please go through the workflow file and suggest any other changes that you may require.

mstimberg:
Many thanks for all the work on this, I wait for the tests to run through and have a final look.

mstimberg:
Hi again. As you can see from the flurry of commits, I battled with Github's yaml syntax to implement what I thought would be a simple thing: have constants for the minimum and maximum Python versions we want to test. I finally managed to get it to work, and I'm now really happy with this PR. I finally also made a slight change compared to what I asked you to implement before: I realized that testing Python 3.8 if we already test 3.9 and 3.7 is a bit of a waste, so now everything is tested with 3.9, and two additional test runs are done for 3.7 to make sure it works in general.
I will wait for the final battery of tests to run through and then go ahead with the merge. Thanks again for your contribution.

mstimberg (approved):

PR #1299, opened 2021-05-04
Fix LaTeX output for newer sympy versions

### LINKED ISSUE #1296: Escaping problem in LaTeX equations
As reported by @schmitts on the [discussion forum](https://brian.discourse.group/t/latex-representation-of-equations/376), something weird is going on in the LaTeX output for equations (escaping of backslashes that should be part of latex commands as `\textbackslash`):
```Python
from brian2 import *
from sympy import latex
G = NeuronGroup(10, 'dv/dt = -v/(10*ms) : volt')
print(latex(G.equations))
```
gives
```
\begin{align*}\mathtt{\text{\textbackslashfrac\{\textbackslashmathrm\{d\}v\}\{\textbackslashmathrm\{d\}t\}}} &= - \frac{v}{10 ms} && \text{(unit of $v$: $\mathrm{V}$)}\end{align*}
```
Individual equastions are rendered correctly, so the problem seems to be in `Equations._latex` where equations are put together.
```Python
print(latex(G.equations['v']))
```
```
\frac{\mathrm{d}\mathtt{\text{v}}}{\mathrm{d}t} = - \frac{v}{10 ms}
```

This seems to be a regression, with very old versions (e.g. 2.1.3.1) it worked.

mstimberg:
Fixes #1296 (see issue for an example of what goes wrong)

Since sympy 1.7, `sympy.latex('...')` will no longer consider a string as "already LaTeX" but rather replace backslashes in it by `\textbackslash` to make it appear as written. With this change, we now only call `sympy.latex` on sympy objects (e.g. Symbol) or our own objects implementing `_latex`.

### DIFF
--- a/brian2/equations/equations.py
+++ b/brian2/equations/equations.py
@@ -458,13 +458,13 @@ class SingleEquation(Hashable, CacheKey):
 
     def _latex(self, *args):
         if self.type == DIFFERENTIAL_EQUATION:
-            return (r'\frac{\mathrm{d}' + sympy.latex(self.varname) + r'}{\mathrm{d}t} = ' +
+            return (r'\frac{\mathrm{d}' + sympy.latex(sympy.Symbol(self.varname)) + r'}{\mathrm{d}t} = ' +
                     sympy.latex(str_to_sympy(self.expr.code)))
         elif self.type == SUBEXPRESSION:
-            return (sympy.latex(self.varname) + ' = ' +
+            return (sympy.latex(sympy.Symbol(self.varname)) + ' = ' +
                     sympy.latex(str_to_sympy(self.expr.code)))
         elif self.type == PARAMETER:
-            return sympy.latex(self.varname)
+            return sympy.latex(sympy.Symbol(self.varname))
 
     def __str__(self):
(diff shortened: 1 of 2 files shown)


thesamovar (approved):
I'm fine with all this as long as it's a little bit backwards compatible?

mstimberg:
> I'm fine with all this as long as it's a little bit backwards compatible?

It should be! For everything where we support LaTeX output (equations, quantities), this fix should make sure you get the same output with sympy < 1.7 and >= 1.7.
The only difference is in output for objects that don't support LaTeX, e.g. calling `sympy.latex(group)` for a `NeuronGroup` would just return a string like`NeuronGroup(clock=Clock(dt=100. * usecond, name='defaultclock'), when=start, order=0, name='neurongroup')` previously, but now it returns a string that can be included in a LaTeX math environment: `\mathtt{\text{NeuronGroup(clock=Clock(dt=100. * usecond, name='defaultclock'), when=start, order=0, name='neurongroup')}}` which a) makes sense and b) is purely a change in `sympy`, so there's not much we can do about it.

thesamovar:
OK!

PR #1301, opened 2021-05-19
Simulation-based inference example

mstimberg:
This adds a new example demonstrating the use of the [sbi](https://www.mackelab.org/sbi/) package with Brian. Developed together with @ybernaerts and @michaeldeistler and very similar to the [Hodgkin-Huxley tutorial](https://www.mackelab.org/sbi/examples/00_HH_simulator/) in the `sbi` documentation – except for using Brian to do the simulation :)

Note that this is a bit of a bare-bones example with not much explanation. I'll convert the more detailed notebook version into a blog post for our website soon.

### DIFF
--- a/examples/advanced/modelfitting_sbi.py
+++ b/examples/advanced/modelfitting_sbi.py
@@ -1,0 +1,174 @@
+"""
+Model fitting with simulation-based inference
+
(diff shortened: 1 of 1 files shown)


thesamovar (approved):
Looks cool! Happy to add this.

PR #1304, opened 2021-05-20
Fix automatic g++ compiler detection on linux

jangmarker:
Disclaimer 1: I'm no expert on make, I just had to fix a local linking error because my system compiler was mixed with the conda environment's standard library and want to contribute this.
Disclaimer 2: I haven't tested this particular change, but I made the same edits to a makefile of an already generated standalone build and it worked there. Relying on continuous integration here :)

From the commit message:

(Ana)conda environments allow to install gcc/g++ compilers and standard libraries.
The compilers are not linked to `gcc` or `g++` but rather preserve their platform dependent naming.
For automatic compiler detection by tools conda sets environment variables.
Details: https://docs.conda.io/projects/conda-build/en/latest/resources/compiler-tools.html#using-the-compiler-packages

The Linux makefile of brian's standalone mode ignores these environment variables because of two reasons:
- CC is overwritten to `@g++`
- CC is used instead of CXX (which is for the C compiler)

Instead, remove the default on CC and use CXX everywhere.
This automatically detects the right C++ compiler in anaconda environments.

For added benefit, use .SILENT instead of the @ prefix.

### DIFF
--- a/brian2/devices/cpp_standalone/templates/makefile
+++ b/brian2/devices/cpp_standalone/templates/makefile
@@ -4,9 +4,8 @@ SRCS = {{source_files}}
 H_SRCS = {{header_files}}
 OBJS = ${SRCS:.cpp=.o}
 OBJS := ${OBJS:.c=.o}
-CC = @g++
 OPTIMISATIONS = {{ compiler_flags }}
-CFLAGS = -c -Wno-write-strings $(OPTIMISATIONS) -I. {{ openmp_pragma('compilation') }} {{ compiler_debug_flags }}
+CXXFLAGS = -c -Wno-write-strings $(OPTIMISATIONS) -I. {{ openmp_pragma('compilation') }} {{ compiler_debug_flags }}
 LFLAGS = {{ openmp_pragma('compilation') }} {{ linker_flags }} {{ linker_debug_flags }}
 DEPS = make.deps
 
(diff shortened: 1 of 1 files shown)


mstimberg:
Many thanks for the PR, I've run into similar problems with the Anaconda compiler but never got around to fix this properly... This looks all good to me, but I'm not entirely convinced we should use `.SILENT`. When we call `make` from Brian, we hide the std output/error by default and only show it when an error occurs. If that is the case, the commands that were executed could be helpful.

Or did this not work for you, and you saw unwanted output when running a standalone simulation?

jangmarker:
Thanks for your reply! I've only included `.SILENT` since in the original code the calls to the compiler were actively silenced with the `@` (see [here](
https://github.com/brian-team/brian2/pull/1304/files#diff-25cbbedeaf35e46a68927bad7aadf44f7ca0b7b3dbeed9ef4313bc2003132de2L7)): 

```
CC = @g++
```

I can remove the `.SILENT` if you want me to, I have no opinion on this.

mstimberg:
> Thanks for your reply! I've only included `.SILENT` since in the original code the calls to the compiler were actively silenced with the `@`

Ah, I see. I think displaying the commands is indeed better, as I said it gets hidden when run from Brian. I'd therefore prefer to remove the `.SILENT`. Could you also merge latest master into your branch? We just switched over the testing infrastructure to GitHub Actions which entails changes to the containers/gcc versions etc. we use for testing.

I can also do these things myself, but you'd have to enable the "Allow edits from maintainers" option.

jangmarker:
I've removed `.SILENT` and also rebased on current master.

mstimberg (approved):
Great, thanks again. All looks good to me, I'll just wait for the tests to run through before merging.

PR #1308, opened 2021-06-07
How to plot model functions

schmitts:
Cf. https://brian.discourse.group/t/plotting-of-arbitrary-functions/399﻿


mstimberg (approved):
Many thanks, this is a nice addition for the docs – I have been thinking of re-organizing things a bit with a "cookbook" section for common tasks, this would be a great fit for that section. But well, one thing at a time. I'll merge this now, but I'll do some additional minor changes (e.g. adding the actual plot to the docs) directly in `master`.

PR #1309, opened 2021-06-08
Hodgkin-Huxley neuron simulation with approximations for gating variable steady-states and time constants

schmitts:
Follows exercise 4, chapter 2 of Eugene M. Izhikevich: Dynamical Systems in Neuroscience


mstimberg (approved):
Great example, covering a lot of advanced Brian usage. The only change I made is moving it to the advanced category, since it is doing quite a lot more than our basic examples.

schmitts:
Thanks! Are the examples part of CI?

schmitts:
Binder seems to be updated as well. https://hub.gke2.mybinder.org/user/brian-team-brian2-binder-oemz28pt/tree/examples/advanced/COBAHH_approximated.ipynb is not available. Anything I can do?

mstimberg:
> Thanks! Are the examples part of CI?

Not in the sense that they are tested automatically (we might have a student working on this later this year, it was supposed to be part of a GSoC project, but unfortunately we did not get the slot). They get only tested when running the `run_examples.py` script locally. It gets automatically rendered as part of the documentation website, though, you can find it here (but you probably saw this already): https://brian2.readthedocs.io/en/latest/examples/advanced.COBAHH_approximated.html

> Binder needs to be updated as well. https://hub.gke2.mybinder.org/user/brian-team-brian2-binder-oemz28pt/tree/examples/advanced/COBAHH_approximated.ipynb is not available. Anything I can do?

Hmm, this is something that we should fix... Binder is caching the full image that `brian2-binder` generates, but one of the steps of the build generation is to clone brian2 so that it gets the latest examples/tutorials. I made an empty commit on the `brian2-binder` repository to force binder to rebuild the image. This seems to have worked, you can access the example here: https://mybinder.org/v2/gh/brian-team/brian2-binder/master?filepath=examples/advanced/COBAHH_approximated.ipynb

However, I just realized that the `argparse` part will not work on binder, maybe it is worth removing the command line handling from the example and just have it show the image?

schmitts:
I could quickly put `run_examples.py` in a github actions? Having examples CIed would be very good. Nothing more frustrating for a new user than a not working example.

I can remove the argparse handling if you wish. Should I?

mstimberg:
> I could quickly put `run_examples.py` in a github actions? Having examples CIed would be very good. Nothing more frustrating for a new user than a not working example.

It is not that straightforward, unfortunately. We don't want to run the examples with every commit, running them all takes  > 2h. Some of the examples also have additional dependencies. One simple approach would be to keep a manual list (or have some kind of marker in the files) of examples that should be excluded from the CI runs. Another option would be to rather have them only, say, once a week.

> I can remove the argparse handling if you wish. Should I?

Thanks, yes, I'd prefer that – it would be nice to not break the binder integration for this.

schmitts:
I've removed the command line parsing in #1314.

What about running the examples at least nightly?

mstimberg:
> What about running the examples at least nightly?

I'd be happy to merge something that runs the examples nightly, but it would have to take care of installing the additional dependencies (I'd have to check, the [OpenCV example](https://brian2.readthedocs.io/en/stable/examples/advanced.opencv_movie.html) might well be the only one). Of course, just running the examples is a very basic test (i.e. it just makes sure that it does not fail with an error)  – but I agree that this is better than nothing.

PR #1311, opened 2021-06-10
fix generate_examples.py: len(title) -> len(file)

obfusk:


mstimberg (approved):
Ha, good catch, thanks!

PR #1312, opened 2021-06-10
make auto_find_examples() output deterministic

obfusk:


mstimberg (approved):
Thanks, makes perfect sense!

PR #1315, opened 2021-06-18
Fix for exc.errisinstance with pytest.raises

mstimberg:
In contrast to what I claimed in #1196, the `ExceptionInfo.errisinstance` check does not actually work with chained exceptions.
Something like
```Python
assert exc_info.errisinstance(exc, KeyError)
```
will fail if exc_info is a `BrianObjectException` caused by a `KeyError`.

We have plenty of tests written like this
```Python
with pytest.raises(BrianObjectException) as exc:
    net.run(0*ms)
    assert exc.errisinstance(UnsupportedEquationsException)
```
These tests run just fine, but only because the test leaves the `pytest.raises` context as soon as an exception is raised, i.e. the `assert` is never triggered. This is mentioned in the [pytest documentation](https://docs.pytest.org/en/6.2.x/reference.html#pytest.raises)...

This PR fixes this by introducing our own testing function which recurses into the causing exceptions. This is not quite as elegant as I'd like (e.g. the error message if the assertion fails is not great), but it should do for now. It is certainly better than assertions that are never evaluated :)

### DIFF
--- a/brian2/parsing/bast.py
+++ b/brian2/parsing/bast.py
@@ -57,7 +57,7 @@ def brian_dtype_from_value(value):
 # so we don't use numpy.issubdtype but instead a precalculated list of all
 # standard types
 
-bool_dtype =numpy.dtype(numpy.bool)
+bool_dtype = numpy.dtype(bool)
 def is_boolean_dtype(obj):
     return numpy.dtype(obj) is bool_dtype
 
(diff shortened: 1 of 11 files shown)


mstimberg:
I'll go ahead with the merge on my own since there's nothing controversial in this PR and it is only affecting the test suite.

PR #1316, opened 2021-06-21
Make clearer that membrane voltage is constant

schmitts:
Remove membrane voltage ODE as voltage is constant
in the sweep.


mstimberg (approved):
Yepp, makes sense!

PR #1317, opened 2021-06-21
Clarify why one wants to add (event-driven)

schmitts:


mstimberg (approved):
Clarifying things is always good :smile:

PR #1320, opened 2021-07-02
Use github-calc-nep29 action instead of hardcoding Python versions

mstimberg:
Small change to use the github action I created in https://github.com/mstimberg/github-calc-nep29 (still needs to be properly documented...). Instead of hardcoding the Python versions we support, it calculates them according to [NEP29](https://numpy.org/neps/nep-0029-deprecation_policy.html).

### DIFF
--- a/.github/workflows/testsuite.yml
+++ b/.github/workflows/testsuite.yml
@@ -3,18 +3,21 @@ name: TestSuite
 on: [push, pull_request]
 
 jobs:
-  set_python_versions:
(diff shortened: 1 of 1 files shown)


thesamovar (approved):
I don't know if you need a review on this or not but it seems fine to me. 😉

mstimberg:
> I don't know if you need a review on this or not but it seems fine to me. wink

always appreciated :smile:

PR #1321, opened 2021-07-06
Fix wrong preference example in documentation

denisalevi:
The preference `codegen.c.compiler` does not exist, guess it was renamed to `codegen.cpp.compiler`?


mstimberg (approved):
Thanks! There were two more instances of `codegen.c` and I also took the liberty to change the example from `gcc` to `unix` :)

Issue #1322, opened 2021-07-06
Setting C++ compiler does not work as documentation would suggest


denisalevi:
The documentation says:
> codegen.cpp.compiler = ''
>
>    Compiler to use (uses default if empty)
>
>    Should be gcc or msvc.

Setting
```
prefs.codegen.cpp.compiler = 'gcc'
```
fails with
```
/home/denis/.conda/envs/brian2-fork/lib/python3.8/site-packages/setuptools/distutils_patch.py:25: UserWarning: Distutils was imported before Setuptools. This usage is discouraged and may exhibit undesirable behaviors or errors. Please use Setuptools' objects directly or at least import Setuptools first.
  warnings.warn(
ERROR      Brian 2 encountered an unexpected error. If you think this is a bug in Brian 2, please report this issue either to the discourse forum at <http://brian.discourse.group/>, or to the issue tracker at <https://github.com/brian-team/brian2/issues>. Please include this file with debug information in your report: /tmp/brian_debug__rfpvc21.log  Additionally, you can also include a copy of the script that was run, available at: /tmp/brian_script_l8q0qgoh.py Thanks! [brian2]
Traceback (most recent call last):
  File "/home/denis/.conda/envs/brian2-fork/lib/python3.8/distutils/ccompiler.py", line 1008, in new_compiler
    (module_name, class_name, long_description) = compiler_class[compiler]
KeyError: 'gcc'

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "example2_gchi_astrocyte.py", line 147, in <module>
    run(duration, report='text')
  File "/home/denis/tubcloud/home/projects/brian2cuda/brian2-fork/brian2/units/fundamentalunits.py", line 2430, in new_f
    result = f(*args, **kwds)
  File "/home/denis/tubcloud/home/projects/brian2cuda/brian2-fork/brian2/core/magic.py", line 373, in run
    return magic_network.run(duration, report=report, report_period=report_period,
  File "/home/denis/tubcloud/home/projects/brian2cuda/brian2-fork/brian2/core/magic.py", line 231, in run
    Network.run(self, duration, report=report, report_period=report_period,
  File "/home/denis/tubcloud/home/projects/brian2cuda/brian2-fork/brian2/core/base.py", line 277, in device_override_decorated_function
    return getattr(curdev, name)(*args, **kwds)
[...]

PR #1323, opened 2021-07-06
Update documentation for C++ compiler prefs

denisalevi:
See #1322

I expected to be able to set the actual compiler binary with this preferences (which is not the case). So I added a sentence on how to set a specific compiler binary (using environment variable `CXX`). But if you think this doesn't belong here, feel free to remove it again.

And I never used Windows makefiles so I couldn't tell from the win_makefile template if there was a Windows equivalent?


mstimberg (approved):
Thanks, this should be useful to others!

PR #1324, opened 2021-07-06
Fix deactivation of spiking objects

### LINKED ISSUE #1319: PoissonGroup continues to fire with active=False
As reported on the [discussion forum](https://brian.discourse.group/t/poissongroup-cannot-be-deactivated-properly/419/2), there seems to be a bug when a `PoissonGroup`'s `active` attribute is set to `False`. If I am not mistaken, units that fired in the very last active time step continue to fire every time step in the deactivated state.
Here's the example provided by Shencong to reproduce the issue:
```Python
group_input_test = PoissonGroup(100,300*Hz)

spk_sti_test = SpikeMonitor(group_input_test, record = True)

net_test = Network(group_input_test, spk_sti_test)

group_input_test.active = False
net_test.run(1000 * ms)
group_input_test.active = True
net_test.run(1000 * ms)
group_input_test.active = False
net_test.run(1000 * ms)

plt.figure()
plt.plot(spk_sti_test.t/ms, spk_sti_test.i,'|')
plt.xlim([0,3000])
plt.xlabel('ms')
plt.ylabel('neuron',fontsize=12)
```

Need to investigate further, but as a first guess I'd say that the data structure storing spikes is not cleared correctly, and since the thresholder is no longer active it will not update the information.

mstimberg:
This fixes #1319 (a group continues to spike after it is deactivated with `active=False`). The reason was quite easy to identify: the threshold notes the spikes in the `spikespace` data structure at every time step, but it never clears it up. So if it has spiked in the last time step of a previous run, deactivating it will simply mean that the `spikespace` never gets cleared, and the object therefore repeats the spikes every time step. 
The fix is relatively straightforward: I added a cleanup step (setting the number of spikes to 0) to the `after_run` part of the thresholder template. Unfortunately, this did not work quite flawlessly, which maybe shouldn't come as a complete surprise since we've never actually used the `after_run` slots (we only needed `before_run` so far)... In particular, `MagicNetwork` (which uses its `after_run` method to clean its internal list of objects to avoid memory leaks), did not call `Network.after_run`, and another issue prevented standalone mode from working correctly.
But well, tests pass now, all looks good :)

### DIFF
--- a/brian2/codegen/runtime/cython_rt/templates/common_group.pyx
+++ b/brian2/codegen/runtime/cython_rt/templates/common_group.pyx
@@ -86,6 +86,9 @@ def main(_namespace):
 {{ cython_directives() }}
 {{ imports() }}
 
+# support code
+{{support_code_lines | autoindent}}
+
 def main(_namespace):
     {{ load_namespace | autoindent }}
     if '_owner' in _namespace:
--- a/brian2/codegen/runtime/cython_rt/templates/threshold.pyx
+++ b/brian2/codegen/runtime/cython_rt/templates/threshold.pyx
@@ -32,3 +32,8 @@
     {{_eventspace}}[N] = _cpp_numevents
     
 {% endblock %}
+
+{% block after_code %}
+    {% set _eventspace = get_array_name(eventspace_variable) %}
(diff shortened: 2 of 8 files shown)


thesamovar (approved):
All looks fairly straightforward I think. Nice fix!

PR #1328, opened 2021-07-13
Improve dependency checks for clock-driven ←→ event-driven

mstimberg:
Before merging #1274, the checks for event-driven vs. clock-driven and summed variables weren't strict enough. Unfortunately, I realized that the current code is too strict and does not allow e.g. a dependency of an event-driven variable on a parameter (which should obviously work). I tried fixing this in the existing code, but things got complicated quickly. I decided instead to take a bit more basic approach and add a `dependencies` property to the `Equations` class, which calculates the dependencies of all variables, annotated with additional information if this dependency is indirect via a subexpression.
@rohithvarma3000 apologies for removing some of your code with this commit, I should have thought of these corner cases when working with you on #1274. If you have time, could I ask you to have a look at the current code and/or run it to see whether 1) the code is understandable and 2) whether it fails to work for situations where your previous code was working? Thanks (and of course no worries if you don't have the time)!


rohithvarma3000:
Sure @mstimberg. I will test this code next week and let you know.

brian2/equations/equations.py
```diff
@@ -930,6 +931,55 @@ def _sort_subexpressions(self):
...
                 eq.update_order = len(sorted_eqs) + 1
 
+    @property
+    def dependencies(self):
+        '''
+        Calculate the dependencies of all differential equations and
+        subexpressions.
+        '''
+        deps = {}
+        for eq in self._equations.values():
```
rohithvarma3000, line 941: Please add a few comments to understand why you have used this for loop and the purpose of putting a " ' " for the differential equations

rohithvarma3000 (commented):
@mstimberg I have tested the code against the test cases I had worked on before and the code works as expected. Initially, I had a bit of difficulty in understanding how the code for the dependencies works, but after I printed the data returned by the dependencies function I have a got a basic understanding of how it works. I have put a comment in one location where you can add a few more comments to improve the readability of the code. Otherwise, the code is good to be merged.

mstimberg:
Thanks for the reivew @rohithvarma3000 and sorry for taking so long to get back to this (I submitted this PR right before going away for a long vacation :) ). I added a long comment to the admittedly complicated code and also added a specific test for the new `Equations.dependencies` function (previously it was only tested indirectly via its use for checking event-driven equations).
I'll therefore go ahead with the merge, thanks again for looking at this :+1:

PR #1333, opened 2021-08-06
Fix Python system prefix appending in C++ Standalone

denisalevi:
The `sys.prefix` was previously prepended to the `include_dirs`, `libraries` and `runtime_library_dirs`. But I think it should be appended instead, right? Otherwise setting `prefs.codegen.cpp.include_dirs/libraries/runtime_library_dirs` would have no effect if a library with the same name exists in the `lib/include` folder of the `sys.prefix` path.

I also updated the docs for `runtime_library_dirs`. But the `sys.prefix` is only added for linux systems, right? Not sure if you want that in the docs.

### DIFF
--- a/brian2/codegen/cpp_prefs.py
+++ b/brian2/codegen/cpp_prefs.py
@@ -172,6 +172,9 @@ prefs.register_preferences(
         default=[],
         docs='''
         List of directories to search for C/C++ libraries at run time.
+        Note that in on Linux platforms, ``$prefix/lib`` will be appended to the end
+        automatically, where ``$prefix`` is Python's site-specific directory prefix as
+        returned by `sys.prefix`.
         '''
     ),
     libraries=BrianPreference(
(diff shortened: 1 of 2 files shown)


mstimberg:
The reason we add `sys.prefix` is that it would otherwise not pickup the libraries/header files from a conda installation. Not quite sure why we prepended it instead of appending it, though. I'm wondering: probably we should simply set `sys.prefix` as the default, but have any settings by the user overwrite it. Presumably, if you have a reason to set these preferences, you know what you are doing and are not relying on some automatic addition.

denisalevi:
Yes, I figured that is what it should do (picking up the libraries in the conda environment). I ran into problems when using nvcc configured to use the system gcc, which then uses the gcc libraries/headers from the conda environment, which might have a different gcc version installed than the system, leading to errors.

I managed to configure brian2cuda to work well in conda environments. For brian2genn, some changes would be needed to make it work. I found some workaround by modifying PATH and CPATH to get brian2genn run in my conda environments and didn't pursue it further. But I think with a few changes brian2genn could also play nicely with conda environment, but that is maybe a different topic.

I agree that being able to removing the libraries in `sys.prefix` would be a good solution as well. In fact, it would have solved the problem I had even better than adding additional library paths :D

PR #1336, opened 2021-08-13
Add standalone project directory path to pytest report

denisalevi:
The standalone project directory is not deleted for failed tests. This commit adds a section to the pytest error report that prints this directory. Helps with debugging.

This checks first if the project directory is even set, which is not the case for RuntimeDevices and standalone tests that failed before code generation. I first considered checking the test markers to decide when to print this information, but I guess checking for `project_dir is not None` is the easiest?

### DIFF
--- a/brian2/conftest.py
+++ b/brian2/conftest.py
@@ -6,7 +6,7 @@ import re
 import numpy as np
 import pytest
 
-from brian2.devices import reinit_devices
+from brian2.devices import reinit_devices, get_device
 from brian2.units import ms
 from brian2.core.clocks import defaultclock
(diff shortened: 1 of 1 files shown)


denisalevi:
I just saw now that the project directory is already appended to the `RuntimeError` raised by `CPPStandaloneDevice.run`, but only if the run itself fails. The section I added will also be included in the test report if the standalone run succeeded but the tests failed.

mstimberg (approved):
This is a really nice feature. I just changed the access to `project_dir`, since this attribute doesn't exist in `RuntimeDevice` and this would lead to a nasty "internal error" in the test suite for failed runtime tests.

PR #1338, opened 2021-08-30
Add prefs to change make and run commands in C++ Standalone

denisalevi:
I am running my test suite on a cluster with a job scheduler. Being able to change the `make` and `run` commands allows me to speed up testing significantly e.g. by using `qmake`, which dynamically allocates jobs on the cluster for compilation or by submitting `./main` to a GPU node via `qsub <optiona> main`, which allows me to use non-GPU nodes for compilation. `qmake` and `qsub` are command from the grid engine that does the job scheduling. But this could of course be used similarly for other job schedulers such as `slurm`.

I could of course do this for brian2CUDA specifically, but since I am inheriting the `run` and `compile_source` functions from `CPPStandaloneDevice`, I thought I'll add it there instead.

Should be backward compatible changes. Feel free to change anything you want. :)

### DIFF
--- a/brian2/devices/cpp_standalone/device.py
+++ b/brian2/devices/cpp_standalone/device.py
@@ -65,6 +65,21 @@ prefs.register_preferences(
         neurons. Now, its value is ignored.
         '''
         ),
+    make_cmd_unix=BrianPreference(
+        default='make',
+        docs='''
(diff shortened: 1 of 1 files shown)


mstimberg (approved):
Very nice!

PR #1342, opened 2021-09-02
Fix `prefs.reset_to_defaults()` which is currently failing with `TypeError`

denisalevi:
I modified the `test_globalprefences` test to test `prefs.reset_to_defaults()`.

The function call resulted in a `TypeError` due to using `BytesIO` (change introduced in 01050fc0c). Changed it back to `StringIO`, which fixes it.

Not sure if this is an issue specific to the Python version, but the test suite will show I guess. I was running with Python 3.9.6.

EDIT:
The `TypeError` was the following:
```
  File "<string>", line 13, in <module>
  File "/path/to/brian2cuda/tests/features/cuda_configuration.py", line 308, in before_run
    brian2.prefs.reset_to_defaults()
  File "/path/to/brian2/brian2/core/preferences.py", line 460, in reset_to_defaults
    self.read_preference_file(BytesIO(self.defaults_as_file))
TypeError: a bytes-like object is required, not 'str'
```

### DIFF
--- a/brian2/core/preferences.py
+++ b/brian2/core/preferences.py
@@ -6,7 +6,7 @@ object ``prefs``.
 import re
 import os
 from collections.abc import MutableMapping
-from io import BytesIO
+from io import StringIO
 
 from brian2.utils.stringtools import deindent, indent
 from brian2.units.fundamentalunits import have_same_dimensions, Quantity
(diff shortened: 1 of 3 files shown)


mstimberg:
I was suprised that I never fixed this before because we *do* reset the preferences at the beginning of the test suite. Turns out, the tests uses `prefs.read_preference_file(StringIO(prefs.defaults_as_file))` instead of calling `reset_to_defaults` :smile:. I'll change this for less redundancy.

mstimberg (approved):

PR #1345, opened 2021-09-16
Remove erroneous list operation

adiehl96:
With the current version "Brian2==2.4.2", the last cell doesn't properly execute, as weights now has to have the same shape as x, which it can only have if it's an np array.

### DIFF
--- a/tutorials/1-intro-to-brian-neurons.ipynb
+++ b/tutorials/1-intro-to-brian-neurons.ipynb
@@ -788,7 +788,7 @@
     "\n",
     "run(duration)\n",
     "\n",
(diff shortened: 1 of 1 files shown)


mstimberg (approved):
Many thanks for the fix! I think the story is slightly more complex, though: the weights argument accepts anything "array-like", so a list instead of an array should be fine in general. I think the issue here has to do with the `hist` feature to receive multiple groups of data as arrays in a list. Either way, I don't quite remember why the `list` was there in the first place, and of course we want this to work with modern matplotlib versions.
Thanks again, merging this now.

PR #1347, opened 2021-09-17
This example reproduces Jansen and Rit's 1995 single-column, firing-rate model

rat-h:
Equations are the system of differential equations number (6) in the original paper.
The rate parameters $a=100 s^{-1}$ and $b=200 s^{-1}$ were changed to excitatory $\tau_e = 1000ms/a =10ms$ and inhibitory $\tau_i = 1000ms/b =20ms$ time constants.
Units were removerd from parameters $e_0$, $v_0$, $r_0$, $A$, $B$, and $p$ to stop Brian's confusion.


mstimberg:
Thanks again! Note that you need to git commit the new reference to the submodule as well (the link to the submodule is basically the name of a commit in a special file). But the whole process is a mess, I will soon open a bunch of issues to look for contributors for our examples/documentation infrastructure.

Note that your example uses Markdown in the documentation, which does not render correctly at the moment in the online documentation (it would have to use restructured text syntax for that). On the other hand, markdown *does* render correctly in the jupyter notebook that you can launch with the "binder" button, and restructured text wouldn't work there, so there isn't a great solution. Part of the revamp of our documentation infrastructure would be using something like `nbsphinx` for our examples, this would take care of rendering things correctly.

PR #1353, opened 2021-09-28
Compiler dirs prefix

mstimberg:
See discussion in #1333. The `sys.prefix` setting was prepended to the preferences for `include_dirs`/`library_dirs`/`runtime_library_dirs` in standalone mode, but appended in Cython (and `runtime_library_dirs` was completely ignored in Cython).

This is now uniform, the `sys,prefix` setting is simply the default for the respective preference, so the user can overwrite it completely if they prefer.

### DIFF
--- a/brian2/codegen/cpp_prefs.py
+++ b/brian2/codegen/cpp_prefs.py
@@ -111,6 +111,11 @@ else:
     else:
         default_buildopts = ['-w']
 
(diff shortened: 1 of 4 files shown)


mstimberg:
@denisalevi This should work for you, right?

denisalevi:
Yes, works for me! Clean solution :)

PR #1354, opened 2021-09-29
Remove default preferences file

mstimberg:
This PR removes the `default_preferences` file, that we previously created during the release process and included in the source tarballs (unless we forgot, e.g. version `2.3.0.2` doesn't have it). The original idea was that one could use this file to change the installation-wide preferences (similar to files in `/etc` on Unix-like file systems).
There are two reasons for not shipping/reading this file any more:
1. When running Brian from its source directory during development, you could get a mismatch between the preference definitions in the Brian source code and in the `default_preferences` file. This would either lead to unexpected values (for changes in the defaults), or to error messages (for added/removed preferences).
2. The generated `default_preferences` file is generated during build-time, which makes it more difficult than necessary to have installation-specific default values (e.g. depending on the operating systems or things like `sys.prefix`). E.g., it does not work with the simplifications I made in PR #1353.

I don't see many use cases for installation-wide preferences, in the first place. If one wants to set a preference for several projects, one can use the user preferences in `~/.brian`. The only use case I see for preferences bound to the Brian installation is for compiler-related settings in conda environments, where for some of them you might like to use the compilers from within the environment (our default) and in others you'd like to use the system compiler. But since such settings will be overwritten with an update of Brian (since pip/conda do not do configuration file management) you could as well change the default settings directly in the code... I strongly doubt that many users relied on that feature.

Either way, this change should not introduce any backwards-incompatibility apart from breaking code that relies on non-default settings in `default_preferences` with the update. But then, this would be true for *any* update, since previously we'd have simply overwritten `default_preferences`. To be safe, I also added a warning if a `default_preferences` file exists.

Finally, one of the possible uses of the `default_preferences` file was to have a template file that you could copy as `~/brian/user_preferences` and then customize it. As an alternative, I now display a full preference file in the documentation (folded by default with HTML `<details>`) which one could use for copy&pasting.


mstimberg:
As I said in my previous comment, this change should not break default preferences any more than any other release. I'll therefore go ahead with the merge.

PR #1377, opened 2021-12-06
Avoid that formatted strings discard units of quantities

mstimberg:
The big string formatting unification in d3ae59251c753ae introduced a regression in a few places, because the default formatting for quantities used in f-strings or `.format` is to display it as a floating point number, discarding its units:
```pycon
>>> start = 0*second
>>> duration = 1 * second
>>> print("Starting simulation at t=%s for a duration of %s" % (start, duration))
Starting simulation at t=0. s for a duration of 1. s
>>> print(f"Starting simulation at t={start} for a duration of {duration}")
Starting simulation at t=0.0 for a duration of 1.0
```
In most places, I worked around this by using something like `"... t={start!s}"` to manually force a conversion to a string, but I think it would be better if we changed quantities' default for "formatting without format spec" (which would also be more consistent with a simple `print(start)`). Luckily, we can do this easily using the `__format__` function. With this PR, both print statements above give the same output.

### DIFF
--- a/brian2/tests/test_units.py
+++ b/brian2/tests/test_units.py
@@ -258,6 +258,15 @@ def test_str_repr():
 
 
 @pytest.mark.codegen_independent
+def test_format_quantity():
+    # Avoid that the default f-string (or .format call) discards units when used without
+    # a format spec
+    q = 0.5*ms
+    assert f"{q}" == f"{q!s}" == str(q)
+    assert f"{q:g}" == f"{float(q)}"
+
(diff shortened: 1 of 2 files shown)


mstimberg:
I'm going ahead with the merge as it should be low risk (and fixes the confusing output in a number of places, e.g. for `run(...report='text')`, or `print(profiling_summary())`. It would also be trivial to revert if we find any issue with it.

PR #1378, opened 2021-12-15
use graphlib2 for toplogical sorting

adriangb:
this is out of personal curiosity, I want to see if CI passes and if it has an impact on performance, I don't necessarily propose that this get merged or anything like that

### DIFF
--- a/brian2/utils/topsort.py
+++ b/brian2/utils/topsort.py
@@ -1,4 +1,4 @@
-from graphlib import TopologicalSorter
+from graphlib2 import TopologicalSorter
 
 
 def topsort(graph):
--- a/brian2/utils/topsort.py
+++ b/brian2/utils/topsort.py
@@ -1,4 +1,4 @@
-from copy import copy
+from graphlib import TopologicalSorter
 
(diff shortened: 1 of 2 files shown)


mstimberg:
Hi @adriangb. You don't actually have to run a pull request to run CI, the tests are run on your own fork as well: https://github.com/adriangb/brian2/actions. The "Build and Publish..." workflow currently fails (since it cannot upload a brian2 package to pypi), we should probably make it fail gracefully on forks.

adriangb:
Yep yep, I had to enable it manually but then it ran. I this used anywhere performance critical?

mstimberg:
> I this used anywhere performance critical?

No. It is used at the beginning of a simulation to sort variables in equations according to their dependencies, but models typically only have "a few" (let's say 1–20) variables. E.g. for the most complex models in our examples the call to toposort currently takes ~0.1ms, so I don't think this is  something we need to optimize as a priority :-)

adriangb:
Cool. It was great to add least be able to test wheels on all of the platforms you test on. Thanks!

PR #1381, opened 2022-01-12
Fixed missing return in StateMonitor.__repr__

kernfel:
This bug was introduced in 9e862585 and caused string representations of StateMonitors (and Networks containing them, consequently) to fail.


kernfel:
Sneakily adding another minor repr fix... missing closed parenthesis.

mstimberg (approved):
Many thanks!

PR #1385, opened 2022-02-08
Firing patterns in the adaptive exponential integrate-and-fire model

schmitts:


mstimberg:
Rendered version is now online: https://brian2.readthedocs.io/en/latest/examples/frompapers.Naud_et_al_2008_adex_firing_patterns.html

PR #1386, opened 2022-02-08
Link to possible values for when

schmitts:


schmitts:
Do we like that? If yes, I could add it to the other occurences of `when=`.

mstimberg (approved):
Thanks, looks great. As a very minor point: for future commits/PRs that only concern documentation, consider adding `[ci skip]` to the commit message (ideally in a new line so it does not clutter the main message) – this way it won't needlessly run our full test suite.

PR #1390, opened 2022-03-02
Add `before_run` & `after_run` code slots for `insert_code` mechanism

denisalevi:
I used these for my benchmarks, would like to get this out of the patch file, also for future benchmarking. Any objections? :)

### DIFF
--- a/docs_sphinx/user/computation.rst
+++ b/docs_sphinx/user/computation.rst
@@ -143,12 +143,7 @@ something like this may not behave as you would like::
 
 The current C++ standalone code generation only works for a fixed number of `~Network.run` statements, not with loops.
 If you need to do loops or other features not supported automatically, you can do so by inspecting the generated
-C++ source code and modifying it, or by inserting code directly into the main loop as follows::
-
-    device.insert_code('main', '''
-    cout << "Testing direct insertion of code." << endl;
-    ''')
-
+C++ source code and modifying it, or by inserting code directly into the main loop as described below.
 
 Variables
 ~~~~~~~~~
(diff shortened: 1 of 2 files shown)


mstimberg:
No general objection of course, but I am a bit worried about potential confusion around the names. We already use `before_run` and `after_run` for the code blocks for individual code objects that are executed once for each run (`Network.before_run` calls `BrianObject.before_run` for each object in the network which in turn calls `CodeObject.before_run` – in standalone mode this generates the code in the `before_code` slots). 
Maybe it would be clearer to call the `insert_code` slots `before_network_run` and `after_network_run`? It *might* also be a good idea to actually document the [`insert_code`](https://brian2.readthedocs.io/en/stable/reference/brian2.devices.cpp_standalone.device.CPPStandaloneDevice.html#brian2.devices.cpp_standalone.device.CPPStandaloneDevice.insert_code) mechanism, but I'm not sure you want to take up that task :wink:

denisalevi:
`before_network_run` makes sense. I can write a quick section in the docs. Where would you like to have it?

mstimberg:
> I can write a quick section in the docs. Where would you like to have it?

I think it would make most sense here: https://brian2.readthedocs.io/en/stable/user/computation.html#standalone-code-generation (i.e. in `docs_sphinx/user/computation.rst`).

A new section "custom code injection" which mentions `device.insert_code` and the supported slots.

denisalevi:
I made a suggestion. Maybe a bit too elaborate? Feel free to modify to your liking. Hope I got the formatting right with the `rst` files. I also added links to examples in Brian2CUDA and Brian2GeNN benchmarks.

docs_sphinx/user/computation.rst
```diff
@@ -179,6 +174,58 @@ threads used in the simulation.
 
 .. _standalone_custom_build:
```
mstimberg, line 175: This will associate the `standalone_custom_build` reference with the new "Custom code injection" section – it should move down to the respective section.

mstimberg (approved):
All looks good to me, only waiting for the tests to run through.

denisalevi:
@mstimberg I just realized that in brian2GeNN, it is also called `before/after_run`... I think you used that in the brian2genn benchmarks. Should probably be renamed there as well to `before/after_network_run`?

PR #1391, opened 2022-03-02
Add timers for compilation and binary execution

denisalevi:
I used these for benchmarks as well. I find this information quite useful. I would even go as far as printing it out explicitly when running with `profile=True`. Or even when not running with it. At least for Brian2CUDA this could be quite important to evaluate the usefulness of using a GPU or not. Or to track down why things go slow.

I just copied what I used in the patch file. It adds a bit of noise to the code and it adds the overhead of 4 `time.time()` calls. If you want, I can make this optional, only when `profile == True` or anything like that? But I doubt this generates much simulation overhead, no?

Oh, and I only added it to the CPPStandaloneDevice I just realized. If you want this, I can also add it to the other devices of course?

### DIFF
--- a/brian2/devices/cpp_standalone/device.py
+++ b/brian2/devices/cpp_standalone/device.py
@@ -7,12 +7,12 @@ import shutil
 import subprocess
 import sys
 import inspect
-import struct
 from collections import defaultdict, Counter
 import itertools
 import numbers
 import tempfile
 from distutils import ccompiler
+import time
 
 import numpy as np
 
@@ -198,6 +198,12 @@ class CPPStandaloneDevice(Device):
                            'after_start': [],
                            'before_end': [],
                            'after_end': []}
+
+        #: Dictionary storing compile and binary execution times
+        self.timers = {'run_binary': None,
+                       'compile': {'clean':None,
+                                   'make': None,
+                                   'all': None}}
 
         self.clocks = set([])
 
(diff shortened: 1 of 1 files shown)


mstimberg:
I don't mind "once-per-run" `time.time()` calls, these won't add any measurable overhead. We cannot really add the same functionality to the runtime devices, though, since there is no separate compilation/execution phase – each code block gets compiled on-demand when needed (I mean we *could* measure this, but it would mean that this information gets propagated back from the depths of the code generation machinery). I would not add any print output by default, but it could definitely be a debug message. I won't have time to look into this further this week, but the binary execution time is also equivalent to the `_last_runtime` attribute of `CPPStandaloneDevice` I guess?

brian2/devices/cpp_standalone/device.py
```diff
@@ -199,6 +199,12 @@ def __init__(self):
                            'before_end': [],
                            'after_end': []}
 
+        #: Dictionary storing compile and binary execution times
+        self.timers = {'run_binary': None,
+                       'compile': {'clean':None,
```
mstimberg, line 204:
```suggestion
                       'compile': {'clean': None,
```
I see you did not use `black` here :wink:

brian2/devices/cpp_standalone/device.py
```diff
@@ -954,13 +960,24 @@ def compile_source(self, directory, compiler, debug, clean):
...
                         if clean:
+                            start_time = time.time()
                             os.system(f'{make_cmd} clean > NUL 2>&1')
+                            self.timers['compile']['clean'] = time.time() - start_time
+                        start_time = time.time()
                         x = os.system(f'{make_cmd} {make_args}>>winmake.log 2>&1')
+                        self.timers['compile']['make'] = time.time() - start_time
+                    self.timers['compile']['all'] = sum(
+                        t for t in self.timers['compile'].values() if t is not None
+                    )
```
mstimberg, line 980: This seems a bit unnecessarily complex to sum up the two values. I don't think we need to store "all" in the dictionary, since it is easy to calculate by summing up "clean" and "make", and in most cases "make" is what we are interested in mostly.

mstimberg (changes requested):
This all looks good (2 minor comments), thanks again. I'd add a `DEBUG` level log output to the end, stating the values. In the long run, it would probably make sense to integrate everything into the profiling info.

brian2/devices/cpp_standalone/device.py
denisalevi: I'm not using black yet because it would reformat all of Brian's code base and this commit would have been a bit of a mess :smile:. But you know, if there was some automatic code formatting like black in place, this would have not happened 😜

brian2/devices/cpp_standalone/device.py
denisalevi: Sure, just did this for the benchmarks (had to happen somewhere). I'll make changes next week!

mstimberg: > I'll make changes next week!

:+1: Nothing very urgent, obviously.

mstimberg (approved):
I quickly finished up this PR with the things we discussed. While doing this I noticed that the PR measured make/make clean times only for Unix, and binary execution time only for Windows :upside_down_face: I think all is working correctly now, but maybe have a quick look again @denisalevi .

denisalevi:
Ah, thanks Marcel! Sorry, lost track of this one.. Looks good!

PR #1392, opened 2022-03-02
Fix docstrings of `event_trains` and `spike_trains`

SimonAltrogge:
The descriptive part of the docstrings of the `EventMonitor.event_trains` and `SpikeMonitor.spike_trains` methods wrongly indicated that a dictionary mapping _event_/_spike_ indices to event/spike times is returned. What is actually returned instead is a dictionary mapping _neuron_ indices to event/spike times. This commit fixes these mistakes by changing "_event_/_spike_ indices" to "_neuron_ indices".


mstimberg (approved):
Makes perfect sense, thanks!

PR #1393, opened 2022-03-08
Fix indentation

bdevans:


mstimberg:
Hi @bdevans, just a quick check: this is just a formatting issue, right? It's not that the docker file does not work like that?

bdevans:
That's correct - I'm just being fussy about a convention!

mstimberg:
Ok :)

PR #1395, opened 2022-03-16
Fix using indexed morphology for threshold_location

mstimberg:
As @romainbrette noticed, the `threshold_location` argument of `SpatialNeuron` did not accept a morphology, as in 
```Python
neuron = SpatialNeuron(..., threshold_location=morpho.axon[15*um])
```
We explicitly mention this syntax in the [documentation](https://brian2.readthedocs.io/en/stable/user/multicompartmental.html#detecting-spikes), so it should work.
The workaround was to either directly provide the compartment number or use `morpho.axon.indices[15*um]`.

This PR fixes the issue, and also adds a corresponding test to make sure this error does not come up again.


mstimberg:
Going ahead with the merge myself, since it is a quite safe fix, and it is unlikely to break anything that was not already broken.

PR #1398, opened 2022-04-04
Add comment that x and y are user-defined parameters

schmitts:
[ci skip]


mstimberg (approved):
Good point, sometimes user got indeed confused about this.

PR #1402, opened 2022-04-06
Fix documentation wrt. refractory period as variable

schmitts:
NeuronGroup(1, model='''refractory : second''', refractory='refractory')

yields:

SyntaxError: The name 'refractory' is used in the refractory mechanism  and should not be used as a variable name.

But

G = NeuronGroup(1, model='''ref : second''', refractory='ref')

works.


mstimberg (approved):
Indeed, thanks for fixing this! I guess we did not reserve the `refractory` name yet when we wrote this part of the documentation…

PR #1405, opened 2022-04-24
Mention the option to install Brian via Spack

schmitts:
[ci skip]


mstimberg (approved):
Looks good in general, thanks! Does spack also support updating packages to a newer version, or is this something that is against the spirit (e.g. you should create a new environment in that case)? In either case, it would be good to have a note about this in the section below.

schmitts:
You would install a new package and load it. You can have multiple versions "installed" in parallel and load on-demand the version you would like to have. I thought about writing more about that, but whoever wants to use Spack is better advised to read the current  documentation of Spack itself.

mstimberg:
> You would install a new package and load it. You can have multiple versions "installed" in parallel and load on-demand the version you would like to have

Ok, I see. I agree it is probably best to simply not mention it in that case.

PR #1406, opened 2022-04-26
Make profiling documentation even more obvious

schmitts:


mstimberg (approved):
Thanks @schmitts.
I slightly changed the docs, since I found "your magic network goes here" a bit confusing, since the paragraph above introduces the code as an example for the CUBA network. I therefore moved the comment about the magic network into the text. For the `Network` example, we need to `# doctest: +SKIP` the `>>> Network(...)` line since it will otherwise try to execute it.

From my side this is ready to merge, but let me know if you find something unclear.

schmitts (commented):
Looks good to me! Thanks, @mstimberg!

PR #1408, opened 2022-04-29
Reproducing figure 2 of Maass, Natschlaeger and Markram (2002)

schmitts:
Takes about 10 minutes on 8 i7 cores @ 3 GHz

### DIFF
--- a/examples/frompapers/Maass_Natschlaeger_Markram_2002.py
+++ b/examples/frompapers/Maass_Natschlaeger_Markram_2002.py
@@ -143,11 +143,11 @@ def get_neurons():
         N_NEURONS,
         """
         tau_mem : second (shared, constant)
-        tau_refrac : second (shared, constant)
+        tau_refrac : second (constant)
         v_reset : volt (shared, constant)
         v_thresh : volt (shared, constant)
         I_b : ampere (shared, constant)
-        tau_stimulus : second (shared, constant)
+        tau_stimulus : second (constant)
         I_syn_ee_synapses : ampere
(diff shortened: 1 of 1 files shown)


schmitts (commented):
That plot does not look right. I'll attach what it looks for me:
[Maas_Natschlaeger_Markram_2002.pdf](https://github.com/brian-team/brian2/files/8619798/Maas_Natschlaeger_Markram_2002.pdf)

mstimberg:
Yes, apologies. I noticed my mistake after pushing and before heading out for lunch and now a meeting. I'll fix it later today (I marked a few variables as `shared` that I should not have – they only have scalar values, but they are different for excitatory and inhibitory cells, so still need to be stored as an array).

mstimberg:
I updated the code and figures, looks good to me now. Just a comment on the two changes I did: if parameters in the equations are representing scalar constants (which I agree can be more convenient than using `namespace`, etc.), it is best to use the `(scalar, constant)` flag. For an equation like `dv/dt = -v/tau`, marking `tau` as a shared variable means that Brian can calculate `exp(-dt/tau)` once, instead of doing it once for every neuron (since each neuron could have a different value for `tau`). I don't think it makes any measurable difference here, but it is a good thing to keep in mind (same as for simply `(constant)` for values that don't change during a simulation). The second change was adding an `if __name__ == '__main__'` block – this is necessary when using multiprocessing on Windows, since every process imports the main script, which would lead to an infinite recursion.
All that said, thanks again for a nice example!

schmitts (commented):
LGTM!

PR #1410, opened 2022-05-05
Fix and sync report_callback documentation

schmitts:
[ci skip]


mstimberg (approved):
Oh, that was indeed outdated, thanks! I also removed the mention of the `graphical` option, which never existed in Brian 2...

PR #1411, opened 2022-05-05
Reproducing figure 2, panel DP of Graupner, Brunel (2012)

schmitts:


mstimberg (approved):
Thanks, looks good to me :+1:

PR #1412, opened 2022-05-05
Handle .ndim and [:, np.newaxis] on VariableView

### LINKED ISSUE #1409: [bug] error while plotting spikes with SpikeMonitor
Running an example from the [tutorial](https://brian2.readthedocs.io/en/stable/resources/tutorials/1-intro-to-brian-neurons.html):

```python
from brian2 import *
start_scope()

N = 100
tau = 10*ms
eqs = '''
dv/dt = (2-v)/tau : 1
'''

G = NeuronGroup(N, eqs, threshold='v>1', reset='v=0', method='exact')
G.v = 'rand()'

spikemon = SpikeMonitor(G)

run(50*ms)

plot(spikemon.t/ms, spikemon.i, '.k')
xlabel('Time (ms)')
ylabel('Neuron index')
```

I get an error on plotting

```
    497 if x.shape[0] != y.shape[0]:
    498     raise ValueError(f"x and y must have same first dimension, but "
    499                      f"have shapes {x.shape} and {y.shape}")
--> 500 if x.ndim > 2 or y.ndim > 2:
    501     raise ValueError(f"x and y can be no greater than 2D, but have "
    502                      f"shapes {x.shape} and {y.shape}")
    503 if x.ndim == 1:

AttributeError: 'VariableView' object has no attribute 'ndim'
```

my brian2 installed from conda on mac m1 is on version `2.5.0.3`.

mstimberg:
These are used when handing a `VariableView` directly to matplotlib, e.g. for `spike_mon.i` (an expression like `spike_mon.t/ms` directly results in an `ndarray` and is therefore not concerned). This problem was introduced with matplotlib 3.5.2, previous versions converted `VariableView` along the way with a call to `np.asanyarray`.

The fix is a minimal workaround, but we have plenty of examples, including the tutorials, that will fail plotting, so fixing this is urgent. For a proper fix I'd like to think a bit more about how (and whether) to support features like adding array dimensions, in particular for synaptic variables which support a form of 2d-indexing which is not the same as for a 2d `ndarray`.

Fixes #1409

### DIFF
--- a/brian2/core/variables.py
+++ b/brian2/core/variables.py
@@ -1,4 +1,3 @@
-
 """
 Classes used to specify the type of a function, variable or common
 sub-expression.
@@ -1091,6 +1090,16 @@ class VariableView(object):
         elif (isinstance(item, slice) and item == slice(None)
               and self.index_var == '_idx'):
             indices = slice(None)
+        # Quick fix for matplotlib calling 1-d variables as var[:, np.newaxis]
+        # The test is a bit verbose, but we need to avoid comparisons that raise errors
+        # (e.g. comparing an array to slice(None))
+        elif (isinstance(item, tuple) and
+              len(item) == 2 and
+              isinstance(item[0], slice) and
+              item[0] == slice(None) and
+              item[1] is None and
+              self.index_var == '_idx'):
+            return variable.get_value()[item]
         else:
             indices = self.indexing(item, self.index_var)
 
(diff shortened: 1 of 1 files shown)


mstimberg:
I am going ahead merging this, even though it is a bit of an ugly workaround, but I keep running into this issue when running Brian's development version. I opened #1422 to keep track of this issue.

schmitts:
Can we have  a patch release that includes the fix?

mstimberg:
I'm out of office until end of May, so I can't do it before then. Note that you can install the development package from TestPyPi though.

PR #1415, opened 2022-05-05
Izhikevich 2003

aesagtekin:


examples/frompapers/Izhikevich_2003.py
```diff
@@ -0,0 +1,179 @@
+#!/usr/bin/env python
+# coding: utf-8
+
+# In[1]:
```
schmitts, line 4: Thanks for the PR. Please remove these left-over lines (from the conversion from notebook to plain python) here and also below.

examples/frompapers/Izhikevich_2003.py
```diff
@@ -0,0 +1,179 @@
...
+# Since there are few neural groups and connections in this simulation, 
+# all the neuron groups and synaptic connections are explicitly coded line by line, 
+# without even using any for loops. 
+
+############################
+
+from brian2 import *
+from brian2tools import *
+import numpy as np
+np.random.seed(1337)
```
schmitts, line 23: explain why the seed is necessary

examples/frompapers/Izhikevich_2003.py
```diff
@@ -0,0 +1,179 @@
...
+
+get_ipython().run_line_magic('matplotlib', 'inline')
```
schmitts, line 25: remove

examples/frompapers/Izhikevich_2003.py
```diff
@@ -0,0 +1,179 @@
...
+eqs = ''' dv/dt = (0.04/ms/mV)*v**2 + (5/ms)*v + 140*mV/ms - u/ms + I*mV/(amp*ms) : volt
+          du/dt = a*(b*v - u)/ms  : volt
+          I : amp
+          a : 1
+          b : 1
+          c : volt
+          d : volt '''
+
+# synaptic parameters are fine-tuned in order to see the activity in the paper.
+# in paper, they used -1*amp for inhibitory and 0.5*amp for excitatory.
```
schmitts, line 39: do you know why the change is necessary? are the numbers in the paper a typo? is the model different?

examples/frompapers/Izhikevich_2003.py
```diff
@@ -0,0 +1,179 @@
...
+we = 2*amp        # weights for excitatory synapses
+wi = -4*amp       # weights for inhibitory synapses
+EL = -65 * mV     # initial value of membrane potentials
+dly = 0*ms        # delay
+
+
+# In[2]:
+
+
+start_scope()
```
schmitts, line 50: remove

examples/frompapers/Izhikevich_2003.py
```diff
@@ -0,0 +1,179 @@
...
+Si1.w = 'we*rand()'
+
+#from inh to inh.
+Si2 = Synapses(N_inh, N_inh, 'w : amp', delay=dly, on_pre='I_post += w') # inhibitory neurons' synapses
+Si2.connect(p=1)
+Si2.w = 'wi*rand()'
+
+
+##input. source: https://www.izhikevich.org/publications/spikes.htm
+N_exc.run_regularly('I = 5*amp*randn()', dt=1*ms)
```
schmitts, line 115: is this supposed to be some white noise random current? if yes, please cf. https://brian2.readthedocs.io/en/stable/user/models.html?highlight=white%20noise#noise

schmitts (commented):
Thanks. Please see the comments.

examples/frompapers/Izhikevich_2003.py
aesagtekin: Sorry, forget to remove the first comments before pushing it.

examples/frompapers/Izhikevich_2003.py
aesagtekin: Just used during fine-tuning for weights, removed.

examples/frompapers/Izhikevich_2003.py
aesagtekin: Done

examples/frompapers/Izhikevich_2003.py
aesagtekin: Done

examples/frompapers/Izhikevich_2003.py
aesagtekin: Indeed it is. I used run_regularly since the noise is given for every dt=1*ms, not for all the dt (0.5*ms). How should I change it?

examples/frompapers/Izhikevich_2003.py
aesagtekin: I don't know unfortunately. I experimentally find out that multiplying these two weights with 4 gives more similar results with MATLAB code of the paper. Maybe specifying the noise as you suggest will fix it,so just waiting for your response about noise.

examples/frompapers/Izhikevich_2003.py
aesagtekin: Here is the paper and MATLAB code: https://www.izhikevich.org/publications/spikes.htm

In code, he updated the 'v' twice at each for loop (a interesting way to do so, and also he did not updated the 'u' variable twice since it is a slow variable and still converges even for 1ms update), while introducing the noise for each dt. So I guess there is a relationship between the dt, gauss, and the single neuron equation.

schmitts: Thanks. I'll have a look asap.

schmitts: @aesagtekin I haven't forgotten. Will do in the next days.

aesagtekin: Thanks!

schmitts: The `run_regularly` is not correct. All neurons will see the same background current. I haven't found a way to fix it yet.

mstimberg: Hi everyone. With the definition `I : amp`, there should be an independent `I` variable per neuron, and therefore `I = 5*amp*randn()` will give a different value for each neuron. The function ["auto-vectorizes"](https://brian2.readthedocs.io/en/stable/advanced/functions.html#functions-with-context-dependent-return-values) according to the context. Or do you have observed that this does not work here?

schmitts: You're right. I had a mistake in my minimal example. So back to square 1 in the comparison with the Matlab implementation.

schmitts:
Please cf. https://brian.discourse.group/t/implemented-izhikevichs-2003-paper-in-brian2-want-to-contribute-to-the-brian2-documentation/692/6

aesagtekin:
Thanks! I will look at it in a few days.

aesagtekin:
I just looked at it; great work. I noticed that you also introduced a 1*ms delay for the inhibitory neurons. Thank you! Is there anything left to do?

aesagtekin:
I just looked at it. Great work, thanks! Is there anything left to do?

schmitts:
> I just looked at it. Great work, thanks! Is there anything left to do?

If you think it's ok, we can put it through.

aesagtekin:
> > I just looked at it. Great work, thanks! Is there anything left to do?
> 
> If you think it's ok, we can put it through.

Yes, I think it is OK.

schmitts:
> > > I just looked at it. Great work, thanks! Is there anything left to do?
> > 
> > 
> > If you think it's ok, we can put it through.
> 
> Yes, I think it is OK.

Great. Then please update the file and we can continue with the PR.

schmitts:
> OK, I've updated the file. But couldn't understand what you mean by PR.

PR stands for "pull request". Now we can wait for @mstimberg or some other maintainer to approve our work.

aesagtekin:
> > OK, I've updated the file. But couldn't understand what you mean by PR.
> 
> PR stands for "pull request". Now we can wait for @mstimberg or some other maintainer to approve our work.

OK, thanks again!

mstimberg:
Many thanks for your work on this @aesagtekin and @schmitts :pray: . I did a minor change to the authorship information on the top of the file, since the reference to the `Izhikevich_2003.py` would be a bit opaque to readers of the documentation. Hope that works for you? As a small consolation for the long waiting time, I also updated the plot a bit to make it look even more like Izhikevich's original figure (see below) :) If you are both happy with it, I will trigger the merge.
Figure_1

Issue #1348, opened 2021-09-22
Spurious `-.o` files are created on Linux


mstimberg:
Brian users on Linux will have a couple of spurious `-.o` lying around – I finally figured out where they come from and will open a PR soon.

PR #1416, opened 2022-05-06
Avoid creating temp files in current dir

mstimberg:
Somewhat similar to #1348, our check whether the compiler supports certain flags created the compiled object files in a `tmp` directory in the current directory. This PR:
1. Creates these files in the temporary directory instead, and
2. Removes them after compilation

### DIFF
--- a/brian2/codegen/cpp_prefs.py
+++ b/brian2/codegen/cpp_prefs.py
@@ -239,7 +239,11 @@ def _determine_flag_compatibility(compiler, flagname):
     with tempfile.NamedTemporaryFile('w', suffix='.cpp') as f, std_silent():
         f.write('int main (int argc, char **argv) { return 0; }')
         try:
-            compiler.compile([f.name], extra_postargs=[flagname])
+            object_files = compiler.compile([f.name],
+                                            output_dir=os.path.dirname(f.name),
+                                            extra_postargs=[flagname])
+            for object_file in object_files:
+                os.unlink(object_file)
         except CompileError:
             logger.warn(f"Removing unsupported flag '{flagname}' from "
                         f'compiler flags.')
--- a/brian2/codegen/cpp_prefs.py
+++ b/brian2/codegen/cpp_prefs.py
@@ -236,14 +236,13 @@ prefs.register_preferences(
 def _determine_flag_compatibility(compiler, flagname):
     import tempfile
     from distutils.errors import CompileError
-    with tempfile.NamedTemporaryFile('w', suffix='.cpp') as f, std_silent():
-        f.write('int main (int argc, char **argv) { return 0; }')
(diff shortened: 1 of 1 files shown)


brian2/codegen/cpp_prefs.py
```diff
@@ -239,7 +239,11 @@ def _determine_flag_compatibility(compiler, flagname):
     with tempfile.NamedTemporaryFile('w', suffix='.cpp') as f, std_silent():
         f.write('int main (int argc, char **argv) { return 0; }')
         try:
-            compiler.compile([f.name], extra_postargs=[flagname])
+            object_files = compiler.compile([f.name],
+                                            output_dir=os.path.dirname(f.name),
+                                            extra_postargs=[flagname])
+            for object_file in object_files:
```
schmitts, line 245: Just curious. Why's the unlink necessary? They should go out of scope by themselves?

mstimberg: By "go out of scope" do you mean that they get automatically deleted when the `with` context manager ends? This is true for the `.cpp` file that we create, but the `.o` file (should be only one in this case) as the results of the compilation is not deleted automatically.

schmitts: Thanks. What about having a temporary directory instead? Then one would not have to unlink individual files?

mstimberg: Ah, I see. Yes, that might be more elegant. Let me give it a quick try.

mstimberg:
I updated the PR to use a `TemporaryDirectory` instead, @schmitts what do you think? Note that due to the way `distutils` works, it will create a directory `/tmp/brian_test_flag_..../` where it puts the `.cpp` file, and then duplicate the hierarchy again for the object file, i.e. it ends up in `/tmp/brian_test_flags_.../tmp/brian_test_flag_.../`. This is a bit ugly, but everything gets deleted anyway.

brian2/codegen/cpp_prefs.py
```diff
@@ -236,14 +236,13 @@
 def _determine_flag_compatibility(compiler, flagname):
     import tempfile
     from distutils.errors import CompileError
-    with tempfile.NamedTemporaryFile('w', suffix='.cpp') as f, std_silent():
-        f.write('int main (int argc, char **argv) { return 0; }')
+    with tempfile.TemporaryDirectory(prefix='brian_flag_test_') as temp_dir, std_silent():
+        fname = os.path.join(temp_dir, 'flag_test.cpp')
+        with open(fname, 'wt') as f:
```
schmitts, line 241: t is default, but idc if explicitly put here.

schmitts (approved):
Looks good to me!

mstimberg:
Thanks for the feedback @schmitts, going ahead with the merge.

PR #1417, opened 2022-05-06
Deterministic code generation

mstimberg:
Two main issues led to code files changing for identical models (i.e., even when running a model again in a new process where the problems with regard to names as discussed in #680 do not play a role):
1. Variable declarations and other "unordered" code was generated in a different order for every run, and
2. The file names written to the result directory used `hash()` to deal with the problem of case-insensitivity on Windows (otherwise `i` and `I` would be stored to the same file). For security reasons, the `hash()` function changes its output for every run.

The fix in this PR is to:
1. Use `sorted()` whenever iterating over variables that are stored in sets, and 
2. Use `zlib.crc32` instead of `hash` to calculate the hash value for the filenames. Strictly speaking, `crc32` is a checksum and not a hash value and is more prone to lead to collisions, but this should not matter for us since we only need to disambiguate in very specific circumstances (names that are identical except for case).

I did not set up a formal test for this, but I verified with fairly complex examples (e.g. `frompapers/Kremer_et_al_2011_barrel_cortex.py`) that the generated is identical over runs.

### DIFF
--- a/brian2/codegen/codeobject.py
+++ b/brian2/codegen/codeobject.py
@@ -349,7 +349,7 @@ def create_runner_codeobj(group, code, template_name,
         needed_variables = []
     # Resolve all variables (variables used in the code and variables needed by
     # the template)
-    variables = group.resolve_all(identifiers | set(needed_variables) | set(template_variables),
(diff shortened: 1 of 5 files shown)


thesamovar (approved):
Good idea!

PR #1419, opened 2022-05-11
Logging with multiprocessing

### LINKED ISSUE #1418: Log files and multiprocessing
Our current logging infrastructure is not compatible with running multiple processes via Python's `multiprocessing` module. I guess working on examples that use `multiprocessing` prompted @schmitts' question about switching off the log file (#1407)?
A minor issue is that the log file is not very useful with multiple processes, since all processes write in no specific order. A bigger issue is that our log file rotation will raise an error if the file becomes big enough to be rotated – several processes try to move/delete the file, leading to a `FileNotFoundError`. I don't think there's a need to deal with this in a perfect way (e.g. by writing a multiprocess-safe logging handler, moving the log rotation into a single process, or similar), but we should at least mention this issue in the documentation. Or, given that few users actually find such details in the documentation, we could raise a warning in the logging code when 1) a file log is used and 2) it is accessed from multiple processes. A downside of this warning would be that I don't see an easy way to only raise it once, instead of once for every process launched by `multiprocessing`. But I guess this is ok, since we really want the user to do something about it.
There are three workarounds that we can recommend to users:
1. Use a preference file to switch off file logging
2. Set `BrianLogger.tmp_log` to `None` to switch off file logging in an ad-hoc way
3. Call `BrianLogger.initialize()` at the beginning of each spawned process, which will make it use its own log file

mstimberg:
See the discussion in #1418. This PR changes the default file log level to DEBUG, switches off file logging when the logger is accessed from a process different from the one that called `BrianLogger.initialize()`, and updated the documentation accordingly.

@schmitts & @thesamovar, let me know if anything is unclear or could be improved.

Fixes #1418

### DIFF
--- a/dev/continuous-integration/run_test_suite.py
+++ b/dev/continuous-integration/run_test_suite.py
@@ -67,7 +67,7 @@ if __name__ == '__main__':
                              test_in_parallel=in_parallel,
                              reset_preferences=reset_preferences,
                              float_dtype=float_dtype,
-                             test_GSL=True, 
+                             test_GSL=True,
                              sphinx_dir=sphinx_dir)
 
     if not result:
--- a/dev/continuous-integration/run_test_suite.py
+++ b/dev/continuous-integration/run_test_suite.py
@@ -12,59 +12,60 @@ import numpy as np
 
 import brian2
 
-split_run = os.environ.get('SPLIT_RUN', None)
-standalone = os.environ.get('STANDALONE', 'no').lower() in ['yes', 'true']
-python_version = os.environ.get('PYTHON_VERSION', os.environ.get('PYTHON'))
-# If TRAVIS_OS_NAME is not defined, we are testing on appveyor
-operating_system = os.environ.get('AGENT_OS', 'unknown').lower()
-cross_compiled = os.environ.get('CROSS_COMPILED', 'FALSE').lower() in ['yes', 'true']
-report_coverage = os.environ.get('REPORT_COVERAGE', 'no').lower() in ['yes', 'true']
-dtype_32_bit = os.environ.get('FLOAT_DTYPE_32', 'no').lower() in ['yes', 'true']
-src_dir = os.environ.get('SRCDIR')
+if __name__ == '__main__':
+    split_run = os.environ.get('SPLIT_RUN', None)
+    standalone = os.environ.get('STANDALONE', 'no').lower() in ['yes', 'true']
+    python_version = os.environ.get('PYTHON_VERSION', os.environ.get('PYTHON'))
+    # If TRAVIS_OS_NAME is not defined, we are testing on appveyor
+    operating_system = os.environ.get('AGENT_OS', 'unknown').lower()
+    cross_compiled = os.environ.get('CROSS_COMPILED', 'FALSE').lower() in ['yes', 'true']
+    report_coverage = os.environ.get('REPORT_COVERAGE', 'no').lower() in ['yes', 'true']
+    dtype_32_bit = os.environ.get('FLOAT_DTYPE_32', 'no').lower() in ['yes', 'true']
+    src_dir = os.environ.get('SRCDIR')
 
-if split_run == '1':
-    targets = ['numpy']
-    independent = True
-elif split_run == '2':
-    targets = ['cython']
-    independent = False
-else:
-    targets = None
-    independent = True
+    if split_run == '1':
+        targets = ['numpy']
+        independent = True
+    elif split_run == '2':
+        targets = ['cython']
+        independent = False
+    else:
+        targets = None
+        independent = True
 
-if operating_system == 'windows' or standalone:
-    in_parallel = []
-else:
-    in_parallel = ['codegen_independent', 'numpy', 'cpp_standalone']
+    if operating_system == 'windows' or standalone:
+        in_parallel = []
+    else:
+        in_parallel = ['codegen_independent', 'numpy', 'cpp_standalone']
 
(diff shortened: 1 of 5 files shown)


mstimberg:
There were a few issues with running the multiprocessing tests under OS X and Windows (of course there were), but now all tests pass on all platforms.

mstimberg:
Hi @bdevans, @thesamovar any comments on the PR? I think it is rather safe to merge in the sense that it should clearly improve things, but if you could have a quick look I'd be grateful. I'd prefer to rather merge it soon, so I propose to turn this into a "merge unless someone objects" instead of waiting for a positive review :blush:

thesamovar:
Have you tried running actual examples (not tests) in Windows? I feel like with multiprocessing stuff that's important because there are all sorts of weirdnesses and I don't know how well the test suite reproduces that?

mstimberg:
I did run the [multiprocessing examples](https://brian2.readthedocs.io/en/stable/examples/index.html#multiprocessing) and [Graupner & Brunel (2012)](https://brian2.readthedocs.io/en/latest/examples/frompapers.Graupner_Brunel_2012.html). They ran successfully – I did not verify in detail whether it actually switches off the log, though. I guess if the corresponding test passes, there is no reason this should not work?
The [Maass et al. (2002)](https://brian2.readthedocs.io/en/latest/examples/frompapers.Maass_Natschlaeger_Markram_2002.html) example also uses multiprocessing, but I recently realized it does not work under Windows (a function should be declared on the module level instead of in the `__main__` block, I think). But it does not work in `master`, either, so this problem is unrelated to this PR.

docs_sphinx/advanced/logging.rst
```diff
@@ -28,14 +28,31 @@ useful for debugging. Each log message has one of the following "log levels":
...
+and all messages of level ``INFO`` and above are displayed on the console. To change
+what messages are displayed, see below.
 
 .. note:: By default, the log file is deleted after a successful simulation run,
    i.e. when the simulation exited without an error. To keep the log around,
    set the `logging.delete_log_on_exit` preference to ``False``.
 
+Logging and multiprocessing
+---------------------------
+Brian's logging system is not designed for multiple parrallel Brian processes
```
bdevans, line 41: Typo in parallel

dev/continuous-integration/run_test_suite.py
```diff
@@ -2,70 +2,73 @@
...
+                             test_GSL=True,
+                             sphinx_dir=sphinx_dir)
+    else:
+        result = brian2.test(targets,
+                             test_codegen_independent=independent,
+                             test_standalone=None,
+                             test_in_parallel=in_parallel,
+                             reset_preferences=reset_preferences,
+                             float_dtype=float_dtype,
+                             test_GSL=True,
```
bdevans, line 70: There's a cheeky extra space here!

bdevans, line 70: There's a cheeky extra space here!

docs_sphinx/advanced/logging.rst
```diff
@@ -28,14 +28,31 @@ useful for debugging. Each log message has one of the following "log levels":
...
+started via Python's `multiprocessing` module (see the
+:ref:`multiprocessing examples <multiprocessing>`). Log messages that get printed
+from different processes to the console are not printed in a well-defined order and
+do not contain any indication about which processes they are coming from. You might
+therefore consider to use e.g. `BrianLogger.log_level_error()` to only show error
```
bdevans, line 46: "wish to use" or "consider using" would be slightly more natural here.

bdevans, line 46: "wish to use" or "consider using" would be slightly more natural here.

bdevans (changes requested):
A few minor tweaks to make but otherwise looks fine. I might lean towards generating a separate log file for each process, calling them something like `f"{log_file_basename}_{pid}.log"` but that might generate too much output so having just one log file for the master process is fine.

mstimberg:
Thanks @bdevans for the review, I think the issues should be fixed now.

> I might lean towards generating a separate log file for each process, calling them something like `f"{log_file_basename}_{pid}.log"` but that might generate too much output so having just one log file for the master process is fine.

Yes, I agree that this would make sense as well. On the other hand, the logged information in the subprocesses will most likely be very similar (if not identical), and if you launch many of them, this could become annoying. IMO, the vast majority of users do not care about the log, and for those who do, the manual method gives them a way to have individual logs per subprocess.

bdevans:
Thanks Marcel, it looks good to me now but for some reason I can't approve the changes (I can only seem to dismiss or re-request the review) - any ideas what's gone wrong?

thesamovar:
> I did run the [multiprocessing examples](https://brian2.readthedocs.io/en/stable/examples/index.html#multiprocessing) and [Graupner & Brunel (2012)](https://brian2.readthedocs.io/en/latest/examples/frompapers.Graupner_Brunel_2012.html). They ran successfully – I did not verify in detail whether it actually switches off the log, though. I guess if the corresponding test passes, there is no reason this should not work? The [Maass et al. (2002)](https://brian2.readthedocs.io/en/latest/examples/frompapers.Maass_Natschlaeger_Markram_2002.html) example also uses multiprocessing, but I recently realized it does not work under Windows (a function should be declared on the module level instead of in the `__main__` block, I think). But it does not work in `master`, either, so this problem is unrelated to this PR.

As long as it's not breaking I guess it's OK! If Ben is happy with the code I'm happy too.

bdevans (approved):

PR #1421, opened 2022-05-18
Document how to set weights from given weight matrix

schmitts:
Cf. https://brian.discourse.group/t/using-custom-synaptic-weight-matrix/485/5


mstimberg (approved):
Thanks, should have added a long time ago :)
I just removed the `mV` units, since the code above defines `w : 1`.

PR #1423, opened 2022-05-26
Reproducing figure 2 of scholarpedia article on the Ornstein-Uhlenbeck process

schmitts:


mstimberg (approved):
Simple but useful, thanks :blush:

PR #1426, opened 2022-06-20
More reasonable LaTeX output for arrays

### LINKED ISSUE #1425: Avoid extremely long LaTeX output for arrays
Brian 2 supports converting `Quantity` objects to LaTeX, which is also used as a "rich representation" in interfaces that support them, e.g. jupyter notebooks. This is kind of a nice gimmick, but not crazily useful in general. There is a major problem with its current implementation, though: it simply passes things on to `sympy`'s `Matrix` object, and this object always prints out the complete content. A seemingly harmless statement like:
```python
G = NeuronGroup(10000, 'v : volt')
G.v[:]
```
in a jupyter notebook will then create a huge LaTeX column vector which makes MathJax work hard to render the equation and blocks the browser.

mstimberg:
This makes the LaTeX output for `Quantity` arrays more sensible (see #1425) . Before, things looked like this, and big arrays could make the browser work hard to do the MathJax rendering:

image

With this PR, vectors/matrices will only be printed up to a reasonable size to fit on the screen:
image

The implementation reuses `np.array2string` and then does a few replacements. Not the most elegant solution, but it means that it follows most of numpy's "print options" automatically. I also finally settled on using `np.get_printoptions['threshold']//100` as the value for the threshold. This is of course quite arbitrary, but this way we don't need a new preference for a niche feature, and users can still influence the behaviour if they really want to:
image

Fixes #1425

### DIFF
--- a/brian2/tests/test_units.py
+++ b/brian2/tests/test_units.py
@@ -17,6 +17,7 @@ from brian2.units.fundamentalunits import (UFUNCS_DIMENSIONLESS,
                                            Unit,
                                            have_same_dimensions,
                                            get_dimensions,
+                                           is_dimensionless,
                                            is_scalar_type,
                                            DimensionMismatchError,
                                            check_units,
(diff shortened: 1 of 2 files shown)


thesamovar (approved):
Great! Big improvement.

PR #1430, opened 2022-07-28
Add missing volume (inherited from morphology) to SpatialNeuron and test

schmitts:


mstimberg (approved):
Oh my, thanks a lot! That's a small but important mistake (the `volume` attribute was available but always zero).

PR #1431, opened 2022-08-19
Change 'is not' to '!=' expressions.py

appukuttan-shailesh:
For a script that I have been running, I am getting the error:
> Expression 'v - vr' uses inconsistent units ('v' has unit V; 'vr' has unit V).

I was certain that the units had been correctly assigned, and therefore examined the source code, and found them to be idential when printed. Changing it to `!=` worked for me (and still did the expected checking of units). If there are no strong motivations for using `is not` for the comparison, I propose using `!=` here.

**==** determines if the values of two objects are equal, 
while **is** determines if they are the exact same object.

I would have liked to provide a dummy script that could reproduce this, but my current one is quite bulky and in a bit of a hurry. I could try to develop a simpler demo later if required.

### DIFF
--- a/brian2/parsing/expressions.py
+++ b/brian2/parsing/expressions.py
@@ -360,7 +360,7 @@ def parse_expression_dimensions(expr, variables, orig_expr=None):
         right_dim = parse_expression_dimensions(expr.right, variables, orig_expr=orig_expr)
         if op in ['Add', 'Sub', 'Mod']:
             # dimensions should be the same
-            if left_dim != right_dim:
+            left_dim is not right_dim and left_dim != right_dim:
                 op_symbol = {'Add': '+', 'Sub': '-', 'Mod': '%'}.get(op)
                 left_str = NodeRenderer().render_node(expr.left)
                 right_str = NodeRenderer().render_node(expr.right)
--- a/brian2/parsing/expressions.py
+++ b/brian2/parsing/expressions.py
@@ -360,7 +360,7 @@ def parse_expression_dimensions(expr, variables, orig_expr=None):
         right_dim = parse_expression_dimensions(expr.right, variables, orig_expr=orig_expr)
         if op in ['Add', 'Sub', 'Mod']:
             # dimensions should be the same
-            if left_dim is not right_dim:
+            if left_dim != right_dim:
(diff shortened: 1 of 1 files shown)


mstimberg:
Hi @appukuttan-shailesh thanks for the report. If you'd be able to provide something simpler (or e.g. send me a link to your full code in a private message), that would be very helpful. I've never encountered this issue, and I am quite curious how this can happen. On a wild guess: does it maybe involve pickled values somewhere?
In general, dimensions are implemented as singletons, i.e. there should be only a single `Dimension` object for voltages, and a simple `is` check should therefore be enough. We do a large number of these comparisons, and therefore made this small optimization.
Your fix would certainly work, but I'd prefer to find the root cause. If that is not possible, I'd probably go with something like `if left_dim is not right_dim and left_dim != right_dim`. This should be as fast as the previous line if `left_dim` matches `right_dim` – the failure case is unimportant performance-wise.

appukuttan-shailesh:
Hi @mstimberg ... I will try to develop a simpler example and send you sometime next week. If not, will share the original model itself (I intend to make it publicly available in the coming days) with instructions to reproduce the error. It involved making use of multiprocessing, which implictly involved pickling... so that could be the issue as you suggested.

I like your proposal for making the change (`left_dim is not right_dim and left_dim != right_dim`)... will help keep the optimized behavior for regular use cases, and also handle this other situation. We can decide once I share the example with you.

appukuttan-shailesh:
Hi @mstimberg, sorry for the super long delay in getting back to this _(some changes on the work front kept me away)_
I finally have a demo that demonstrates the original issue:
[issue_demo.zip](https://github.com/brian-team/brian2/files/9933203/issue_demo.zip)

The environement required for above can be created via:
> pip install brian2
pip install dill==0.3.5.1
pip install pathos==0.2.9
_(latest version of dill gives additional unrelated warnings)_ 

and then just run the init.py file. I get an error such as:
image

Your proposed change:
`left_dim is not right_dim and left_dim != right_dim`
should resovel this situation. I have made that change in this PR now.

mstimberg:
Many thanks for the example, that makes the issue very easy to reproduce. I did find the underlying issue and will fix it instead of making the comparison more robust as proposed in this PR – I think there are several other places where the same problem can occur. Thanks again for reporting it and for following up with the example.

PR #1437, opened 2022-09-29
Apply dict comprehension

boneyag:
Dict comprehension is faster than invoking the constructor

### DIFF
--- a/brian2/core/variables.py
+++ b/brian2/core/variables.py
@@ -86,9 +86,8 @@ def get_dtype_str(val):
 
 def variables_by_owner(variables, owner):
     owner_name = getattr(owner, 'name', None)
-    return dict([(varname, var) for varname, var in variables.items()
-                 if getattr(var.owner, 'name', None) is owner_name])
-
+    return {varname: var for varname, var in variables.items()
+          if getattr(var.owner, 'name', None) is owner_name}
 
 class Variable(CacheKey):
     """
--- a/examples/advanced/stochastic_odes.py
+++ b/examples/advanced/stochastic_odes.py
@@ -57,7 +57,7 @@ dts = [1*ms, 0.5*ms, 0.2*ms, 0.1*ms, 0.05*ms, 0.025*ms, 0.01*ms, 0.005*ms]
 
 rows = floor(sqrt(len(dts)))
 cols = ceil(1.0 * len(dts) / rows)
-errors = dict([(method, zeros(len(dts))) for method in methods])
+errors = {method: zeros(len(dts)) for method in methods}
 for dt_idx, dt in enumerate(dts):
     print('dt: %s' % dt)
     trajectories = {}


mstimberg:
@boneyag Thanks for your contribution (and apologies for the terrible late reply :flushed:...).
Can you say why you selected those two usages of the `dict` constructor? I think our code base has many more of them. I basically agree with this change, but rather from the point of view of code style. I don't think the speed matters in these instances (whether that code uses, say, 0.5ms instead of 0.6ms isn't really changing anything).

boneyag:
> @boneyag Thanks for your contribution (and apologies for the terrible late reply flushed...). Can you say why you selected those two usages of the `dict` constructor? I think our code base has many more of them. 

I am trying to understand whether a tool that could transform such changes would help the developers. I think these two specific instances were detected based on frequent code change patterns mined from other projects. Since my checker only checks for particular patterns of using dict constructor, it is possible that I did not detect other instances. 

> I basically agree with this change, but rather from the point of view of code style. I don't think the speed matters in these instances (whether that code uses, say, 0.5ms instead of 0.6ms isn't really changing anything).
Yes. for small dicts, I don't think the speed improvement is not significant. I was not focusing on Python idioms, because it did not seem like a meaningful PR.

mstimberg:
Thanks for getting back to us. I gather you wrote your own tool for detection and are testing it out ? In case you are not aware, the [`pylint` tool](https://pylint.pycqa.org) has support for these kind of checks, see e.g. its output for [`R1717`](https://pylint.pycqa.org/en/stable/user_guide/messages/refactor/consider-using-dict-comprehension.html):
```console
$ pylint --disable=all --enable=R1717 brian2
************* Module brian2.monitors.statemonitor
brian2/monitors/statemonitor.py:325:34: R1717: Consider using a dictionary comprehension (consider-using-dict-comprehension)
************* Module brian2.equations.equations
brian2/equations/equations.py:322:22: R1717: Consider using a dictionary comprehension (consider-using-dict-comprehension)
brian2/equations/equations.py:1025:21: R1717: Consider using a dictionary comprehension (consider-using-dict-comprehension)
************* Module brian2.core.variables
brian2/core/variables.py:92:11: R1717: Consider using a dictionary comprehension (consider-using-dict-comprehension)
```

boneyag:
I write my own checkers when pylint does not have a particular checker. For this specific use case, I used pylint's 'consider-using-dict-comprehension'. I ran it on the repo that I cloned in August. For some reason, it only returned the two instances. When I ran it now I received 3 more instances that I did not find before: `brian2.monitors.statemonitor` and 2 instances in `brian2.equations.equations`. I need to check back why pylint did not find these 3 before.

mstimberg:
Thanks again for your contribution, and again apologies for the long time it took me to reply initially. We finally ended up using `pyupgrade` in another PR (#1446) which among many other things also took care of replacing dict constructor calls by dict comprehensions.

Issue #610, opened 2015-12-14
Use sphinx-gallery for examples


mstimberg:
Nothing for right now, but I noticed that scikits-learn made their example documentation system into a separate extension so we could base our examples on that. This would make a few of our own scripts obsolete (e.g. linking to examples from the reference docs) and deal with issue #171

thesamovar:
Looks great!

mstimberg:
Forgot to put the link actually: http://sphinx-gallery.readthedocs.org/

PR #1440, opened 2022-11-09
New example with animated GIF

mstimberg:
Following a [discussion on the forum](https://brian.discourse.group/t/network-dynamics-visualizer/828/14), I created a simple example of coupled oscillators ([Kuramoto model](https://en.wikipedia.org/wiki/Kuramoto_model)). I realized that the results are very boring if they are not animated, so I created an animated GIF instead of the usual static plot (the code for the simulation is very short, but the plotting code isn't...). I adapted the code that includes examples in the documentation so that it accepts GIFs in addition to the usual PNG files.

Unfortunately, storing the GIF is not compatible with the approach we take in our `run_examples.py` script, so I added it manually for now. I don't see an easy way to automate this ATM, but hopefully we'll move over to an automated system like `sphinx-gallery` at some point (#610).


brian2/sphinxext/generate_examples.py
```diff
@@ -139,8 +139,8 @@ def main(rootpath, destdir):
         output += '\n'.join(['    ' + line for line in afterdoccode.split('\n')])
         output += '\n\n'
 
-        eximgpattern = os.path.join(eximgpath, '%s.*.png' % exname)
-        images = glob.glob(eximgpattern)
+        eximgpattern = os.path.join(eximgpath, '%s.*' % exname)
```
bdevans, line 142: Let's make this an f-string!

bdevans (changes requested):
There's an artefact in the produced gif whereby the subplots appear to jump (at the beginning or possibly end) when the animation loops. It would be good to fix the figure size / axis locations (or whatever is causing it) for every frame to prevent this.

mstimberg:
> There's an artefact in the produced gif whereby the subplots appear to jump (at the beginning or possibly end) when the animation loops. It would be good to fix the figure size / axis locations (or whatever is causing it) for every frame to prevent this.

Thanks, weird that I overlooked this... The problem seems to be the interaction between `layout='tight'` and the writer that saves the animation – the animatino as shown by `plt.show()` does not show the jump. I now use `fig.subplots_adjust` to set everything manually.

brian2/sphinxext/generate_examples.py
mstimberg: It's weird that it wasn't an f-string already, actually... I changed the remaining `%` strings in the file as well.

mstimberg:
> Thanks, all good +1

You look through my changes faster than I can comment on them :yum:  Thanks, for the review :smile:

PR #1442, opened 2022-11-16
Replace deprecated distutils package

bdevans:
Replace distutils.version.LooseVersion with packaging.version.Version.

### DIFF
--- a/brian2/__init__.py
+++ b/brian2/__init__.py
@@ -73,7 +73,7 @@ from brian2.only import test
 
 # Check for outdated dependency versions
 def _check_dependency_version(name, version):
-    from distutils.version import LooseVersion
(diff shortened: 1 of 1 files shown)


mstimberg (approved):
Many thanks, one less `DeprecationWarning`…

mstimberg:
I am going ahead with the merge, but I'll also open a new issue about removing these version checks altogether. I'm not sure anymore whether it is really a good thing to have, but if we want to keep it, we should probably fill in the version requirements automatically from a single source. The requirements hardcoded in `brian2.__init__.py` will easily go out of sync (e.g. it only requires numpy 1.10, whereas our `setup.py` requires numpy 1.17…)

Issue #180, opened 2013-12-06
Code generation syntax for reductions?


thesamovar:
One thing we would like to do in monitors and potentially in other situations is allow for reductions, e.g. `mean(V)` for monitoring. We already do this for summed synaptic variables, but we only allow the sum. In that case, it seemed sufficient, but for monitors and possibly more widely there might be a case for a more general syntax.

One option is to look at existing packages that include reductions, including Theano, numexpr and pycuda:
- http://deeplearning.net/software/theano/library/scan.html
- http://documen.tician.de/pycuda/array.html#module-pycuda.reduction
- http://code.google.com/p/numexpr/wiki/UsersGuide#Supported_reduction_operations

Theano goes for a complicated general system, pycuda just lets you write your own code basically, and numexpr has a small list of just two reductions. My initial feeling is that the pycuda option is the best with maybe a few defaults like sum and prod built in.

mstimberg:
I know this is a very old topic, but some of this came up again in #1445, where @schmitts used a network operation to take the dot product between two neuronal variables (as a linear readout to get a scalar variable). I realized that a straightforward extension of our existing `(summed)` syntax could allow for `(summed, shared)` in neuronal equations, i.e.
```python
eqs = '''
x : 1
y : 1
z = x*y : 1 (summed, shared)
'''
```
would have `z` store the dot product of `x` and `y` (not sure whether we need the extra `shared` – it is redundant but more explicit). I also thought that by making `summed` a special-case of more general reductions, we could get away by only changing the syntax for equations in `NeuronGroup` and `Synapses`, and leave things like `StateMonitor` untouched (it could simply record `z` in the above example). Here's what I could imagine: `(summed)` would be a special case of `(reduction: sum)` or `(reduced: sum)`. I'm not sure "reduction"/"reduced" is the most intuitive term for less technical people not familiar with map/reduce, etc. It might also be something like "merge", "combine", "aggregate", etc.
We could then (both in `NeuronGroup` and `Synapses`) have `(reduce: max)`, etc., with a common basic template that initializes with the neutral value (0 for `sum`, `-∞` for `max`, etc.) and then applies the respective function. It would be nice to support e.g. `argmin`, `argmax` as well, but it would be slightly more complex (the final value is different from the values that are compared).
[...]

PR #1445, opened 2022-11-30
Reproducing Sinusoid oscillator of Supervised learning in spiking neural networks with FORCE training

schmitts:


mstimberg:
Many thanks (again :blush:)! I wonder whether we cannot replace the two network operations with a "creative use" of synapses – this might be a nice showcase for doing this and would make everything more efficient (and enable us to use standalone mode!). It will make things quite a bit less readable, though, not sure...

schmitts:
Yes, I thought the same thing. I was not really happy with the explicit matrix operations. On the other hand it makes the math more accessible/direct. Do you want to draft a version with Synapse operations? It's easier to discuss/compare with an example at hand.

mstimberg:
I have hacked together a solution that only uses `Synapses`, summed variables, and `run_regularly` operations, but… it isn't pretty. In particular, we don't have a nice syntax to set the timing of several summed variable operations, but here they need to be executed in the right order to faithfully reproduce the algorithm in your `network_operation`. Actually, the result doesn't seem to be exactly the same, so there might be a bug as well – but either way, I don't think this should go into an example like this. You can compare the two implementations here: https://github.com/schmitts/brian2/compare/Nicola_Clopath_2017...brian-team:brian2:Nicola_Clopath_2017_no_network_op

The best solution is probably to go with your implementation for now, but also think about how we can improve syntax to make these things more straightforward in the future. For example, I had one idea that would work quite nicely as an extension of our current syntax, I think: we could extend `(summed)` to `NeuronGroup`s, to make it possible to sum expressions into a shared variable. This would allow users to replace this
```python
# linear readout
@network_operation(dt=defaultclock.dt)
def readout(t):
    neurons.z = np.dot(neurons.BPhi, neurons.r)
```
by:
```python
"""...
z = BPhi * r : 1 (summed, shared)
"""
```
in the equations.

schmitts:
Hi!

Thanks for your effort! Did you do some benchmarking between the two implementations?

Allowing for

```
"""...
z = BPhi * r : 1 (summed, shared)
"""
```

would be cool!

Cheers,

Sebastian

mstimberg:
> Did you do some benchmarking between the two implementations?

It was faster, but not dramatically so. Instead of 1m20s, it ran in something like 42s. With C++ standalone mode, it went further down to 22s. But I think the `network_operation` can be slightly sped up as well, I'll try out a few things.

mstimberg:
I tried a few things and nothing makes it significantly faster. In general, I'd recommend to use e.g. `neurons.v[:]` instead of `neurons.v` when passing Brian variable to functions, since `neurons.v[index]` adds quite a bit of overhead to support string-based indexing, etc. In this case, the `numpy` functions deal with this in a decent way, and it only shaved off a second or two here and there. For simplicity, I'll therefore leave it as it is and go ahead with the merge.

We can use this example as a use case for the syntax I mentioned earlier (more general discussion: #180).

mstimberg (approved):
Thanks again :pray:

mstimberg:
And here it is: https://brian2.readthedocs.io/en/latest/examples/frompapers.Nicola_Clopath_2017.html

PR #1449, opened 2023-01-04
flake8 precommit hook

mstimberg:
Continuing the work started with  #1435 to make Brian's code more readable and "best practice" compatible, this adds another precommit hook, this time for `flake8`. It will flag common issues such as local variables or imports that are never used, or minor stylistic issues such as using `not element in list` instead of `element not in list`. I configured it in a conservative way to not be too radical/annoying for Brian's code base, e.g. I excluded all tests and `__init__.py` files, since they use wildcard imports quite extensively.

Not sure anyone cares about this very much, but tagging @oleksii-leonov, @bdevans, @denisalevi just in case :yum:

### DIFF
--- a/.devcontainer/dev-requirements.txt
+++ b/.devcontainer/dev-requirements.txt
@@ -7,3 +7,5 @@ pre-commit == 2.20.*
 black == 22.10.0
 isort == 5.10.1
 pyupgrade == 3.2.3
+flake8 == 4.0.1
+flake8-bugbear == 21.11.29
--- a/.devcontainer/devcontainer.json
+++ b/.devcontainer/devcontainer.json
@@ -19,13 +19,13 @@
 	"settings": {
 		"python.defaultInterpreterPath": "/usr/local/bin/python",
 		"python.linting.enabled": true,
-		"python.linting.pylintEnabled": true,
+		"python.linting.flake8Enabled": true,
 		"python.formatting.provider": "black",
 		"python.formatting.autopep8Path": "/usr/local/py-utils/bin/autopep8",
 		"python.formatting.blackPath": "black",
 		"python.formatting.yapfPath": "/usr/local/py-utils/bin/yapf",
 		"python.linting.banditPath": "/usr/local/py-utils/bin/bandit",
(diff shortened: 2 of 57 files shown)


bdevans (approved):
Looks good to me!
Just one point to mention, there are now several files to keep in sync with versions of dependencies e.g. [.pre-commit-config.yaml](https://github.com/brian-team/brian2/pull/1449/files#diff-63a9c44a44acf85fea213a857769990937107cf072831e1a26808cfde9d096b9) and [.devcontainer/dev-requirements.txt](https://github.com/brian-team/brian2/pull/1449/files#diff-b36f7558d5f7f38cbb7489ef21549dfd0a946358c2692cd2032497ff67cd56d8). Would it be possible to consolidate them all into one location that gets read by the others?

mstimberg:
Thanks @bdevans.


> Just one point to mention, there are now several files to keep in sync with versions of dependencies e.g. [.pre-commit-config.yaml](https://github.com/brian-team/brian2/pull/1449/files#diff-63a9c44a44acf85fea213a857769990937107cf072831e1a26808cfde9d096b9) and [.devcontainer/dev-requirements.txt](https://github.com/brian-team/brian2/pull/1449/files#diff-b36f7558d5f7f38cbb7489ef21549dfd0a946358c2692cd2032497ff67cd56d8). Would it be possible to consolidate them all into one location that gets read by the others?

I was wondering about the same thing, but I did not find an easy solution. We could use an approach like the one described here: https://github.com/pre-commit/pre-commit/issues/945#issuecomment-527603460 , which would mean that `pre-commit` wouldn't install anything but just use what is installed on the system. This would be ok for running the pre-commit hooks in the CI, but for running them locally it would mean that we lose the isolation that `pre-commit` provides. The user could update a tool locally, and thereby potentially change the result of applying the hooks – this would be very confusing if the commit then gets refused by the CI on GitHub.
Going the other way, i.e. somehow parsing `.pre-commit-config.yaml` and using it to `pip` install requirements is probably the better way. If anyone wants to have a go at anything like that, please go ahead :wink:

PR #1450, opened 2023-01-04
Failure to produce documentation with python3.11.

emollier:
In [Debian Bug#1027936], Adrian Bunk noticed that the documentation failed to be produced with python3.11 with the following symptom:

	  File "/usr/lib/python3/dist-packages/sphinx/config.py", line 350, in eval_config_file
	    exec(code, namespace)
	  File "/<<PKGBUILDDIR>>/docs_sphinx/conf.py", line 71, in <module>
	    generate_examples.main(root_dir, target_dir)
	  File "/<<PKGBUILDDIR>>/debian/tmp/usr/lib/python3.11/dist-packages/brian2/sphinxext/generate_examples.py", line 72, in main
	    with codecs.open(fname, 'rU', encoding='utf-8') as f:
	         ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
	  File "<frozen codecs>", line 906, in open
	ValueError: invalid mode: 'rUb'

This patch removes a couple of uses of unified newline "open" invocations, which have no effect since the first version of python3. This may cause changes of behavior with python2, but this interpreter is unmaintained since 2020.

[Debian Bug#1027936]: https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=1027936

Signed-off-by: Étienne Mollier <emollier@debian.org>

### DIFF
--- a/.github/workflows/testsuite.yml
+++ b/.github/workflows/testsuite.yml
@@ -110,3 +110,40 @@ jobs:
         coveralls --finish
       env:
         GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+
+  test_doc_build:
+    needs: get_python_versions
+    name: Test building the documentation
+    runs-on: ubuntu-latest
+    defaults:
+      run:
+        shell: bash -l {0}
+    steps:
+    - name: Checkout Repository
+      uses: actions/checkout@v2
+      with:
+        fetch-depth: 0
+        submodules: true
+
+    - name: Setup Conda and Python
+      uses: conda-incubator/setup-miniconda@v2
+      with:
+        auto-update-conda: true
+        auto-activate-base: false
+        channels: conda-forge,defaults
+        channel-priority: true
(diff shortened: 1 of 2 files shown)


mstimberg:
Many thanks for the report and fix. I was a bit surprised that we still had the `U` flags in there, since I thought that the `pyupgrade` tool had taken care of this. It turns out that it only did for `open` calls, not for `codecs.open`. I think we only used `codecs.open` here since we needed to enforce `utf-8`, but with modern Python versions, we can also do this with `open`. I therefore commited a change that gets rid of the use of `codecs` here. I also updated our test suite so that it tests building the HTML docs with the latest Python version – hopefully that should avoid this type of issue in the future. Thanks again, I'll go ahead with the merge as soon as the tests have passed.

mstimberg (approved):

emollier:
Hi Marcel,
Great, glad this helped!
Have a nice day,  :)
Étienne.

PR #1451, opened 2023-01-12
Create list of package versions

bdevans:
In response to the discussion about synchronising package versions across the whole project (https://github.com/brian-team/brian2/pull/1449#issuecomment-1372476018), as an interim measure, this is a manually curated list of files which contain package versions, enabling us to check them and update as necessary when updating a dependency.

### DIFF
--- a/versions.md
+++ b/versions.md
@@ -1,0 +1,20 @@
+Package versions
+================
+
+This file contains a list of other files where versions of packages are specified so that they can easily be found when upgrading a dependency version, keeping them all in sync. 
+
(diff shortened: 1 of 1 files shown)


bdevans:
I may have missed some and it might be better to tidy it away in the `dev` folder, but it's a start!

mstimberg:
Oh, thanks a lot, that's really useful. It also clearly shows what a mess everything is :grimacing:
And it reminded me that we should get rid of the conda recipe, the actual recipe we use is https://github.com/conda-forge/brian2-feedstock/blob/main/recipe/meta.yaml

> I may have missed some and it might be better to tidy it away in the `dev` folder, but it's a start!

The words "tidy" and "`dev` folder" don't quite work together in our repo ;-)

PR #1452, opened 2023-01-12
Update conda-forge recipe

bdevans:


mstimberg:
Thanks again :)
Minor thing: you can include `[ci skip]` in the commit message (preferrably after an empty line), then it won't trigger the full test suite for changes like this.

PR #1454, opened 2023-01-18
Force UTF-8 encoding for file logs

mstimberg:
Makes non-ASCII characters in log messages work on Windows.

See https://brian.discourse.group/t/logging-error-on-windows/872/3


mstimberg:
Merging my own PR. This is a minor fix, but potentially a very annoying bug for users that run into it. In addition to the CI, I tested manually in a Windows VM (the new test triggers a failure without the fix).

PR #1460, opened 2023-03-14
Remove numpy-fallback for SpatialNeuron

### LINKED ISSUE #1459: SpatialNeuron (spatialstateupdate) fails with numpy without scipy
As reported in the [discussion forum](https://brian.discourse.group/t/about-the-problems-of-brian2-test-in-the-test-process-failures-test-spatialneuron-timearray/931), the no-scipy-fallback for `SpatialNeuron` fails with current `master`. This should either be fixed or alternatively we simply declare scipy as a hard dependency? If we fix it, we should also make sure that the CI tests cover this situation.

mstimberg:
Running `SpatialNeuron` with `numpy` now requires the `scipy` package – the previous "pure-numpy" fallback has been removed. It was broken and complicated the code in general; as discussed in #1459, `scipy` is a standard dependency and maintaining an alternative, scipy-independent approach seems unnecessary.

Closes #1459

### DIFF
--- a/brian2/codegen/generators/base.py
+++ b/brian2/codegen/generators/base.py
@@ -120,7 +120,7 @@ class CodeGenerator:
         used for example by the `CPPCodeGenerator` to set up all the supporting
         code
         """
-        raise NotImplementedError
+        return {}
 
     def translate_one_statement_sequence(self, statements, scalar=False):
         raise NotImplementedError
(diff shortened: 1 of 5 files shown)


mstimberg:
I'll go ahead with merging this as it is. See discussion in #1459 for more details. For now, this simply fails with an error if scipy is not installed, but adding it as a dependency would be trivial, of course.

PR #1467, opened 2023-05-01
Fix typo that causes an error

pjsph:


mstimberg (approved):
Many thanks! Typos like this should ideally be discovered by our test suite of course – I added a test covering the code in question now.

PR #1468, opened 2023-05-15
Update devcontainer

bdevans:
This includes a few fixes and bumps to the developer packages used.


mstimberg (approved):
Thanks for the version updates and fixes. The newer versions made the pre-commit hook fail, though, so I committed a few additional changes. The newer flake8-bugbears version flagged some (false positive) issues, I corrected/ignored them in the source code. The newer black version made a large number of changes to multi-line strings (since we use the `--preview` flag). I'd rather not commit these changes, since it might change again in the near future. For now, I'd suggest that we stay with the previous version of black until the "preview style" becomes the new stable style (planned for beginning of 2024). I also installed the flake8 extension in the dev container so that Visual Code stops nagging me about it ;-) If all that looks ok for you, happy to merge.

PR #1473, opened 2023-06-13
Fix numpy warnings

mstimberg:
This fixes a number of issues that will raise warnings in the forthcoming numpy version (1.25).

### DIFF
--- a/brian2/codegen/optimisation.py
+++ b/brian2/codegen/optimisation.py
@@ -38,12 +38,12 @@ def evaluate_expr(expr, ns):
 
     Example
     -------
-    >>> assumptions = {'sin': DEFAULT_FUNCTIONS['sin'].pyfunc,
-    ...                'pi': DEFAULT_CONSTANTS['pi'].value}
+    >>> assumptions = {'exp': DEFAULT_FUNCTIONS['exp'].pyfunc,
+    ...                'inf': DEFAULT_CONSTANTS['inf'].value}
     >>> evaluate_expr('1/2', assumptions)
     (0.5, True)
-    >>> evaluate_expr('sin(pi/2)', assumptions)
-    (1.0, True)
+    >>> evaluate_expr('exp(-inf)', assumptions)
+    (0.0, True)
     >>> evaluate_expr('sin(2*pi*freq*t)', assumptions)
     ('sin(2*pi*freq*t)', False)
     >>> evaluate_expr('1/0', assumptions)
--- a/brian2/core/variables.py
+++ b/brian2/core/variables.py
@@ -371,6 +371,9 @@ class Constant(Variable):
     def get_value(self):
         return self.value
 
+    def item(self):
+        return self.value
+
 
 class AuxiliaryVariable(Variable):
     """
@@ -502,6 +505,12 @@ class ArrayVariable(Variable):
 
     def get_value(self):
         return self.device.get_value(self)
+
+    def item(self):
+        if self.size == 1:
+            return self.get_value().item()
+        else:
+            raise ValueError("can only convert an array of size 1 to a Python scalar")
 
     def set_value(self, value):
         self.device.fill_with_array(self, value)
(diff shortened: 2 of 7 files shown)


brian2/core/variables.py
```diff
@@ -957,15 +966,9 @@ def set_item(self, item, value, level=0, namespace=None):
...
-                    # length-1 arrays are also convertible to float, but we
-                    # don't want the repr used later to be something like
-                    # array([...]).
-                    value = value[0]
-                except (IndexError, TypeError):
-                    # was scalar already apparently
-                    pass
+                if isinstance(value, str):
+                    raise TypeError  # Will be dealt with below
+                value = np.asanyarray(value).item()
```
thesamovar, line 971: Not sure I understand these changes, can you summarise?

brian2/codegen/optimisation.py
```diff
@@ -38,12 +38,12 @@ def evaluate_expr(expr, ns):
 
     Example
     -------
-    >>> assumptions = {'sin': DEFAULT_FUNCTIONS['sin'].pyfunc,
-    ...                'pi': DEFAULT_CONSTANTS['pi'].value}
+    >>> assumptions = {'exp': DEFAULT_FUNCTIONS['exp'].pyfunc,
+    ...                'inf': DEFAULT_CONSTANTS['inf'].value}
```
thesamovar, line 42: Any reason for change from sin to exp etc.? Not that it matters much, just curious.

thesamovar (commented):
Can you briefly summarise what's going on here? I have to say I don't really understand the changes.

mstimberg:
> Can you briefly summarise what's going on here? I have to say I don't really understand the changes.

Sure. The main issue that is adressed here is that in a few places, we used code like this:
```python
int(self.variables["N"].get_value())
```
The result of `self.variables["N"].get_value()` is sometimes a scalar value (e.g. for `NeuronGroup` where `N` is a constant), and sometimes a length-1 array (e.g. for `Synapses`, where the value can change). By wrapping it in `int`, we were sure about always getting a scalar Python int. Starting with numpy 1.25 (currently a release candidate), this will raise a warning, and in future versions it will raise an error. To replace this code, I added an `item()` function to `ArrayVariable`, `Constant`, and `Quantity` (the latter already had it inherited from `numpy`, but discarded the units). This mirrors the built-in function that `numpy` offers:
```pycon
>>> import numpy as np
>>> np.array(3).item()
3
>>> np.array([3]).item()
3
>>> np.array([3, 4]).item()
  File "<stdin>", line 1, in <module>
ValueError: can only convert an array of size 1 to a Python scalar
```

brian2/codegen/optimisation.py
mstimberg: I briefly explain this in the commit message: our docttest relied on the fact that `np.sin(np.pi/2) == 1.0`, which is reasonable but not guaranteed (results are only guaranteed to be < 4 epsilons away from the correct value). With the previous versions it matched 1.0 exactly, but with the newer, faster implementations in `numpy` 1.25 this is no longer the case (even with older versions, `np.sin(0.0)` or `np.sin(np.pi)` did not return 0.0). So I replaced it with `exp(-inf)` which I think is guaranteed to be exactly zero – at least it works with both numpy 1.24 and numpy 1.25 :)

brian2/core/variables.py
mstimberg: This change is a slightly more complex issue related to what I wrote in my earlier comment: in future versions, we can no longer use `float(x)` as a check that `x` is either a scalar number or a length-1 array with a number (and not for example a string) – it will wail on length-1 arrays.

brian2/monitors/spikemonitor.py
```diff
@@ -357,7 +357,7 @@ def all_values(self):
                 "them."
             )
         indices = self.i[:]
-        sort_indices = np.argsort(indices)
```
mstimberg, line 360: BTW: This was always wrong (we need a stable sort here), but it only now got triggered with numpy 1.25 (and only on GitHub, not on my local machine).

thesamovar (approved):
OK, all that makes sense - thanks!

PR #1475, opened 2023-06-27
Upgrade build infrastructure

mstimberg:
This modernizes our build infrastructure by moving all metadata to `pyproject.toml` and switching from `versioneer` to `setuptools_scm`. The latter was mostly triggered by the fact that I didn't get the most recent release to build on `conda-forge`. In general, `setuptools_scm` seems to be more "mainstream" and better supported.


bdevans:
Just to check - does [versions.md](https://github.com/brian-team/brian2/blob/master/versions.md) need updating following these changes?

PR #1480, opened 2023-07-11
Better logging for extension modules

mstimberg:
Brian configures the logging infrastructure under the `brian2` logger root, so all uses of the logger need to have a `brian2.something` name.  This leads extension packages like `brian2cuda` to use names that do not correspond to actual module names, e.g.:

https://github.com/brian-team/brian2cuda/blob/f240d34c45daa858716da9176936cd642f17ccb2/brian2cuda/cuda_generator.py#L24C1-L24C64

This can be potentially confusing for users, but also for the test suite: in several places the test suite checks for specific warnings (or the absence of such warnings). If `brian2cuda` logs a warning in one of those tests, the test will fail even though the warning is not relevant to the test.

With this PR, packages can use Brian's logger with `get_logger` using their own name (typically `get_logger(__name__)`). When interacting with Python's `logging` module internally, this name will be prefixed with `brian2`, so that all the configuration (file log etc.) is still applicable. However, when log messages are printed or written to a file, they appear without the prefix. Also, the `catch_logs` context manager will by default ignore messages that do not have names starting with `brian2`, so that tests are not affected by warnings printed by extension packages.

### DIFF
--- a/brian2/utils/logger.py
+++ b/brian2/utils/logger.py
@@ -311,6 +311,18 @@ class NameFilter:
         return self.name != record_name
 
 
+class RemoveBrian2Filter(logging.Filter):
+    """
+    A class for removing the ``brian2`` prefix from log messages.
+    Will be used for extension packages that use the Brian logging system.
+    """
+
+    def filter(self, record):
+        assert record.name.startswith("brian2."), record.name
+        record.name = record.name[7:]
+        return True
+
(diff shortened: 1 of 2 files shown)


mstimberg:
Given that this is a fairly low-stakes internal feature, I'll go ahead with merging this without review.

PR #1484, opened 2023-09-14
Numpy 2 compatibility

mstimberg:
This fixes a few issues for compatibility with the upcoming (backwards-compatibililty breaking) 2.0 version of numpy. It also switches off the GSL tests for now – GSL is broken on C++ standalone for unrelated reasons (and by chance/bad luck, the way it broke made it skip the tests so I only realized recently…). 

I still need to check whether any of the changes requires us to update our minimum version requirement for numpy, but I don't think it is.


mstimberg:
> This fixes a few issues for compatibility with the upcoming (backwards-compatibililty breaking) 2.0 version of numpy. It also switches off the GSL tests for now – GSL is broken on C++ standalone for unrelated reasons (and by chance/bad luck, the way it broke made it skip the tests so I only realized recently…).
> 
> I still need to check whether any of the changes requires us to update our minimum version requirement for numpy, but I don't think it is.

All good – only the Cython dependency needed an update.

PR #1492, opened 2023-11-06
Setuptools related improvements

sanjayankur31:
I'm going to preface this PR by saying that I do not know why the current upstream configuration works. From what I see in the various docs, it shouldn't, but there's so much going on in the Python packaging ecosystem that I'm no longer confident I understand it all properly :joy: 

While building on Fedora, the current pyproject.toml etc., doesn't work---it only includes the top level brian2 package in the generated wheels. This seems to be the correct behaviour because as the setuptools documentation notes (or suggests rather), only including the top level package name in the package list is not enough. One either has to list all sub-packages, or use the discovery tools.

https://setuptools.pypa.io/en/latest/userguide/package_discovery.html#package-discovery-and-namespace-packages

So, to ensure that all sub-packages and the necessary template/data is included, I had to tweak these files as noted in the PR. (the tests from my local build are still running, so there may be more data files that are needed to be added to the manifest etc.).

### DIFF
--- a/MANIFEST.in
+++ b/MANIFEST.in
@@ -23,3 +23,12 @@ exclude .coveragerc
 exclude .gitattributes
 exclude .gitmodules
 include brian2/_version.py
+
+include brian2/devices/cpp_standalone/templates/*
+include brian2/devices/cpp_standalone/templates_GSL/*
+include brian2/tests/test_templates/fake_package_1/templates/*
+include brian2/tests/test_templates/fake_package_2/templates/*
+include brian2/codegen/runtime/numpy_rt/templates/*
+include brian2/codegen/runtime/GSLcython_rt/templates/*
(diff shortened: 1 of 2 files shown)


mstimberg:
Hi @sanjayankur31 thanks for your PR! I have to say I am now more confused than ever – if I remember correctly, I only arrived at the current solution by trial and error, as you say, the documentation seems to suggest it shouldn't work. But then, our [`cibuildwheel`](https://cibuildwheel.readthedocs.io/)-based wheels do contain all the subpackages (and templates, etc.), and the same for the source package we build with `python -m build`. Weird.
But of course everything still works with your additions, and if it fixes building on Fedora, I'm of course happy to merge it.

mstimberg (approved):

PR #1497, opened 2023-12-28
Extend compiler cpu optmizations to loongarch64.

emollier:
This patch extends the just-in-time compiler options in brian2/codegen/cpp_prefs.py to optimize aggressively on loongarch64 CPU, similarly to numerous other platforms already supported by brian2.

It was originally published in [Debian bug #1059428] and provided by Loongson.

[Debian bug #1059428]: https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=1059428


mstimberg (approved):
LGTM!

PR #1504, opened 2024-02-13
Document how to record population averages in a memory efficient way

paulbrodersen:
Hi, following our discussion [here](https://brian.discourse.group/t/how-to-implement-a-populationstatemonitor/1138), I added a small section to the documentation that outlines how to continuously record population averages in a memory efficient way (so I can find the information next time I need it). 

The section is based on the example given by @mstimberg [here](https://brian.discourse.group/t/is-there-a-way-to-monitor-the-average-membrane-potential-of-a-group-of-neurons/658/3).
 
```python
group = NeuronGroup(..., 'dv/dt = ...', ...)

# Dummy group to store the average membrane potential at every time step
vm_container = NeuronGroup(1, 'average_vm : volt')
# Synapses averaging the membrane potential of all neurons
vm_averager = Synapses(group, average_vm, 'average_vm_post = v_pre/N_pre : volt (summed)')
vm_averager.connect()
# Monitor recording the average membrane potential
vm_mon = StateMonitor(vm_container, 'average_vm', record=0)
```

I made the following changes:

1. I indicated that the units of the variable and the variable average need to match: `group = NeuronGroup(..., 'dv/dt = ... : volt', ...)`
2. I corrected a typo: `vm_averager = Synapses(group, vm_container, ...)`
3. I set `record=True` to match the examples in the same section.

### DIFF
--- a/docs_sphinx/user/recording.rst
+++ b/docs_sphinx/user/recording.rst
@@ -288,4 +288,4 @@ group and the `Synapse` class' :ref:`summed variable syntax
     vm_averager.connect()
 
     # Monitor recording the average membrane potential
-    vm_monitor = StateMonitor(vm_container, 'average_vm', record=0)
+    vm_monitor = StateMonitor(vm_container, 'average_vm', record=True)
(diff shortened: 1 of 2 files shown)


mstimberg (approved):
Looks great to me, many thanks for this contribution! I changed a minor detail: if I am not mistaken, the memory usage in your example would be even worse since each value is a double, i.e. takes 8 bytes. So recording 1000 × 100 × 10000 (neurons × duration × timesteps per second) values would take up ~ 8 Gigabytes :grimacing:

paulbrodersen:
Apparently, I hadn't had enough coffee this morning, because I thought a float**64** would only need 8 bits not 8 bytes... 
Thanks for the correction!

PR #1513, opened 2024-03-13
Add dependabot configuration file

mstimberg:
This should activate dependabot for github-actions and the devcontainer, let's see how this works.

CC @bdevans

https://docs.github.com/en/code-security/dependabot/dependabot-version-updates/about-dependabot-version-updates


mstimberg:
I'll merge this right away since it won't do anything by itself/in a branch.

PR #1517, opened 2024-03-15
New array copy semantics

mstimberg:
In numpy 1.x, `array(x, copy=False)` meant : do not copy if possible. In numpy 2.0, it will raise an error if copying is not possible. This commit replaces the use of `array(x, copy=False)` by `asarray(x)`, which has the "do not copy if possible" in both numpy 1.x and numpy 2.0.


mstimberg:
The changes in this PR should hopefully make tests pass against numpy 2.0: https://github.com/brian-team/brian2/actions/runs/8294727053

PR #1534, opened 2024-05-31
Fix reference doc issues

### LINKED ISSUE #1532: Issue: Tutorial Part 1 documentation for StateMonitor
Screenshot from 2024-05-28 15-07-19

Issue with Tutorial Part 1, [Link](https://brian2.readthedocs.io/en/stable/resources/tutorials/1-intro-to-brian-neurons.html#)

For the above example, `record=True` parameter is passed to StateMonitor.

But according to documentation of StateMonitor class [here](https://brian2.readthedocs.io/en/stable/reference/brian2.monitors.statemonitor.StateMonitor.html#statemonitor-class) , 

`record` parameter is used to specify the index of a particular Neuron (in case of population of Neuron)

mstimberg:
When fixing warnings about unused variables and similar issues quite a while back (commit 0344b00), I inadvertently broke a few features of the reference documentation – partly because some of the warnings were somewhat misleading (e.g. some function call return values weren't used, but the functions updated global variables). This should fix the issue dicussed in #1532. 

I've built the updated documentation here, from a quick glance it seems to be working correctly: https://brian2.readthedocs.io/en/fix_doc_reference_issues/

Closes #1532

### DIFF
--- a/brian2/sphinxext/docscrape.py
+++ b/brian2/sphinxext/docscrape.py
@@ -148,7 +148,7 @@ class NumpyDocString:
                 stop = i
                 break
 
-        return doc[start:-stop]
+        return doc[start : len(doc) - stop]
 
     def _read_to_next_section(self):
         section = self._doc.read_to_next_empty_line()
--- a/brian2/sphinxext/examplefinder.py
+++ b/brian2/sphinxext/examplefinder.py
@@ -64,8 +64,12 @@ def auto_find_examples(obj, headersymbol="="):
     make mistakes but is usually going to be correct).
     """
     name = obj.__name__
-    examples = sorted(the_examples_map[name])
-    tutorials = sorted(the_tutorials_map[name])
+    # These will only go through the examples/tutorials once. The dictionaries are cached in the
+    # module variables `the_examples_map` and `the_tutorials_map`.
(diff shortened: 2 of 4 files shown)


thesamovar (approved):
This all looks pretty reasonable to me!

mstimberg:
With hindsight, I'm really embarrassed about having changed `doc[start : len(doc) - stop]` into `doc[start:-stop]`. Like, oh, these dummies don't know that you can simply use negative indices for that in Python. But well, this works well unless `stop` is 0 :grimacing:

thesamovar:
I've shot myself in the foot with that one too. Increasingly think that notation might have been a mistake for Python. Some language allows you to write ``end-1`` or something like that, which is maybe better.

mstimberg:
Actually in this specific case, making `-0` different from `0` would have worked, too :-P

thesamovar:
Hah! Maths has things to say about that though. ;-)

Although I think in floating point representations -0.0 is different from +0.0.

mstimberg:
> Although I think in floating point representations -0.0 is different from +0.0.

It is indeed! and things like $\frac{1}{-\infty} = -0$ are even required by the IEEE standard, I think. A can of worms :laughing:

PR #1535, opened 2024-06-02
Fix Error in Brunel_Wang_2001.py

guy-singer:
Deleted "eqs_pre_ext" in rhe Brunel-Wang 2001 example implementation.

These equations aren't used anywhere in the simulation. 

Their inclusion is likely an oversight. 

Removed to make the code more clear.


mstimberg (approved):
Makes sense! I think this equation was necessary previously, when the model still used less efficient `PoissonGroup` instead of `PoissonInput`. Thanks for spotting this.

Issue #1499, opened 2024-01-09
Returned value needs to be retrieved for object to work


jlubo:
When defining a `PoissonInput` object, the value returned by the constructor needs to be retrieved for the object to work. While I think that this is not a big issue, it can be confusing because one wouldn't expect the retrieval of the object to make any difference. The same issue might exist for other object types like `NeuronGroup`, `SpikeGeneratorGroup`, etc, but for those the practical relevance should be small because the returned object is usually needed by the subsequent code.

Here is an example of what works and what doesn't (altered code blocks for [this](https://github.com/brian-team/brian2/blob/04c91b824ef6d8c9872e80f8d14154ee815be058/examples/frompapers/Brunel_2000.py) example):

```
# Case 1: retrieval of returned object -- WORKS
external_poisson_input = PoissonInput(
    target=neurons, target_var="v", N=C_ext, rate=nu_ext, weight=J
)
```
```
# Case 2: no retrieval of returned object -- DOES NOT WORK
PoissonInput(
    target=neurons, target_var="v", N=C_ext, rate=nu_ext, weight=J
)
```
```
# Case 3: retrieval of returned object with placeholder -- WORKS
_ = PoissonInput(
    target=neurons, target_var="v", N=C_ext, rate=nu_ext, weight=J
)
```

I guess a solution to the issue may be to simply throw a warning if the returned value is not retrieved (in case 2).

mstimberg:
Hi Jannik, 
I'd say that this is working as intended, actually. A `PoissonInput` is an element of a network (as a `NeuronGroup`, etc.) and therefore needs to be added to the network. If you call `run`, it uses the visible variables (i.e. the variables in the current scope) to decide which objects to include in the network. The other option is to explicitly construct your `Network` object yourself, but in this case you'd also need a variable storing the `PoissonInput` object, or at least do something like this:
```python
net = Network()
# ...
net.add(PoissonInput(...))
```
See https://brian2.readthedocs.io/en/stable/user/running.html#networks


During the early days of Brian 2 we decided for this mechanism and against a system that would do what you expected here, i.e. something that tracks all the objects that have been created and adds them to the network. One of the reasons is that it makes it more straightforward to have independent simulations, such as in:
```python
def sim1():
   group1 = NeuronGroup(...)
   run(...)
def sim2():
   group2 = NeuronGroup(...)
   run(...)

# Run simulations
sim1()
sim2()
```
[...]

PR #1536, opened 2024-06-04
Unused object warning

mstimberg:
This addresses the issue in #1499, by raising a warning if an object gets garbage collected but has never been included in any `Network`. Instead of dealing with the complexities of weak references and the `InstanceTracker`, as initially proposed in #1499, this simply uses the `__del__` function in `BrianObject`. This seems to be working well, and running it against our examples I did not encounter any issues. There are false positives for "artificial" examples like https://brian2.readthedocs.io/en/stable/examples/synapses.efficient_gaussian_connectivity.html, since they create `Synapses` but do not ever run a simulation. In principle, we could have another check that only triggers these warnings when there has been a `run` at some point, but I am not sure whether it is worth the complication.
This code from the forum (where the `define_network` function only returns the `NeuronGroup`s, but not the `Synapses`) is a good example for the kind of code that will trigger a warning: https://brian.discourse.group/t/synapses-not-functioning/1118

### DIFF
--- a/brian2/__init__.py
+++ b/brian2/__init__.py
@@ -71,7 +71,7 @@ except ImportError:
         )
         __version_tuple__ = tuple(int(x) for x in __version__.split(".")[:3])
     except ImportError:
-        logging.getLogger("brian2").warn(
+        logging.getLogger("brian2").warning(
             "Cannot determine Brian version, running from source and "
             "setuptools_scm is not installed."
         )
--- a/brian2/core/base.py
+++ b/brian2/core/base.py
@@ -180,6 +180,22 @@ class BrianObject(Nameable):
     #: dependent objects such as `StateUpdater`
     add_to_magic_network = False
 
+    def __del__(self):
+        # For objects that get garbage collected, raise a warning if they have
+        # never been part of a network
+        if (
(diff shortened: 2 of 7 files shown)


thesamovar (approved):
This seems fine to me, although I have to admit I didn't entirely understand the ``__del__`` logic.

mstimberg:
> This seems fine to me, although I have to admit I didn't entirely understand the `__del__` logic.

The reasoning is: objects fill in their `_network` attribute when they are simulated as part of a `Network` – this is used for the errors when an object has been simulated as part of different networks. We therefore raise the warning when an object gets garbage collected when an object has its attribute `_network` still as `None` (for some reason it can happen that the attribute does not exist at all, but we can simply ignore that).

The second check is to avoid that objects such as `StateUpdater`, `Resetter`, etc. also raise this warning. It's a bit of a hack, we simply assume that everything that has a `group` attribute is something dependent on other objects and not worth warning about. The other alternative would be to overwrite the `__del__` method in all the subclasses. We actually need to do this for `Subgroup`, because it stores its reference to the main group as `source`, not as `group` :upside_down_face:

PR #1548, opened 2024-07-14
Add Diehl and Cook (2015) small model

bjourne:
Hello Brian2 team,

Code below is my attempt at making a modest contribution to the Brian2 simulator. In Brian2's Discourse there has been some interest in this model so adding it as an example seems like a good idea. The code is a cleaned up and Python 3-compatible version of [Diehl and Cook's original implementation](https://github.com/peter-u-diehl/stdp-mnist/tree/master) and [Bruce Tian's Brian2 implementation](https://github.com/tianjihua254/stdp-mnist-brian2).


mstimberg:
Many thanks! This model is discussed quite often on the forum, so adding it to the documentation would clearly make many users happy. I won't be able to have a closer look at your code before next week, but it is on my TODO list :-) Please do not hesitate to ping me if you don't hear anything from me…

mstimberg:
Many apologies for taking so long to reply/review. One reason is that I started to port your example to standalone mode to make it run faster :blush: It now works, but it is still a bit ugly, I am considering changing and option in Brian's standalone mode first, and then I might write a blog post about this, since the general approach might be interesting to many users. In any case, thanks again for your contribution, it is a really nice example of a type of usage that we don't currently document in our examples.

I've made a few minor changes to your code, which should not change anything but make the code a bit more "idiomatic".

I've only two questions for some aspects of the model I do not quite understand (but they might be inherited from the original implementation):
- what is the need for the `timer` variable defined as `dtimer/dt = 0.1 : second` and used in the threshold with `... and (timer > 5*ms)`? This seems to be redundant with the `refractory` functionality that is also used. And given the definition of `dtimer/dt`, the `timer` variable will be at 5ms after 50ms, which is a bit confusing…
- why is the initial value for the membrane potential `V_EXC_REST - 40 *mV` and not just `V_EXC_REST` (same for inhibitory neurons)? It doesn't matter here since this is only used once at the very beginning of the train/observe/test period, but it still feels off (and makes the very first sample demonstration quite different from the other ones).

bjourne:
Hi Marcel!

No need to apologize. The model is indeed quite slow. On my computer simulating 3200 excitatory neurons takes twice as long as real-time. So simulating one hour takes two hours, for example. I tried myself to compile the model with Brian2's C++, but it was complicated and I couldn't get it to work.

You are right, the `dtimer/dt` and `dv/dt` mechanics are confusing. The intent is to keep current at zero for 5 ms and also prevent neurons from spiking for 50 ms. It is not covered in Diel and Cook's article, but is present in their source code. And if you don't do it like that the model doesn't work. Same story with the incorrect scaling of `dtimer/dt`. It's confusing, but it is in their source code and (afaict) the model misbehaves if you change it. My guess is that this is some workaround for feature Brian1 missed.

I honestly don't remember if using `V_EXC_REST - 40 *mV` over `V_EXC_REST` made any difference. It probably doesn't.

Btw, feel free to change the code in any way you see fit.

mstimberg (approved):
Thanks for the clarifications, and thanks again for the example! I have a mostly working standalone version (I used it to generate the figure), but there's still something odd going on, so I won't add it to the documentation right now.

PR #1556, opened 2024-08-14
potential VariableView.shape fix

### LINKED ISSUE #1555: Confusing VariableView behavior: reports shape of parent Quantity
I expect to be able to use VariableView like a Quantity, but I get some surprising behavior where `shape` reflects the parent `Group` object. `len()` is not affected

```python
import brian2.only as b2
ng = b2.NeuronGroup(10, 'v : 1')

print(ng[:5].v) 
# <neurongroup_subgroup.v: array([0., 0., 0., 0., 0.])>

print(len(ng[:5].v))
# 5

# 🚨 the weird part 🚨
print(ng[:5].v.shape)
# (10,)

# this works as expected, but it's a new `Quantity` as opposed to a `VariableView`
print(ng.v[:5].shape)
# (5,)
```

kjohnsen:
I don't know if this behavior is desired, but if it is and tests pass, this seems to work.

closes #1555

### DIFF
--- a/brian2/core/variables.py
+++ b/brian2/core/variables.py
@@ -1573,10 +1573,7 @@ class VariableView:
     # Get access to some basic properties of the underlying array
     @property
     def shape(self):
-        if self.ndim == 1:
-            return (self.variable.size,)
-        else:
-            return self.variable.size
+        return self[:].shape
 
     @property
     def ndim(self):
--- a/brian2/core/variables.py
+++ b/brian2/core/variables.py
@@ -1573,7 +1573,18 @@ class VariableView:
     # Get access to some basic properties of the underlying array
     @property
     def shape(self):
-        return self[:].shape
+        if self.ndim == 1:
+            if not self.variable.scalar:
+                # This is safer than using the variable size, since it also works for subgroups
+                # see GitHub issue #1555
+                size = self.group.stop - self.group.start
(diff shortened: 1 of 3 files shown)


kjohnsen:
The only test that failed on my machine was
```
E               AssertionError: Transformation for statement 'x = x + x' gave an unexpected result: RHS expressions '2.00000000000000' and '2' are not identical
E               assert 2.00000000000000 == 2

brian2/tests/test_codegen.py:488: AssertionError
```

mstimberg:
Many thanks for the PR. Unfortunately, this fix would break standalone mode in some (corner-)cases. E.g. the following
```pycon
>>> import brian2.only as b2
>>> b2.set_device('cpp_standalone')
>>> ng = b2.NeuronGroup(10, 'v : 1')
>>> ng.v = 'rand()'
>>> ng.v.shape
(10,)
```
would raise an error, since it tries to access the values of `ng.v`, which are not known in standalone mode before running the simulation. I will have a closer look, I am sure there is a reasonably simple solution.

mstimberg:
@kjohnsen I committed a slight variant of the previous code (plus some tests to make sure that everything works). I am going to merge it now, but please let me know if you run into any issues.

kjohnsen:
Groovy. Thank you!

PR #1559, opened 2024-09-10
Use Mersenne Twister from C++ standard library

### LINKED ISSUE #1483: Random number generation on C++ standalone is slow
I only just now realized that random number generation on C++ standalone is slow compared to Cython/numpy. E.g., with a trivial example like:
```python
G = NeuronGroup(10000, 'x : 1')
G.run_regularly('x = rand()')
```
the `run_regularly` code object uses ~2.5s on Cython and numpy, and ~7s on C++ standalone. My hunch was that this is because in Cython/numpy, we create a buffer of random numbers at once using `numpy`'s `rand` function, whereas in C++ standalone, we call `rk_double` for each number we need. But it might be more fundamental: numpy switched to a different random number generator a couple of years ago, while in C++ standalone, we are still using the Mersenne-Twister "legacy code"…

Not sure how feasible/easy it is to switch to the same generator in C++ standalone. 

This could also be the time to tackle #19, but I don't see any obvious/simple solution.

mstimberg:
When I opened #1483, I did not realize that we have a simple solution – C++11 (not *that* recent!) comes with PRNGs and it has a Mersenne Twister implementation that is about 6 times as fast as our "radomkit" implementation. So we can delete a lot of code and get faster execution for free :+1: 

Now, there is one thing that I am not quite sure about: given that we keep using the same algorithm, the quality of the random numbers, the way of how the `seed` function, multi-threading, etc. works stays the same *but* you can no longer reproduce the same random numbers simply by setting the same seed, since the implementations for generating the uniform/Gaussian distributions from the raw random numbers differ (and even if they happened to be the same, there are no guarantees how they are implemented).

As an alternative, we can keep our old `rk_double` and `rk_gauss` functions around, and just use the new MT generator, which keeps almost all the performance improvement and allows us to exactly reproduce previous numbers. The cost is that the code gets a bit messier, since we still need to ship our own functions – and getting things to work correctly with threads and seeds is slightly more complicated. But maybe worth it? Being able to reproduce the same random numbers as before would be quite nice to guarantee. I think one could say we imply this feature in our docs:

>  In general, we therefore only guarantee the use of the same numbers if the code generation target and the number of threads (for C++ standalone simulations) is the same.

https://brian2.readthedocs.io/en/stable/introduction/compatibility.html#random-numbers

Closes #1483


thesamovar:
6x speed improvement definitely good! I guess reproducibility is probably important for some people, so yeah I would agree with your solution.

mstimberg:
Yeah, while writing this I convinced myself that this would be good to have ;-) I'll try to refactor a code a bit so that the RNG is a bit more encapsulated than what it is now – this should also make it easier to switch to a different RNG in the future.

mstimberg:
This now looks good from my side, and it reproduces the same random numbers as before. When we generate `randn` and `rand` numbers, the speed improvement is less spectacular than by looking only at the random bit generation, but a simple example like this:
```python
G = NeuronGroup(10000, """x : 1
                          y : 1""")
G.run_regularly('x = rand()', name="rand")
G.run_regularly('y = randn()', name="randn")
```
still goes from
```
randn_codeobject                        3.26 s    52.15 %
poisson_spike_thresholder_codeobject    1.51 s    24.20 %
rand_codeobject                         1.48 s    23.65 %
```
to
```
randn_codeobject                        1.18 s    63.96 %
poisson_spike_thresholder_codeobject    0.35 s    18.76 %
rand_codeobject                         0.32 s    17.29 %
```
i.e. `randn()` is about twice as fast, and `rand()` and `PoissonGroup` (which is using `rand()` every time step) are about five times as fast.

PR #1561, opened 2024-09-12
Python3.13

mstimberg:
Nothing particular interesting here, but one doctest failed with Python 3.13 (output of `ast.dump` changed slightly). The remaining changes only prevent deprecation warnings, nothing would break yet.


mstimberg:
Tests pass with Python 3.13: https://github.com/brian-team/brian2/actions/runs/10830134032

PR #1581, opened 2025-01-23
Add an example of a homeostatic, inhibitory STDP rule with a global f…

paulbrodersen:
…actor

Contribution requested here:
https://brian.discourse.group/t/valueerror-equations-of-type-parameter-cannot-have-a-flag-linked-only-the-following-flags-are-allowed-constant-shared/1373/4


mstimberg (approved):
Thanks!

PR #1584, opened 2025-01-30
Extend linked variables functionality

mstimberg:
Previously, "linked variables" could not be used with `Synapses`, since the dynamic number of synapses makes a static mapping fragile. With this PR, `Synapses` gains the possibility to link to scalar variables and to variables in groups of size 1, where there is no ambiguity. In addition, all groups (including `Synapses`) can link to other variables using a dedicated index variable (which could even change its values over time, which might be useful for some corner cases). Since this variable would grow together with the number of synapses, it will always be of the right size. However, note that using an index variable means that simulations can crash due to out-of-bounds access. Actually, the `linked_var` documentation already stated:

> index : str or [ndarray](https://numpy.org/doc/stable/reference/generated/numpy.ndarray.html#numpy.ndarray), optional
>
>    An indexing array (or the name of a state variable), providing a mapping from the entries in the link source to the link target.

But before this PR, providing the "name of a state variable"  raised an error... I'm not 100% sure whether this never worked or whether this feature disappeared by accident with some other change :shrug:

### DIFF
--- a/brian2/core/variables.py
+++ b/brian2/core/variables.py
@@ -767,6 +767,12 @@ class LinkedVariable:
     """
 
     def __init__(self, group, name, variable, index=None):
+        if isinstance(variable, DynamicArrayVariable):
+            raise NotImplementedError(
+                f"Linking to variable {variable.name} is "
+                "not supported, can only link to "
+                "state variables of fixed size."
+            )
         self.group = group
         self.name = name
         self.variable = variable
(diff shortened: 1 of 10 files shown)


thesamovar:
Do you need any input from me on this? Am a bit slammed at the moment but maybe if there are specific points you'd like feedback on I can prioritise?

mstimberg:
> Do you need any input from me on this? Am a bit slammed at the moment but maybe if there are specific points you'd like feedback on I can prioritise?

Of course always happy to have input, but don't need it particularly here – it's a bit of a niche feature and also a straightforward extension of we already have (i.e. no new syntax).

PR #1602, opened 2025-03-24
Fix issue run at

Samuele-DeCristofaro:


mstimberg:
Apologies, I did not yet have time to have a look at the PR, but I will do so tomorrow !

brian2/core/clocks.py
```diff
@@ -61,8 +61,98 @@ def check_dt(new_dt, old_dt, target_t):
...
+            self.variables["t"].set_value(self.times[start_idx])
+            self._i_end = end_idx - 1 
+        else:
+
+            pass
+            
+    def __lt__(self, other):
+        return self.variables["t"].get_value().item() < other.variables["t"].get_value().item()
+
+    def __eq__(self, other):
```
mstimberg, line 130: While I get the appeal of using `__eq__` here, I don't think it is the right approach – also, it breaks everything :) The reason is that we don't want to consider two `Clock`s to be equal just because their time is the same. E.g. there are places where we check whether two objects use the same clock, and those would always return `True` at the start of a simulation when those clocks are at 0ms. Also, we have places where we put clocks into a `set`, and Python objects that have `__eq__`  but not `__hash__` are not hashable (and the hash of an object is not allowed to change and needs to be consistent with `__eq__` so we cannot simply add a `__hash__` function. I think the best approach would be to add a specific function to compare them, say `same_time(self, other)`

brian2/core/network.py
```diff
@@ -1035,21 +1035,17 @@ def after_run(self):
...
+        
         curclocks = {
             clock
-            for clock, time, dt in clocks_times_dt
-            if (
-                time == min_time
-                or abs(time - min_time) / min(minclock_dt, dt) < Clock.epsilon_dt
-            )
+            for clock in self._clocks
+            if clock == minclock
```
mstimberg, line 1044: See my comment about `__eq__` above

brian2/core/network.py
```diff
@@ -1035,21 +1035,17 @@ def after_run(self):
                 obj.after_run()
 
     def _nextclocks(self):
-        clocks_times_dt = [
-            (c, self._clock_variables[c][1][0], self._clock_variables[c][2][0])
-            for c in self._clocks
-        ]
-        minclock, min_time, minclock_dt = min(clocks_times_dt, key=lambda k: k[1])
+
+        minclock = min(self._clocks)
```
mstimberg, line 1039: While this (i.e. defining `__lt__` and `__le__`) is less of an issue than the equality check, for consistency/readability we should probably also have an explicit function here, or simply use something along the lines of `min(self._clocks, key=lambda c: c.t)`

brian2/core/network.py
```diff
@@ -1035,21 +1035,17 @@ def after_run(self):
...
-                time == min_time
-                or abs(time - min_time) / min(minclock_dt, dt) < Clock.epsilon_dt
-            )
+            for clock in self._clocks
+            if clock == minclock
         }
+        
         return minclock, curclocks
-
+
```
mstimberg, line 1048: Nothing to worry about before the PR is finalized, but have a look at our documentation on how to enable `pre-commit` so that it will make sure that the source code is formatted consistently: https://brian2.readthedocs.io/en/stable/developer/guidelines/style.html#code-style

brian2/core/clocks.py
mstimberg, line 1: `RegularClock` and `EventClock` should be added to `__all__` as well.

mstimberg (changes requested):
This looks really good and very close to what I had in mind :blush: Unfortunately, it also breaks everything :sweat_smile:

As explained in the comments, I get the appeal of the `__eq__` approach, but I don't think it is the right solution here.

Samuele-DeCristofaro:
Thanks a lot for taking the time to guide me through my mistakes, I must have overlooked the mention about the __eq__, I'll get to work and get back to you with news soon :).

Samuele-DeCristofaro:
I made some changes

Samuele-DeCristofaro:
Hi @mstimberg, I've made all the changes I think are needed, but I'm not too sure if they are correct, looking for your feedback.

mstimberg:
Hi @Samuele-DeCristofaro. I did not go through your updated code in detail yet, but I'd suggest that you'd try running a simple example that uses your branch. If I do this for the [CUBA example](https://brian2.readthedocs.io/en/stable/examples/CUBA.html), I get an error which strongly suggests that there is still something to fix :blush: For more thorough testing, please look at the documentation on how to run the test suite: https://brian2.readthedocs.io/en/stable/developer/guidelines/testing.html#running-the-test-suite

Here's the full error I am getting:
```pytb
  File "/home/mstimberg/git/brian2/examples/CUBA.py", line 17, in <module>
    from brian2 import *
  File "/home/mstimberg/git/brian2/brian2/__init__.py", line 91, in <module>
    from brian2.only import *
  File "/home/mstimberg/git/brian2/brian2/only.py", line 59, in <module>
    set_device(all_devices["runtime"])
  File "/home/mstimberg/git/brian2/brian2/devices/device.py", line 684, in set_device
    _do_set_device(device, build_on_run, **kwargs)
  File "/home/mstimberg/git/brian2/brian2/devices/device.py", line 698, in _do_set_device
    active_device.activate(build_on_run=build_on_run, **kwargs)
  File "/home/mstimberg/git/brian2/brian2/devices/device.py", line 378, in activate
    self.defaultclock = Clock(dt=0.1 * ms, name="defaultclock")
                        ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/mstimberg/git/brian2/brian2/core/clocks.py", line 300, in __init__
    super().__init__(dt, name)
  File "/home/mstimberg/git/brian2/brian2/core/clocks.py", line 186, in __init__
    super().__init__(times, name=name)
  File "/home/mstimberg/git/brian2/brian2/core/clocks.py", line 76, in __init__
    self.times = sorted(times)
                 ^^^^^^^^^^^^^
  File "/home/mstimberg/git/brian2/brian2/core/clocks.py", line 70, in __getitem__
    return self.clock.dt * timestep
           ^^^^^^^^^^^^^
  File "/home/mstimberg/git/brian2/brian2/core/clocks.py", line 257, in <lambda>
    fget=lambda self: Quantity(self.dt_, dim=second.dim),
                               ^^^^^^^^
  File "/home/mstimberg/git/brian2/brian2/core/clocks.py", line 245, in _get_dt_
    return self.variables["dt"].get_value().item()
           ~~~~~~~~~~~~~~^^^^^^
  File "/home/mstimberg/git/brian2/brian2/core/variables.py", line 1646, in __getitem__
    return self._variables[item]
           ~~~~~~~~~~~~~~~^^^^^^
KeyError: 'dt'
```

Samuele-DeCristofaro:
Hi @mstimberg thanks for the suggestion, I'll get back to you with some more useful feedback as soon as possile.

Samuele-DeCristofaro:
Hi @mstimberg , I was working on this issue but it seemed like I just miss typed self.dt=_dt meanwhile in the ClockArray class was trying to use self.clock.dt, but then another error came up, which is basically about using self.dt=float(dt) which is fine when i call self.variables.addarray("dt") because it expects a float but when the function '_set_dt' is called also  expectes a quantity with unit second for argument 'dt' but got '0.0001' (unit is 1). Tomorrow I will try to get things to work also because I don't think it's hard to fix it just felt reasonable to update you.

Samuele-DeCristofaro:
I feel like the solution involves using the original dt, but it's a bit tricky.

Samuele-DeCristofaro:
I also think that if we wanto to keep the _dt we should also add something to reset it aswell if needed.

Samuele-DeCristofaro:
Hi @mstimberg  thanks, now it's all more clear,  I really did not understand where the issue was 😅, I was also wondering if something like this would be useful aswell:
`    @check_units(dt_=1)
    def _set_dt_(self, dt_):
        self._old_dt = self._get_dt_()
        self.variables["dt"].set_value(dt_)
        self._dt=dt_
        `

mstimberg:
Hi @Samuele-DeCristofaro This seems to be because the time calculation in `ClockArray` uses `Clock._dt` which does not get stored/restored by the usual mechanism. Is it really necessary to have it, i.e. couldn't you use `Clock.dt` instead there? Maybe, if the reason is in the `EventClock` → `Clock` inheritance, going back to the previously discueed design of having an abstract `BaseClock` (with the two types of clocks inheriting from it) would be better?

Samuele-DeCristofaro:
Hi @mstimberg  before changing the approach I tried to use the initial_dt only to set the dt the first time:
```
class ClockArray:
    def __init__(self, clock):
        self.clock = clock
        

    def __getitem__(self, timestep):
        return self.clock.dt * timestep


class EventClock(VariableOwner):
    def __init__(self, times, name="eventclock*"):
        Nameable.__init__(self, name=name)
        self.variables = Variables(self)
        self.variables.add_array(
            "timestep", size=1, dtype=np.int64, read_only=True, scalar=True
        )
        self.variables.add_array(
            "t",
            dimensions=second.dim,
            size=1,
            dtype=np.float64,
            read_only=True,
            scalar=True,
        )
        self.variables["timestep"].set_value(0)
        if isinstance(times, ClockArray):
            self.times = times  # Don't sort, don't check for duplicates
            self.variables.add_array(
                "dt",
                dimensions=second.dim,
                size=1,
                values=times.clock.initial_dt,
                dtype=np.float64,
                read_only=True,
                constant=True,
                scalar=True,
            )
        else:
            self.times = sorted(times)
            if len(self.times) != len(set(self.times)):
                raise ValueError(
                    "The times provided to EventClock must not contain duplicates"
                )
        self.variables["t"].set_value(self.times[0])
```


, when it was time to set up the t, and now we can simply always use dt even if it's not the cleanest way, in the test_network, it didn't give any errors but it took like 30 s to simulate, then I started making some adjustments in the network file and it's getting faster, now it takes 25 s by simply changing little things. Do you still think that maybe this is not the best way ? I can try to optimize things even more but I don't know how much faster it can get.

Samuele-DeCristofaro:
I also forgot to mention that with this new clock implementation, I tried to use the old network file and the test_network worked perfectly without any changes or error in the expected output. Which means that at least I did not broke anything in the clocks.

mstimberg:
Hi @Samuele-DeCristofaro I feel that this is not the right way: if the `EventClock`'s `__init__` file needs to have a special case for how it is called from the subclass (with a `ClockArray` for `times`), then there is something wrong in the overall inheritance structure. If `Clock` (i.e. `RegularClock`) and `EventClock` both are independent (inheriting from the same base class), then the code will look much simpler and easier to understand, we won't even need `ClockArray` anymore, since we can directly implement the `t` access differently in both cases.

Samuele-DeCristofaro:
Hi @mstimberg, alright I will then apply these changes requested, but what about the network file is there something to avoid using that could lead to some performance issues as I said before.

mstimberg:
Please update the PR first. I did not quite understand the performance issue, and it is much easier for me to simply try out the code and see for myself. Thanks!

Samuele-DeCristofaro:
PR updated.

mstimberg:
Thanks for updating the PR. I tried it out, and the performance of some test cases is indeed much worse than before. A good/extreme example is the `test_network_different_clocks` test from `test_network.py`, because it does not really do anything during a run, but simulates one clock for 1 million, and another one for 100000 time steps. If you run this test with a profiler (e.g. Python's built-in cProfile, you can visualize the results with tools like *snakeviz*), you see that a lot of time is spent in `ClockArray.__getitem__`. The reason is that this accesses `self.clock.dt`, which creates a new `Quantity` object for each access to provide the values with physical units. Since the value is only used to set the `t` value via `set_value`, you do not actually need units, a faster access would therefore be to use `self.clock.dt_` or `self.clock.variables['dt'].get_value().item()`. A similar but more subtle issue occurs in the `same_time` function because it uses `hasattr(..., 'dt')`: this will actually *get* the value of `dt` (i.e., including units) to see whether it raises an error... A faster check would be to use `isinstance`, but by implementing this function separately for `EventClock` and `Clock`, you'd also have one check less (since you know for the class itself whether it has a `dt` or not). I'd make things "reasonably" fast for now by fixing these big issues, but it is not necessary to make as fast as it was previously. This would require keeping direct references to the arrays storing `t`, `dt`, and `timestep` around instead of going through `variables[...],get_value()` every time, but this approach is also a bit more error-prone.

Samuele-DeCristofaro:
PR updated again, please let me know if more changes are needed.

Samuele-DeCristofaro:
PR updated ready to be checked, I also added some simple checks to test_clocks.py for the new event_clock, let me know if some checks could be added up or if it's not needed at all. As you expected this new implementation is slightly slower then the initial one but I don't think there are any big issues now.

Samuele-DeCristofaro:
Also if we manage to merge this PR, I would be open to help more with everything if needed.

mstimberg:
Hi @Samuele-DeCristofaro. I think this looks good so far, we'll keep the final performance tweaking for the end. The new test looks good in general, but I'd tweak the values a bit to have irregularly spaced events (which is kind of the point of `EventClock` ;-) ), and also add a test that combines several clocks in a run.
Finally, we need to add the `run_at` method! This should be almost trivial now, we can have one general `run_code` function (not sure what the best name would be…) that is basically identical to the current `run_regularly` function, but does not allow the `dt` argument, and a `run_at` function that has a `times` argument instead of `dt` and `clock`, and calls `run_code` with `clock=EventClock(times)`.
It would be more consistent if `run_regularly` would only have a `dt` argument and no `clock` (in the same way that `run_at` only has `times`), but that might break someone's code…

mstimberg:
Hi @Samuele-DeCristofaro. Sorry, it took me a while to get back to this PR. Regarding the `Quantity` support ­– I realized that the `EventClock` (and the `times` argument of `run_at`) should *only* support a `Quantity` with the correct units to be consistent with how we handle units elsewhere. Therefore, there should be a `times = Quantity(times)` (to support lists of quantities) and a `fail_for_dimension_mismatch(...)` somewhere at the top.

Regarding the tests: I think your tests cover the essentials, but instead of the `network_operation`, I'd use the `NameLister` that is used elsewhere in `test_network`. E.g. if you run a network with
```python
 x = NameLister(name="x", dt=0.1 * ms, order=0)
 y = NameLister(name="y", clock=EventClock(times=[0*ms, 0.5*ms, 1.5*ms]), order=1)
```
It should give `"xyxxxxxyxxxxxxxxxxyxxxx" for a 2 second run.

brian2/groups/group.py
```diff
@@ -1185,6 +1185,51 @@ def run_regularly(
         # contained objects would no be enough, since the CodeObject needs a
         # reference to the group providing the context for the operation, i.e.
         # the subgroup instead of the parent group. See github issue #922
+
+        return self.run_custom_clock(code, clock, when, order, name, codeobj_class)
+
+    def run_custom_clock(
```
mstimberg, line 1191: I find this name a bit confusing, how about `run_on_clock`?

Samuele-DeCristofaro:
Hi @mstimberg , PR updated, let me know if something else should be fixed, the only issue for me is that sometimes I get this error when I run the new test:
  File "C:\Users\ssamu\OneDrive\Desktop\github-brian2\brian2\brian2\tests\test_clocks.py", line 111, in test_combined_clocks_with_run_at
    x_count == expected_x_count
AssertionError: Expected 5 x's, got 4

Samuele-DeCristofaro:
HI @mstimberg , I tried your tip but the test it's still not always working 😭 .

mstimberg:
Hi @Samuele-DeCristofaro The problem was that `i_end` was off by one, i.e. the added `inf` was basically ignored. I fixed this and added another test, things seem to work for now. I'll have a closer look tomorrow!

Samuele-DeCristofaro:
Got it! Thanks for the help! 😃 I got confused because sometimes the test worked, so I thought there was something wrong with the run method.

brian2/core/clocks.py
```diff
@@ -99,64 +89,252 @@ def __init__(self, dt, name="clock*"):
...
+        for time in self._times:
+            if float(time) in seen:
+                duplicates.append(time)
+            else:
+                seen.add(float(time))
+        if duplicates:
+            raise ValueError(
+                "The times provided to EventClock must not contain duplicates. "
+                f"Duplicates found: {duplicates}"
+            )
```
mstimberg, line 189: The error message will be slightly less useful, but I think I'd prefer a simple
```python
if len(self._times) != len(set(self._times)):
  raise ValueErrror(...)
````

File: brian2/core/clocks.py [hunk 7]
```diff
@@ -99,64 +89,252 @@ def __init__(self, dt, name="clock*"):
...
+    Parameters
+    ----------
+    dt : float
+        The time step of the simulation as a float
+    name : str, optional
+        An explicit name, if not specified gives an automatically generated name
+
+    Notes
+    -----
+    Clocks are run in the same Network.run iteration if ~Clock.t is the
```
mstimberg, line 292: Missing backticks
```suggestion
    Clocks are run in the same `Network.run` iteration if `~Clock.t` is the
```

brian2/tests/test_clocks.py
```diff
@@ -57,10 +59,69 @@ def test_set_interval_warning():
...
+    # Create and run the network
+    net = Network(regular_lister, event_lister)
+    net.run(5 * ms)
+
+    # Get update string
+    updates = "".join(NameLister.updates)
+
+    # Expected output: "x" at 0,1,2,3,4ms = 5 times
+    # "y" at 0.5, 2.5, 4.0ms = 3 times
+    # We don't care about exact timing here, just the sequence
```
mstimberg, line 101: I think we *do* care about the exact timing :smile: I'd directly test for the expected `xyxxyxxy` value

mstimberg (changes requested):
Looks good, we are getting there! I've left several nit-picking comments, but those are mostly minor issues. The main remaining task is to get C++ standalone support in place (but I can take care of that if you prefer), and to mention the new features in the documentation.

Samuele-DeCristofaro:
I'm very happy we are getting there !, I think I can work on the cpp standalone part as well, and I'll fix these minor issues ASAP, I have no clue about what happened to those quotes 🥲.

brian2/core/clocks.py
Samuele-DeCristofaro: I tried your solution but set uses __eq__ methods to compare the times so it returns an error with Quantity types.

brian2/core/clocks.py
```diff
@@ -99,64 +89,252 @@ def __init__(self, dt, name="clock*"):
...
+
+        logger.diagnostic(f"Created event clock {self.name}")
+
+    def advance(self):
+        """
+        Advance to the next time in the sequence.
+        """
+        new_ts = self.variables["timestep"].get_value().item()
+        if self._i_end is not None and new_ts + 1 > self._i_end:
+            return
```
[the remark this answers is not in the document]
Samuele-DeCristofaro: Yes and I also fixed the test_event_clock accordingly

brian2/core/clocks.py
```diff
@@ -36,7 +37,7 @@ def check_dt(new_dt, old_dt, target_t):
     ------
     ValueError
         If using the new dt value would lead to a difference in the target
-        time of more than `Clock.epsilon_dt` times ``new_dt`` (by default,
+        time of more than Clock.epsilon_dt times `new_dt (by default,
```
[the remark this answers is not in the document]
Samuele-DeCristofaro: fixed

brian2/core/clocks.py
```diff
@@ -62,31 +63,20 @@ def check_dt(new_dt, old_dt, target_t):
         )
 
 
-class Clock(VariableOwner):
+class BaseClock(VariableOwner):
```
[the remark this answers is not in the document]
Samuele-DeCristofaro: fixed

brian2/core/clocks.py
```diff
@@ -99,64 +89,252 @@ def __init__(self, dt, name="clock*"):
...
 
         Returns
         -------
+        bool
+            True if both clocks are at the same time
+        """
+        t1 = self.variables["t"].get_value().item()
+        t2 = other.variables["t"].get_value().item()
+
+        return abs(t1 - t2) < self.epsilon
```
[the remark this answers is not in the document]
Samuele-DeCristofaro: fixed

brian2/core/clocks.py
```diff
@@ -99,64 +89,252 @@ def __init__(self, dt, name="clock*"):
...
+        times = Quantity(times)
+        from brian2.units.fundamentalunits import fail_for_dimension_mismatch
+
+        fail_for_dimension_mismatch(
+            times,
+            second.dim,
+            error_message="'times' must have dimensions of time, got %(dim)s",
+            dim=times,
+        )
+        self._times = sorted(times)
```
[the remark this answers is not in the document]
Samuele-DeCristofaro: I created the times array, does it make sense to use a numpy array instead of a list there?

brian2/core/clocks.py
```diff
@@ -99,64 +89,252 @@ def __init__(self, dt, name="clock*"):
...
+
+    def __init__(self, times, name="eventclock*"):
+        super().__init__(name=name)
+        times = Quantity(times)
+        from brian2.units.fundamentalunits import fail_for_dimension_mismatch
+
+        fail_for_dimension_mismatch(
+            times,
+            second.dim,
+            error_message="'times' must have dimensions of time, got %(dim)s",
```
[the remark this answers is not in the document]
Samuele-DeCristofaro: fixed

brian2/core/clocks.py
```diff
@@ -99,64 +89,252 @@ def __init__(self, dt, name="clock*"):
...
-        new_t = new_i * self.dt_
-        if new_t == target_t or np.abs(new_t - target_t) / self.dt_ <= Clock.epsilon_dt:
-            new_timestep = new_i
+        return self._times[timestep]
+
+    def same_time(self, other):
+        """
+        Check if two clocks are at the same time.
+
+        For comparisons with Clock objects, uses the Clock's dt and epsilon_dt.
```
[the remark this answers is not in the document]
Samuele-DeCristofaro: fixed

brian2/tests/test_clocks.py
Samuele-DeCristofaro: Yep my bad this was a reminding for me going through the investigation about that error.

brian2/core/clocks.py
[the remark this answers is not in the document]
mstimberg: When you create the Brian variable `times` below, it will already be stored internally as a numpy array. To not duplicate things, I'd not store `self._times` at all, e.g. you could do `times = sorted(times)` and `times.append(np.inf*ms)` before passing `values=times` to the `add_array` function.

mstimberg (changes requested):
Thanks for the updates, I made a few more minor comments. But as I said earlier, the big thing to do now would be the C++ standalone support – let me know if you need further guidance!

De-Cri:
I'm sorry if I'm not updating this pr with the rest of the work but I gotta take those exams!  😢 , I will get back to this in a couple weeks.

mstimberg:
Hi @De-Cri. Thanks for the small updates you did lately, they look good to me. Do you need guidance to work on the standalone support, and do you think you will be able to work on this in the next weeks? I'd otherwise might take over the standalone part to push this PR further along.

De-Cri:
Hi @mstimberg, I understand that it's taking quite some time, and I think it's better if you could take over the rest . Are we going to close this PR?

mstimberg:
> Hi @mstimberg, I understand that it's taking quite some time, and I think it's better if you could take over the rest . Are we going to close this PR?

Understood. I've done the basic changes to support C++ standalone mode, and it seems to work! I'll do some more testing and documentation soon and then merge it. This is a really nice feature, many thanks for your work on it!

De-Cri:
It has been really fun and I've learnt many things, so it was a pleasure! I will for sure contribute more as soon as I have more time.

mstimberg (approved):
:shipit:

PR #1603, opened 2025-03-24
Added before_run codeblock in numpy and C++ templates

mahipalimkar:
Moved the Python code in` SpikeGeneratorGroup.before_run` into the `before_code` blocks of the numpy and C++ templates.

### DIFF
--- a/brian2/codegen/runtime/numpy_rt/templates/spikegenerator.py_
+++ b/brian2/codegen/runtime/numpy_rt/templates/spikegenerator.py_
@@ -1,5 +1,32 @@
-{# USES_VARIABLES { _spikespace, neuron_index, _timebins, _period_bins, _lastindex, t_in_timesteps } #}
-{% extends 'common_group.py_' %}
+{% extends 'common_group.py.jinja2' %}
+
+{% block before_code %}
+# Copy of the SpikeGeneratorGroup.before_run code
+dt = {{dt.item()}}  # Always copy dt
+period = {{period_}}  # Always copy period
+
+# Always recalculate timesteps
+current_t = {{t}}[0]
+timesteps = ({{_spike_time}} / dt).astype(np.int32)
+current_step = int(current_t / dt)
+
+# Always update _lastindex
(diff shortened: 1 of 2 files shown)


mstimberg:
Apologies, I did not yet have time to have a look at the PR, but I will do so tomorrow !

brian2/codegen/runtime/numpy_rt/templates/spikegenerator.py_
```diff
@@ -1,25 +1,48 @@
...
+# Always recalculate timesteps
+from brian2 import defaultclock  # Use Brian 2's clock instead of 't'
+current_t = defaultclock.t
+timesteps = ({{_spike_time}} / dt).astype(np.int32)
+current_step = int(current_t / dt)
+
+# Always update _lastindex
+in_the_past = np.nonzero(timesteps < current_step)[0]
+if len(in_the_past):
+    {{_lastindex}}[0] = in_the_past[-1] + 1
```
mstimberg, line 17: For scalar variables, Brian already applies the `[0]` automatically.
```suggestion
    {{_lastindex}} = in_the_past[-1] + 1
```

mstimberg (changes requested):
This is a good start! I only looked at the Python code so far, and there are a number of issues that prevent the code from running – I cannot yet judge whether it will actually work after fixing these issues, but let's take it one step at a time :blush:

mstimberg:
Working with this mix of Python code, Jinja templates, Brian's idiosyncratic variable system, etc. is really far from straightforward – I wouldn't expect anyone to get this right on the first try! For testing whether things work, try running any example that uses `SpikeGeneratorGroup` and add `prefs.codegen.target = 'numpy'` so that it uses the Python target instead of the default Cython.

mahipalimkar:
Sorry for the gap in communication @mstimberg ! I have started incorporating all the changes that you suggested. Will do this at the earliest.

mstimberg:
Hi @mahipalimkar, do you think you will have time to pick up the work on this PR any time in the upcoming weeks?

PR #1609, opened 2025-04-02
[Fix #1608] : Raises a ValueError for negative duration in run() function

### LINKED ISSUE #1608: Raise an error for negative duration in `run()` function
Currently, calling `run(duration)` with a negative duration does not raise an error; instead, it simply runs for 0 seconds.

Proposed Fix: 

- Modify `run()` to raise an appropriate error (e.g., ValueError) when a negative duration is passed.
- The case of `run(0*ms)` should remain unchanged, as it is valid and useful in tests.

PalashChitnavis:
Fixes #1608

Added a condition to check if the duration is negative and raise a ValueError accordingly.

### DIFF
--- a/brian2/core/network.py
+++ b/brian2/core/network.py
@@ -1102,7 +1102,16 @@ class Network(Nameable):
         -----
         The simulation can be stopped by calling `Network.stop` or the
         global `stop` function.
+
+        Raises
+        ------
+        ValueError
+            Error raised when duration passed in run function is negative.
         """
+        if duration < 0:
+            raise ValueError(
+                f"Function 'run' expected a non-negative duration but got '{duration}'"
+            )
         # This will trigger warnings for objects that have not been included in a network
         gc.collect()
         device = get_device()  # Do not use the ProxyDevice -- slightly faster
--- a/brian2/devices/cpp_standalone/device.py
+++ b/brian2/devices/cpp_standalone/device.py
@@ -1770,6 +1770,11 @@ class CPPStandaloneDevice(Device):
         level=0,
         **kwds,
     ):
+        if duration < 0:
+            raise ValueError(
+                f"Function 'run' expected a non-negative duration but got '{duration}'"
+            )
+
         self.networks.add(net)
         if kwds:
             logger.warn(
--- a/brian2/tests/test_cpp_standalone.py
+++ b/brian2/tests/test_cpp_standalone.py
@@ -1048,13 +1048,15 @@ def test_header_file_inclusion():
         run(defaultclock.dt)
     assert_allclose(S.rate_copy[:], np.arange(len(G)) * 42 * Hz)
 
+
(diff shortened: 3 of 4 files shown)


mstimberg (changes requested):
Hi @PalashChitnavis. This looks very good, but could you please move it into `Network.run` (which will be called by the `magic_network.run` line below your check)? There are two ways to run a network: `run(...)` and `net = Network(...); net.run(...)` (see https://brian2.readthedocs.io/en/stable/user/running.html#networks) – both are affected by this issue in the same way, and since both go through `Network.run` it would be best to fix it there.
Could you also please add test for this check in `brian2/tests/test_network.py`? Please mark it as `@pytest.mark.codegen_independent`, since it does not have to be run for each code generation target.

Actually, now that I am thinking about it: you might have to add the check to the `network_run` method of `brian2/devices/cpp_standalone/device.py` as well. It is not super obvious, but it uses Brian's builtin "device_override" mechanism, which means that `Network.run` will instead call `CPPStandaloneDevice.network_run` if the standalone device is activated.

PalashChitnavis:
@mstimberg , I have shifted the condition from `magic.py` to `network.py`. Also I have added 3 tests (one for magic run, one for explicit network run and one for cpp_standalone run)

mstimberg (changes requested):
Thanks again, this is what I had in mind. Just a small request with respect to the tests:
1. Could you please move the test for C++ standalone to the `test_cpp_standalone.py` file (and mark it as `cpp_standalone` and `standalone_only` as the other tests in there) – this will make sure that it only gets executed when someone asks to test the standalone device
2. There's nothing wrong with testing the exception message, but I'd rather prefer removing it to avoid the tests breaking for some small change in the wording or formatting.

PalashChitnavis:
@mstimberg 

> There's nothing wrong with testing the exception message, but I'd rather prefer removing it to avoid the tests breaking for some small change in the wording or formatting.

Should I completely remove the assert statement and just check if it raises a ValueError?

Something like this: 
```
@pytest.mark.codegen_independent
def test_negative_duration_in_run():
    G = NeuronGroup(1, "v:1")
    with pytest.raises(ValueError) as exc_info:
        run(-1 * second)
```

or a simple assert message: 
```
@pytest.mark.codegen_independent
def test_negative_duration_in_run():
    G = NeuronGroup(1, "v:1")
    with pytest.raises(ValueError) as exc_info:
        run(1 * second)
    assert "expected a non-negative duration" in str(exc_info.value), "Negative duration check failed in magic network run()"

```

PalashChitnavis:
> Either would work for me, I have a slight preference for the simpler version (i.e. without assert).

@mstimberg done 👍

mstimberg (approved):
Looks great to me – I'll wait for the tests to run through and then merge.

Issue #868, opened 2017-08-19
Refractoriness for custom events


thesamovar:
We should probably allow for this possibility. Syntax should be straightforward, allow for a dict instead of string/quantity.

mstimberg:
Agreed, makes sense.

maverick4code:
@thesamovar @mstimberg I think we can tweak `refractory` parameter in `NeuronGroup` to accept a dictionary. This would allow us to specify refractory period for `custom events` and default `spike events`.

Here is a simple example([reference](https://brian2.readthedocs.io/en/stable/advanced/custom_events.html)):
' ' ' Python
from brian2 import *
"""Creates a NeuronGroup with a spike event (v > 1, 5ms) and a custom event (x > 1, x < 0.5)."""
eqs = '''
dv/dt = -v / (10*ms) : 1
dx/dt = -x / (20*ms) : 1
'''
G = NeuronGroup(1, eqs, threshold='v > 1', reset='v = 0', events={'custom_event': 'x > 1'}, 
                refractory={'spike': 5*ms, 'custom_event': 'x < 0.5'})

The old code with just `5*ms` would still work for `spike`.

If any additional thoughts, please let me know.

maverick4code:
@thesamovar @mstimberg,  I have reviewed the current implementation in [brian2/groups/neurongroup.py](https://github.com/brian-team/brian2/blob/master/brian2/groups/neurongroup.py) and tried to get a solution. 

1)  `NeuronGroup`
I have updated the `__init__` method in [brian2/groups/neurongroup.py](https://github.com/brian-team/brian2/blob/master/brian2/groups/neurongroup.py) to support the `refractory` parameter.

```python
class NeuronGroup(Group):
    def __init__(self, N, model, threshold=None, reset=None, refractory=None, events=None, ):
        
        if events is None:
            events = {}
        self.events = {'spike': threshold} if threshold else {}
        self.events.update(events)

        if refractory is not None:
            if isinstance(refractory, (str, Quantity)):
                refractory = {'spike': refractory}
            elif isinstance(refractory, dict):
              
                for event in refractory:
                    if event not in self.events:
                        raise ValueError(f"Unknown event '{event}' in refractory dictionary.")
            else:
                raise TypeError("refractory must be a string, Quantity, or dictionary")
        else:
            refractory = {}
        self._refractory = refractory

        # calling parent class, setting up variables
       . . .



2) `Variables setup`

```python
def _create_variables(self):
[...]

PR #1617, opened 2025-04-07
Update neurongroup.py

maverick4code:
Hi, here’s the draft PR for review.

### DIFF
--- a/brian2/groups/neurongroup.py
+++ b/brian2/groups/neurongroup.py
@@ -518,7 +518,7 @@ class NeuronGroup(Group, SpikeSource):
         method_options=None,
         threshold=None,
         reset=None,
-        refractory=False,
+        refractory=None, # Updated to None instead of False
         events=None,
         namespace=None,
         dtype=None,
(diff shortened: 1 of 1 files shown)


maverick4code:
@mstimberg Are the changes correct ?

Samuele-DeCristofaro:
Hi @maverick4code, in the meanwhile you wait for the pr to be checked you might want to check out the black formatting tool, since I see here it's giving you an error, which can help before you commit something 😄 .

mstimberg:
Hi @maverick4code. From what I can see, the code still adds a `not_refractory_spike` variable in addition to the `not_refractory` variable that is kept for background compatibility. I would also not have the `unless refractory` flag be responsible to all event types, but rather only to the `spike` event. We might extend the syntax later. Having said all this, I cannot run your current branch, since there are indentation errors in `neurongroup.py` – these need to be fixed first.

PS: This is a minor issue since I know what this PR is about, but the current title and description is not describing what the PR is aiming for (which would be "Refractoriness for custom events"), but rather describes the latest changes *to* this PR. Also, it links to itself instead of to #868.

mstimberg:
Hi @maverick4code. Thanks for your work on this so far, just checking: are you still interested/do you have the time to work on this?

maverick4code:
Hi @mstimberg, yes I’m definitely still interested! I will make the suggested fixes and updates soon.

mstimberg:
Happy to hear that!

PR #1643, opened 2025-06-12
Eliminate Python Layer Calls in `Synapses` Codegen Templates for SpikeQueue Operations

### LINKED ISSUE #1642: Performance Bottleneck in `SpikeQueue` : Python Method Call in Critical Path
---

### Description

Brian2’s current handling of spike propagation introduces significant **performance overhead** by invoking **Python methods** inside the **tight simulation loop**. This is caused by the `synapses_push_spikes.pyx` template calling `_owner.push_spikes()`—a Python method that indirectly triggers C++ code.

---

### Root Cause

Brian2 uses a **template-based code generation system** (Jinja2) to emit simulation code. The spike propagation step currently looks like this:

```jinja2
{% block maincode %}
_owner.push_spikes()
{% endblock %}
```

This call chain for **every timestep**, **every pathway**, results in:

```
Cython → Python (_owner.push_spikes) → Python (queue.push) → C++ (actual push)
```

Each transition introduces:

* Python interpreter overhead
* Object lookups & dynamic dispatch
* Type checking & conversions

This significantly slows down simulation for large networks and also the unnecessary large codegen of generated cpp files ...

---

Legend101Zz:
##  Description

This PR removes indirect Python method calls for interacting with the spike queue during synaptic spike propagation and replaces them with **direct C++ access** via Cython, improving runtime performance.

### Problem

Previously, the `synapses_push_spikes.pyx` template (and related code objects) relied on Python methods like:

* `_owner.push_spikes()`
* `_owner._queue.peek()`
* `_owner._queue.advance()`

This created an inefficient call chain:

```
Cython → Python → C++ (via method call)
```

---

### This PR Introduces

A **C++-only access path** through PyCapsules and minimal C++ interfaces:

| Operation       | Old (Python method)    | New (Direct C++ via Cython) |
| --------------- | ---------------------- | --------------------------- |
| Push spikes     | `_owner.push_spikes()` | `cpp_queue->push(...)`      |
| Peek from queue | `_queue.peek()`        | `cpp_queue->peek()`         |
| Advance queue   | `_queue.advance()`     | `cpp_queue->advance()`      |

* The `CSpikeQueue*` pointer is passed via a `PyCapsule` into the Cython template.
* Only minimal C++ methods (`push`, `peek`, `advance`) are declared and used.
* NumPy or Python overhead is completely bypassed in the inner simulation loop.

Closes #1642

### DIFF
--- a/brian2/synapses/cythonspikequeue.pyx
+++ b/brian2/synapses/cythonspikequeue.pyx
@@ -54,6 +54,20 @@ cdef class SpikeQueue:
     def _full_state(self):
         return self.thisptr._full_state()
 
+    def getptr(self):
+        """
+        Returns the raw memory address of the underlying C++ CSpikeQueue object.
+
+        This is used to pass the queue directly into Cython templates for fast,
+        low-overhead access—bypassing Python calls entirely.
+
+        1. self.thisptr is a C++ pointer (CSpikeQueue*) managed by Cython
+        2. <long> cast converts the pointer to a Python integer (memory address)
+        3. The returned address can be cast back to CSpikeQueue* in templates
+        4. Templates then call C++ methods directly without Python involvement
+        """
+        return <long>self.thisptr
+
     cdef object __weakref__  # Allows weak references to the SpikeQueue
 
     def _restore_from_full_state(self, state):
@@ -80,7 +94,7 @@ cdef class SpikeQueue:
                                  <int32_t*>sources.data,
                                  sources.shape[0], dt)
 
-    def push(self, np.ndarray[int32_t, ndim=1, mode='c'] spikes):
+    cpdef push(self, np.ndarray[int32_t, ndim=1, mode='c'] spikes):
         self.thisptr.push(<int32_t*>spikes.data, spikes.shape[0])
 
     def peek(self):
--- a/brian2/synapses/cythonspikequeue.pyx
+++ b/brian2/synapses/cythonspikequeue.pyx
@@ -5,7 +5,6 @@
 from libcpp.vector cimport vector
 from libcpp.pair cimport pair
 from libcpp.string cimport string
-from libc.stdint cimport uintptr_t
 import cython
 from cython.operator import dereference
 from cython.operator cimport dereference
--- a/brian2/synapses/cythonspikequeue.pyx
+++ b/brian2/synapses/cythonspikequeue.pyx
@@ -98,7 +98,7 @@ cdef class SpikeQueue:
                                  <int32_t*>sources.data,
                                  sources.shape[0], dt)
 
-    cpdef push(self, np.ndarray[int32_t, ndim=1, mode='c'] spikes):
+    def push(self, np.ndarray[int32_t, ndim=1, mode='c'] spikes):
         self.thisptr.push(<int32_t*>spikes.data, spikes.shape[0])
 
     def peek(self):
--- a/brian2/synapses/cythonspikequeue.pyx
+++ b/brian2/synapses/cythonspikequeue.pyx
@@ -5,12 +5,11 @@
 from libcpp.vector cimport vector
 from libcpp.pair cimport pair
 from libcpp.string cimport string
-
+from libc.stdint cimport uintptr_t
 import cython
 from cython.operator import dereference
 from cython.operator cimport dereference
-from libc.stdint cimport uintptr_t
-
+from cpython.pycapsule cimport PyCapsule_New, PyCapsule_GetPointer
 
 cimport numpy as np
 import numpy as np
@@ -69,6 +68,9 @@ cdef class SpikeQueue:
         4. Templates then call C++ methods directly without Python involvement
         """
         return <uintptr_t>self.thisptr
+
+    def get_capsule(self):
+        return PyCapsule_New(<void*>self.thisptr, "CSpikeQueue", NULL)
 
     cdef object __weakref__  # Allows weak references to the SpikeQueue
 
--- a/brian2/synapses/cythonspikequeue.pyx
+++ b/brian2/synapses/cythonspikequeue.pyx
@@ -9,7 +9,8 @@ from libcpp.string cimport string
 import cython
 from cython.operator import dereference
 from cython.operator cimport dereference
-from cpython.pycapsule cimport PyCapsule_New, PyCapsule_GetPointer
+from libc.stdint cimport uintptr_t
+
 
 cimport numpy as np
 import numpy as np
--- a/brian2/synapses/cythonspikequeue.pyx
+++ b/brian2/synapses/cythonspikequeue.pyx
@@ -8,7 +8,7 @@ from libcpp.string cimport string
 import cython
 from cython.operator import dereference
 from cython.operator cimport dereference
-from cpython.pycapsule cimport PyCapsule_New, PyCapsule_GetPointer
+from cpython.pycapsule cimport PyCapsule_New
 
 cimport numpy as np
 import numpy as np
--- a/brian2/synapses/cythonspikequeue.pyx
+++ b/brian2/synapses/cythonspikequeue.pyx
@@ -55,21 +55,13 @@ cdef class SpikeQueue:
     def _full_state(self):
         return self.thisptr._full_state()
 
-    def getptr(self):
-        """
-        Returns the raw memory address of the underlying C++ CSpikeQueue object.
-
-        This is used to pass the queue directly into Cython templates for fast,
-        low-overhead access—bypassing Python calls entirely.
-
-        1. self.thisptr is a C++ pointer (CSpikeQueue*) managed by Cython
-        2. <uintptr_t> cast converts the pointer to a platform-appropriate integer
-        3. The returned address can be cast back to CSpikeQueue* in templates
-        4. Templates then call C++ methods directly without Python involvement
-        """
-        return <uintptr_t>self.thisptr
-
     def get_capsule(self):
+        """
+        Returns a PyCapsule object wrapping the underlying C++ CSpikeQueue pointer.
+
+        PyCapsules are used to safely pass raw C/C++ pointers between Python modules
+        or C extensions without exposing the actual implementation details to Python.
+        """
         return PyCapsule_New(<void*>self.thisptr, "CSpikeQueue", NULL)
 
     cdef object __weakref__  # Allows weak references to the SpikeQueue
--- a/brian2/synapses/cythonspikequeue.pyx
+++ b/brian2/synapses/cythonspikequeue.pyx
@@ -9,6 +9,7 @@ from libcpp.string cimport string
 import cython
 from cython.operator import dereference
 from cython.operator cimport dereference
+from cpython.pycapsule cimport PyCapsule_New, PyCapsule_GetPointer
 
 cimport numpy as np
 import numpy as np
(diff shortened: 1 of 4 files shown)


Legend101Zz:
---
Hi @mstimberg,

Thanks for your patience! Apologies for the delay — I had less time this week than I’d hoped and ran into some complexity while working through the bottleneck fix. Whenever you have a chance, I’d really appreciate it if you could take a look. Thanks!

Looking forward to your feedback :)

brian2/synapses/cythonspikequeue.pyx
```diff
@@ -54,6 +54,20 @@ cdef class SpikeQueue:
...
+
+        This is used to pass the queue directly into Cython templates for fast,
+        low-overhead access—bypassing Python calls entirely.
+
+        1. self.thisptr is a C++ pointer (CSpikeQueue*) managed by Cython
+        2. <long> cast converts the pointer to a Python integer (memory address)
+        3. The returned address can be cast back to CSpikeQueue* in templates
+        4. Templates then call C++ methods directly without Python involvement
+        """
+        return <long>self.thisptr
```
Legend101Zz: ---

Hi @mstimberg, quick update on this —

I went with a very crude initial approach just to validate things: I simply converted the raw memory address to an `int`. It’s definitely not ideal, and I’m not a fan of this method — so I’m currently exploring using `PyCapsule` for a cleaner and safer solution.

That said, I did a quick benchmark. Running an *empty* template like this:

```python
{% extends 'common_group.pyx' %}

{% block before_code %}
    _owner.initialise_queue()
{% endblock %}

{% block maincode %}
# Do nothing
{% endblock %}
```

...generates **11,682 lines** of C++ code.

In comparison, the new version with the template logic produces **11,784 lines**, so only \~100 lines added — and the previous yellow highlighting is also gone:
Screenshot 2025-06-17 at 18 31 45

Let me know if this direction looks good to you. In the meantime, I’ll continue working on integrating `PyCapsule`.

---

mstimberg:
Hi, thanks, looks great, this is what I had in mind :+1: . I used the following code to test:
```python
from brian2 import *
prefs.codegen.runtime.cython.delete_source_files = False
clear_cache("cython")
G = NeuronGroup(10000, "v : 1", threshold="True")
S = Synapses(G, G, on_pre="v += 1")
S.connect()
run(100*ms, profile=True, report="text")
print(profiling_summary())
print("To make sure something actually happened:", G.v[0])
```
It sometimes crashed but when it ran it showed a dramatic speed improvement – however, it turned out that this improvement was mostly because it did not do anything :rofl: (this is why I added the print in the end…). I had to use
```
cdef void* _void_ptr = <void*>{{queue_ptr}}L
```
(added `L` in the end), to make sure that the pointer value does not get truncated. This then seems to work, and gives a minor speed improvement (from 118s to 111s in my case). Maybe when we'll replace the `peek` and `advance` in the same way (in `synapses.pyx`), this gets a bit more significant? In either case, let's not forget that the long-term goal here is to work towards a full C++-based runtime, the speed improvement is only a nice bonus.

Legend101Zz:
> Maybe when we'll replace the `peek` and `advance` in the same way (in `synapses.pyx`), this gets a bit more significant?

Ahh — yes I forgot that .... silly me , totally forgot about potential truncation there. Thank you :)
I think the `PyCapsule` patch I'm working on might make this workaround unnecessary anyway ... 

> Maybe when we'll replace the peek and advance in the same way (in synapses.pyx), this gets a bit more significant?

Yes, that was exactly the plan. Do you think it makes sense to include those changes in this PR itself, or would it be cleaner to separate them into a follow-up PR?

also we need one more PR to remove the python spikeQueue version ... so how do we handle these PR's ?

mstimberg:
In this case, I'd prefer to either have everything in this PR, or open a separate PR for the Python `SpikeQueue` removal (which we'd merge before this PR). The main constraint is that we want `master` to be in a working state at all times, and the current PR does not work with the Python `SpikeQueue` (even though no test actually catches this :grimacing: ).

For me, the `advance` and `peek` changes are very similar to what we are doing here, so they should be part of the same PR (feel free to change its title, though :wink: )

Legend101Zz:
---

Hi @mstimberg,

I did a bit of digging into our earlier theory — that passing the pointer directly into the template would cause issues due to caching behavior. We assumed that if the Python code changed slightly (e.g., by adding a print statement), the caching mechanism might reuse an old compiled module with a stale pointer address, while the actual `spikeQueue` object had changed.

Turns out, it’s actually the opposite.

###  What’s really happening:

The caching mechanism **does not kick in at all** in this case — because the pointer value is embedded directly into the rendered Cython template, and that pointer changes every run. This causes the cache key to differ every time.

Take a look at the relevant section of the cache logic here:

https://github.com/brian-team/brian2/blob/d7a39ec68018df5d0f8a8eb001fb56b5132edae3/brian2/codegen/runtime/cython_rt/extension_manager.py#L113-L133

The key point is: the **`code`** used in the cache key is the fully rendered template — meaning after `{{queue_ptr}}` substitution.

So if the pointer changes between runs, so does the rendered code → so does the cache key → resulting in a fresh compilation every time.

#### Example:

```python
# Python template kwarg:
python_template_kwds = {
    "queue_ptr": self.queue.getptr(),  # e.g., 105553165423776
}
```

Template renders to:

```
Run 1: cdef void* _void_ptr = <void*>105553165423776L
Run 2: cdef void* _void_ptr = <void*>105553165423680L
```

Cache keys:

```
Run 1 → "...cdef void* _void_ptr = <void*>105553165423776L..."
Run 2 → "...cdef void* _void_ptr = <void*>105553165423680L..."
```

Result: Different rendered code → different cache key → new `.so` every time → **no caching benefit at all**.

---

###  So, in short (TLDR) :

The original concern (stale pointers in reused cache) doesn't apply — but ironically, embedding the pointer like this **disables caching entirely**, so we are back again to what we wanted ... 

So we still need to move the pointer from **compile-time (template substitution)** to **runtime (namespace variable)** as we originally wanted but for a different reason😅

---

Legend101Zz:
---
Hi @mstimberg, quick update from my side:

I've successfully figured out how to inject the CSpikeQueue* pointer into the namespace at runtime instead of compile time. This sidesteps the caching issues we discussed earlier :)

I'm now finalizing the peek() and advance() methods and should have them ready today.

In parallel, I experimented with both the PyCapsule and direct pointer casting approaches. Below is a summary of the differences I've observed so far. Would love your thoughts on which direction makes the most sense going forward:

---

| **Aspect**                    | **PyCapsule Approach**                                                         | **Pointer Casting Approach**                                                       |
|------------------------------|----------------------------------------------------------------------------------|-------------------------------------------------------------------------------------|
| **Generated C Code Size**     | ~11,774 lines (smaller by ~2.4%)                                                | ~12,058 lines                                                                       |
| **HTML Annotation**          | Screenshot 2025-06-24 at 17 46 52  | Screenshot 2025-06-24 at 18 01 45   |
| **Binary Size**               | ~110 KB                                                                         | ~110 KB                                                                             |
| **Per-call Overhead**         | Every method call must do a PyCapsule_GetPointer(capsule, "CSpikeQueue") and error-check it via the Python C-API. That’s at least one function call into Python land plus branch checks every time we touch the queue.  | Once we hold a raw CSpikeQueue* , then All subsequent calls go straight to C++ with zero Python-API invocation.|
| **Python API Calls**          | `PyCapsule_GetPointer()` invoked on every method access                         | No Python API calls after one-time setup                                            |
| **Memory Access Pattern**     | Indirect via Python object + validation                                         | Direct pointer dereference (via integer cast)                                       |
| **Runtime Safety**            | Safer — automatic runtime validation of capsule type and pointer validity       | Unsafe — invalid pointer results in segmentation fault; must be manually validated  |
| **Cross-platform Behavior**   | Fully abstracted by Python internals                                            | Safe if cast via `uintptr_t`; relies on platform-compatible pointer size            |

mstimberg:
Hi @Legend101Zz, thanks for the update! I think I'd prefer the `PyCapsule` approach, since it is more clear about what the variable contains. I don't think this "once-per-timestep" unpacking makes any difference here. If we want to improve on this in an even more general way, we'd replace the current approach where our generated Cython code is:
```
def main(_namespace):
   ...
```
by something along the lines of
```
class CodeObject:
   def __init__(_namespace):
       ...
    def run():
```
All the type conversions etc. would only happen once at the start of a new run, from then on we'd only run "pure C++" code that works on the internally stored pointers. But this would be a bigger change in our infrastructure, so don't worry about it for this PR. But something to keep in mind for the bigger project of replacing Cython!

Legend101Zz:
Ahh Also I refactored the `peek` and `advance` methods of the **`SpikeQueue`**  as we discussed last week to eliminate Python-layer overhead and directly be called so all the interactions with Class pointer now use a `PyCapsule` to access the underlying C++ `CSpikeQueue` pointer directly from Cython.

Here's how the updated method annotations look with the new `PyCapsule`-based access pattern for `peek` and `advance` methods:
Screenshot 2025-06-26 at 23 37 39

brian2/synapses/cythonspikequeue.pyx
```diff
@@ -54,6 +55,23 @@ cdef class SpikeQueue:
     def _full_state(self):
         return self.thisptr._full_state()
 
+    def getptr(self):
```
Legend101Zz: @mstimberg please tell me know if we need this method now or should we remove it altogether

Legend101Zz: @mstimberg I have also removed the log statements that were accidentally added ... please let me know what to do of the above cython function and any other changes you'd like to :)

Legend101Zz: @mstimberg bump on this :)

mstimberg: Given that the function is not used anywhere, I think you can indeed remove it (or rather not add it, given that it has been added in this PR ;-) )

Legend101Zz: great thanks :)
changes done https://github.com/brian-team/brian2/pull/1643/commits/c3ca88dff77910818466a041927238c0087098ee

Legend101Zz:
@mstimberg Do you want me to merge the updated master branch into this branch first, or is it okay to proceed as is? I don’t see any warnings or conflicts, and CI seems to be passing cleanly :)

brian2/synapses/cythonspikequeue.pyx
```diff
@@ -5,10 +5,10 @@
 from libcpp.vector cimport vector
 from libcpp.pair cimport pair
 from libcpp.string cimport string
-
 import cython
 from cython.operator import dereference
 from cython.operator cimport dereference
+from cpython.pycapsule cimport PyCapsule_New, PyCapsule_GetPointer
```
mstimberg: Yet another nitpick: I don't think we use `PyCapsule_GetPointer` anywhere :blush:

mstimberg:
> @mstimberg Do you want me to merge the updated master branch into this branch first, or is it okay to proceed as is? I don’t see any warnings or conflicts, and CI seems to be passing cleanly :)

No, no need to do a merge.

brian2/synapses/cythonspikequeue.pyx
Legend101Zz: ahh yes 😅 ... changes done https://github.com/brian-team/brian2/pull/1643/commits/9d8fc7c9b7deff744b256ccfaaedf24dde404f4b , thanks for pointing that out :)

Issue #1245, opened 2020-10-19
Error! C:\Users\USER\Desktop\FAITH\output\results\_dynamic_array_spikemonitor_2_t_-5040102116673421125 is not UTF-8 encoded Saving disabled. 


ramapati166:
During simulation of Spiking Neural Network using Brian2 at cpp standalone mode "set_device('cpp_standalone', directory='STDP_standalone')
device.reinit()
device.activate()"
after the simulation in the generated output folder in the result section given the list of dynamic file of SpikeMonitor such as 
"_dynamic_array_spikemonitor_2_t_-5040102116673421125".But when I am trying to open the file it showing the above described error.As an alternative if I want to access the spike information such as spike time(t) and corresponding neuron( i) by using SpikeMonitor command it is not showing the complete information.Say I have considered 3 layered neural netwok(first layer for poisson input,hidden layer 28 neurons and output layer 12 neurons) so when I am trying to accessing the spike time information using SpikeMonitor command at the cosole I am getting the result like 
<spikemonitor_2.i: array([ 0,  1,  2, ..., 25, 26, 27])>
<spikemonitor_2.t: array([  3.2,   3.2,   3.2, ..., 499.9, 499.9, 499.9]) * msecond>
How to get the intermediate values which are denoted by "dot"?If anyone can help regarding this issues I will be highly obliged to you.Thanksin advance.

mstimberg:
[In the future, please direct questions or error reports where you are not sure whether it is "Brian's fault" to Brian's [discussion forum](https://brian.discourse.group), we try to keep the github issues for bugs and feature requests.]

The files in the outputs/results folder are binary files used internally by Brian to exchange the values between the C++ simulation and the Python script. They are not supposed to be opened directly, and in particular not with a text editor which I think is what you were trying to do.

The `SpikeMonitor`'s `i` and `t` attribute contain all the values. You can plot them or analyze them in any way you want with the same techniques you'd use for any `numpy` array. Since there are more than 1000 values, not all values get printed on the console by default. If you really want to print all the values, you can [use numpy's `set_printoptions` function](https://stackoverflow.com/questions/1987694/how-to-print-the-full-numpy-array-without-truncation).

ramapati166:
Thank you very much for your reply.I am very sorry sir,from next time I will send the issues to Brian's discussion forum.

mstimberg:
[...]

PR #1644, opened 2025-06-12
Remove runtime dependency-version checks from __init__.py

### LINKED ISSUE #1443: Remove version checks in `brian2.__init__.py`?
When importing the `brian2` package, we check a few packages for outdated versions:
https://github.com/brian-team/brian2/blob/4886c46e92d9197cbc7401fd2ab6debe9116134d/brian2/__init__.py#L74-L100

I'm not entirely convinced anymore that this is a good idea, so maybe we should simply remove these checks. But it would be good to have a look first how/if other packages are handling this. If we keep this, we should at least make sure that the checked requirements stay in sync with the actual requirements, by reading them in from some canonical source (`requirements.txt`, `pyproject.toml`, ...).

Legend101Zz:
Closes #1443
---
### Description 

 This PR removes the dependency version checks from `brian2/__init__.py`.

 These checks are:

 * Redundant (handled by `pyproject.toml` and pip)
 * Inconsistent (e.g., required `numpy>=1.10` at runtime vs. `>=1.23.5` in `pyproject.toml`)
 * Confusing to users
 * A maintenance burden

 **How other major packages handle this:**

 * NumPy, Pandas, Matplotlib, scikit-learn: no runtime checks, rely on install-time resolution
 * SciPy: uses runtime version checks, but it’s caused numerous user complaints

**Documented problems from SciPy’s runtime checks:**

- **[GitHub Issue #16964](https://github.com/scipy/scipy/issues/16964)**: "BUG: scipy 1.7.3 wheels on PyPI require numpy<1.23 in contradiction with pyproject.toml and documentation"  
- **[Kaggle Docker Issue #1245](https://github.com/Kaggle/docker-python/issues/1245)**: "Numpy version (1.23.5) exceeds Scipy version (1.7.3) requirement"
- **[Intel Community Post](https://community.intel.com/t5/Intel-Distribution-for-Python/Incompatible-numpy-and-scipy-versions-in-the-Intel-channel/m-p/1482452)**: Intel users getting warnings even with tested MKL-optimized versions
- **[Bioconda Issue #45942](https://github.com/bioconda/bioconda-recipes/issues/45942)**: Runtime test failures due to SciPy version warnings
- **[Allsky Discussion #3130](https://github.com/AllskyTeam/allsky/discussions/3130)**: Nightly warning spam from SciPy checks
- **[Stack Overflow Question](https://stackoverflow.com/questions/73072257/resolve-warning-a-numpy-version-1-16-5-and-1-23-0-is-required-for-this-versi)**: Multiple users confused by contradictory version requirements

These examples show that runtime checks often **do more harm than good** — better to let users get a clean ImportError or install-time resolution failure.

This PR removes `_check_dependency_versions()` and all related logic from the codebase.

---

### DIFF
--- a/brian2/__init__.py
+++ b/brian2/__init__.py
@@ -92,7 +92,6 @@ __docformat__ = "restructuredtext en"
 from brian2.only import *
 from brian2.only import test
 
-
 # Initialize the logging system
 BrianLogger.initialize()
 logger = get_logger(__name__)
--- a/brian2/__init__.py
+++ b/brian2/__init__.py
@@ -93,38 +93,6 @@ from brian2.only import *
 from brian2.only import test
 
 
-# Check for outdated dependency versions
-def _check_dependency_version(name, version):
-    import sys
-
-    from packaging.version import Version
-
-    from .core.preferences import prefs
-    from .utils.logger import get_logger
-
-    logger = get_logger(__name__)
-
-    module = sys.modules[name]
-    if not isinstance(module.__version__, str):  # mocked module
-        return
-    if not Version(module.__version__) >= Version(version):
-        message = (
-            f"{name} is outdated (got version {module.__version__}, need version"
-            f" {version})"
-        )
-        if prefs.core.outdated_dependency_error:
-            raise ImportError(message)
-        else:
-            logger.warn(message, "outdated_dependency")
-
-
-def _check_dependency_versions():
-    for name, version in [("numpy", "1.10"), ("sympy", "1.2"), ("jinja2", "2.7")]:
-        _check_dependency_version(name, version)
-
-
-_check_dependency_versions()
-
 # Initialize the logging system
 BrianLogger.initialize()
 logger = get_logger(__name__)
--- a/brian2/core/core_preferences.py
+++ b/brian2/core/core_preferences.py
@@ -18,6 +18,20 @@ def default_float_dtype_validator(dtype):
     return dtype in [float32, float64]
 
 
+def outdated_dependency_error_validator(value):
+    if value is not None:
+        import warnings
+
+        warnings.warn(
+            "The 'core.outdated_dependency_error' preference is no longer "
+            "used and will be removed in a future version. Brian2 now "
+            "relies on package managers for dependency management.",
+            UserWarning,
+            stacklevel=2,
+        )
+    return True  # We always accept the value user provides, but warn them
+
+
 prefs.register_preferences(
     "core",
     "Core Brian preferences",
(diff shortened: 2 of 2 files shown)


mstimberg:
Hi @Legend101Zz. Thanks for the change and the research on how other packages handle it. The changes look good (they are rather straightforward ;-)), but we should ideally also do something about the now useless `core.outdated_dependency_error` preference. I think the best approach is to change its default value ([here](https://github.com/brian-team/brian2/blob/54e782609e7bce33e79eab3e34b557c7a484c0af/brian2/core/core_preferences.py#L39-L45)) to `None` and add a `validator` (which preferences can use to check for valid preference values) which will raise a warning whenever the value is set to something that is not `None`, i.e. when the user explicitly sets this preference. And of course, we'd note in the documentation that this preference no longer serves any purpose.

mstimberg:
Not a big deal, but to avoid that our CI complains about coding style (in this case, a missing empty line…), make sure to install the `pre-commit` tool (https://pre-commit.com/#install), which will check/format before it lets you commit.

Legend101Zz:
> Not a big deal, but to avoid that our CI complains about coding style (in this case, a missing empty line…), make sure to install the `pre-commit` tool (https://pre-commit.com/#install), which will check/format before it lets you commit.

ahh .. thanks will do , totally forgot about that  `:)

Legend101Zz:
> Hi @Legend101Zz. Thanks for the change and the research on how other packages handle it. The changes look good (they are rather straightforward ;-)), but we should ideally also do something about the now useless `core.outdated_dependency_error` preference. I think the best approach is to change its default value ([here](https://github.com/brian-team/brian2/blob/54e782609e7bce33e79eab3e34b557c7a484c0af/brian2/core/core_preferences.py#L39-L45)) to `None` and add a `validator` (which preferences can use to check for valid preference values) which will raise a warning whenever the value is set to something that is not `None`, i.e. when the user explicitly sets this preference. And of course, we'd note in the documentation that this preference no longer serves any purpose.



> Hi @Legend101Zz. Thanks for the change and the research on how other packages handle it. The changes look good (they are rather straightforward ;-)), but we should ideally also do something about the now useless `core.outdated_dependency_error` preference. I think the best approach is to change its default value ([here](https://github.com/brian-team/brian2/blob/54e782609e7bce33e79eab3e34b557c7a484c0af/brian2/core/core_preferences.py#L39-L45)) to `None` and add a `validator` (which preferences can use to check for valid preference values) which will raise a warning whenever the value is set to something that is not `None`, i.e. when the user explicitly sets this preference. And of course, we'd note in the documentation that this preference no longer serves any purpose.



> Hi @Legend101Zz. Thanks for the change and the research on how other packages handle it. The changes look good (they are rather straightforward ;-)), but we should ideally also do something about the now useless `core.outdated_dependency_error` preference. I think the best approach is to change its default value ([here](https://github.com/brian-team/brian2/blob/54e782609e7bce33e79eab3e34b557c7a484c0af/brian2/core/core_preferences.py#L39-L45)) to `None` and add a `validator` (which preferences can use to check for valid preference values) which will raise a warning whenever the value is set to something that is not `None`, i.e. when the user explicitly sets this preference. And of course, we'd note in the documentation that this preference no longer serves any purpose.

Thanks for the

Legend101Zz:
> Hi @Legend101Zz. Thanks for the change and the research on how other packages handle it. The changes look good (they are rather straightforward ;-)), but we should ideally also do something about the now useless `core.outdated_dependency_error` preference. I think the best approach is to change its default value ([here](https://github.com/brian-team/brian2/blob/54e782609e7bce33e79eab3e34b557c7a484c0af/brian2/core/core_preferences.py#L39-L45)) to `None` and add a `validator` (which preferences can use to check for valid preference values) which will raise a warning whenever the value is set to something that is not `None`, i.e. when the user explicitly sets this preference. And of course, we'd note in the documentation that this preference no longer serves any purpose.


Thanks for pointing that out! I’ve made the required changes [here](https://github.com/brian-team/brian2/pull/1644/commits/3c0e79a8ff207c6afa6cc70ea090ca9b1091ed93) ..
Also, sorry about the accidental PR closure — that was a misclick on my end 😅

PR #1649, opened 2025-06-26
Remove Python SpikeQueue fallback and make Cython version mandatory

Legend101Zz:
**Description:**

This PR removes the Python implementation of SpikeQueue and makes the C++/Cython version mandatory. Previously, Brian2 would fall back to a pure Python implementation when the Cython extension failed to build, but this fallback is no longer needed since Cython wheel distribution has become more reliable.

**Rationale:**
The Python fallback was originally added when Cython wheel builds were complex and unreliable. Since Cython distribution has improved significantly, maintaining both implementations adds unnecessary complexity and potential inconsistencies.

**Breaking Changes:**
- Installations without properly built Cython extensions will now fail with a clear error message instead of silently falling back to slower Python implementation
- Code that explicitly imported from `brian2.synapses.spikequeue` will need to import from `brian2.synapses.cythonspikequeue`

---


Legend101Zz:
Hi @mstimberg ...

So I raised the PR to remove the Python `spikeQueue fallback`, I'll rebase this once we merge in #1643 :)

Please let me know if there are any other areas in the codebase I need to handle to ensure we smoothly remove the Python `spikeQueue fallback`

Legend101Zz:
---

Hi @mstimberg,

I’ve been experimenting with transitioning Brian2's build system to use only `pyproject.toml` for building Cython extensions, aiming to eliminate `setup.py`. However, I’ve hit a roadblock related to NumPy header inclusion.

Specifically, the C extension build fails with the following error due to missing NumPy headers:

```bash
brian2/synapses/cythonspikequeue.cpp:1176:10: fatal error: 'numpy/arrayobject.h' file not found
 1176 | #include "numpy/arrayobject.h"
      |          ^~~~~~~~~~~~~~~~~~~~~
compilation terminated.
```

Screenshot 2025-06-30 at 18 57 39

In the legacy `setup.py`, we handle this via:

```python
import numpy
numpy.get_include()
```

However, `pyproject.toml` is static and doesn’t support dynamic evaluation like `numpy.get_include()`. I also came across this discussion:
👉 [https://github.com/pypa/setuptools/discussions/4154#discussioncomment-12670172](https://github.com/pypa/setuptools/discussions/4154#discussioncomment-12670172)
which confirms that dynamically specifying `include_dirs` (e.g., for NumPy) is currently not possible using only TOML.

Given this limitation, the most pragmatic approach seems to be **keeping a minimal `setup.py` shim**, solely for handling Cython extension builds that require dynamic paths like NumPy's headers.

Let me know your thoughts. Would you be okay retaining a lightweight `setup.py` just for this purpose?

---

mstimberg:
Of course. I know that the support for extensions in `pyproject.toml` is still experimental, so if getting the include dir is not possible, we cannot use it yet. A minimal `setup.py` would be the best solution then.

Legend101Zz:
> Of course. I know that the support for extensions in `pyproject.toml` is still experimental, so if getting the include dir is not possible, we cannot use it yet. A minimal `setup.py` would be the best solution then.

Thanks , I did the needed changes https://github.com/brian-team/brian2/pull/1649/commits/2489a31bfe2c4273365d08cb013b00c4d1d958d4

dev/benchmarks/benchmark_spikequeue.py
```diff
@@ -6,7 +6,7 @@
 GENERAL_SETUP =  ['import numpy as np',
                   'from brian2.tests.test_spikequeue import create_all_to_all, create_one_to_one',
                   'from brian2.units.stdunits import ms',
-                  'from brian2.synapses.spikequeue import SpikeQueue']
+                  'from brian2.synapses.cythonspikequeue import SpikeQueue']
```
mstimberg, line 9: These import changes are of course fine, but I think for maximum backwards compatbility it would be good to keep the `brian2/synapses/spikequeue.py` file, but have it do a
```python
from .cythonspikequeue import SpikeQueue
```
(potentially this could also be the place to give a good error message, telling users that if they intend to run from the source directory, they need to do a `pip install -e .` to do an editable install)

setup.py
```diff
@@ -5,98 +5,114 @@
...
 from setuptools import setup, Extension
 from setuptools.command.build_ext import build_ext
-from setuptools.errors import CompileError, PlatformError as DistutilsPlatformError
+
+
 
 try:
     from Cython.Build import cythonize
-    cython_available = True
+    CYTHON_AVAILABLE = True
```
mstimberg: I appreciate the effort of keeping the .cpp→.so shortcut, but given that users are not usually building the package themselves, I don't think we really need it? From my side, the whole setup.py file could only consist of the `setup` call with `cythonize`, adding the `numpy` include dir. Oh, we might actually not need to call `cythonize` ourselves, `setuptools` will take care of it: https://setuptools.pypa.io/en/latest/userguide/ext_modules.html#distributing-extensions-compiled-with-cython

mstimberg (changes requested):

dev/benchmarks/benchmark_spikequeue.py
Legend101Zz: understood , thanks ...  :)
changes done https://github.com/brian-team/brian2/pull/1649/commits/351ce08f2c09b08e3fcd960ee3ecbcd46e948aac

setup.py
Legend101Zz: how does this look to you https://github.com/brian-team/brian2/pull/1649/commits/b5c91ecd24de2582c7561947a75793f4837342a3 ?

mstimberg: This looks better, but I think we can simplify it even more drastically. If there's no major reason to have additional logic, `setup.py` should only be a few lines, containing nothing more than imports and something along the lines of:
```python
setup(ext_modules=Extension("brian2.synapses.cythonspikequeue",
                            ["brian2/synapses/cythonspikequeue.pyx"],
                            include_dirs=numpy.get_include_dirs())
```

Legend101Zz: ---

Great, thanks — I've implemented the changes as we discussed in our conversation here:
🔗 [https://github.com/brian-team/brian2/pull/1649/commits/2285edfaf9dff1a1c1be63ee61631a327252d72a](https://github.com/brian-team/brian2/pull/1649/commits/2285edfaf9dff1a1c1be63ee61631a327252d72a)

I had a quick follow-up question though:

Should we explicitly account for the caveat that `setuptools` does **not** parse `# distutils:` directives in `.pyx` files — only `Cython` does?

For example, in this file:
[https://github.com/brian-team/brian2/blob/eafcc708b9a9623f1435cfaaf17fa986a95bf338/brian2/synapses/cythonspikequeue.pyx#L1-L7](https://github.com/brian-team/brian2/blob/eafcc708b9a9623f1435cfaaf17fa986a95bf338/brian2/synapses/cythonspikequeue.pyx#L1-L7)

If the line:

```python
# distutils: sources = brian2/synapses/cspikequeue.cpp
```

was missing, and we didn't manually pass `"brian2/synapses/cspikequeue.cpp"` in the `Extension(..., sources=[...])` in `setup.py`, then the `.cpp` wouldn't be compiled — which could lead to fail build (though I guess we as always can catch that beforehand , but still wanted to ask ).

So do you think it’s safer to always declare the `.cpp` explicitly in `setup.py` ? 

like 
```python
# A Helper function to require cython extension
def require_cython_extension(module_path, module_name):
    """
    Create a cythonized Extension object from a .pyx source.
    """
    # File paths
    base_path = os.path.join(*module_path)
    pyx_file = os.path.join(base_path, f"{module_name}.pyx")
    cpp_file = os.path.join(base_path, f"{module_name}.cpp")

    # Module name for setuptools
    full_module_name = ".".join(module_path + [module_name])

    ext = Extension(full_module_name, [pyx_file,  cpp_file], include_dirs=[
        numpy.get_include()],)
    return ext
```

PR #1657, opened 2025-07-22
Feat: Introduce abstract RateMonitor class for unified rate analysis

### LINKED ISSUE #1368: Add monitor to estimate rates of each neuron in group
It's often useful to estimate a smoothed estimate of firing rate for each neuron in a group.

However, the current options are: 
1. `PopulationRateMonitor` which has a `smooth_rate()` method, but averages across the population
    - in theory you could connect one of these to each neuron in a group individually but this is somewhat clunky
2. `SpikeMonitor` which can return an array of `spike_trains()` one for each neuron in a group, but doesn't have a built-in `smooth_rate()` method

it would be great to extend `SpikeMonitor` to be able to call `smooth_rate()` for each neuron.

as a bonus, it would be nice for an option of returning a vector of spike counts binned at a certain interval (this could be thought of as `smooth_rate(window='flat', width=dt, stride=dt)` but may also be clearer as it's own function)

Legend101Zz:
Closes #1368 

Based on the great discussions and instructions in the issue (thanks, they were really helpful!), I've started the implementation for a unified rate-handling mechanism.

So far, I've created the abstract `RateMonitor` base class and refactored `PopulationRateMonitor` to inherit from it. I've also added the new `binned()` method to `PopulationRateMonitor` to allow for re-binning of the recorded population rate.

So I guess , `EventMonitor` is a bit more complex since it handles generic events, not just spikes. I'll need to look more closely at how to pass the correct variable information to the template for that implementation.

**Did:**
- Added the abstract `RateMonitor` class.
- Refactored `PopulationRateMonitor` to inherit from `RateMonitor`.
- Implemented the `binned()` method for `PopulationRateMonitor`.

**ToDo:**
- Implement the `binned()` method for `EventMonitor` to calculate binned rates from event times.
- Thoroughly test the new functionality.
- Write unit tests for the new methods.

### DIFF
--- a/brian2/monitors/ratemonitor.py
+++ b/brian2/monitors/ratemonitor.py
@@ -2,8 +2,11 @@
 Module defining `PopulationRateMonitor`.
 """
 
+from abc import ABC, abstractmethod
+
 import numpy as np
 
+from brian2.core.clocks import Clock
 from brian2.core.variables import Variables
 from brian2.groups.group import CodeRunner, Group
 from brian2.units.allunits import hertz, second
@@ -14,6 +17,114 @@ __all__ = ["PopulationRateMonitor"]
 
 
 logger = get_logger(__name__)
+
+
+class RateMoniter(CodeRunner, Clock, ABC):
+    """
+    Abstract base class for monitors that record rates.
+    """
+
+    @abstractmethod
+    @check_units(bin_size=second)
+    def binned(self, bin_size):
+        """
+        Return the rate calculated in bins of a certain size.
+
+        Parameters
+        -------------
+        bin_size : `Quantity`
+            The size of the bins in seconds. Should be a multiple of dt.
+
+        Returns
+        -------
+        bins : `Quantity`
+            The midpoints of the bins.
+        binned_values : `Quantity`
+            The binned values. For EventMonitor subclasses, this is a 2D array
+            with shape (neurons, bins). For PopulationRateMonitor, this is a 1D array.
+        """
+        raise NotImplementedError()
+
+    @check_units(width=second)
+    def smooth_rates(self, window="gaussian", width=None):
+        """
+        Returns a smooted out version of the firing rate(s).
+
+        Parameters
+        ----------
+        window : str, ndarray
+            The window to use for smoothing. Can be a string to chose a
+            predefined window(`flat` for a rectangular, and `gaussian`
+            for a Gaussian-shaped window).
+
+            In this case the width of the window
+            is determined by the `width` argument. Note that for the Gaussian
+            window, the `width` parameter specifies the standard deviation of
+            the Gaussian, the width of the actual window is `4*width + dt`
+            (rounded to the nearest dt). For the flat window, the width is
+            rounded to the nearest odd multiple of dt to avoid shifting the rate
+            in time.
+            Alternatively, an arbitrary window can be given as a numpy array
+            (with an odd number of elements). In this case, the width in units
+            of time depends on the ``dt`` of the simulation, and no `width`
+            argument can be specified. The given window will be automatically
+            normalized to a sum of 1.
+        width : `Quantity`, optional
+            The width of the ``window`` in seconds (for a predefined window).
+
+        Returns
+        -------
+        rate : `Quantity`
+            The smoothed firing rate(s) in Hz. For EventMonitor subclasses,
+            this returns a 2D array with shape (neurons, time_bins).
+            For PopulationRateMonitor, this returns a 1D array.
+            Note that the rates are smoothed and not re-binned, i.e. the length
+            of the returned array is the same as the length of the binned data
+            and can be plotted against the bin centers from the ``binned`` method.
+        """
+        if width is None and isinstance(window, str):
+            raise TypeError("Need a width when using a predefined window.")
+        if width is not None and not isinstance(window, str):
+            raise TypeError("Can only specify a width for a predefined window")
+
+        if isinstance(window, str):
+            if window == "gaussian":
+                width_dt = int(np.round(2 * width / self.clock.dt))
+                # Rounding only for the size of the window, not for the standard
+                # deviation of the Gaussian
+                window = np.exp(
+                    -np.arange(-width_dt, width_dt + 1) ** 2
+                    * 1.0
+                    / (2 * (width / self.clock.dt) ** 2)
+                )
+            elif window == "flat":
+                width_dt = int(np.round(width / (2 * self.clock.dt))) * 2 + 1
+                used_width = width_dt * self.clock.dt
+                if abs(used_width - width) > 1e-6 * self.clock.dt:
+                    logger.info(
+                        f"width adjusted from {width} to {used_width}",
+                        "adjusted_width",
+                        once=True,
+                    )
+                window = np.ones(width_dt)
+            else:
+                raise NotImplementedError(f'Unknown pre-defined window "{window}"')
+        else:
+            try:
+                window = np.asarray(window)
+            except TypeError:
+                raise TypeError(f"Cannot use a window of type {type(window)}")
+            if window.ndim != 1:
+                raise TypeError("The provided window has to be one-dimensional.")
+            if len(window) % 2 != 1:
+                raise TypeError("The window has to have an odd number of values.")
+
+        # Get the binned rates at the finest resolution
+        _, binned_values = self.binned(bin_size=self.clock.dt)
+
+        # The actual smoothing
+        smoothed = np.convolve(binned_values, window * 1.0 / sum(window), mode="same")
+        return Quantity(smoothed, dim=hertz.dim)
 
 
 class PopulationRateMonitor(Group, CodeRunner):
--- a/brian2/monitors/ratemonitor.py
+++ b/brian2/monitors/ratemonitor.py
@@ -13,7 +13,7 @@ from brian2.units.allunits import hertz, second
 from brian2.units.fundamentalunits import Quantity, check_units
 from brian2.utils.logger import get_logger
 
-__all__ = ["PopulationRateMonitor"]
+__all__ = ["PopulationRateMonitor", "RateMoniter"]
 
 
 logger = get_logger(__name__)
--- a/brian2/monitors/ratemonitor.py
+++ b/brian2/monitors/ratemonitor.py
@@ -19,7 +19,7 @@ __all__ = ["PopulationRateMonitor"]
 logger = get_logger(__name__)
 
 
-class RateMoniter(CodeRunner, Clock, ABC):
+class RateMoniter(CodeRunner, Group, Clock, ABC):
     """
     Abstract base class for monitors that record rates.
     """
@@ -127,7 +127,7 @@ class RateMoniter(CodeRunner, Clock, ABC):
         return Quantity(smoothed, dim=hertz.dim)
 
 
-class PopulationRateMonitor(Group, CodeRunner):
+class PopulationRateMonitor(RateMoniter):
     """
     Record instantaneous firing rates, averaged across neurons from a
     `NeuronGroup` or other spike source.
@@ -210,83 +210,6 @@ class PopulationRateMonitor(Group, CodeRunner):
         Clears all recorded rates
         """
         raise NotImplementedError()
-
-    @check_units(width=second)
-    def smooth_rate(self, window="gaussian", width=None):
-        """
-        smooth_rate(self, window='gaussian', width=None)
-
-        Return a smooth version of the population rate.
-
-        Parameters
-        ----------
-        window : str, ndarray
-            The window to use for smoothing. Can be a string to chose a
-            predefined window(``'flat'`` for a rectangular, and ``'gaussian'``
-            for a Gaussian-shaped window). In this case the width of the window
-            is determined by the ``width`` argument. Note that for the Gaussian
-            window, the ``width`` parameter specifies the standard deviation of
-            the Gaussian, the width of the actual window is ``4*width + dt``
-            (rounded to the nearest dt). For the flat window, the width is
-            rounded to the nearest odd multiple of dt to avoid shifting the rate
-            in time.
-            Alternatively, an arbitrary window can be given as a numpy array
-            (with an odd number of elements). In this case, the width in units
-            of time depends on the ``dt`` of the simulation, and no ``width``
-            argument can be specified. The given window will be automatically
-            normalized to a sum of 1.
-        width : `Quantity`, optional
-            The width of the ``window`` in seconds (for a predefined window).
-
-        Returns
-        -------
-        rate : `Quantity`
-            The population rate in Hz, smoothed with the given window. Note that
-            the rates are smoothed and not re-binned, i.e. the length of the
-            returned array is the same as the length of the ``rate`` attribute
-            and can be plotted against the `PopulationRateMonitor` 's ``t``
-            attribute.
-        """
-        if width is None and isinstance(window, str):
-            raise TypeError("Need a width when using a predefined window.")
-        if width is not None and not isinstance(window, str):
-            raise TypeError("Can only specify a width for a predefined window")
-
-        if isinstance(window, str):
-            if window == "gaussian":
-                width_dt = int(np.round(2 * width / self.clock.dt))
-                # Rounding only for the size of the window, not for the standard
-                # deviation of the Gaussian
-                window = np.exp(
-                    -np.arange(-width_dt, width_dt + 1) ** 2
-                    * 1.0
-                    / (2 * (width / self.clock.dt) ** 2)
-                )
-            elif window == "flat":
-                width_dt = int(width / 2 / self.clock.dt) * 2 + 1
-                used_width = width_dt * self.clock.dt
-                if abs(used_width - width) > 1e-6 * self.clock.dt:
-                    logger.info(
-                        f"width adjusted from {width} to {used_width}",
-                        "adjusted_width",
-                        once=True,
-                    )
-                window = np.ones(width_dt)
-            else:
-                raise NotImplementedError(f'Unknown pre-defined window "{window}"')
-        else:
-            try:
-                window = np.asarray(window)
-            except TypeError:
-                raise TypeError(f"Cannot use a window of type {type(window)}")
-            if window.ndim != 1:
-                raise TypeError("The provided window has to be one-dimensional.")
-            if len(window) % 2 != 1:
-                raise TypeError("The window has to have an odd number of values.")
-        return Quantity(
-            np.convolve(self.rate_, window * 1.0 / sum(window), mode="same"),
-            dim=hertz.dim,
-        )
 
     def __repr__(self):
         classname = self.__class__.__name__
--- a/brian2/monitors/ratemonitor.py
+++ b/brian2/monitors/ratemonitor.py
@@ -89,12 +89,15 @@ class RateMoniter(CodeRunner, Group, Clock, ABC):
 
         if isinstance(window, str):
             if window == "gaussian":
-                width_dt = int(np.round(2 * width / self.clock.dt))
-                # Rounding only for the size of the window, not for the standard
+                # basically Gaussian theoretically spans infinite time, but practically it falls off quickly,
+                # So we choose a window of +- 2*(Standard deviations), i.e 95% of gaussian curve
+                width_dt = int(
+                    np.round(2 * width / self.clock.dt)
+                )  # Rounding only for the size of the window, not for the standard
                 # deviation of the Gaussian
                 window = np.exp(
                     -np.arange(-width_dt, width_dt + 1) ** 2
-                    * 1.0
+                    * 1.0  # hack to ensure floating-point division :)
                     / (2 * (width / self.clock.dt) ** 2)
                 )
             elif window == "flat":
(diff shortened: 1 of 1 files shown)


Legend101Zz:
@mstimberg as initially anticipated, adding rate analysis directly to `EventMonitor` might be a bit more involved. Since EventMonitor  supports general events (not just spikes), we’d need to pass in information about which variable should be used for rate calculation to the code generation templates.

That said, I might be overthinking it—perhaps we can restrict it for spikes only :)

Legend101Zz:
@mstimberg bump on this :)

mstimberg:
Hi @Legend101Zz. Sorry, I did not yet have time to look into the code in detail, but regarding the `EventMonitor`: it shouldn't actually change anything here with respect to `SpikeMonitor`. A few quick remarks:
- the `spikemonitor` templates have their name for historical reasons only, they are actually implementing the monitor for arbitrary events (they get the name of an `eventspace` passed to the template, which stores the events that were triggered in the current time step. In the case of a `SpikeMonitor`, these events are spikes, but it does not make any different if they are something else – the `eventspace` is always a list of neuron indices.
- for the added binning/smoothing functions we are discussing here, the templates do not come into play, since the whole analysis happens on the Python side after the simulation – the monitors will still record everything in the same way as before.
- For this analysis, it does not matter what the events are, we only need the spike/event trains that give us the times of the spikes/events per neuron. A `SpikeMonitor` has a `spike_trains` method because that is the term that is commonly used, but all it does is return `self.event_trains()`...

Legend101Zz:
@mstimberg hope you are doing well :) 

So I add a very peculiar bug in the code I wrote and it a while to catch that , phew ...

for a very brief context here for the binned fn for eventmonitor 

```python
        if (bin_size / self.clock.dt) % 1 > 1e-6:
            raise ValueError("bin_size has to be a multiple of dt.")

        # Get the total duration and number of bins
        duration = self.clock.t
         num_bins = int(duration / bin_size)
  ```
  I had a slip and had `duration = self.clock.dt` which led to weird behaviour as I had just a single bin everytime , that said ... now  this seems to be working good sharing some snippets , what do you think ? 
  
  
Screenshot 2025-09-14 at 17 47 22
Screenshot 2025-09-14 at 17 50 00

Screenshot 2025-09-14 at 17 59 16


Next I'll add the tests and also check why are the current tests failing ... :)

**Comment by Legend101Zz** [comment]
also @mstimberg can you pleaase guide me here , why do we get this error 

`AttributeError: 'Word' object has no attribute 're_match'`
It seems unrelated to the rate monitoring implementation , happening in this PR. ?

File: brian2/tests/test_monitor.py [hunk 1]
```diff
@@ -588,7 +588,16 @@ def test_rate_monitor_smoothed_rate():
     assert_allclose(smoothed[index - 2 : index + 3], 0.2 / defaultclock.dt)
     with catch_logs(log_level=logging.INFO):
         smoothed2 = r_mon.smooth_rate(window="flat", width=5.4 * defaultclock.dt)
-        assert_array_equal(smoothed, smoothed2)
+
+        # The window size should round up to 7.
```
Legend101Zz, line 592: needed to fix this test

mstimberg:
Regarding test failures: at least partly this is explained by a bug in pyparsing 3.2.4 (see https://github.com/pyparsing/pyparsing/issues/618) – consider merging in `master` or cherry-picking its last commit (which simply changes the requirement to `<3.2.4` as a workaround).

Legend101Zz:
> Regardingg w test failures: at least partly this is explained by a bug in pyparsing 3.2.4 (see [pyparsing/pyparsing#618](https://github.com/pyparsing/pyparsing/issues/618)) – consider merging in `master` or cherry-picking its last commit (which simply changes the requirement to `<3.2.4` as a workaround).

@mstimberg what is strange I just did this change :

Screenshot 2025-09-16 at 23 38 27

and everything works now 😵‍💫 

maybe it can be because the master branch was merged here ... or something related to not having assert statements inside classes , expect for tests ...

But all testsuites pass now 🥳

mstimberg:
> maybe it can be because the master branch was merged here ... or something related to not having assert statements inside classes , expect for tests ...

The PR tests actually run on the merged version, i.e. the workaround from `master` was used, that's why you don't see the failure any more.

Legend101Zz:
> > maybe it can be because the master branch was merged here ... or something related to not having assert statements inside classes , expect for tests ...
> 
> The PR tests actually run on the merged version, i.e. the workaround from `master` was used, that's why you don't see the failure any more.

Ahh understood thanks , that makes more sense , also can you please test and let me know if this was what you expected from the newly added `smooth_rate` ( for spike monitor) or if we'd have to change things , because for me the challenge was dealing with `np.convolve` for EventMonitor/SpikeMonitor case as they return a binned 2D array ( neurons * bins) and so we had to convolve each neuron seperately , to me that seemed right , but please let me know if that makes sense :)

brian2/monitors/ratemonitor.py
```diff
@@ -2,21 +2,154 @@
...
 
-__all__ = ["PopulationRateMonitor"]
+__all__ = ["PopulationRateMonitor", "RateMonitor"]
 
 
 logger = get_logger(__name__)
 
 
-class PopulationRateMonitor(Group, CodeRunner):
+class RateMonitor(CodeRunner, Group, Clock, ABC):
```
mstimberg: A bit confused: why does `RateMonitor` has to inherit from `Clock`? (It has a clock, but it isn't a clock)

brian2/monitors/ratemonitor.py
```diff
@@ -2,21 +2,154 @@
...
+            of time depends on the ``dt`` of the simulation, and no `width`
+            argument can be specified. The given window will be automatically
+            normalized to a sum of 1.
+        width : `Quantity`, optional
+            The width of the ``window`` in seconds (for a predefined window).
+
+        Returns
+        -------
+        rate : `Quantity`
+            The smoothed firing rate(s) in Hz. For EventMonitor subclasses,
```
mstimberg, line 96: I think this docstring is slightly incorrect: the smoothing function does not use any binning, the resulting array will always have the size of the original number of time steps.

brian2/monitors/ratemonitor.py
```diff
@@ -100,82 +233,48 @@ def reinit(self):
...
-        return Quantity(
-            np.convolve(self.rate_, window * 1.0 / sum(window), mode="same"),
-            dim=hertz.dim,
-        )
+        num_bins = int(self.clock.t / bin_size)
+        bins = (
+            np.arange(num_bins) * bin_size + bin_size / 2.0
+        )  # as we want Bin centers (not edges)
+
+        t_indices = (self.t / self.clock.dt).astype(int)
```
mstimberg: Actually, let's discuss this during our meeting, the issue is slightly more complicated than this.

brian2/monitors/ratemonitor.py
```diff
@@ -100,82 +233,48 @@ def reinit(self):
...
+        binned_values = np.zeros(num_bins)  # to store total firing rate values per bin
+        bin_counts = np.zeros(num_bins)  # to store how many samples went into each bin
+
+        np.add.at(binned_values, bin_indices, self.rate)
+        np.add.at(bin_counts, bin_indices, 1)
+
+        # Avoid division by zero for empty bins
+        non_empty_bins = bin_counts > 0
+        binned_values[non_empty_bins] /= bin_counts[non_empty_bins]
+        return bins, Quantity(binned_values, dim=hertz.dim)
```
mstimberg, line 327: I feel that this complicates things a bit too much – we shouldn't have to count the number of values that go into each bin (each bin should have `bin_size/dt` values), and there shouldn't be any empty bins. But then, we can actually have a much simpler calculation for binned rates in the `PopulationRateMonitor`. Let's say our rates are: `[1, 2, 3, 4, 4, 4, 3, 2, 1, 0, 0, 0]`, and we want to have bins of 3 timesteps each. Then reshaping this array into a matrix so that it has three values in every row, and then taking the average value over columns gives as the mean firing rate in each bin:
```pycon
>>> print(ar.reshape(-1, 3).mean(axis=1))
[2. 4. 2. 0.]
```

brian2/monitors/spikemonitor.py
```diff
@@ -383,6 +385,71 @@ def event_trains(self):
...
+            #     [0.0, 2.0, 0.0, 0.0, 1.0]   # Neuron 2: 2 in bin 1, 1 in bin 4
+            # ]
+            for event_idx, neuron_idx in enumerate(event_indices):
+                if 0 <= neuron_idx < num_neurons:  # sanity check
+                    bin_idx = bin_indices[event_idx]
+                    if bin_idx < num_bins:  # To handle edge case at the end
+                        binned_values[neuron_idx, bin_idx] += 1
+
+            # Convert counts to rates (Hz)
+            binned_values = binned_values / float(bin_size)
```
mstimberg, line 447: Something I didn't think about earlier: if we don't return spike counts but rates here (which probably makes sense given that we are inheriting from **Rate**Monitor, and that this is the behaviour in the `PopulationRateMonitor.binned` method), then we should probably call this method `binned_rate` – this makes it nicely parallel with `smooth_rate`.

brian2/monitors/spikemonitor.py
```diff
@@ -383,6 +385,71 @@ def event_trains(self):
...
+            # binned_values = [
+            #     [2.0, 0.0, 1.0, 0.0, 0.0],  # Neuron 0: 2 in bin 0, 1 in bin 2
+            #     [0.0, 1.0, 0.0, 1.0, 0.0],  # Neuron 1: 1 in bin 1, 1 in bin 3
+            #     [0.0, 2.0, 0.0, 0.0, 1.0]   # Neuron 2: 2 in bin 1, 1 in bin 4
+            # ]
+            for event_idx, neuron_idx in enumerate(event_indices):
+                if 0 <= neuron_idx < num_neurons:  # sanity check
+                    bin_idx = bin_indices[event_idx]
+                    if bin_idx < num_bins:  # To handle edge case at the end
+                        binned_values[neuron_idx, bin_idx] += 1
```
mstimberg: If we do not care about half-filled bins in the end of the array (and I think we should not), then I think all of this can be replaced by a call to numpy's `bincount` function (https://numpy.org/doc/stable/reference/generated/numpy.bincount.html) :blush:

brian2/tests/test_monitor.py
```diff
@@ -684,6 +693,188 @@ def test_rate_monitor_subgroups_2():
     assert_allclose(rate_3.rate, 1 / defaultclock.dt)
 
 
+# ====== Tests to test the added rate monitoring functionality in monitors =====
+
+
+@pytest.mark.codegen_independent
+def test_population_rate_monitor_binning():
+    """Test binning functionality for PopulationRateMonitor."""
+    # Create a group with regular spiking
```
mstimberg: The group below has random instead of regular spiking. It would probably be better to use regular spiking, though (e.g. `SpikeGeneratorGroup`), in that way we could have a precise test instead of only assuring that the values are in a reasonable range.

brian2/tests/test_monitor.py
```diff
@@ -684,6 +693,188 @@ def test_rate_monitor_subgroups_2():
...
+
+
+@pytest.mark.codegen_independent
+def test_smooth_rates_spike_monitor():
+    """Test smoothing functionality for SpikeMonitor."""
+    # Create a neuron with a burst of spikes
+    times = np.arange(40, 61) * ms  # Burst from 40-60ms
+    indices = np.zeros(21, dtype=int)
+
+    G = SpikeGeneratorGroup(1, indices, times)
```
mstimberg: I think it would be important here to test with more than one neuron in the group.

brian2/monitors/ratemonitor.py
```diff
@@ -2,21 +2,154 @@
...
+        # So we need to handle for both 1D (PopulationRateMonitor) and 2D (EventMonitor) cases separately as `np.convolve()` only works with 1D arrays
+        if binned_values.ndim == 1:
+            # PopulationRateMonitor case - 1D convolution
+            smoothed = np.convolve(binned_values, window, mode="same")
+        else:
+            # EventMonitor/SpikeMonitor case - convolve each neuron and then we return the smoothed 2D array ( neuron * bins )
+            num_neurons, num_bins = binned_array.shape
+            smoothed = np.zeros((num_neurons, num_bins))
+            for i in range(num_neurons):
+                smoothed[i, :] = np.convolve(binned_array[i, :], window, mode="same")
```
mstimberg, line 167: We previously depended on `scipy` which has a function `convolve2d`, but without it this is probably the best approach.

brian2/monitors/ratemonitor.py
Legend101Zz: ahh ... yes that can be removed ... I thought that because we had `self.clock` so we needed the inheritance ...

brian2/monitors/ratemonitor.py
Legend101Zz: does this sound good to you ?

`"Note that the rates are smoothed at the original time step resolution (dt), not re-binned. The length of the returned array equals the number of recorded time steps and can be plotted against the original time values (e.g., self.t)."`

mstimberg: Yes :+1:

brian2/monitors/ratemonitor.py
```diff
@@ -100,82 +233,48 @@ def reinit(self):
...
-                raise TypeError(f"Cannot use a window of type {type(window)}")
-            if window.ndim != 1:
-                raise TypeError("The provided window has to be one-dimensional.")
-            if len(window) % 2 != 1:
-                raise TypeError("The window has to have an odd number of values.")
-        return Quantity(
-            np.convolve(self.rate_, window * 1.0 / sum(window), mode="same"),
-            dim=hertz.dim,
-        )
+        num_bins = int(self.clock.t / bin_size)
```
mstimberg:
```python
bin_timesteps = timestep(bin_size, dt)
num_bins = self.clock.timestep // bin_timesteps
timesteps = num_bins * bin_timesteps
```

brian2/monitors/ratemonitor.py
Legend101Zz: @mstimberg quick question here ... then are we good if we skip incomplete bins , like :

```bash
say dt = 0.1 * ms
bin_size = 0.3 * ms 
If we had 13 values: [1, 2, 3, 4, 4, 4, 3, 2, 1, 0, 0, 0, 5]
num_bins = 13 // 3 = 4
num_values = 4 * 3 = 12
rate_to_bin = [1, 2, 3, 4, 4, 4, 3, 2, 1, 0, 0, 0]  # so we drop the last value then here right ? 
```

brian2/monitors/ratemonitor.py
Legend101Zz: changes done https://github.com/brian-team/brian2/pull/1657/commits/88b0a49867ba731cd974e3bc08c5b4a3c3a61316 thanks

brian2/monitors/ratemonitor.py
Legend101Zz: changes done https://github.com/brian-team/brian2/pull/1657/commits/88b0a49867ba731cd974e3bc08c5b4a3c3a61316 thanks

brian2/monitors/ratemonitor.py
Legend101Zz: please tell me know fif this looks good to you https://github.com/brian-team/brian2/pull/1657/commits/a1d99aa6a0c14703faf9b59e2a70428c4bffa1e1

brian2/monitors/ratemonitor.py
Legend101Zz: please tell me know fif this looks good to you https://github.com/brian-team/brian2/pull/1657/commits/a1d99aa6a0c14703faf9b59e2a70428c4bffa1e1

brian2/monitors/ratemonitor.py
Legend101Zz: please tell me know fif this looks good to you https://github.com/brian-team/brian2/pull/1657/commits/a1d99aa6a0c14703faf9b59e2a70428c4bffa1e1

brian2/monitors/spikemonitor.py
Legend101Zz: done https://github.com/brian-team/brian2/pull/1657/commits/d588872096344bf3b533a2b6cd12359a1663f807 thanks

brian2/monitors/spikemonitor.py
Legend101Zz: changes done thanks https://github.com/brian-team/brian2/pull/1657/commits/2fcdc4c8a8f8a07b99a99b9f99b1ad2d01a65bc6

brian2/tests/test_monitor.py
Legend101Zz: done https://github.com/brian-team/brian2/pull/1657/commits/a812206a26d6c2fe29fbe70ac41242bc5e7de10d

brian2/tests/test_monitor.py
Legend101Zz: please let me know how do these look to you https://github.com/brian-team/brian2/pull/1657/commits/ad4c00968c7da74c63c205f1c002e9efaff81ad9 :)

Legend101Zz:
@mstimberg thanks alot for the reviews , I am done with resolving them , please let me know if incase I forgot to implement anything that we discussed  :)

brian2/monitors/ratemonitor.py
mstimberg: I'll try out some examples, but I think yes, this looks good!

mstimberg (commented):
Hi. This all looks good, we are almost there I think! I made a quite drastic simplification to the `SpikeMonitor.binned_rate` code in 1ff7214d99 – the tests still seem to pass, so I think it is equivalent (the `ufunc.at` trick is not used very often, but exactly what we need here 😊 ).

While having a final look over the code, I realized something, though: we are a tiny bit inconsistent with respect to the times/bins that we return. For the times that are recorded and returned by the monitors directly, we are using the beginning of the time step. I.e., a spike at time `t` actually means "there was a spike in [t, t+dt[". When we bin the spikes, your code returns bin centres, though (so in the previous example `t + dt/2`). I understand the reasoning, and I think using the bin centres is generally better, e.g. for plotting, but I think for consistency we should return the start of the bins everywhere. I think at least for the corner case where we use `binned_rate` with `bin_size=dt`, this is the most intuitive behaviour, in the sense that "binning with size=dt" doesn't change the times. Actually, while all this is a decision that we can take one way or the other, I would consider `PopulationRateMonitor.binned_rate` buggy right now, since with `bin_size=dt` it returns the originally recorded times (bin starts), and with any other `bin_size` it returns bin centres. What do you think?

Legend101Zz:
> Hi. This all looks good, we are almost there I think! I made a quite drastic simplification to the `SpikeMonitor.binned_rate` code in [1ff7214](https://github.com/brian-team/brian2/commit/1ff7214d99ff8d3079de657813e981c53e0f61c0) – the tests still seem to pass, so I think it is equivalent (the `ufunc.at` trick is not used very often, but exactly what we need here 😊 ).
> 
> While having a final look over the code, I realized something, though: we are a tiny bit inconsistent with respect to the times/bins that we return. For the times that are recorded and returned by the monitors directly, we are using the beginning of the time step. I.e., a spike at time `t` actually means "there was a spike in [t, t+dt[". When we bin the spikes, your code returns bin centres, though (so in the previous example `t + dt/2`). I understand the reasoning, and I think using the bin centres is generally better, e.g. for plotting, but I think for consistency we should return the start of the bins everywhere. I think at least for the corner case where we use `binned_rate` with `bin_size=dt`, this is the most intuitive behaviour, in the sense that "binning with size=dt" doesn't change the times. Actually, while all this is a decision that we can take one way or the other, I would consider `PopulationRateMonitor.binned_rate` buggy right now, since with `bin_size=dt` it returns the originally recorded times (bin starts), and with any other `bin_size` it returns bin centres. What do you think?

Hmm now that you mention it , I realise that for `PopulationRateMonitor.binned_rate()` we are pretty inaccurate with the logic of :

```python
def binned(self, bin_size):
    # ...
    if bin_size == self.clock.dt:
        return self.t[:], self.rate  # Returns STARTS (because self.t uses starts)
    
    # For any other bin_size:
    bins = np.arange(num_bins) * bin_size + bin_size / 2.0  # Returns CENTERS
    # ...
    return bins, binned_rates
 ```
 
 so essentially we have a function which eturns different time representations depending on the bin size 😅 ...
 I agree in general that users can always add bin_size/2 if they really want centers for plotting, but the default should match the rest of Brian2's conventions , we have ...
 
 So should I go ahead and do the changes for this then ?

mstimberg:
> So should I go ahead and do the changes for this then ?

Yes please :blush: Maybe add a "Note" to the docstring mentioning that for plotting it might make more sense to plot bin centres (in particular for large bins).

Legend101Zz:
> add a "Note" to the docstring mentioning that for plotting it might make more sense to plot bin centres (in particular for large bins).

changes done https://github.com/brian-team/brian2/pull/1657/commits/26bf58cc0af4704b5e38931c7b5a9367dd08a26f :)

mstimberg (approved):
LGTM!

PR #1663, opened 2025-07-30
Only use group.start and group.stop to calculate shape if they exist

### LINKED ISSUE #1662: NeuronMonitor indices .shape -> AttributeError: No attribute with name stop
calling `.shape` on the indices of a `NeuronMonitor` gives an attribute error:

```python
spikes.i.shape

KeyError                                  Traceback (most recent call last)
File ~/.pyenv/versions/3.13.0/lib/python3.13/site-packages/brian2/groups/group.py:382, in VariableOwner.state(self, name, use_units, level)
    381 try:
--> 382     var = self.variables[name]
    383 except KeyError as exc:

File ~/.pyenv/versions/3.13.0/lib/python3.13/site-packages/brian2/core/variables.py:1646, in Variables.__getitem__(self, item)
   1645 def __getitem__(self, item):
-> 1646     return self._variables[item]

KeyError: 'stop'

The above exception was the direct cause of the following exception:

KeyError                                  Traceback (most recent call last)
File ~/.pyenv/versions/3.13.0/lib/python3.13/site-packages/brian2/groups/group.py:420, in VariableOwner.__getattr__(self, name)
    419         use_units = True
--> 420     return self.state(name, use_units)
    422 except KeyError:

File ~/.pyenv/versions/3.13.0/lib/python3.13/site-packages/brian2/groups/group.py:384, in VariableOwner.state(self, name, use_units, level)
    383 except KeyError as exc:
--> 384     raise KeyError(f"State variable {name} not found.") from exc
    386 if use_units:

KeyError: 'State variable stop not found.'

AttributeError                            Traceback (most recent call last)
Cell In[4], line 1
----> 1 spikes.i.shape

File ~/.pyenv/versions/3.13.0/lib/python3.13/site-packages/brian2/core/variables.py:1586, in VariableView.shape(self)
   1582 if self.ndim == 1:
   1583     if not self.variable.scalar:
   1584         # This is safer than using the variable size, since it also works for subgroups
   1585         # see GitHub issue #1555
-> 1586         size = self.group.stop - self.group.start
   1587         assert size <= self.variable.size
   1588     else:

File ~/.pyenv/versions/3.13.0/lib/python3.13/site-packages/brian2/groups/group.py:423, in VariableOwner.__getattr__(self, name)
    420     return self.state(name, use_units)
    422 except KeyError:
--> 423     raise AttributeError(f"No attribute with name {name}")

AttributeError: No attribute with name stop
```

This seems to be caused by #1556 (this issue is not present for older brian2 <2.8.0 before this change)

lkeegan:
- resolves #1662

### DIFF
--- a/brian2/core/variables.py
+++ b/brian2/core/variables.py
@@ -1579,6 +1579,10 @@ class VariableView:
     # Get access to some basic properties of the underlying array
     @property
     def shape(self):
+        if isinstance(self.variable, DynamicArrayVariable):
+            # For dynamic variables, we need to make sure that their size is up-to-date in standalone mode
+            # We can trigger the update by asking for the values, even though we don't need them here
+            self.variable.get_value()
         if self.ndim == 1:
             if (
                 hasattr(self, "group")


mstimberg (approved):
This looks very reasonable to me, I'll wait for the test suite to pass and will then merge. Thanks again!

brian2/core/variables.py
```diff
@@ -1580,7 +1580,7 @@ def __hash__(self):
     @property
     def shape(self):
         if self.ndim == 1:
-            if hasattr(self, "group.start") and hasattr(self, "group.stop"):
+            if hasattr(self, "group") and hasattr(self.group, "start") and hasattr(self.group, "stop"):
```
mstimberg: Hah, saved by the testsuite, I did not spot this just by reading :) Usually I write this as `hasattr(getattr(self, "group", None), "start")`, but this would be cumbersome here, since `group` is used twice – I think your approach is the best/clearest.

mstimberg:
@lkeegan sorry, this is a bit annoying for small-ish contributions: we require the code to be black-formatted and our test suite fails if it is not the case (see https://brian2.readthedocs.io/en/stable/developer/guidelines/style.html#code-style). I reformatted your changes to make our linter happy.

lkeegan:
@mstimberg thanks for the quick review and fixes!

PR #1665, opened 2025-07-31
docs(README): Add neuromorphic computing badge - Collaboration Network

neural-loop:
Thank you so much for joining the Collaboration Network! Here is a preview of the badge I've added:

image

Again, you are welcomed and encouraged to reach out if there is any thing we can do to signal boost your significant efforts! We are at your service.


mstimberg:
Thanks :blush:

PR #1674, opened 2025-08-30
Add cppyy Runtime Code Generation Target for Brian2

Legend101Zz:
### Problem

The current Brian2 code generation pipeline suffers from a fundamental performance bottleneck. The issue is not tied to the specific tools we use, but rather to the **Ahead-of-Time (AOT) compilation paradigm** itself.

Regardless of whether we use Cython (our current approach) or manual C-extensions, the workflow remains slow and cumbersome:

* Generate large C++ source files on disk
* Invoke an external compiler (e.g., g++, clang) with significant overhead
* Wait for compilation to complete (often 15–40 seconds, which disrupts interactivity)
* Dynamically load the compiled result through a complex process

In other words, the bottleneck lies in the **file-based, external-compiler, AOT workflow**.

---

### Proposed Solution: JIT Compilation with `cppyy`

This PR introduces `cppyy` as a new runtime code generation target, shifting from AOT to **Just-in-Time (JIT) compilation**.

With `cppyy`, C++ code is compiled in-memory using the Cling C++ interpreter, which eliminates:

* File I/O overhead
* External compiler process spawning
* Long compilation waiting times
* Complex dynamic loading procedures

---

### Current Status

 Implemented and functional in this initial proof of concept:

* End-to-end JIT compilation pipeline
* Basic neuron group simulations
* State updates, thresholds, and resets
* Template system for different operations
* Integration with the device layer

---

### Next Steps

This PR serves as a starting point and contains some hardcoded components that will be refined in future iterations. A follow-up issue will be created to present the detailed roadmap and outline how we plan to transition from this POC to a fully integrated solution for Brian2 :) 

---


Legend101Zz:
closing in favour of #1769

Issue #1686, opened 2025-09-17
GSL integration with CPP/standalone


mdepitta:
Hi folks,
What's the implementation status of GSL integrators with CPP/standalone?
I have a tricky HH-type model that, if I integrate by Euler in Brian2, it shows firing. In contrast, in a separate implementation outside Brian2, it fires -- and further analysis proves that it is supposed to fire (periodically). 

I tried to run my code using GSL integrators, but they don't work:

````
from brian2 import *
from brian2.devices.device import reinit_devices
from brian2.utils.logger import clean_up_logging
code_dir = './codegen'
prefs.GSL.directory = '/usr/include/'
set_device('cpp_standalone',directory=code_dir,build_on_run=False)
prefs.devices.cpp_standalone.openmp_threads = 4
prefs.logging.file_log = False
prefs.logging.delete_log_on_exit = True`
##
device.delete(force=True)
## Build the neuron model and monitors
params = lpc5_parameters(model='hh-neuron',T_ramp=duration*second,**kwargs)
cell = lpc5_neuron(1,params,name='HH',dt=1*us) #--> This return a NeuronGroup where internally, method='gsl'
cell.v = -70*mV
v_mon = StateMonitor(cell,variables='v',record=True,dt=0.1*ms,name='vmon')
network = Network([cell,v_mon])
## Run the simulator
network.run(duration=duration*second,report='text')
device.build(directory=code_dir, clean=True)
```

The above code generates:

```
Traceback (most recent call last):
  File "/home/xxx/Python3.10-VENV/lib/python3.10/site-packages/brian2/core/network.py", line 1003, in before_run
    obj.before_run(run_namespace)
  File "/home/xxx/Python3.10-VENV/lib/python3.10/site-packages/brian2/groups/group.py", line 1266, in before_run
    self.create_code_objects(run_namespace)
  File "/home/xxx/Python3.10-VENV/lib/python3.10/site-packages/brian2/groups/group.py", line 1259, in create_code_objects
    code_object = self.create_default_code_object(run_namespace)
  File "/home/xxx/Python3.10-VENV/lib/python3.10/site-packages/brian2/groups/group.py", line 1240, in create_default_code_object
    self.codeobj = create_runner_codeobj(
  File "/home/xxx/Python3.10-VENV/lib/python3.10/site-packages/brian2/codegen/codeobject.py", line 484, in create_runner_codeobj
    return device.code_object(
[...]

PR #1687, opened 2025-09-19
Fix GSL for C++ standalone mode

mstimberg:
The GSL implementation did not catch up with recent changes in C++ standalone mode, leading it to raise `NotImplementedError`. Unfortunately, the GSL test suite skips tests that raise `NotImplementedError`, since this is supposed to be a signal for use of the `numpy` backend, which doesn't have GSL support (this is probably not a good approach…). The failure therefore went unnoticed. A few small additional fixes were necessary, probably related to changes in our compiler flags.

### DIFF
--- a/brian2/tests/test_GSL.py
+++ b/brian2/tests/test_GSL.py
@@ -21,6 +21,8 @@ def skip_if_not_implemented(func):
             prefs.codegen.target == "auto" and auto_target().class_name == "numpy"
         ):
             pytest.skip("GSL support for numpy has not been implemented yet")
+        else:
+            return func()
 
     return wrapped
 


brian2/codegen/codeobject.py
```diff
@@ -42,7 +43,7 @@ def constant_or_scalar(varname, variable):
         return f"{varname}"
 
 
-class CodeObject(Nameable):
+class CodeObject(Nameable, ABC):
```
Legend101Zz, line 46: yay ... now we use the ABC class :)

mstimberg: :laughing:

mstimberg:
Ok, this took me quite a while to finalize – GSL on C++ standalone did not work on Windows and needed adding the DLL directory to the PATH. There was also an issue with incompatibility of compiled Cython files across versions, and finally a stupid bug introduced by me that meant that all GSL tests were passing… without being executed. All this should now be fixed and this is finally ready to merge.

Fixes #1686.

PR #1688, opened 2025-09-19
Fix backwards compatibility for Python spike queue stored states

### LINKED ISSUE #1661: How to deal with stored states using Python spikequeue
In this [discussion on our discourse forum](https://brian.discourse.group/t/synapses-cannot-be-simulated-after-using-set-states/1475/3), I realized that we have a problem for a not very common corner case: if someone used the Python spike queue fallback and stored the state of the network to disk with `Network.store(filename=...)`, this file can no longer be loaded, since the Cython spike queue uses a slightly different format to store its internal state (I don't quite remember why, though…). Of course, using a Python spike queue state with the Cython spike queue was not working before, either, but now there is no way to run Brian without the Cython spike queue, so this can be considered a regression. Here's a simple example that stores a network with a spike queue to disk:
```python
from brian2 import *

G = NeuronGroup(2, 'x: 1', threshold="True", reset="")
S = Synapses(G, G, on_pre = "x+=1")
S.connect(j="i")
S.delay = [2*defaultclock.dt, 4*defaultclock.dt]
run(defaultclock.dt)
store(filename="test.pkl")
```
We can load the stored state with pickle:
```pycon
>>> import pickle
>>> with open("test.pkl", "rb") as f:
...     d = pickle.load(f)
... 
>>> d["default"]["synapses_pre"]["_spikequeue"]
```
With the Python spike queue, this gives:
```pycon
(0.0001, array([[1, 0],
       [3, 1]]), (5, 4))
```
And with the Cython spike queue:
```pycon
(1, [[], [], [0], [], [1]])
```
Not surprisingly, loading one format with the other version fails. If you try to load the Python spike queue state with the Cython version, the error looks like this:
```pytb
  File "/home/mstimberg/scratch/pickle_test.py", line 6, in <module>
    restore(filename="test.pkl")
  File "/home/mstimberg/git/brian2/brian2/core/magic.py", line 467, in restore
    magic_network.restore(
  File "/home/mstimberg/git/brian2/brian2/core/magic.py", line 273, in restore
    super().restore(
  File "/home/mstimberg/git/brian2/brian2/core/base.py", line 346, in device_override_decorated_function
    return func(*args, **kwds)
           ^^^^^^^^^^^^^^^^^^^
  File "/home/mstimberg/git/brian2/brian2/core/network.py", line 706, in restore
    obj._restore_from_full_state(state[obj.name])
  File "/home/mstimberg/git/brian2/brian2/synapses/synapses.py", line 452, in _restore_from_full_state
    self.queue._restore_from_full_state(queue_state)
  File "brian2/synapses/cythonspikequeue.pyx", line 72, in brian2.synapses.cythonspikequeue.SpikeQueue._restore_from_full_state
    if real_delays.dtype == np.float32:
  File "<stringsource>", line 189, in pair.from_py.__pyx_convert_pair_from_py_int__and_std_3a__3a_vector_3c_std_3a__3a_vector_3c_int32_t_3e____3e___
ValueError: too many values to unpack (expected 2)
```

Changing the format of Cython spike queue to use the same internal format as the Python spike queue would break backwards compatibility in an even worse way, since most users have been using the Cython spike queue.  Therefore, I think the cleanest way is probably to special case this situation here: https://github.com/brian-team/brian2/blob/31f945d4e4f03ae5c9513f06efd596f8f97af500/brian2/synapses/synapses.py#L452 and manually convert the state from the "Python format" to the "Cython format" (which is luckily quite straightforward to detect, since it consists of a 3-tuple instead of a 2-tuple).

Ping @Legend101Zz 😊 This is not a high priority right away, but it would be could to deal with this before the release that removes the Python fallback.

Legend101Zz:
Closes #1661 

#### Problem

Users who previously stored network states using the Python spike queue fallback (before Cython spike queue became mandatory) cannot restore these files with current Brian2 versions. This creates a backwards compatibility regression where valid saved states become unusable.

**Error when loading Python spike queue states:**
```
ValueError: too many values to unpack (expected 2)
```

**Root cause:** The two spike queue implementations use different internal state formats:
- **Python format:** `(time, spike_array, dimensions)` - stores relationships as `[delay_bin, synapse_index]` pairs
- **Cython format:** `(offset, spike_lists)` - stores spike lists per delay slot in circular buffer

Fixed now : 

Screenshot 2025-09-19 at 18 47 47

cc : @mstimberg

### DIFF
--- a/brian2/synapses/synapses.py
+++ b/brian2/synapses/synapses.py
@@ -459,9 +459,68 @@ class SynapticPathway(CodeRunner, Group):
         super()._restore_from_full_state(state)
         if self.queue is None:
             self.queue = get_device().spike_queue(self.source.start, self.source.stop)
-        self.queue._restore_from_full_state(queue_state)
+
+        if queue_state is not None:
+            converted_queue_state = self._convert_queue_state_if_needed(queue_state)
+            self.queue._restore_from_full_state(converted_queue_state)
+        else:
+            # Handle None case (empty queue)
+            self.queue._restore_from_full_state(None)
+
         # Put the spike queue state back for future restore calls
         state["_spikequeue"] = queue_state
+
+    def _convert_queue_state_if_needed(self, queue_state):
+        """
+        Convert spike queue state from Python format to Cython format if needed.
+
+        Python format: (time, spike_indices_array, queue_capacities)
+        Cython format: (offset, list_of_spike_lists)
+
+        Returns the queue_state in Cython format.
+        """
+        # Check if this is the old Python format (3-tuple vs 2-tuple) , also note this is sort of a hack
+        # As it's completely based on Python format being a 3 tuple ...
+        if isinstance(queue_state, tuple) and len(queue_state) == 3:
+            # This is Python spike queue format, need to convert
+            return self._python_to_cython_format(queue_state)
+        else:
+            return queue_state
+
+    def _python_to_cython_format(self, python_state):
+        """
+        Convert Python spike queue format to Cython format.
+
+        Python format: (current_time, spike_array, queue_sizes)
+        - current_time: float representing current simulation time
+        - spike_array: 2D array where spike_array[delay_slot][spike_index] = neuron_id
+        - queue_sizes: tuple of queue capacities
+
+        Cython format: (offset, spike_lists)
+        - offset: int representing current position in circular buffer
+        - spike_lists: list of lists, where spike_lists[i] contains spikes for time slot i
+        """
+        _, spike_array, __ = python_state
+
+        # The Cython format uses a circular buffer approach
+        # So We need to extract the actual spikes and organize them by time slots
+        max_delay_slots = spike_array.shape[0] if len(spike_array.shape) > 0 else 1
+        spike_lists = [[] for _ in range(max_delay_slots)]
+
+        # Convert the 2D spike array to lists of spikes per time slot
+        if spike_array.size > 0:
+            for delay_slot in range(spike_array.shape[0]):
+                for spike_idx in range(spike_array.shape[1]):
+                    neuron_id = spike_array[delay_slot][spike_idx]
+                    if (
+                        neuron_id >= 0
+                    ):  # Valid neuron ID (negative values often indicate empty slots)
+                        spike_lists[delay_slot].append(int(neuron_id))
+
+        # The offset in Cython format represents the current position in the circular buffer
+        # For a newly converted queue, we typically start at offset 0
+        offset = 0
+        return (offset, spike_lists)
 
     def push_spikes(self):
         # Push new events (e.g. spikes) into the queue


brian2/synapses/synapses.py
```diff
@@ -459,10 +459,69 @@ def _restore_from_full_state(self, state):
...
+            for delay_slot in range(spike_array.shape[0]):
+                for spike_idx in range(spike_array.shape[1]):
+                    neuron_id = spike_array[delay_slot][spike_idx]
+                    if (
+                        neuron_id >= 0
+                    ):  # Valid neuron ID (negative values often indicate empty slots)
+                        spike_lists[delay_slot].append(int(neuron_id))
+
+        # The offset in Cython format represents the current position in the circular buffer
+        # For a newly converted queue, we typically start at offset 0
```
Legend101Zz, line 516: @mstimberg let me know if this assumption is correct

mstimberg:
Hi @Legend101Zz. Many thanks for getting the ball rolling! Your assumption that the offset represents the current position in the circular buffer is correct – when you restore it, you can always set it to zero. I tried out what the current code does, and I don't think the reconstruction is correct as it is. Here is a little example of what the spike queue looks like/should look like in Cython that hopefully makes things a bit clearer. Imagine you have two spikes triggering two different synapses at time 0 (note that for the spike queue it does not change anything if this means that there are e.g. two neurons spiking targetting the same neuron, or one neuron spiking targetting two different neurons – all that matters is that it concerns two synapses). The first synapse has a delay of 2 time steps and the second of 4 time steps. After simulating the first time step, the spike queue will look like this:
```
(1, [[], [], [0], [], [1]])
```
This means:
|Time | synapses that should spike|
|------|----------------------------------|
| t=0 | - |
| t=1 | 0 |
| t=2 | - |
| t=3 | 1 |

At the next timestep, the offset will advance by one (and will then remove the index zero from the respective slot), etc.
As I said before, the offset is not in itself important, the exact spike queue state can be represented by:
```
(0, [[], [0], [], [1], []])
```
(The last empty list is also optional, the total length of this circular buffer is determined by the maximal delay that has been encountered, but can always grow dynamically).
Just to make the list structure clearer: if both cells had a delay of 4 time steps, the queue would look like:
```
(0, [[], [], [], [0, 1], []])
```

Now, in the Python version, you have three entries in the tuple: `(dt, spikes, X.shape)` The `X.shape` is the size of the internal data structure, but not actually needed here. `dt` is also stored elsewhere and not required here. So we only have to deal with `spikes`.  It stores exactly the same information as the representation above, but in a different way: instead of
```
(0, [[], [0], [], [1], []])
```
it stores
```
array([[1, 0],
       [3, 1]])
```
Which means: at time t=1, synapse 0 should activate, and at time t=3, synapse 1 should activate. So something like this should do the translation:
```python
new_queue = [] * np.max(spikes[:, 0])
for spike in spikes:
    new_queue[spike[0]].append(spike[1])
```
Hope that clears things up a bit :blush: !

brian2/tests/test_network.py
```diff
@@ -1159,6 +1159,62 @@ def test_store_restore():
     assert defaultclock.t == 10 * ms
     assert net.t == 10 * ms
 
+    # Now we do a regression test for backwards compatibility to test restore from Python spike queue format
```
Legend101Zz: @mstimberg let me know if this test looks good to you ... what I essentially did was create a network using the old pythonSpikequeue ( by going back in time using git😅 , honestly git still surprises me , how good of a tool it is ) ... and then ran the network there and stored that to the pickle file you'll see in the data folder now ... this test essentially just mimicks what we were doing in the same test above ... but just here we use the saved state from the pickle file

mstimberg: In general, I'm happy with this approach – it is a bit fragile due to the nature of Pickle files, but it is a test for a quite specific situation, and we will probably remove the conversion feature at some point in the future. There are two small things I'd like to change, though:
- could you split up the test function so that this new test has its own function?
- I think it is good to have the `if os.path.exists` instead of failing when the file isn't there, but to make the situation clearer, I'd use something like `pytest.skip("Cannot test restore from a pickle file with old SpikeQueue state - file does not exist.")`

Legend101Zz: makes sense ... changes done now https://github.com/brian-team/brian2/pull/1688/commits/df1d55559a4d97e3968353d45e08ccce985ebf1c , thank you :)

PR #1712, opened 2025-10-30
Use ruff

mstimberg:
Somewhat big PR due to many whitespace and a few formatting changes, but using ruff makes the pre-commit hook much faster, which is nice for local development. The linter also pointed to a few issues that were worth fixing.


thesamovar:
I don't know if you wanted me to look at this but I won't have any time for a while!

mstimberg:
No need to take a look - I would have tagged you otherwise ;) Also, this would be the most boring PR you ever reviewed, it is full of whitespace and formatting changes!

PR #1717, opened 2025-11-09
Eliminate Python overhead from synapses_create_array and synapses_create_generator templates _resize() and _update_synapse_numbers() methods

Legend101Zz:
### What was the problem?

In our larger effort to remove python indirection from cython runtime , there were some changes left for the synapses_create_array and synapses_create_generator templates. 

Every time we created new synapses, the fast Cython code had to stop and call back to Python (`_owner._resize()` and `_owner._update_synapse_numbers()`). This Python step was in charge of resizing all the synaptic arrays (like weights, delays, etc.) and updating counters like `N_incoming`. This back-and-forth between Cython and Python is super slow and was a real bottleneck.

### What's the fix?

I've moved all that logic directly into the Cython templates, just like how standalone mode already does it.

This means when you create synapses now:

* All the registered arrays get resized immediately in C++ (`varname_ptr.resize(...)`).
* The `N_incoming`, `N_outgoing`, and total `N` counters are updated right there in Cython.
* The `synapse_number` (for multisynaptic connections) is also calculated in Cython.

Note : This is still WIP as I need to verify a few cases and also do the changes ofr the synapses_create_generator template

### DIFF
--- a/brian2/synapses/synapses.py
+++ b/brian2/synapses/synapses.py
@@ -2013,18 +2013,6 @@ class Synapses(Group):
             f"'{self.target.name}', using pre-defined arrays)"
         )
 
-        needed_variables.extend(
-            [
-                "N_incoming",
-                "N_outgoing",
-                "N",
-                "N_pre",
-                "N_post",
-                "_source_offset",
-                "_target_offset",
-            ]
-        )
-
         codeobj = create_runner_codeobj(
             self,
             abstract_code,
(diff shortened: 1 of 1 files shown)


Legend101Zz:
@mstimberg thanks for the wait ... 

This PR is finalised now , please take a gander at it whenever you get the chance :) 

Some testing I did for create array template : 

Screenshot 2025-11-20 at 08 26 11

for generator template : 

Screenshot 2025-11-20 at 08 42 02

mstimberg (approved):
LGTM!

PR #1721, opened 2025-11-14
Standalone: Do not delete sourcefiles.txt with make clean

mstimberg:
This file is needed for making `main.exe`


mstimberg:
See https://brian.discourse.group/t/project-compilation-failed-error-code-2/1507/4

Issue #1214, opened 2020-08-10
Google colab links in docs


thesamovar:
We currently have Binder links, but it would be good to have Colab links too.

Installing dependencies quietly (could be auto-deleted from Binder and docs versions):

```
%%capture
!pip install brian2 
```

How would we handle examples? At the moment on Binder they're generated from Python files by Docker build script. I guess we could move everything over to ipynb instead of our current Python system, which could have some other advantages (some nice formatting). Not sure that's warranted though.

shriya7ise:
Hello Sir ,
I'm Shriya Ise, this is my first contribution to Brian 2 and I hope my suggestion proves helpful.

We can retain examples in .py format for consistency while dynamically generating 'Open in Colab' links in the documentation. Additionally, we can modify the examples to detect Colab and handle dependencies automatically. 
Would this approach align with the project's goals? Looking forward to your feedback.

nishantraghuvanshi:
Hi @thesamovar!
I'd like to work on this issue as my first contribution to Brian2. I noticed that there have not been any PRs related to this issue after the previous comment, and the issue is still open.

Background

I'm Nishant, a student at BITS Pilani Goa, interested in computational neuroscience and machine learning. I have experience with Jupyter notebooks, cloud platforms (Colab, AWS, Azure), and Python frameworks (PyTorch, TensorFlow). I'm exploring Brian2 for GSoC 2026.

## Proposed Approach

I propose adding Google Colab integration alongside the existing Binder setup:

### Changes:
1. Add Colab badges to README.md and docs_sphinx/index.rst (next to Binder badges)
2. Update the 3 tutorial notebooks in /tutorials/ with:
  - Colab environment detection
  - Auto-installation of Brian2 when running in Colab

### Implementation:
- Use direct GitHub→Colab linking (no separate repo needed like brian2-binder)
- Focus on existing tutorial notebooks (no conversion of examples)
- Keep .py examples as-is 

### Files to modify:
- README.md - Add Colab badge in Quickstart section
- docs_sphinx/index.rst - Add Colab badge alongside Binder
- tutorials/*.ipynb - Add Colab detection cell

### Questions

1. Should the main Colab badge point to Tutorial 1, or create an index notebook?
2. Should this PR convert examples to notebooks, or keep them as .py files?
3. Any Colab-specific features/limitations I should know about?

I can have a PR ready by January 23. Would appreciate any guidance!

Issue #1219, opened 2020-08-27
Handling of += expressions by Equations()


smalltimer:
Hi, 

While defining equations using Equations() can be very useful when defining models which can then be fine-tuned by substituting variables or values, this method does not seem to work for on_pre and on_post events. If there is a variable which is only used during the events, it will not feature in the model equation (or can be included there as a declaration like 'x : 1'). However, such a declaration cannot be substituted using Equations(), thus limiting the use of Equations() in on_pre and on_post to define flexible models which can be adjusted for various synaptic groups.

Any workarounds/tips would be really helpful. I am not sure if this is strictly an issue, but I think it does limit the use of the very handy Equations() structure. 

Cheers

thesamovar:
Interesting idea. The same would be true of threshold, reset in ``NeuronGroup``. Having the same substitution options for these would be nice, and not a complicated feature to implement. Thanks for the suggestion. I don't suppose you'd like to implement it would you? We'd be very happy to help.

thesamovar:
The first thing to do would be to decide the syntax. I suspect ``Equations`` might not be easy to adapt to work with this because it's really specific to differential equations. We could have a new object (e.g. ``Statements``) with the same syntax for substitutions, or try to add in sequences of statements as an additional option for ``Equations``. Any thoughts @mstimberg ?

smalltimer:
That was fast, thanks @thesamovar ! I am afraid I have only been working from the wiki as a reference, and have not explored the Equations() code. I am writing some models in collaboration with a group of experimentalists, so making the code as easy to tune as possible for non-programmers was what led me to raise the issue.

thesamovar:
No worries! It's a really interesting and useful idea, and we should have done this from the beginning actually.

Maybe the simplest thing is to just have a nice function that just does the same clever substitution stuff as Equations does.

smalltimer:
I may have found a hack-y solution which might be helpful in the meanwhile. Assuming an instantaneous synapse with scaled updating, the following does not work. 
```
m = '''dg/dt = -g/tau : siemens'''
m_onpre = '''g += k*w'''
S = Synapses( N_pre, N_post, model =  Equations(m, tau = 3*ms), on_pre = Equations(m_onpre, k = 0.3) )
```

A possible work-around is to do this is to include ALL parameters in the model, even the ones which are strictly event-driven. 
```
m = '''
dg/dt = -g/tau : siemens
k = k_parameter : 1
'''
[...]

PR #1754, opened 2026-01-19
Add Google Colab integration to documentation and tutorials

nishantraghuvanshi:
This commit adds Google Colab support alongside existing Binder integration,
  making it easier for users to try Brian2 in the cloud without local installation.

  Changes:
  - Added Google Colab badge to README.md in Quickstart section
  - Added Google Colab badge to docs_sphinx/index.rst alongside Binder
  - Updated text to mention both Binder and Google Colab services
  - Added Colab environment detection to all 3 tutorial notebooks
  - Notebooks now auto-install Brian2 when running in Google Colab

  The implementation uses direct GitHub→Colab linking (no separate repository
  needed like brian2-binder). Tutorial notebooks detect the Colab environment
  and automatically install Brian2, providing a seamless user experience.

  Addresses #1214


mstimberg:
Hi @nishantraghuvanshi, thanks for your work on this! I need to have a closer look at this, but I think I'd prefer having this a bit more automized. In our current documentation, we have links to open the three tutorials on binder.org: https://brian2.readthedocs.io/en/stable/resources/tutorials/index.html and we also add an automatic banner to open each example, see e.g. https://brian2.readthedocs.io/en/stable/examples/COBAHH.html
With your approach, how would you actually open e.g. Tutorial 2 in collab?

The advantage of a separate repository is that we can add the header with the `!pip install brian2` automatically instead of including it manually in every notebook. And this part could also be added when we convert the examples from `.py` to `.ipynb`. There might also be existing things we can reuse, see e.g. https://sphinx-book-theme.readthedocs.io/en/stable/content/launch.html and the discussion here: https://github.com/pydata/pydata-sphinx-theme/discussions/1422
But as I said, I need to think about this a bit more, I don't have the perfect solution in my sleeve :blush:

nishantraghuvanshi:
Hi @mstimberg, thanks a lot for the detailed feedback. You bring up some great points, so I gave this a bit more thought. Here's what I think could work well:

## 1. Proposed solution: two-part automation

My current idea is to automate both the badges and the installation step.

### Part 1: automated badges (via Sphinx)
- Use either `sphinx-book-theme` or the PyData theme launch buttons.
- Configure this once in `conf.py` so Colab badges are generated automatically for all tutorial and example pages. That way, users can open any notebook in Colab directly from the docs.

### Part 2: automated installation (separate repository)
- Create a `brian2-colab` repository, similar to `brian2-binder`
- Set up a small pipeline that:
  1. pulls notebooks from the main `brian2` repo.  
  2. injects a `!pip install brian2` cell at the top.
  3. pushes the result to `brian2-colab`.
- The Colab badges would then point to this repo instead of the main one.

## 2. Alternative: build-time injection

If maintaining a separate repository feels like too much overhead, I could instead look into injecting the installation cell during the existing `.py → .ipynb` conversion or as part of the docs build process. That said, the separate-repo approach feels quite clean and consistent with how `brian2-binder` already works.

## 3. A few questions before I proceed

1. **Separate repo:** would you be open to creating a `brian2-colab` repository? I’m happy to set it up and handle the automation.
2. **Sphinx integration:** would you prefer `sphinx-book-theme`, or should I try to integrate this with the current Brian2 theme?
3. **Scope for this PR:** would you rather I
   - implement the full solution (separate repo + Sphinx badges), or  
   - start with a minimal version and iterate from there?

I’m flexible and happy to go with whatever fits best into Brian2's workflow, just wanted to run this by you first.

Thanks again for taking the time to review this!

P.S. I'm also working on issue #1219 (Statements substitution), please do give your suggestions on that if you have the time. Thanks

mstimberg:
Hi again, sorry for not being more reactive on this issue. I'm still a bit hesitant, since we are thinking about overhauling the documentation system more globally (see e.g. #610) , and I don't want you to work on something that becomes obsolete soon. Having a separate repository is certainly a possibility, but it would be a bit different from the brian2-binder repository: for collab, we'd actually need the notebooks in the repo, whereas the brian2-binder repository only creates the notebooks as part of the [`postBuild`](https://repo2docker.readthedocs.io/en/latest/configuration/actions/#postbuild-run-code-after-installing-the-environment). Sorry, I still need to think about ithe best approach for a bit…

PR #1768, opened 2026-02-10
Implement substitution functionality for Statements class

### LINKED ISSUE #1219: Handling of += expressions by Equations()

nishantraghuvanshi:
- Add __init__ method with **substitutions parameter to Statements class
- Implement _substitute static method for identifier replacement
- Support both name substitution (string -> string) and value substitution (identifier -> value)
- Use regex with word boundaries to respect identifier boundaries
- Add comprehensive test suite with 22 tests covering:
  - Basic value and name substitution
  - Multiple simultaneous substitutions
  - Word boundary handling
  - Brian2 units support
  - Multi-line statements
  - Integration with Synapses and NeuronGroup
  - Backward compatibility
- All existing tests pass (no regression)

This enables users to create reusable statement templates for on_pre, on_post, threshold, and reset parameters, similar to how Equations supports substitution for differential equations.

Fixes #1219

### DIFF
--- a/brian2/equations/codestrings.py
+++ b/brian2/equations/codestrings.py
@@ -109,13 +109,11 @@ class Statements(CodeString):
         for identifier, replacement in substitutions.items():
             if isinstance(replacement, str):
                 # Replace identifier with another identifier
-                new_code = re.sub(r'\b' + identifier + r'\b', replacement, new_code)
+                new_code = re.sub(r"\b" + identifier + r"\b", replacement, new_code)
             else:
                 # Replace identifier with a value
                 new_code = re.sub(
-                    r'\b' + identifier + r'\b',
-                    '(' + repr(replacement) + ')',
-                    new_code
+                    r"\b" + identifier + r"\b", "(" + repr(replacement) + ")", new_code
                 )
 
         return new_code
--- a/brian2/equations/codestrings.py
+++ b/brian2/equations/codestrings.py
@@ -74,11 +74,11 @@ class Statements(CodeString):
     >>> Statements('g += w')
     Statements('g += w')
     >>> Statements('g += k*w', k=0.3)
-    Statements('g += 0.3*w')
+    Statements('g += (0.3)*w')
     >>> Statements('g += k*w', g='g_ampa')
     Statements('g_ampa += k*w')
     >>> Statements('g += k*w', g='g_ampa', k=0.3)
-    Statements('g_ampa += 0.3*w')
+    Statements('g_ampa += (0.3)*w')
     """
 
     def __init__(self, code, **substitutions):
--- a/brian2/equations/codestrings.py
+++ b/brian2/equations/codestrings.py
@@ -58,23 +58,67 @@ class CodeString(Hashable):
 
 class Statements(CodeString):
     """
-    Class for representing statements.
+    Class for representing statements with support for substitution.
 
     Parameters
     ----------
     code : str
         The statement or statements. Several statements can be given as a
         multi-line string or separated by semicolons.
+    **substitutions
+        Substitutions to apply to the code. Can be either strings (to replace
+        a name with another name) or values (to replace a name with a value).
 
-    Notes
-    -----
-    Currently, the implementation of this class does not add anything to
-    `~brian2.equations.codestrings.CodeString`, but it should be used instead
-    of that class for clarity and to allow for future functionality that is
-    only relevant to statements and not to expressions.
+    Examples
+    --------
+    >>> Statements('g += w')
+    Statements('g += w')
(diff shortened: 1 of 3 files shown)


mstimberg:
@nishantraghuvanshi Thank you for the PR! I did not yet have time to look at it in detail, but I will so next week.

nishantraghuvanshi:
@mstimberg Do you think I should make any more changes to this? Thanks for the review.

brian2/equations/codestrings.py
```diff
@@ -58,23 +58,65 @@ def __hash__(self):
...
+            The original code string
+        substitutions : dict
+            Dictionary mapping identifiers to replacements (strings or values)
+
+        Returns
+        -------
+        new_code : str
+            Code with substitutions applied
+        """
+        import re
```
mstimberg: Very minor point, but this move to the imports at the top of the file (we sometimes have to import things inside functions to avoid circular imports, but this isn't the case here).

mstimberg (changes requested):
Thanks for the PR, and apologies for taking so long to look into it. The code looks good to me in general (made a minor point about an import as a comment), but there is one slightly tricky issue that isn't taken care of: in contrast to `Equations`, the `Statements` class will happily accept a substitution that leads to incorrect code:
```pycon
>>> Statements("v += x : volt", v=3*mV)
Statements('(3. * mvolt) += x : volt') 
```
This is not a *huge* issue (after all, it will lead to an error later), but for consistency with `Equations`, it would be nice to handle this with a nice error message. Probably some ad-hoc handling of this is OK, otherwise we could also use the `parse_statement` function that is used later in the code generation process. Related to this, in principle a statement can also include a comment, and *probably*, replacements should not affect the comment. I'm thinking of examples like this:
```pycon
>>> Statements("x += weight # Use a small weight", weight=1*nS)
Statements('x += (1. * nsiemens) # Use a small (1. * nsiemens)')
```
On the other hand, in this example the substitution makes perfect sense:
```pycon
>>> Statements("x += weight # x is the post-synaptic target variable", x="y")
Statements('y += weight # y is the post-synaptic target variable')
```
Not quite sure what the best approach here is, maybe defaulting to doing substitutions also in the comment part for names but not for values? And having an option to change this behaviour? I'm worried that we are making things too complicated…

There are also a few more things to do before this feature becomes really useful to users, but I'd also be happy to merge the PR before that (given its narrowly scoped title):
- the same feature should also be added to `Expression` (for threshold conditions, synaptic connections, etc.)
- finally, neither `Statements` nor `Expression` can currently be used in the places they are meant for. E.g. the original issue was about using the substitution mechanism to provide a statement to `on_pre` of `Synapses`, but it currently does not accept a `Statements` object (actually, it silently ignores everything that is not a string or a mapping, but that's a bug 😬 ). So this should be handled in the same way that a `NeuronGroup` accepts either a string or an `Equations` object.

Related to the last point: this PR's description states that the test suite covers "Integration with Synapses and NeuronGroup", but this isn't the case. Also, it mentions that users could use the functionality for "on_pre, on_post, threshold, and reset parameters", but apart from that not working yet, `threshold` does not work with statements but with expressions.

brian2/equations/codestrings.py
nishantraghuvanshi: Hey. I implemented this. Thanks for the review. 
Please tell me if there are any more changes that need to be implemented.

mstimberg:
Hi @nishantraghuvanshi, could you please fix the source code formatting so that the test suite can run? You can install the `pre-commit` hook to fix such issues before committing: https://brian2.readthedocs.io/en/stable/developer/guidelines/style.html#code-style

nishantraghuvanshi:
@mstimberg I implemented the fix in formatting. Please check it out.

mstimberg:
Hi @nishantraghuvanshi are you still interested in/have time for working on this? Note that the PR is still stuck on a code formatting issue – to avoid this, please install the pre-commit tool locally as detailed in the developer documentation.

nishantraghuvanshi:
Hi @mstimberg 
Yes, I am interested in working on this. Just could not get the time. 
Will fix the code formatting issue over the weekend.

PR #1769, opened 2026-02-13
Experimental cppyy

Legend101Zz:
### Problem

The current Brian2 code generation pipeline suffers from a fundamental performance bottleneck. The issue is not tied to the specific tools we use, but rather to the **Ahead-of-Time (AOT) compilation paradigm** itself.

Regardless of whether we use Cython (our current approach) or manual C-extensions, the workflow remains slow and cumbersome:

* Generate large C++ source files on disk
* Invoke an external compiler (e.g., g++, clang) with significant overhead
* Wait for compilation to complete (often 15–40 seconds, which disrupts interactivity)
* Dynamically load the compiled result through a complex process

In other words, the bottleneck lies in the **file-based, external-compiler, AOT workflow**.

---

### Proposed Solution: JIT Compilation with `cppyy`

This PR introduces `cppyy` as a new runtime code generation target, shifting from AOT to **Just-in-Time (JIT) compilation**.

With `cppyy`, C++ code is compiled in-memory using the Cling C++ interpreter, which eliminates:

* File I/O overhead
* External compiler process spawning
* Long compilation waiting times
* Complex dynamic loading procedures

---

### Current Status

* End-to-end JIT compilation pipeline
* Basic neuron group simulations
* State updates, thresholds, and resets
* Template system for different operations
* Integration with the device layer

---

### Next Steps

Fix for dynamic arrays and spikequeue and synapses
---


Legend101Zz:
@mstimberg the initial PR I'll keep working on this branch itself and finally make the synapses too working on this , cheers :)

Legend101Zz:
@mstimberg I added a introspection to do the stuff we were discussing yesterday , like to view the cppy code and codeobjects and even change it over the fly , I have added an attached jupyter sample for reference which you can use and play around with :)

Attaching a few screenshots of how it looks :

Screenshot 2026-02-14 at 16 01 02
Screenshot 2026-02-14 at 16 01 23

brian2/codegen/runtime/cppyy_rt/cppyy_rt.py
```diff
@@ -0,0 +1,636 @@
...
+                self.namespace[f"_num{name}"] = var.get_len()
+
+                # Scalar constants also get a plain-name entry with the unwrapped value
+                if var.scalar and var.constant:
+                    self.namespace[name] = value.item()
+            else:
+                self.namespace[name] = value
+
+            # ── Dynamic arrays: store BOTH the data view AND the capsule ──
+            # The data view (_ptr_array_*) gives C++ direct pointer access
```
Legend101Zz, line 715: So I realised something while coding this , the way we coded dynamic arrays in runtime to be available for runtime mode , we'll have to change that for cppyy as `RuntimeDevice` currently creates its dynamic arrays through Cython wrappers. Those Cython wrappers own the underlying `DynamicArray1D<T>*`. If we want to go fully cppyy-native  or have multiple backends only , we'll need to change how the RuntimeDevice stores dynamic arrays in cppyy codegen backend — either replacing the Cython wrappers with cppyy-managed objects, as that is what would be best ...

Legend101Zz:
here is a snippet I used to test things :

```python
import time

import numpy as np

from brian2 import *

prefs.codegen.target = 'cppyy'
prefs.codegen.runtime.cppyy.enable_introspection = True # Enable introspection

# prefs.codegen.target = 'cython'
# prefs.codegen.runtime.cython.cache_dir = 'cythontmp/'
# prefs.codegen.runtime.cython.delete_source_files = False



# Hodgkin-Huxley neuron model

num_neurons = 100
duration = 500*ms

# Parameters
area = 20000*umetre**2
Cm = 1*ufarad*cm**-2 * area
gl = 5e-5*siemens*cm**-2 * area
El = -65*mV
EK = -90*mV
ENa = 50*mV
g_na = 100*msiemens*cm**-2 * area
g_kd = 30*msiemens*cm**-2 * area
VT = -63*mV

eqs = Equations('''
dv/dt = (gl*(El-v) - g_na*(m*m*m)*h*(v-ENa) - g_kd*(n*n*n*n)*(v-EK) + I)/Cm : volt
dm/dt = 0.32*(mV**-1)*4*mV/exprel((13.*mV-v+VT)/(4*mV))/ms*(1-m)-0.28*(mV**-1)*5*mV/exprel((v-VT-40.*mV)/(5*mV))/ms*m : 1
dn/dt = 0.032*(mV**-1)*5*mV/exprel((15.*mV-v+VT)/(5*mV))/ms*(1.-n)-.5*exp((10.*mV-v+VT)/(40.*mV))/ms*n : 1
dh/dt = 0.128*exp((17.*mV-v+VT)/(18.*mV))/ms*(1.-h)-4./(1+exp((40.*mV-v+VT)/(5.*mV)))/ms*h : 1
I : amp
''')

group = NeuronGroup(num_neurons, eqs,
                    threshold='v > -40*mV',
                    refractory='v > -40*mV',
                    method='exponential_euler')
group.v = El
group.I = '0.7*nA * i / num_neurons'


# SpikeMonitor: records spike times and indices (dynamic arrays)
spike_mon = SpikeMonitor(group)

# StateMonitor: records v for a few neurons every timestep (2D dynamic array)
state_mon = StateMonitor(group, 'v', record=[0, 25, 50, 75, 99])


print(f"Running {num_neurons} HH neurons for {duration}...")
t_start = time.perf_counter()
run(duration)
t_elapsed = time.perf_counter() - t_start
print(f"Done in {t_elapsed:.2f}s")

print(f"\nTotal spikes: {spike_mon.num_spikes}")
print(f"StateMonitor recorded {state_mon.t.shape[0]} timesteps "
      f"for {len(state_mon.record)} neurons")

# --- Now use the introspector ---
from brian2.codegen.runtime.cppyy_rt.introspector import get_introspector

intro = get_introspector()

# ---- 1. List all compiled code objects ----
print("=" * 60)
print("LIST OBJECTS")
print("=" * 60)
print(intro.list_objects())

# ---- 2. Inspect the state updater ----
print("\n" + "=" * 60)
print("INSPECT STATE UPDATER")
print("=" * 60)
# Using glob pattern — "stateupdater*" matches the full name
print(intro.inspect("*stateupdater*"))

# ---- 3. View just the params ----
print("\n" + "=" * 60)
print("PARAMS")
print("=" * 60)
print(intro.params("*stateupdater*"))

# ---- 4. View the namespace ----
print("\n" + "=" * 60)
print("NAMESPACE")
print("=" * 60)
print(intro.namespace("*stateupdater*"))

# ---- 5. View C++ globals ----
print("\n" + "=" * 60)
print("C++ GLOBALS")
print("=" * 60)
print(intro.cpp_globals())

# ---- 6. Evaluate a C++ expression ----
print("\n" + "=" * 60)
print("EVAL C++")
print("=" * 60)
print(f"M_PI = {intro.eval_cpp('M_PI')}")
print(f"sizeof(double) = {intro.eval_cpp('sizeof(double)', 'size_t')}")
print(f"_brian_mod(7, 3) = {intro.eval_cpp('_brian_mod(7, 3)', 'int32_t')}")

```

Legend101Zz:
Hi @mstimberg 

A follow-up , after our last sync on the `run_block` hot path, I kept pulling on that thread and the result turned out much better than I expected — I want to walk you through what I did, how I measured it, and what came out.

### Headline

The cppyy backend is now **faster than Cython on the simulation loop itself** for the three realistic workloads we've been measuring, while keeping its existing big lead on the first-time build-and-compile cost.

### What the problem was

You'll remember the breakdown we walked through and discussed on last meeting , on every timestep, for every code object, `run_block()` was paying four kinds of cost:

- **(A)** a `logger.diagnostic(...)` call that ran `inspect.stack()` and `logging.getLogger().log(...)` even when diagnostics were filtered out,
- **(B)** an unconditional `np.ascontiguousarray(val)` on every ndarray parameter — even though Brian2 arrays are already C-contiguous in practice,
- **(C)** re-allocating `args = []` and re-walking `param_mapping` with a `dict.get` per parameter on every call,
- **(D)** the cppyy dispatch itself, which extracts the buffer pointer from each ndarray and calls the C++ function.

We agreed (A), (B), (C) were Python-level overhead that could be safely reduced, while (D) was the cppyy floor.

### What I changed

**First commit (`dd216629`)** is the cleanup you and I discussed:

- Removed the per-call `logger.diagnostic`. This was the single biggest contributor — even filtered out, the call cost ~6 µs of logger plumbing per invocation.
- Guarded `np.ascontiguousarray` behind `not val.flags.c_contiguous`. The fallback still handles the rare non-contiguous case (e.g. some subgroup sliced views), but the common case is now a single flag check.
- Cached the normalized args tuple per block. The cache is invalidated from `update_namespace()` only when `nonconstant_values` is non-empty — i.e. only when a dynamic-array resize could have changed an object reference. For static-namespace code objects the cache is built once and reused forever.

That commit alone roughly halved the per-timestep simulation cost. But once those obvious bits were gone, profiling showed the remaining `run_block` call was still ~3 µs, and three things I hadn't expected were dominating it:

1. `run_block` was doing three `template_name == "..."` string compares **on every call** to decide which return-value path to take, even though `template_name` is fixed at construction time.
2. Even with the args tuple cached, `run_block` was still doing a `self.compiled_code.get(block)` lookup, then a `self._cached_args.get(block)` lookup, then a fallback build path — two dict lookups and a branch for what could be a single load.
3. When a workflow calls `Network.run()` more than once in the same Python process (parameter sweeps, store/restore, optimizer loops — common in user notebooks), every iteration was paying the full `cppyy.cppdef()` cost again for code whose source was byte-identical to a previous compile.

**Second commit (`e7f65bc6`)** addresses those three:

- **Eager per-block dispatch binding.** At the end of `compile_block()`, for code objects whose namespace has no `nonconstant_values` (the common case), I now eagerly compute the args tuple and store `(compiled_func, args)` in `self._fast_dispatch[block]`. The top of `run_block()` then becomes a single dict lookup and a tuple-unpacking call:

  ```python
  fast = self._fast_dispatch.get(block)
  if fast is not None and self._return_kind is None:
      func, args = fast
      func(*args)
      return
  ```

  The original lazy build path is preserved for the minority of code objects that need it (those with dynamic-array references, or those that return a value).

- **Precomputed return-kind.** The three `template_name == "group_get_indices" / "group_variable_get" / "group_variable_get_conditional"` compares are folded into a single `self._return_kind` attribute set once in `__init__`.

- **Process-level Cling compile cache.** A module-level dict keyed on `sha256(canonical_source)` is consulted before `cppyy.cppdef()`. On a hit, we reuse the JIT-compiled function proxy from the previous compile. `_set_user_func_globals` is still called per-codeobject on every miss *and* every hit, so each code object continues to see its own TimedArray / Binomial globals — which is what makes the reuse correct.

I also tried `compiled_func.__overload__('signature')` pre-resolution to skip cppyy's per-call overload lookup. On a tight microbench it saves ~13 ns/call; on the real workloads the delta disappears into noise. I removed it before committing.

### How I measured this

- **Benchmark harness.** Three workloads kept the same throughout: `small_lif` (2 000 neurons, 200 ms simulated), `coba` (4 000 neurons, 500 ms simulated), `kremer3` (Kremer barrel-cortex network, 500 ms simulated). Each measurement runs in a fresh subprocess so the build-and-compile cost and the simulation-loop cost are cleanly separated. Every reported number is the median of at least 8 samples.
- **Fair comparison flags.** `EXTRA_CLING_ARGS=" -O2"` set for cppyy, so its compiled code uses the same optimisation level Cython's extension modules get. The Cython on-disk cache is wiped before any first-run measurement so we're never comparing a re-used artefact against a fresh build.


Environment: macOS arm64, Python 3.13, cppyy 3.5.0.

### Numbers

#### Simulation-loop time (the per-timestep cost — the metric we'd been chasing)

| Workload | Cython | cppyy (before this work) | cppyy (now) | cppyy / Cython |
|---|---:|---:|---:|---:|
| small_lif — 2 000 neurons, 200 ms | 0.038 s | 0.104 s (1.93× slower) | **0.039 s** | **1.02× — parity** |
| coba — 4 000 neurons, 500 ms | 0.097 s | 0.283 s (2.14× slower) | **0.089 s** | **0.92× — cppyy 8 % faster** |
| kremer3 barrel cortex — 500 ms | 0.343 s | 0.681 s (1.79× slower) | **0.302 s** | **0.88× — cppyy 12 % faster** |

The stage-by-stage journey on the same metric:

| Stage | small_lif | coba | kremer3 |
|---|---:|---:|---:|
| Baseline (before any of this work) | 0.104 s | 0.283 s | 0.681 s |
| After the `run_block` cleanup (commit `dd216629`) | 0.058 s | 0.116 s | 0.369 s |
| After the dispatch + cache changes (commit `e7f65bc6`) | **0.039 s** | **0.089 s** | **0.302 s** |
| Total reduction vs baseline | −63 % | −69 % | −56 % |

#### First-time build-and-compile (the existing cppyy lead — unchanged)

This is the cost of building a brand-new network and getting it ready to simulate in a fresh Python process. cppyy was already much faster here, and nothing in this work made it worse.

| Workload | Cython first-time setup | cppyy first-time setup | cppyy advantage |
|---|---:|---:|---:|
| small_lif | 3.66 s | 0.09 s | 41× faster |
| coba | 4.30 s | 0.10 s | 43× faster |
| kremer3 barrel cortex | 8.31 s | 0.32 s | 26× faster |

#### Repeated-`run()` pattern (where the compile cache pays off)

The kind of workflow where a user calls `Network.run()` multiple times on the same network — parameter sweeps, store/restore, optimizer loops, anything in a Jupyter notebook that re-runs the same simulation cell. Five iterations of the same 2 000-neuron COBA network in one Python process, stable group names so the generated source is byte-identical across iterations:

| Iteration | Cython (build + setup + sim) | cppyy (build + setup + sim) |
|---:|---:|---:|
| 0 (initial) | 6.31 + 8.23 + 0.15 = **14.69 s** | 0.17 + 0.10 + 0.08 = **0.35 s** |
| 1 | 0.02 + 0.09 + 0.10 = 0.20 s | 0.02 + 0.03 + 0.04 = 0.09 s |
| 2 | 0.02 + 0.08 + 0.10 = 0.20 s | 0.01 + 0.03 + 0.04 = 0.08 s |
| 3 | 0.02 + 0.08 + 0.09 = 0.19 s | 0.01 + 0.03 + 0.04 = 0.08 s |
| 4 | 0.02 + 0.08 + 0.10 = 0.19 s | 0.01 + 0.03 + 0.04 = 0.09 s |
| **5-iteration total** | **15.47 s** | **0.69 s — 22× faster end-to-end** |

The compile cache turns iterations 1–4 of cppyy into near-no-ops for the build/setup phases, while Cython still pays its small per-iteration overhead (and a much larger first-iteration cost).


 **Anything else you'd want to see before this lands?** Different hardware ? A specific stress test?


Thanks for steering this , cppyy is now looking more hopeful :)

Legend101Zz:
warm_sim_progression
warm_ratio_progression
cold_total_comparison

Sharing the benchmark results

cppyy_optimization_work/scripts/README.md
```diff
@@ -0,0 +1,275 @@
+# cppyy Optimization Benchmark Scripts
```
Legend101Zz, line 1: @mstimberg please find these scripts I used to benchmark , cython and cppyy codegens , please let me know how they play out in your system :)

brian2/codegen/runtime/cppyy_rt/cppyy_rt.py
Legend101Zz: note this is old comment

PR #1789, opened 2026-03-12
Add prefix() and postfix() methods to Equations class

nishantraghuvanshi:
This PR fixes issue #1050, adding `prefix()` and `postfix()` methods to the `Equations` class. 
These methods allow users to rename all variable names in a set of equations, which is useful for avoiding name conflicts in multi-population models.

## Example
```python
eqs = Equations('''
    dv/dt = -v/tau : volt
    I = g*E : amp
''')

exc_eqs = eqs.prefix('exc_')  # Variables become: exc_v, exc_I
inh_eqs = eqs.postfix('_inh')  # Variables become: v_inh, I_inh
```

## Key Design Decisions

- **Recursive behavior**: Methods apply to current variable names, so `eqs.prefix('a_').prefix('b_')` produces `b_a_v` (nested)
- **Comprehensive validation**: Full Python identifier validation with keyword checks
- **Built-in protection**: Variables like `t`, `dt`, `xi`, `i`, `N` are never renamed
- **Scope**: Only explicitly defined variables are renamed (diff eqs, parameters, subexpressions), not external references

Please suggest any changes that I need to make in the above implementation.
@thesamovar  @mstimberg

### DIFF
--- a/brian2/equations/equations.py
+++ b/brian2/equations/equations.py
@@ -737,6 +737,276 @@ class Equations(Hashable, Mapping):
     def substitute(self, **kwds):
         return Equations(list(self._substitute(kwds).values()))
 
+    def prefix(self, prefix_str):
+        """
+        Return a copy of the equations with all variable names prefixed.
+
+        Parameters
+        ----------
+        prefix_str : str
+            String to prepend to variable names
+
+        Returns
+        -------
+        Equations
+            New Equations object with prefixed variables
+
+        Raises
+        ------
+        ValueError
+            If prefix_str is not a valid Python identifier
+
+        Notes
+        -----
+        This method prefixes all user-defined variables (differential equations,
+        parameters, subexpressions) while protecting built-in variables (t, dt, xi,
+        i, N, etc.) and external namespace references.
+
+        The prefixing is recursive, so `eqs.prefix('a_').prefix('b_')` would
+        result in variables prefixed with `b_a_` (nested).
+
+        Examples
+        --------
+        >>> eqs = Equations('''
+        ...     dv/dt = -v/tau : volt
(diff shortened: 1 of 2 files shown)


mstimberg:
Hi @nishantraghuvanshi, thank you for the PR. This issue is a good way of getting familiar with Brian's code and our contribution process, but please be aware that we most likely won't merge it. We should have probably closed #1050, since our discussions since then have converged onto a more general solution (see #1244). So, I'd be happy to continue to work with you on this PR, but I also understand if you want to focus on things that will actually make it into Brian's codebase.

nishantraghuvanshi:
Thanks for the transparency, @mstimberg  
I really appreciate you pointing me toward #1244.

I've started reading through the issue thread (it's quite long, so I'm still digesting everything), but I can already see the templating approach with `{...}` syntax is much more flexible than the simple prefix/postfix methods.

A few questions as I'm learning about this:

1. From what I've read so far, it looks like the template approach can handle the prefix/suffix use case (by doing things like `{name}_{suffix}`). Is that right, or am I understanding it incorrectly?

2. How far along is #1244? Is there a specific area where I can help?

I'm definitely interested in contributing to something that's going into Brian2. Whether that means pivoting to help with #1244 or continuing with this PR as a learning exercise to better understand the codebase, I'm excited to contribute wherever makes the most sense. Would love your guidance on the best path forward.

mstimberg:
Hi @nishantraghuvanshi. I will close this PR, but if you are still interested in contributing to #1244, please make a comment over in that PR. I haven't looked into that branch in quite some time, so I'm not exactly sure how much work it still needs (also, for now, it does not apply cleanly to master).

Issue #1129, opened 2019-11-06
Improve test suite


mstimberg:
After switching to pytests (#1127), we can potentially improve our test suite in various ways, to make it more useful during development:

* [ ] Use [`pytest.mark.parametrize`](https://docs.pytest.org/en/latest/parametrize.html#pytest-mark-parametrize-parametrizing-test-functions) to parametrize tests instead of having loops within the test function. The advantage is that we see all failed test cases, not only the first one.
* [ ] Currently, we call `pytest` several times during a run of `brian2.tests` (to run for each of the code generation targets, etc.) – investigate whether it is possible to avoid this. e.g. by having custom decorators that wrap around `pytest.mark.parametrize` and fixtures and run the same test for all codegen targets. This would make it possible to fully make use of `pytest` features (e.g. re-run only the tests that failed) and lead to better integration in IDEs.

PR #1790, opened 2026-03-12
Docs: Update developer workflow to prioritize Devcontainers and local setup using uv

tharun-mk:
->Added the standard command `uv add --dev --editable .[docs,test]` so developers can resolve dependencies directly from the existing `pyproject.toml` without needing to maintain redundant dependency files.
->ocal setup instructions with a modern `uv` workflow.
->Restructured the environment setup guide to prominently feature the VS Code Devcontainer approach as the #1 recommended method for new contributors.

### DIFF
--- a/docs_sphinx/developer/index.rst
+++ b/docs_sphinx/developer/index.rst
@@ -5,12 +5,41 @@ This section is intended as a guide to how Brian functions internally for
 people developing Brian itself, or extensions to Brian. It may also be of some
 interest to others wishing to better understand how Brian works internally.
 
-If you use `VS code <https://code.visualstudio.com/>`_ as your development
-environment, it will offer to automatically build a Brian development
-`Docker <https://www.docker.com/>`_ container when you open the repository,
-with all the required dependencies installed and configured. Further
-`documentation <https://github.com/brian-team/brian2/blob/master/.devcontainer/README.md>`_
-for this approach can be found in the ``.devcontainer`` directory.
+Setting up a development environment
+-------------------------------------
+
+Option 1: Devcontainers (Highly Recommended)
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+The easiest and most reliable way to set up a development environment is to use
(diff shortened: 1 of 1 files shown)


mstimberg:
Hi @tharun-mk, thanks for the PR. The doc looks mostly good, even though the "``uv`` resolves all dependencies directly from the existing ``pyproject.toml`` file, keeping the repository clean without needing external dependency files" is not very meaningful, I think. Regarding the `uv add...` command: the idea was not to include this command in the documentation, but to use it to update the `pyproject.toml` file. The documentation would have to state how to actually run the code in the environment created by `uv`.

tharun-mk:
Hi @mstimberg,
I have rewritten the "Option 2" section to use uv venv and uv pip install so it safely installs the existing dependencies without modifying the pyproject.toml. I also added a step showing how to execute the test suite within that environment.
Here is the updated text for this section. Let me know if this looks good to you and is it okay push commit!
Option 2: Local Setup using ``uv``
~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~

If you prefer a local setup without Docker, you can use
`uv <https://docs.astral.sh/uv/>`_ to create a virtual environment and install Brian in development mode.

1. `Install uv <https://docs.astral.sh/uv/getting-started/installation/>`_ if
   you haven't already.
2. From the root of the repository, create a virtual environment and install the package along with its testing and documentation dependencies:

   .. code-block:: bash

      uv venv
      uv pip install -e ".[docs,test]"

3. To run code or tests within this environment, use ``uv run`` (which automatically detects the virtual environment). For example, to run the test suite:

   .. code-block:: bash

      uv run pytest

mstimberg:
Hi @tharun-mk. Why not change the `pyproject.toml` file? I think if you did, you wouldn't need your point two, and calling `uv run pytest` would automatically install the environment. But also note that calling `pytest` directly is not a good way of calling Brian's test suite (fixing this would be #1129). Instead, there are scripts in `dev/tools` or you can call it programatically from Python (https://brian2.readthedocs.io/en/stable/developer/guidelines/testing.html#running-the-test-suite).

mstimberg:
Hi @tharun-mk are you still interested in/have time for working on this?

PR #1802, opened 2026-03-25
Warn when NetworkOperation is present during Network.store()

Sanchit2662:
I noticed that if you have a NetworkOperation with closure state in your network and call store(), nothing happens to that state , it's silently skipped. Then  after restore(), the operation's internal variables are still at their current values, not the values they had at the time of store. There was no warning about this anywhere.

Added a warning in store() that fires (once) whenever a NetworkOperation is in the network, telling the user that its internal state won't be captured or restored.

### DIFF
--- a/brian2/core/network.py
+++ b/brian2/core/network.py
@@ -18,6 +18,7 @@ from collections.abc import Mapping, Sequence
 
 from brian2.core.base import BrianObject, BrianObjectException
 from brian2.core.clocks import defaultclock
+from brian2.core.operations import NetworkOperation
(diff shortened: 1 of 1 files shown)


mstimberg:
Hi @Sanchit2662. This looks good (state like this should be rare, but it is a reasonable thing to warn about, I think). Could you please fix the code formatting and include a tests that checks that the warning is raised. Thanks :pray:

mstimberg:
Hi @Sanchit2662 are you still interested in/have time for addressing my comment for this PR?

PR #1812, opened 2026-04-01
Draft: Fix #1350 - Preserve exprel, expm1, and log1p during SymPy translation

### LINKED ISSUE #1350: Revisit sympy processing of exprel
SymPy does all kind of transformations of equations to replace constant terms (e.g. it will replace `exp(0)` by `1`), and that is of course desirable in general. However, in the context of the `exprel` function which is not used for its mathematical properties but rather for accuracy reasons, this can be problematic. The situation for this actually got a bit better since we no longer considered all numerical literals as floating point values, but I think it is still problematic. 
Here's a quick example:
```pyconsole
>>> exprel = DEFAULT_FUNCTIONS['exprel']
>>> exprel.sympy_func(sympy.sympify('x + 2'))
exprel(x + 2)
>>> exprel.sympy_func(sympy.sympify('x + 2.0'))
>>> (7.38905609893065*exp(x) - 1)/(x + 2.0)
```
In the second formulation, we lose all the advantage of the numerically more accurate `exprel` function. I guess the same also applies to `expm1` et `log1p`.

ayush4874:
Hi @mstimberg,

Got the `evaluate=False` fix working in `rendering.py`. It intercepts `exprel`, `expm1`, and `log1p` during the AST walk and passes the flag directly. 

This preserves the AST nodes for the C++ codegen without touching the global parser or breaking float evaluations downstream.

Tests are passing locally and I've added a regression test, but opening this as a draft so you can verify the architecture.

### DIFF
--- a/brian2/parsing/rendering.py
+++ b/brian2/parsing/rendering.py
@@ -241,8 +241,15 @@ class SympyNodeRenderer(NodeRenderer):
         elif len(node.args) == 0:
             return self.render_func(node.func)(sympy.Symbol("_placeholder_arg"))
         else:
+            unevaluated_funcs = ['exprel', 'expm1', 'log1p']
+            
+            kwargs = {}
+            if getattr(node.func, 'id', None) in unevaluated_funcs:
+                kwargs['evaluate'] = False
+                
             return self.render_func(node.func)(
-                *(self.render_node(arg) for arg in node.args)
+                *(self.render_node(arg) for arg in node.args),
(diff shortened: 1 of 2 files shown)


mstimberg (approved):
Apologies that it took so long for me to get back to this, but everything looks good to me! I did a minor change in the way the code was written, but this is more a stylistic choice than anything important. I'll merge if the tests pass.
Thanks for the contribution!

PR #1813, opened 2026-04-10
Refactor: Implement Lazy Loading for heavy dependencies and fix version parsing

pradeeep-pal:
Overview
This PR addresses the startup latency issue identified in #1770. The goal was to modernize the package infrastructure by implementing lazy loading for heavy dependencies like matplotlib (pylab), Cython, and pytest, following the SPEC 0001 recommendations for scientific Python libraries.

Key Changes
Lazy Loading for Pylab:

Replaced the top-level from pylab import * with a module-level __getattr__ hook.

This prevents matplotlib from being loaded during the initial import brian2, significantly reducing startup time for non-plotting scripts.

Deferred Cache Initialization:

Postponed the call to _check_caches() and wrapped it in a lazy initialization guard (_ensure_caches_checked).

This avoids early imports of Cython and pytest modules.

Robust Version Parsing:

Refactored the __version_tuple__ logic using re.findall(r'\d+', __version__).

This fixes a ValueError caused by setuptools_scm post-release strings (e.g., '2.10.post7113') which contain non-numeric characters.

Verification
Verified using python -X importtime.

Before: matplotlib and Cython were part of the initial import tree.

After: These modules are only loaded on-demand when pylab or cache-related functions are accessed.

Validated that the version parsing now correctly handles development and post-release version strings.

### DIFF
--- a/brian2/__init__.py
+++ b/brian2/__init__.py
@@ -4,7 +4,7 @@ Brian 2
 
 import logging
 import signal
-
+import re
 
 def _check_dependencies():
     """Check basic dependencies"""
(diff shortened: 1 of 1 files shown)


mstimberg:
Hi @pradeeep-pal. There are various issues with this PR. I will close it for now but please do not hesitate to open a new one if you are still interested in this. Here quickly a list of remarks:
- the code needs to be formatted to pass our pre-commit checks, have a look at our developer guide in the documentation
- the lazy import of pylab will break scripts that rely on a (admittedly ugly) `from brian2 import *` and then directly use `plot`, etc.
- I did not get how the deferred cache check is supposed to work, it simply seems to be skipped, i.e. none of the cache checking code is called?
- the PR removed a few unrelated docstrings
- comments like "FIXED: ..." shouldn't be in the actual code, this is something that should go into a commit message or a changelog

PR #1852, opened 2026-07-28
brian2cuda: Forward optional compiler_kwds from BinomialFunction generators.

SinYita:
An updated generator for `BinomialFunction` now supports returning `(code, dependencies, compiler_kwds)` instead of just `(code, dependencies)`. This fix allows external backends like Brian2CUDA to specify extra compiler settings and include necessary CUDA headers (such as `"rand.h"` or `<curand_kernel.h>`) when generating per-instance code, which was previously unsupported.


mstimberg (approved):
LGTM! Will merge once the tests pass.

PR #1853, opened 2026-07-29
brian2cuda: replace the defualt std c++ version from c++11 to c++17

SinYita:
This modification is brian2cuda relevant Replace the default std version from c++ 11 to c++ 17. [c++11 will eventually not be supported in future CUDA toolkit](https://github.com/NVIDIA/thrust/blob/main/CHANGELOG.md#thrust-1100-nvidia-hpc-sdk-209-cuda-toolkit-112). And thrust is merged to cccl, which only supports c++17 and c++20.


mstimberg (approved):
All good from my side 👍

PR #1854, opened 2026-07-29
Allow deferred preference validation for device categories

### LINKED ISSUE #1340: Brian preference files don't allow setting perferences registered in externel packages
During the import of `brian2`, the validity of preferences set in preference files is checked. At time of the brian2 import, any external package that defines a new preference (e.g. brian2cuda or brian2genn) will not be imported yet. If the preference files sets a preference registered in the external package, one gets a `Preference cateogry ... is unregistered` error.

I think it would be nice to support preference files  brian2cuda / brian2genn as well. One solution would be to move the preference check from `brian2/only.py` into the first `run` call. Alternatively, one could implement a way that brian2 is imported without the check and have brian2cuda / brian2genn be imported before brian2 (they would internally still import brian2 but without the preference check). When brian2 is imported in the main script, the check would run, but external preferences would be registered already.

mstimberg:
This implements a solution to #1340 and brian-team/brian2cuda#281 that doesn't require any fundamental changes: unknown preferences that are in a `devices.*` category do not raise errors but stay unvalidated. They will be validated when the `brian2cuda` package registers the preference.
CC @SinYita 
Closes #1340

### DIFF
--- a/brian2/core/preferences.py
+++ b/brian2/core/preferences.py
@@ -184,8 +184,17 @@ class BrianGlobalPreferences(MutableMapping):
     def __setitem__(self, name, value):
         basename, endname = parse_preference_name(name)
         if basename not in self.pref_register:
+            parts = basename.split(".")
+            if len(parts) >= 2 and parts[0] == "devices":
+                # This is an unknown preference category, but it is possibly defined
+                # for a device such as brian2cuda. Do not raise an error for now.
+                # If this is a preference for a 3rd-party device package it
+                # will be set and validated at the set_device call.
+                # If the category is wrong even after importing all other packages, it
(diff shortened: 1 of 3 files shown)


SinYita:
@mstimberg  nice and gentle fix !  Much simpler than I thought.

mstimberg:
Thanks :blush: I still fixed one small regression. Doing something like
```
prefs.devices.cuda_standalone.incorrect_pref = 0
```
did not raise an error anymore, but in this case we can actually raise one – if the user sets a preference directly, they will have `brian2cuda` already imported.

SinYita (approved):
Everything looks good to me !

Issue #1823, opened 2026-05-06
Replace integration tests by unit tests


mstimberg:
Our current test suite consists mainly of integration tests, even simple tests run the full machinery and often end with a `run(0*ms)` to trigger code generation/compilation. Instead, we should add proper unit tests and use techniques like mocking to only test the relevant part of the code base. This should significantly lower the time taken by our test suite which is currently way to long (up to 2h even with the massively parallel workflows).

nishantraghuvanshi:
Hi @mstimberg 
Starting to work on this issue to replace integration tests with unit tests.
I am planning to measure the current baseline time for running test and then create unit tests and compare timings.

PR #1861, opened 2026-08-13
Skip gc.collect() in Network.run() when unused-object warnings are disabled

AxelNoun:
Recreates #1860, which was closed automatically on 12 Aug when the fork
hosting its branch was detached during an account cleanup on my side —
my mistake, apologies for the noise.

The head commit is byte-identical to the original PR (`f9291df8be`).
All prior context and discussion: https://github.com/brian-team/brian2/pull/1860

---

Skips the explicit `gc.collect()` at the start of `Network.run()` when the `logging.warn_for_unused_objects` preference is disabled.

### Motivation

The only purpose of this collection is to trigger `BrianObject.__del__` for objects that were created but never included in a network, so that the "was never included in a network" warning can be raised. That warning is already gated by `prefs.logging.warn_for_unused_objects` inside `BrianObject.__del__`, so when the preference is `False` the collection has no user visible effect, but it still has a fixed cost: a full pass over all tracked objects (about 140k of them right after `import brian2`, roughly 80 ms per call on my machine).

This fixed cost dominates short runs. On Windows (CPython 3.14), 500 consecutive `run(0*ms)` calls with the warning disabled spent about 37.9 s out of 40 s (more than 90%) inside `gc.collect()`. Since #1823 points out that many tests end with a `run(0*ms)` just to trigger code generation, and since the test runner already sets `logging.warn_for_unused_objects = False` (`brian2/tests/__init__.py`), the whole test suite benefits directly from this change: on a small Linux machine (1 vCPU, CPython 3.12), `gc.collect()` accounted for roughly 45 to 50% of the wall time of the `codegen_independent` run of `test_neurongroup.py` before this change, and only a few percent after. These numbers are indicative and machine dependent; happy to rerun benchmarks if useful.

### Behaviour

* Default behaviour is unchanged: the preference defaults to `True`, so the collection happens exactly as before.
* With the preference disabled, no warning was emitted before and none is emitted now. The only observable difference is that unused objects are collected later, by the regular generational garbage collector, instead of eagerly at the start of `run()`. Users who rely on the eager collection for memory reasons can call `gc.collect()` themselves.
* The existing tests `test_unused_object_warning` and `test_magic_unused_object` pass unchanged (they enable the preference explicitly).

### Changes

* `brian2/core/network.py`: make the collection conditional and explain why in the comment.
* `brian2/utils/logger.py`: document the link in the description of the preference (this also updates the generated preference documentation).
* `brian2/tests/test_network.py`: add `test_conditional_gc_collect`, a regression test checking via `unittest.mock` that the collection happens if and only if the preference is enabled.

### Out of scope

* `MagicNetwork.after_run()` also calls `gc.collect()` after every magic `run()`. That call frees objects that were kept alive by the magic network, so it is not purely warning related and is left untouched here; it could be revisited as part of #1823.
* The broader test suite restructuring of #1823, and `cpp_standalone` run overhead, which is a separate topic.

Supersedes #1843: I had opened the same core change from a branch that has since been deleted, and closed that PR before review. Compared to #1843, this version adds the regression test and the documentation update for the preference.

### DIFF
--- a/brian2/core/network.py
+++ b/brian2/core/network.py
@@ -1106,8 +1106,13 @@ class Network(Nameable):
             raise ValueError(
                 f"Function 'run' expected a non-negative duration but got '{duration}'"
             )
-        # This will trigger warnings for objects that have not been included in a network
-        gc.collect()
+        # Trigger a garbage collection to raise warnings for objects that have
+        # been created but never included in a network (see
+        # `BrianObject.__del__`). A full collection can take a significant
+        # amount of time, in particular for short or repeated runs (see #1823),
+        # so it is only performed when these warnings are enabled.
+        if prefs.logging.warn_for_unused_objects:
+            gc.collect()
         device = get_device()  # Do not use the ProxyDevice -- slightly faster
 
         if profile is None:
--- a/brian2/tests/test_network.py
+++ b/brian2/tests/test_network.py
@@ -1908,6 +1908,31 @@ def test_unused_object_warning():
 
 
 @pytest.mark.codegen_independent
+def test_conditional_gc_collect():
+    # The garbage collection at the start of Network.run only serves to raise
+    # the warning for unused objects. It is costly and therefore skipped when
+    # the warning is disabled (see #1823).
+    import gc
+    from unittest import mock
+
+    group = NeuronGroup(1, "v : 1")
+    net = Network(group)
+    _old_pref = prefs.logging.warn_for_unused_objects
+    try:
+        with catch_logs():
+            net.run(0 * ms)  # warm up (code generation, caches, ...)
+            with mock.patch.object(gc, "collect", wraps=gc.collect) as mocked:
+                prefs.logging.warn_for_unused_objects = False
+                net.run(0 * ms)
+                assert mocked.call_count == 0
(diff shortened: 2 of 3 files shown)


mstimberg (approved):
Thanks for the PR, and apologies for having taken so long to review. This makes perfect sense to me, and everything looks good. I made a tiny change in the comments, and replaced references to #1823 (which is general and does not specifically talk about garbage collection) by #1861 (i.e., this PR).

AxelNoun:
> Thanks for the PR, and apologies for having taken so long to review. This makes perfect sense to me, and everything looks good. I made a tiny change in the comments, and replaced references to #1823 (which is general and does not specifically talk about garbage collection) by #1861 (i.e., this PR).

Thanks for the review and for updating the comments and issue references! I appreciate you taking the time to look it over.

Are there any other tasks or issues I should focus on as a priority right now?

mstimberg:
> Are there any other tasks or issues I should focus on as a priority right now?

I'm currently trying to wrap up things to do a release (it's been way too long…) and I am quite busy with some other things, but I'll give it a thought and get back to you! Thanks for the offer to help out :blush:

PR #1869, opened 2026-09-02
Add dtype keyword to StateMonitor

mstimberg:
Useful in particular to save memory by using float32 to record a float64 variable

I actually thought we had this already (see my comment at https://brian.discourse.group/t/memory-error-insufficient-disc-space/1471/2) …


mstimberg:
Note to self: check whether this breaks Brian2CUDA (it shouldn't, but still)

Issue #1825, opened 2026-05-06
Improve compilation speed in standalone mode


mstimberg:
While compilation in standalone mode is significantly faster than in Cython runtime (which we are already working on replacing completely, see #1769), it is still slower than it could/should be, especially in Brian2CUDA (see brian-team/brian2cuda#179). We should investigate ways of improving this, e.g. by merging code object files together (to have fewer files to compile in total), by pre-compiling header files and library functions, and potentially by using helper tools such as ccache.

PR #1871, opened 2026-09-11
Collect standalone build dependencies during compilation on POSIX

viranovskaya:
Related to #1825.

This replaces the separate whole-project `-MM` pass with per-object dependency files generated during compilation on Linux and macOS.

It also fixes a dependency-target mismatch: the existing output can refer to `foo.o` when the actual target is `code_objects/foo.o`, which can cause header changes to be missed.

The patch records the correct object target, rebuilds objects whose dependency files are missing, and removes generated dependency files during cleanup. C compilation keeps make's standard compiler command and flag precedence. Windows/MinGW retains its existing behavior, and the MSVC template is unchanged.

For this patch, the relevant standalone and Makefile tests passed on macOS: 34 passed, with one existing Apple Silicon skip. Nine portable build-contract tests also passed on Linux/GCC using a host-generated fixture. Seeded baseline and patched builds produced byte-identical outputs, including a four-thread OpenMP smoke test on Linux.

In four alternating build pairs per model on Apple Clang, median compilation time changed from:

- 8.69 to 6.71 seconds for a 13-file model;
- 76.66 to 62.46 seconds for a 94-file model.

These are local measurements; OS caches and background activity were not controlled. Native Windows and CUDA builds have not been tested.

Would you be happy to take this as a small POSIX-only first step, with PCH handled separately?


mstimberg (commented):
Thank you for the PR, @viranovskaya, building the individual .d files seems indeed the right approach. This globally looks good, but I am unclear about two things, which I hope you can clarify for me:
1. You keep the "legacy behavior" around for MinGW on Windows – do you know whether this is necessary, i.e. does MinGW not support `-MMD`? I'm actually not sure whether things currently build with MinGW, we do not test this – in the days of WSL, I'm not sure it is used much.
2. You add a new rule for .c files, and a lot of the testing is around handling both c and c++ files correctly. Would you mind commenting on why you think this is necessary? All the files we generate are .cpp files, as far as I can tell.

viranovskaya:
Thanks, Marcel — both are fair questions.

**1. MinGW:** I kept the existing path because I had not tested the change on native Windows, not because of a known lack of `-MMD` support. GCC-based MinGW should support these dependency-generation flags. I don’t have evidence that it needs a separate dependency strategy, so I’m happy to remove that split and use the same approach in the GNU make template, leaving MSVC unchanged. Native Windows build and cleanup behaviour would still need validation.

**2. C files:** You’re right that Brian’s generated sources are `.cpp`. My reason was to preserve the existing handling of additional C sources: `CPPWriter` already recognises `.c`, and the original Makefile already maps `.c` to `.o`, relying on make’s built-in C compilation rule. The `additional_source_files` docstring does only mention `.cpp`, though, so I should distinguish that existing behaviour from a documented C-source API.

With dependency files now expected for every object, leaving C compilation to the built-in rule would never create its `.d` file. The missing-dependency recovery would therefore rebuild that C object on every invocation. I reproduced this in a small GNU make/Clang example: with the explicit rule, the second build is a no-op; without it, the C source is compiled again and its `.d` remains absent. Using `$(COMPILE.c)` preserves make’s existing C compiler and flag ordering.

I agree the tests give this edge case disproportionate weight. I can make the main incremental-build tests C++-only and retain a focused C compatibility test.

mstimberg:
Thanks for the reply. Regarding the first point, I will look into running the tests on Windows with MinGW – this might not work at all in which case we could simply declare it unsupported for now, and we would not have to do the additional `if auto_dependencies` step. And if it works, we can see whether it works with `-MMD` as well.

Regarding the `.c` support: you are right about `CPPWriter` supporting `.c` files, but I think this is a left-over from when we included a `randomkit.c` file to implement random numbers (which has been replaced by use of the standard library since then). So I think the cleanest solution would be actually to not support `.c` files anywhere and therefore not need those additional rules and checks. Our code generation code is already very complex and fragile, everything we can do to simplify things is welcome :blush:

viranovskaya:
Thanks, that makes sense. I've removed the leftover C-source handling and made the incremental-build tests C++-only. Additional sources now require the documented `.cpp` extension.

I've also updated against current master and moved dependency cleanup into `code_files_to_delete()`. Local standalone tests pass: 35 passed, 1 skipped on macOS/Apple Clang.

I've left the MinGW path unchanged pending your check; I haven't run a native Windows test. The updated workflows are awaiting maintainer approval — could you approve them when convenient?

